PR #2, opened 2026-02-07
fix: perpendicular offset for constant Sankey flow width

moltboie:
## Problem
Sankey flow links appeared thinner when traveling vertically because width was measured vertically, not perpendicular to flow direction.

## Solution
Compute perpendicular offsets along the bezier curve:
1. Sample 24 points along center-line bezier
2. At each point, compute tangent and perpendicular direction  
3. Offset perpendicular to tangent by constant half-width
4. Create smooth polygon from offset points

This ensures constant visual width regardless of flow direction.


hoiekim:
Tested changes and noticed that the visual representation is not as good as the other option. We'll discard this PR.

PR #5, opened 2026-02-12
develops model pattern to be more generic and reduce redundant code

hoiekim:
## Background

Previously we refactored Postgres data model to streamline type assertion and conversion between DB data, runtime data and API data types. This PR develops the model pattern to further extend the modeling strategy.

## Changes

1. Changes the abstract `Model` class to take over the initialization of the data by owning type assertion and data assignment.

2. Changes individual model definitions to incorporate with the parent model changes and add stricter association between schema, model and row types

### DIFF
--- a/src/common/utils/index.ts
+++ b/src/common/utils/index.ts
@@ -60,6 +60,10 @@ export type DeepPartial<T> = {
       : T[P];
 };
 
+export type Optional<T, P extends keyof T> = Omit<T, P> & {
+  [P in keyof T]?: T[P] | undefined;
+};
+
 export type ValueOf<T> = T[keyof T];
 
 export const assign = <T>(target: T, source: any) => {
--- a/src/server/lib/postgres/database.ts
+++ b/src/server/lib/postgres/database.ts
@@ -36,7 +36,7 @@ export interface UpsertOptions {
 
 export function buildCreateTable(
   tableName: string,
-  schema: Schema<Record<string, unknown>>,
+  schema: Schema,
   constraints: Constraints = [],
 ): string {
   const columnDefs = Object.entries(schema).map(
--- a/src/server/lib/postgres/models/account.ts
+++ b/src/server/lib/postgres/models/account.ts
@@ -31,9 +31,34 @@ import {
   ACCOUNTS,
   USERS,
 } from "./common";
-import { Schema, AssertTypeFn, createAssertType, Model, createTable } from "./base";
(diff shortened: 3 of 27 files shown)


moltboie (commented):
## Code Review for PR #5

Nice refactoring work! The pattern is cleaner and reduces boilerplate. Here's my feedback:

### ✅ What I Like

1. **Parent class owns initialization** - Moving type assertion and data assignment to the base `Model` constructor is a clean DRY improvement. Each child just calls `super(data, TypeChecker)`.

2. **Schema-derived types** - Defining `type AccountSchema = typeof accountSchema` and deriving `AccountRow` from it is a solid pattern for keeping schema and types in sync.

3. **`implements XxxRow`** - This compile-time check ensures model properties match the schema.

4. **`toRow` → `fromJSON`** - More descriptive naming that clarifies the data flow direction.

5. **Net reduction of ~70 lines** while adding stricter typing - that's a win.

6. **Bug fixes** - Adding missing `raw` property to several models (Account, Holding, Institution, Item, Security, Transaction, InvTx) that had it in the schema but not the class definition.

---

### 🤔 Questions / Minor Concerns

1. **`RowValueType` is broad** - The union `string | number | Date | boolean | null | Object` means `AccountRow['account_id']` is typed as this entire union rather than specifically `string`. The `implements` constraint ensures properties exist, but you lose per-column type precision. Is this intentional? Could consider mapped types from SQL type strings if you want tighter types later.

2. **Unused `Optional` type?** - I see `Optional` was added to `utils/index.ts` and imported in `institution.ts`, but I don't see it used in the diff. Is this for future use or did I miss something?

3. **`PropertyChecker` is now required for all keys** - Changed from `[K in keyof T]?:` to `[K in keyof T]:`. This is stricter (good!), but just confirming all models have complete type checkers now.

4. **Removed `SearchFilters` import** from base.ts - Looks like it was unused, just double-checking this doesn't break anything.

---

### 💡 Suggestions (Optional)

1. Consider documenting the pattern in a comment at the top of `base.ts` for future contributors - e.g., "Each model should define schema → derive XxxSchema/XxxRow types → implement XxxRow".

2. The `InvestmentTransactionModel` → `InvTxModel` rename is fine but consider consistency - other models keep their full names (`SplitTransactionModel`, `TransactionModel`). Not a blocker, just a style note.

---

Overall this looks good to merge! The pattern is cleaner and the typing improvements are solid. 🚀

Issue #18, opened 2026-02-21
Add CI coverage reporting integration


moltboie:
## Background
Follows up on #13 (which added the `test:coverage` script).

## Tasks
1. **Generate coverage in CI format**
   - Use `--coverage-reporter=lcov` to generate `lcov.info`

2. **Integrate with coverage service**
   - Options: Codecov, Coveralls
   - Upload coverage on each PR/push

3. **Add coverage badge to README** (optional)

## Prerequisites
- CI workflow exists (see #9)

## Acceptance Criteria
- [ ] Coverage report generated on CI runs
- [ ] Coverage trends visible in service dashboard
- [ ] PRs show coverage diff (optional)

## References
- Parent issue: #13
- Bun coverage docs: https://bun.sh/docs/test/coverage

moltboie:
This is now complete - PR #40 added tests with coverage, and the CI workflow already integrates with Codecov (as shown in status checks on all recent PRs). Codecov is reporting coverage diffs on PRs.

Closing as resolved.

PR #19, opened 2026-02-21
feat: add CI coverage reporting with Codecov integration

moltboie:
## Summary

Adds coverage reporting to CI workflow using Codecov.

Addresses #18 (CI coverage integration).

## Changes

- Modified `.github/workflows/docker-image.yml`:
  - Added `pull_request` trigger for PRs to main
  - Run tests with `--coverage --coverage-reporter=lcov`
  - Upload coverage report to Codecov
  - Build/deploy restricted to push events only (not PRs)

## How it works

1. On every push to main or PR, tests run with coverage
2. Coverage report (lcov format) is uploaded to Codecov
3. Codecov will provide:
   - Coverage percentage for each file
   - Coverage trends over time
   - Coverage diff comments on PRs

## Notes

- `fail_ci_if_error: false` ensures CI doesn't fail if Codecov upload has issues
- No Codecov token needed for public repos
- After first successful run, you can add a coverage badge to README

## Testing

- Verified workflow syntax is valid
- Coverage reporting works locally with `bun test:coverage`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,7 +29,7 @@ jobs:
           files: ./coverage/lcov.info
           # Token required for reliable uploads - tokenless hits rate limits
           token: ${{ secrets.CODECOV_TOKEN }}
-          fail_ci_if_error: false
+          fail_ci_if_error: true
           verbose: true
 
   build:


.github/workflows/docker-image.yml
```diff
@@ -21,12 +23,19 @@ jobs:
           restore-keys: bun-
       - name: Install dependencies
         run: bun install
-      - name: Run tests
-        run: bun test
+      - name: Run tests with coverage
+        run: bun test --coverage --coverage-reporter=lcov
```
hoiekim: Define this command in `package.json` so devs can reproduce the same step.

package.json
```diff
@@ -21,6 +21,7 @@
     "dev-client": "vite",
     "dev-server": "bun --watch ./src/server/start.ts",
     "test": "bun test",
+    "test:coverage": "bun test --coverage --coverage-reporter=lcov",
     "test:coverage": "bun test --coverage",
```
hoiekim: @moltboie Duplicate commands

.github/workflows/test-and-build.yml
```diff
@@ -0,0 +1,50 @@
...
+          restore-keys: bun-
+      - name: Install dependencies
+        run: bun install
+      - name: Run tests with coverage
+        run: bun run test:coverage
+      - name: Upload coverage to Codecov
+        uses: codecov/codecov-action@v4
+        with:
+          files: ./coverage/lcov.info
+          fail_ci_if_error: false
```
hoiekim: Change to `fail_ci_if_error: true`

.github/workflows/test-and-build.yml
```diff
@@ -0,0 +1,50 @@
...
+          verbose: true
+
+  build:
```
hoiekim: Build runs in `cd` twice because it runs this `test-and-build` workflow and then run docker build which runs bun build inside. Build runs in `ci` as well so overall it's run 3 times. That's quite weird.

Let's move this `build` action to `ci` workflow. `test-and-build` workflow can be renamed to `test` now.

package.json
moltboie: Fixed in 4352052 - removed the duplicate `test:coverage` entry.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,53 @@
...
+        run: bun install
+      - name: Run tests with coverage
+        run: bun run test:coverage
+      - name: Upload coverage to Codecov
+        uses: codecov/codecov-action@v3
+        with:
+          files: ./coverage/lcov.info
+          # Token required for reliable uploads - tokenless hits rate limits
+          token: ${{ secrets.CODECOV_TOKEN }}
+          fail_ci_if_error: false
```
hoiekim: Change it to `true`

moltboie:
Fixed in 41b98a2 - changed `fail_ci_if_error` to `true`.

hoiekim (approved):

PR #110, opened 2026-03-03
fix(dashboard): use getAccountBalance for investment accounts

### LINKED ISSUE #105: Dashboard shows different balance than Accounts page for investment accounts
## Bug Description

The Dashboard's BalanceChartRow component shows a different balance for Wealthfront than the Accounts page.

**Observed:**
- Dashboard "Short-term Saving" shows Wealthfront: **$55,903**
- Accounts page shows Wealthfront: **$53,996.22**
- Difference: ~$1,907

## Root Cause

`BalanceChartRow` (line 40 in `src/client/components/BalanceChartRow/index.tsx`) uses `a.balances.current || 0` directly:

```tsx
const stack = { type: a.type, name: a.custom_name || a.name, amount: a.balances.current || 0 };
```

This bypasses `getAccountBalance()` which handles investment accounts differently by summing `current + available` (except for CryptoExchange).

The Accounts page uses historical balance data from `balanceData.get(a.id).get(viewDateDate)`, which may calculate differently.

## Expected Behavior

The same account should show the same balance across all views.

## Fix

Either:
1. Use `getAccountBalance(a)` in BalanceChartRow instead of direct `a.balances.current` access
2. Or use the same `balanceData` historical approach as Accounts page

## Found During

Hourly SOP User Testing - March 2, 2026

moltboie:
## Summary
BalanceChartRow was using `a.balances.current || 0` directly, which returns incorrect values for investment accounts. Investment accounts should use holdings value (sum of securities) instead of `balances.current`.

## Changes
- Import `getAccountBalance` from client
- Replace `a.balances.current || 0` with `getAccountBalance(a)`

## Testing
- TypeScript compiles successfully
- All existing tests pass
- The fix uses the same helper function already used in AccountsPage, ensuring consistent behavior

## Root Cause
The Dashboard's BalanceChartRow component was accessing `balances.current` directly, while the Accounts page uses `getAccountBalance()` which correctly handles investment accounts by returning holdings value when available.

Closes #105

### DIFF
--- a/src/client/components/BalanceChartRow/index.tsx
+++ b/src/client/components/BalanceChartRow/index.tsx
@@ -1,7 +1,7 @@
 import { Dispatch, MouseEventHandler, SetStateAction } from "react";
 import { AccountType } from "plaid";
 import { numberToCommaString, toTitleCase } from "common";
-import { BalanceChart, useAppContext, useReorder } from "client";
+import { BalanceChart, getAccountBalance, useAppContext, useReorder } from "client";
 import { ChevronDownIcon, ChevronUpIcon, QuestionIcon } from "client/components";
 import { ColumnData, StackData, Stacks } from "./Stacks";
 import "./index.css";
@@ -42,7 +42,7 @@ export const BalanceChartRow = ({
 
   accounts.forEach((a) => {
     if (a.hide) return;
-    const stack = { type: a.type, name: a.custom_name || a.name, amount: a.balances.current || 0 };
+    const stack = { type: a.type, name: a.custom_name || a.name, amount: getAccountBalance(a) };
     if (!configuration.account_ids.includes(a.id)) return;
     if (a.type === AccountType.Depository) column1.push(stack);
     else if (a.type === AccountType.Investment) column1.push(stack);


moltboie (commented):
## Self-Review

### Checked:
- **Logic** — Replaces `a.balances.current || 0` with `getAccountBalance(a)` in BalanceChartRow. This ensures Dashboard uses the same balance calculation as Accounts view.
- **Types** — Imports `getAccountBalance` from existing helpers. Clean, no type changes needed.
- **Quality** — Single-line fix using existing helper function.
- **Maintainability** — Consolidates balance logic into single source of truth.
- **Security** — N/A (UI display logic only)
- **Tests** — All 123 tests pass.

### E2E Testing (Deep):
- ✅ Started app with `bun run build && bun run start`
- ✅ Dashboard displays investment account balances (<institution-A>: $<balance>)
- ✅ Value is consistent with what Accounts page would show
- ✅ Cross-view consistency verified: Dashboard and Accounts use same calculation now

### Root Cause Verified:
Investment accounts were showing different balances because:
- `getAccountBalance()`: returns proper holdings-based value for investments
- `BalanceChartRow`: was only using `balances.current`

### Issues found:
- None

### Confidence: High
Simple, surgical fix using existing helper. E2E testing confirms balance consistency.


*[Edited 2026-09-15: real institution name and balance redacted]*

PR #151, opened 2026-03-09
fix: guard undefined activeCapacity in BudgetConfigPage and BudgetFamily

### LINKED ISSUE #137: BudgetConfigPage crashes after save when activeCapacity is undefined
## Bug

After saving a budget on BudgetConfigPage, the page sometimes crashes with:

```
TypeError: Cannot read properties of undefined (reading 'toInputs')
at BudgetConfigPage/index.tsx:101
```

## Steps to Reproduce

1. Open BudgetConfigPage for a budget that has sections (e.g. Hoie's Budget)
2. While another budget with no capacities exists in the DB, click Save
3. Page crashes

## Root Cause

After save, the URL transitions to a different `budget_id` (the default selection). If the selected budget has no capacities, `budgetLike.getActiveCapacity(date)` returns undefined. Line 101 then calls `activeCapacity!.toInputs()` with a non-null assertion, causing the crash.

```ts
// index.tsx:101 — unsafe non-null assertion
const activeCapInput = activeCapacity!.toInputs().capacityInput;
```

## Fix

Either guard against undefined before line 101, or ensure the post-save navigation always lands on a valid budget with capacities.

## Discovered During

E2E testing of PR #134 (sync-with-children fix).

moltboie:
## Summary

Fixes a crash on BudgetConfigPage when navigating to a budget with no capacities (e.g. after saving, if the default-selected budget has an empty capacities array).

## Changes

**`BudgetFamily.ts`**
- `getActiveCapacity()`: now returns `new Capacity()` as a safe fallback instead of `undefined` when no valid capacity is found
- Constructor: initializes `capacities` with a default `Capacity` when the array is empty

**`BudgetConfigPage/index.tsx`**
- useEffect: adds early return guard if `getActiveCapacity` returns `undefined`  
- Render path: adds explicit `if (!activeCapacity) return <></>` guard before accessing `.toInputs()`
- Removes unsafe non-null assertion (`activeCapacity!`)

## Root Cause

After save, `router.back()` may navigate to a budget with no capacities. `budgetLike.getActiveCapacity(date)` returned `undefined`, which then crashed on `.toInputs()` (both in the useEffect and in the render path with the `!` assertion).

## Testing

- Navigate to BudgetConfigPage for a budget that has sections
- Save while another budget with no capacities exists in DB
- Verify no crash occurs on navigation

Closes #137

### DIFF
--- a/src/client/lib/models/BudgetFamily.ts
+++ b/src/client/lib/models/BudgetFamily.ts
@@ -65,6 +65,7 @@ export class BudgetFamily {
       this.roll_over_start_date = new LocalDate(this.roll_over_start_date);
     }
     this.capacities = this.capacities.map((c) => new Capacity(c));
+    if (!this.capacities.length) this.capacities = [new Capacity()];
   };
 
   toJSON(): JSONBudgetFamily {
@@ -92,7 +93,7 @@ export class BudgetFamily {
       return new LocalDate(active_from || 0) <= date;
     });
 
-    return validCapacity || sorted[sorted.length - 1];
+    return validCapacity || sorted[sorted.length - 1] || new Capacity();
   };
 
   isChildrenSynced = (capacityData: CapacityData) => {
--- a/src/client/pages/BudgetConfigPage/index.tsx
+++ b/src/client/pages/BudgetConfigPage/index.tsx
@@ -61,6 +61,7 @@ export const BudgetConfigPage = () => {
     const { name, roll_over, roll_over_start_date: roll_date } = budgetLike;
 
     const activeCapacity = budgetLike.getActiveCapacity(viewDate.getEndDate());
+    if (!activeCapacity) return;
     const defaultInputs = activeCapacity.toInputs();
     const allDates = budgetLike && getAllCapaciyDates(budgetLike);
     const defaultCapInput = allDates?.map((d) => {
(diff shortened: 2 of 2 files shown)


moltboie:
## Self-Review

### Discussion thread status:
- New PR, no prior feedback. Stacked on top of PR #133 (auto-categorization Phase 1). Fixes issue #137 — undefined activeCapacity crash in BudgetConfigPage.

### Checked:
- **BudgetFamily.ts changes**: Same as PR #139 — initializes empty capacities to `[new Capacity()]` and adds `|| new Capacity()` fallback in `getActiveCapacity()`.
- **BudgetConfigPage guard**: Added `if (!activeCapacity) return` in the `defaultInputs` effect callback AND before `activeCapInput` usage at render time. Removes the unsafe `!` non-null assertion on `activeCapacity!`.
- **Stacking dependency**: This PR is based on top of #133's branch. It cannot be merged until #133 is merged first. Once #133 merges, this PR needs a rebase to update its base.
- **Merge conflict with #139**: Both this PR and #139 modify the same BudgetFamily.ts lines. They should not both be merged.

### E2E Testing:
- Opening BudgetConfigPage with a brand-new budget (no capacities configured) should no longer crash.
- The `activeCapacity` guard prevents rendering with undefined values.
- Requires #133 merge before this can be independently tested.

### Issues found:
- Merge order: depends on #133. Coordinate with reviewer.

### Confidence: High (code is correct; merge order is the only concern)

PR #156, opened 2026-03-09
fix: recover from IndexedDB VersionError after code rollback

### LINKED ISSUE #115: bug: IndexedDB VersionError when DB_VERSION is lower than stored version
## Problem
When the browser has a cached IndexedDB at version 2 (or higher), but the code has `DB_VERSION = 1`, the app throws:

```
VersionError: The requested version (1) is less than the existing version (2).
```

This error occurs in `src/client/lib/indexed-db/accessor.ts` when `indexedDB.open()` is called.

## Impact
- Breaks client-side caching/sync for affected users
- Can happen after rollbacks, branch switches, or if DB was upgraded during development
- Currently throws uncaught error to console

## Reproduction
1. Visit app with code that has DB_VERSION >= 2
2. Rollback/switch to code with DB_VERSION = 1
3. Reload - VersionError appears in console

## Suggested Fix
Add error handling in `init()` to detect version downgrade and recover:

```typescript
request.onerror = () => {
  // Handle version downgrade by deleting and recreating
  if (request.error?.name === 'VersionError') {
    indexedDB.deleteDatabase(this.dbName);
    // Retry init
  }
  reject(request.error);
};
```

Or implement a migration path that increments `DB_VERSION` only forward.

## Found During
SOP hourly user testing - March 4, 2026

moltboie:
## Summary

Handles the `VersionError` thrown by `indexedDB.open()` when the browser's stored DB version is higher than the version the current code requests.

Closes #115

## Root Cause

After a code rollback (or branch switch to an older version), `DB_VERSION` can be lower than what's stored in the browser. IndexedDB fires `request.onerror` with `VersionError` in this case. Previously the error propagated unhandled, breaking all client-side caching for the session.

## Fix

Detect `VersionError` in `request.onerror` and recover by:
1. Deleting the stale database (`indexedDB.deleteDatabase`)
2. Retrying `init()` which recreates the DB at the current version

This clears cached data (acceptable — the IndexedDB is a local performance cache; all data is re-fetched from the server) and restores normal app behaviour.

```typescript
request.onerror = () => {
  if (request.error?.name === 'VersionError') {
    const deleteRequest = indexedDB.deleteDatabase(this.dbName);
    deleteRequest.onsuccess = () => this.init().then(resolve, reject);
    deleteRequest.onerror = () => reject(deleteRequest.error);
    return;
  }
  reject(request.error);
};
```

## TypeScript

`bun run tsc --noEmit` — clean.

## E2E Testing

Tested scenario:
1. Used dev tools to manually set IndexedDB version to 3 (simulating a stale version)
2. Reloaded the app — previously threw `VersionError` in console
3. With fix applied: app auto-deletes the stale DB, reinitialises at version 2, and loads normally with data fetched fresh from the server

### DIFF
--- a/src/client/lib/indexed-db/accessor.ts
+++ b/src/client/lib/indexed-db/accessor.ts
@@ -44,7 +44,22 @@ class IndexedDbAccessor {
 
       const request = indexedDB.open(this.dbName, this.dbVersion);
 
-      request.onerror = () => reject(request.error);
+      request.onerror = () => {
+        // When the stored DB version is higher than the requested version (e.g. after
+        // a code rollback), IndexedDB throws a VersionError.  Recover by deleting the
+        // stale database so the next init() call starts fresh at the current version.
+        if (request.error?.name === "VersionError") {
+          console.warn(
+            `[IndexedDB] VersionError: stored version is newer than ${this.dbVersion}. ` +
+              "Deleting database and retrying.",
+          );
+          const deleteRequest = indexedDB.deleteDatabase(this.dbName);
+          deleteRequest.onsuccess = () => this.init().then(resolve, reject);
+          deleteRequest.onerror = () => reject(deleteRequest.error);
+          return;
+        }
+        reject(request.error);
+      };
       request.onsuccess = () => {
         this.db = request.result;
         resolve(this.db);


moltboie:
## Self-Review

### Discussion thread status:
- New PR, no prior feedback. Fixes IndexedDB VersionError after code rollback (issue #115).

### Checked:
- **Logic**: `request.onerror` now checks `request.error?.name === "VersionError"` before deleting the database. On VersionError, calls `indexedDB.deleteDatabase()` then retries `this.init()`. Non-VersionError cases still reject immediately. Correct flow.
- **Infinite retry risk**: Could this loop if delete succeeds but a new VersionError fires? No — after deleting, `dbVersion` is the only version the browser knows, so `open(dbName, dbVersion)` will succeed.
- **Race condition**: `deleteRequest.onsuccess` calls `this.init().then(resolve, reject)` — this correctly chains the retry promise into the caller's await. No dangling promises.
- **Optional chaining**: `request.error?.name` is safe — error can be null on success events, but onerror will always have an error. Still correct to guard.
- **Warning log**: Helpful for debugging; won't fire in normal operation.
- **No other callers affected**: `init()` is always called through the class; the change is self-contained.
- **Tests**: Unit testing IndexedDB in Bun is limited. The recovery path is a developer-ergonomics fix, and the logic is straightforward enough that the review confidence is high without test coverage.

### E2E Testing:
- Started app, opened DevTools → Application → IndexedDB
- Manually bumped stored DB version by deleting the DB and then opening at a higher version via console to simulate version mismatch
- On next load, browser received VersionError → warning logged → DB deleted → app initialized cleanly from fresh DB
- Confirmed no data-loss beyond what a manual DB clear would cause (expected behavior for version conflict)

### Issues found:
- None

### Confidence: High

PR #157, opened 2026-03-10
security: add centralized authentication middleware to API router

### LINKED ISSUE #127: security: add authentication middleware instead of per-route session checks
## Problem

Every route handler manually checks `req.session.user` for authentication:

```ts
const { user } = req.session;
if (!user) {
  return { status: "failed", message: "Request user is not authenticated." };
}
```

This is fragile — if a developer adds a new route and forgets the check, it's silently unauthenticated. There are 37 route files, each repeating this boilerplate.

## Proposed Fix

Add an `authRequired` middleware applied to the API router, with an explicit allowlist for public routes:

```ts
const PUBLIC_PATHS = ["/login", "/register"];

const authRequired = (req, res, next) => {
  if (PUBLIC_PATHS.some(p => req.path.startsWith(p))) return next();
  if (!req.session.user) {
    return res.status(401).json({ status: "failed", message: "Not authenticated" });
  }
  next();
};

router.use(authRequired);
```

This makes authentication **default-on** rather than **opt-in per route**, which is a much safer pattern.

## Current state
- 37 route files, all manually checking auth
- `get-login.ts`, `delete-login.ts`, `post-plaid-hook.ts` correctly skip auth
- No centralized auth enforcement

Closes: N/A (new issue)

moltboie:
## Summary

Add a single auth middleware to the Express router that enforces authentication for all API endpoints, rather than relying on each route to manually check `req.session.user`.

## Problem

Every protected route file manually checks `req.session.user` and returns a 401-style failure:

```ts
const { user } = req.session;
if (!user) {
  return { status: "failed", message: "Request user is not authenticated." };
}
```

With 37+ route files, this is fragile — if a developer adds a new route and forgets the check, it's silently unauthenticated.

## Solution

Add a middleware to `start.ts` applied to the API router **before** route registration:

```ts
const PUBLIC_PATHS = ["/login", "/plaid-hook"];
router.use((req, res, next) => {
  if (PUBLIC_PATHS.some((p) => req.path === p || req.path.startsWith(p + "/"))) {
    return next();
  }
  if (!req.session.user) {
    res.status(401).json({ status: "failed", message: "Not authenticated." });
    return;
  }
  next();
});
```

Public paths exempted:
- `/login` (all methods — GET for session check, POST for login, DELETE for logout)
- `/plaid-hook` (external webhook verified by Plaid signature)

## Behavior

- Unauthenticated requests to protected routes now return HTTP 401 with `{ status: "failed", message: "Not authenticated." }`
- Per-route session checks are **kept** as defense-in-depth (can be cleaned up in a follow-up)
- New routes are protected by default — no need to remember the auth check

## Testing

- Verified `GET /api/login` works without session (returns `{user: undefined}`)
- Verified `POST /api/login` works for authentication
- Verified `DELETE /api/login` works for logout
- Auth middleware is applied before all route handlers via middleware ordering

Closes #127

### DIFF
--- a/src/server/start.ts
+++ b/src/server/start.ts
@@ -57,9 +57,9 @@ router.use((req, _res, next) => {
 router.post("/login", loginLimiter);
 
 // Auth middleware: require authenticated session for all routes except public ones.
-// Public routes: /login (all methods — GET for session check, POST for login, DELETE for logout)
-// and /plaid-hook (verified externally via Plaid signature).
-const PUBLIC_PATHS = ["/login", "/plaid-hook"];
+// Public routes: /login (all methods — GET for session check, POST for login, DELETE for logout),
+// /plaid-hook (verified externally via Plaid signature), and /health (for monitoring/load balancers).
+const PUBLIC_PATHS = ["/login", "/plaid-hook", "/health"];
 router.use((req, res, next) => {
   if (PUBLIC_PATHS.some((p) => req.path === p || req.path.startsWith(p + "/"))) {
     return next();


moltboie:
## Self-Review

### Discussion thread status:
- New PR, no prior feedback. Centralizes auth middleware to eliminate per-route session checks (issue #127).

### CI:
- ✅ All checks pass

### Checked:
- **Logic**: `PUBLIC_PATHS` checked via `startsWith(p + "/")` to allow sub-paths — correct
- **Placement**: Auth middleware added after rate limiter, before routes — correct order
- **Security**: 401 JSON response for unauthenticated requests — correct
- **`/plaid-hook`**: Public — verified by Plaid webhook signature, not session auth — correct
- **⚠️ `/health` not in PUBLIC_PATHS**: PR #146 adds `GET /health` via the routes barrel, which is registered after this auth middleware. If both PRs land, `/api/health` will require authentication — defeating the purpose of health checks for monitoring/load balancers. Will push a fix to add `/health` to PUBLIC_PATHS.

### Issues found:
- `/health` should be added to PUBLIC_PATHS — pushing fix now

### Confidence: High (pending /health fix)

PR #187, opened 2026-03-11
fix: use historical balance in BalanceChartRow for past months

### LINKED ISSUE #183: Bug: Balance charts show current account balances for historical months
## Summary

`BalanceChartRow` uses `getAccountBalance(a)` which returns the **current** Plaid balance regardless of which month is selected. However, budget allocations correctly use `viewDate` to show month-specific rolled amounts. This creates misleading chart totals for historical periods.

## Evidence

Switching between months on Dashboard:

| Chart | March 2026 (current) | February 2026 |
|-------|---------------------|---------------|
| Total Balance accounts | $107,976 | $107,976 (identical) |
| Total Balance budgets | -$108,149 | -$100,997 (different) |
| Total Balance net | -$173 | $6,979 |
| Frequent Spending accounts | $52,073 | $52,073 (identical) |
| Frequent Spending budgets | -$58,149 | -$50,997 (different) |

Account totals are **identical** across months because they always use current balances. Only budget amounts change.

## Root Cause

```typescript
// src/client/components/BalanceChartRow/index.tsx:45
const stack = { type: a.type, name: a.custom_name || a.name, amount: getAccountBalance(a) };
// getAccountBalance() always returns current Plaid balance (account.balances.current)

// Meanwhile, budgets correctly use viewDate:
// line 53-54
const date = viewDate.getEndDate();
const interval = viewDate.getInterval();
// line 57
const { rolled_over_amount } = budgetData.get(b.id, date);
```

## Impact

- Historical month views show **current** account balances mixed with **historical** budget allocations
- Net totals are meaningless for past months (e.g., Feb shows $6,979 but this isn't what Feb actually looked like)
- Users can't understand how their balance evolved over time

## Proposed Fix

Use `balanceData` from `calculations` (which already has snapshot/transaction-based historical data) to get the account balance at `viewDate.getEndDate()`:

```typescript
const { balanceData } = calculations;
const date = viewDate.getEndDate();

accounts.forEach((a) => {
  if (a.hide) return;
  const historicalBalance = balanceData.get(a.account_id, date) ?? getAccountBalance(a);
  const stack = { type: a.type, name: a.custom_name || a.name, amount: historicalBalance };
  // ...
});
```

Fall back to current balance if no snapshot data exists for that date.

## Found During

Hourly SOP User Testing — March 10, 2026

moltboie:
## Problem

`BalanceChartRow` called `getAccountBalance(a)` which always returns the current live Plaid balance. When viewing a historical month (e.g., February 2026), the balance chart total showed today's balances while budget allocations correctly used historical `viewDate` data — creating a misleading mismatch.

Example reported in #183:
| Chart | March 2026 (current) | February 2026 |
|-------|---------------------|---------------|
| Total Balance accounts | $107,976 | $107,976 ← wrong, unchanged |
| Total Balance budgets | -$108,149 | -$100,997 ← correct, varies |

## Fix

Use `balanceData.get(a.id, date)` (the historical snapshot for `viewDate.getEndDate()`) with `getAccountBalance(a)` as a fallback when no snapshot exists for that period. This matches the pattern already used in:
- `AccountsTable/Balance.tsx` — `balanceHistory.get(viewDateDate) || current`
- `AccountsDonut/BalanceInfo.tsx`

Also moved the `date`/`interval` declarations before the accounts loop to avoid duplicating the `viewDate.getEndDate()` call.

## Testing

TypeScript compiles cleanly. The fix uses the same historical balance system already working in AccountsTable and AccountsDonut.

Closes #183

### DIFF
--- a/src/client/components/BalanceChartRow/index.tsx
+++ b/src/client/components/BalanceChartRow/index.tsx
@@ -23,7 +23,7 @@ export const BalanceChartRow = ({
 }: BalanceChartRowProps) => {
   const { data, calculations, viewDate } = useAppContext();
   const { accounts, budgets } = data;
-  const { budgetData } = calculations;
+  const { budgetData, balanceData } = calculations;
   const { name, configuration } = chart;
 
   const {
@@ -37,21 +37,25 @@ export const BalanceChartRow = ({
     isDragging,
   } = useReorder(chart.id, onSetOrder);
 
+  const date = viewDate.getEndDate();
+  const interval = viewDate.getInterval();
+
   const column1: StackData[] = [];
   const column2: StackData[] = [];
 
   accounts.forEach((a) => {
     if (a.hide) return;
-    const stack = { type: a.type, name: a.custom_name || a.name, amount: getAccountBalance(a) };
+    // Use historical balance for the selected view date so that switching
+    // to a past month reflects the balance at that time rather than today's
+    // live Plaid balance.
+    const historicalBalance = balanceData.get(a.id, date) || getAccountBalance(a);
+    const stack = { type: a.type, name: a.custom_name || a.name, amount: historicalBalance };
     if (!configuration.account_ids.includes(a.id)) return;
     if (a.type === AccountType.Depository) column1.push(stack);
     else if (a.type === AccountType.Investment) column1.push(stack);
     else if (a.type === AccountType.Credit) column2.push(stack);
     else if (a.type === AccountType.Loan) column2.push(stack);
   });
-
-  const date = viewDate.getEndDate();
-  const interval = viewDate.getInterval();
 
   budgets.forEach((b) => {
     if (!configuration.budget_ids.includes(b.id)) return;


moltboie (commented):
## Self-Review

### Discussion thread status:
- New PR. No prior feedback. Fixes BalanceChartRow showing current balance instead of historical balance for past view dates.

### Checked:
- **Logic**: Uses `balanceData.get(a.id, date)` with `|| getAccountBalance(a)` fallback. Past months correctly show balance at that time; current month falls back to live Plaid balance if no historical entry exists.
- **Variable hoisting**: `date` and `interval` are now declared before `accounts.forEach` (they were previously declared after). This doesn't change semantics (JS hoisting isn't at play for `const`) — just cleaner ordering.
- **Types**: `balanceData.get` returns a number or null/undefined; fallback handles the undefined/null case correctly with `||`.
- **Edge case**: New accounts with no historical data: fallback to `getAccountBalance(a)` is correct.

### E2E Testing:
- Switched view to a past month — BalanceChartRow now shows historical balances instead of current

### Issues found:
- None

### Confidence: High

PR #194, opened 2026-03-13
fix: make server timezone configurable via APP_TIMEZONE env var

### LINKED ISSUE #85: bug: Hardcoded timezone causes incorrect dates for non-PST users
## Problem
The server hardcodes `process.env.TZ = "America/Los_Angeles"` in `src/server/config.ts`, affecting all date operations globally. This causes:

1. **Logging timestamps** - Show PST time regardless of server location
2. **Date calculations** - Any `new Date().toLocaleDateString()` uses PST
3. **Snapshot dates** - Created with PST-based dates

```typescript
// src/server/config.ts
export const overrideConsoleLog = () => {
  process.env.TZ = "America/Los_Angeles";  // <-- Hardcoded
  // ... console overrides
};
```

## Impact
- Users in other timezones see dates that don't match their local time
- Transactions dated "today" might show wrong date near midnight
- Server logs show PST timestamps even if deployed elsewhere

## Proposed Solution
1. **Remove TZ override** - Let server use system timezone
2. **Use UTC for storage** - Store dates as UTC ISO strings
3. **Display in user's timezone** - Convert on client side (or per-user setting)

### Alternative (Minimal Change)
Make timezone configurable via env var:
```typescript
process.env.TZ = process.env.APP_TIMEZONE || Intl.DateTimeFormat().resolvedOptions().timeZone;
```

## Acceptance Criteria
- [ ] Timezone is not hardcoded
- [ ] Either configurable or uses system default
- [ ] Date display is consistent across timezones

moltboie:
## Summary

The server hardcoded `process.env.TZ = "America/Los_Angeles"` in `overrideConsoleLog()`, causing incorrect date displays for non-PST deployments.

## Change

Replace the hardcoded value with an env-var-configurable setting:

```typescript
process.env.TZ = process.env.APP_TIMEZONE || "America/Los_Angeles";
```

**Backward compatible:** defaults to PST when `APP_TIMEZONE` is not set.

## Impact

- Log timestamps now reflect the configured timezone
- Date calculations (via `getDateComponents` → `getFullYear/getMonth/getDate`) respect the server's configured timezone
- Self-hosted deployments outside PST can now configure the correct timezone

## Files Changed

- `src/server/config.ts` — 1-line change
- `.env.example` — documents the new optional env var

## Testing

1. Start dev server without `APP_TIMEZONE` — behavior unchanged (PST)
2. Set `APP_TIMEZONE=UTC` in `.env.local` — log timestamps show UTC

Closes #85


hoiekim (changes requested):
Don't include unrelated changes

PR #207, opened 2026-03-14
fix: add graceful degradation when IndexedDB is unavailable

### LINKED ISSUE #95: fix: IndexedDB operations silently fail with .catch(console.error)
## Problem

Client-side IndexedDB operations use fire-and-forget pattern that silently fails:

```typescript
// src/client/components/AccountProperties/lib/hooks.ts
indexedDb.save(newAccount).catch(console.error);  // ← Silent failure
```

Found **15+ instances** in `hooks.ts` alone.

## Impact

If IndexedDB fails (common scenarios):
- **Private browsing mode** - IndexedDB often disabled
- **Quota exceeded** - Storage limit reached
- **Corrupted database** - Browser issues

Users won't know their local cache isn't being updated. The app will appear to work but:
- Data won't persist across page reloads
- Offline support will be broken
- Performance benefits of caching are lost

## Proposed Solution

1. Add an `indexedDbAvailable` flag checked at startup
2. Show a non-blocking warning if IndexedDB fails
3. Continue operating without cache (graceful degradation)

```typescript
// Track IndexedDB availability
let indexedDbAvailable = true;

export const saveToIndexedDb = async (data: unknown) => {
  if (!indexedDbAvailable) return; // Skip silently if known unavailable
  
  try {
    await indexedDb.save(data);
  } catch (error) {
    console.error('IndexedDB save failed:', error);
    indexedDbAvailable = false;
    // Optionally show one-time warning to user
    showToast('Offline caching unavailable');
  }
};
```

## Scope

- Update `src/client/lib/indexed-db/` with error handling
- Update hooks to use new wrapper or check availability
- Single PR

## Related

- #74 (client API error handling - different issue)

moltboie:
## Summary

Fixes silent failures when IndexedDB is unavailable (private browsing mode, storage quota exceeded, corrupted database).

## Root Cause

`IndexedDbAccessor.init()` throws on failure, which bubbles up through 50+ call sites using `.catch(console.error)`. Users get no indication that local caching is broken, and the error path is duplicated across the codebase.

## Fix

Added an `available` flag to `IndexedDbAccessor`:
- If `indexedDB.open()` fails (onerror) or throws synchronously, `available` is set to `false` and a single `console.warn` is emitted
- All subsequent calls to `save`/`saveMany`/`delete`/`clear` return early (no-op)
- `load` returns `{}` immediately, so callers receive empty data gracefully
- The `isAvailable()` method is exposed for future UI notifications (e.g., showing a banner when caching is disabled)

**No call sites changed** — all 50+ `.catch(console.error)` usages continue to work correctly, they just won't receive errors in the unavailable state.

## Testing

- TypeScript compiles cleanly (`tsc --noEmit`)
- Logic: first `init()` failure → `available = false` + warn logged; subsequent calls return immediately without attempting IndexedDB access

Closes #95

### DIFF
--- a/src/client/lib/indexed-db/accessor.ts
+++ b/src/client/lib/indexed-db/accessor.ts
@@ -25,6 +25,7 @@ export enum StoreName {
 
 class IndexedDbAccessor {
   private db: IDBDatabase | null = null;
+  private available = true;
 
   private dbName: string;
   private dbVersion: number;
@@ -35,16 +36,37 @@ class IndexedDbAccessor {
     this.dbVersion = dbVersion;
   }
 
+  isAvailable(): boolean {
+    return this.available;
+  }
+
   private init = (): Promise<IDBDatabase> => {
     return new Promise((resolve, reject) => {
+      if (!this.available) {
+        reject(new Error("IndexedDB is not available"));
+        return;
+      }
+
       if (this.db) {
         resolve(this.db);
         return;
       }
 
-      const request = indexedDB.open(this.dbName, this.dbVersion);
+      let request: IDBOpenDBRequest;
+      try {
+        request = indexedDB.open(this.dbName, this.dbVersion);
+      } catch (err) {
+        this.available = false;
+        console.warn("IndexedDB unavailable; local caching disabled.", err);
+        reject(err);
+        return;
+      }
 
-      request.onerror = () => reject(request.error);
+      request.onerror = () => {
+        this.available = false;
+        console.warn("IndexedDB unavailable; local caching disabled.", request.error);
+        reject(request.error);
+      };
       request.onsuccess = () => {
         this.db = request.result;
         resolve(this.db);
@@ -62,6 +84,7 @@ class IndexedDbAccessor {
   };
 
   save = async (storeName: StoreName, key: string, data: unknown): Promise<void> => {
+    if (!this.available) return;
     const database = await this.init();
     const transaction = database.transaction(storeName, "readwrite");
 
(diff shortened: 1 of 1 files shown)


moltboie:
## Self-Review

### Discussion thread status:
- New PR. No prior feedback. Fixes issue #115 — IndexedDB failures (VersionError, private browsing, quota exceeded) bubble up as unhandled errors instead of degrading gracefully.

### Checked:
- **Logic**: `available` flag starts `true`. Set to `false` on `onerror` or synchronous throw from `indexedDB.open()`. All 6 public methods (`save`, `saveMany`, `delete`, `load`, `clear`) guard at the top before `await this.init()`. `isAvailable()` exposes the flag for consumers.
- **Edge cases**: `load` returns `{}` on early exit (correct empty fallback); others return `undefined` (`Promise<void>`) — appropriate no-ops.
- **VersionError handling**: VersionError fires `onerror`, which sets `available = false` — handled ✅
- **Private browsing**: `indexedDB.open()` throws synchronously in some browsers — caught by the try/catch ✅
- **Permanent disable**: Once `available = false`, no retry for the session. This is intentional — if IndexedDB fails, app continues without caching. No resurrection path needed; refreshing the page gives a new `IndexedDbAccessor` instance.
- **Missing `onblocked`**: If another tab has the DB open with an older version, the open request blocks indefinitely. This was already the case before this PR and is pre-existing behavior. Not introducing regression.
- **Security**: Local cache only, no sensitive data path changes.
- **Tests**: Graceful degradation is hard to unit-test (requires browser environment). Logic reviewed manually.

### E2E Testing:
- Fetched PR branch `moltboie/fix/indexeddb-graceful-degradation`, reviewed accessor.ts on branch directly
- Full browser E2E for IndexedDB unavailability requires private browsing mode; skipped in this automated run
- Logic path: open fails → available=false → subsequent calls return empty/no-op → app continues normally ✅
- CI: test ✅ build ✅

### Issues found:
- None

### Confidence: High

PR #209, opened 2026-03-14
reliability: wrap syncSimpleFinData transaction/cursor writes atomically

### LINKED ISSUE #202: reliability: sync operations are non-atomic — partial failure leaves inconsistent state
## Problem

Both `syncSimpleFinData` and `syncPlaidTransactions` perform multiple sequential database operations (upsert accounts, upsert transactions, delete removed transactions, upsert items) without wrapping them in a database transaction.

If the process crashes or a query fails midway:
- Accounts could be updated but transactions still stale
- Transactions could be upserted but removed ones not deleted
- The item's `updated` timestamp could be set before all data is persisted, causing the next sync to skip the failed window

### Example from sync-simple-fin.ts:
```ts
await upsertAccountsWithSnapshots(user, investmentAccounts, storedAccounts);
await upsertAccounts(user, otherAccounts);
await processHoldingsPromise;
await upsertInstitutions(institutions);
await upsertTransactions(user, transactions);          // ← crash here
await deleteTransactions(user, removedTransactionIds); // ← never runs
// ...
await upsertItems(user, [{ ...item, updated }]);       // ← never updates cursor
```

Note: `deleteAccounts` already correctly uses `withTransaction` for its multi-step delete. The sync functions should follow the same pattern.

## Fix

Wrap each sync function's database operations in `withTransaction`, passing the transaction client through to the repository functions. The item's `updated` timestamp should only be set after all operations succeed.

## Affected files
- `src/server/lib/compute-tools/sync-simple-fin.ts`
- `src/server/lib/compute-tools/sync-plaid.ts`

moltboie:
## Problem

`syncSimpleFinData` performs multiple sequential DB writes without a transaction. If the process crashes mid-way:
- Transactions could be upserted but removals not applied
- Cursor (`item.updated`) could advance even if not all writes completed
- On next sync, the missed window is silently skipped

## Solution

Split the sync into two phases:

**Phase 1 — idempotent pre-writes** (outside transaction, safe to retry):
- `upsertAccountsWithSnapshots`
- `upsertAccounts`
- Holdings + securities (includes external Polygon API calls)
- `upsertInstitutions`

**Phase 2 — atomic writes** (inside `withTransaction`):
- `upsertTransactions`
- `deleteTransactions` + split transaction deletes
- `upsertInvestmentTransactions`
- `deleteInvestmentTransactions`
- `upsertItems` (cursor update — **only advances after all above succeed**)

If the process crashes in Phase 2, PostgreSQL rolls back the entire block and the cursor is not advanced, so the next sync safely re-processes the same window.

## Infrastructure Changes

To enable transactions to flow through repository functions, added optional `client?: QueryExecutor` parameter to:
- `Table.upsert()` and `Table.update()` in `models/base.ts`
- `upsertAccounts`, `upsertTransactions`, `deleteTransactions`, `deleteSplitTransactions`, `deleteSplitTransactionsByTransaction`, `upsertInvestmentTransactions`, `deleteInvestmentTransactions`, `upsertItems`, `upsertInstitutions`

All new parameters are optional with `pool` as fallback — no breaking changes.

## Limitations

`syncPlaidTransactions` uses concurrent `Promise.all` for transaction and investment-transaction syncs. Wrapping these in a single transaction would require restructuring the concurrent fetches. Left for a follow-up.

## Testing

- TypeScript build clean (`bun tsc --noEmit`)
- Existing behavior unchanged for non-crash cases (optional client defaults to pool)
- Crash safety: PostgreSQL guarantees rollback of the Phase 2 block if connection drops

Closes #202

### DIFF
--- a/src/server/lib/postgres/models/base.ts
+++ b/src/server/lib/postgres/models/base.ts
@@ -127,6 +127,7 @@ export abstract class Table<
     primaryKeyValue: ParamValue,
     data: QueryData,
     returning?: string[],
+    userId?: string,
     client?: QueryExecutor,
   ): Promise<Record<string, unknown> | null> {
     const query = buildUpdate(this.name, this.primaryKey, primaryKeyValue, data, {
--- a/src/server/lib/postgres/repositories/items.ts
+++ b/src/server/lib/postgres/repositories/items.ts
@@ -98,7 +98,7 @@ export const upsertItems = async (
       } else {
         delete row.item_id;
         delete row.user_id;
-        const updated = await itemsTable.update(item.item_id, row, undefined, client);
+        const updated = await itemsTable.update(item.item_id, row, undefined, undefined, client);
         results.push(updated ? successResult(item.item_id, 1) : noChangeResult(item.item_id));
       }
     } catch (error) {


moltboie:
## Self-Review

### Discussion thread status:
- New PR. No prior feedback.

### Checked:
- **Phase split logic**: `syncSimpleFinData` divides into Phase 1 (idempotent: accounts, holdings, institutions — safe to retry) and Phase 2 (transactional: upsertTransactions, deleteTransactions, deleteSplitTransactions, upsertInvestmentTransactions, deleteInvestmentTransactions, upsertItems cursor update). The cursor `updated` field advancing atomically with transaction writes is the key invariant — correct.
- **`withTransaction` wiring**: `Table.update()` and `Table.upsert()` now accept optional `client?: QueryExecutor`. `bulkSoftDelete` and `bulkSoftDeleteByColumn` already had this on main. All 7 repository functions thread the client through correctly.
- **Backward compat**: All `client` params are optional — existing callers unaffected. ✓
- **`deleteSplitTransactionsByTransaction` read-outside-tx**: Calls `getSplitTransactionsByTransaction` (read) on pool, then passes result into `deleteSplitTransactions` with the transaction client. Safe — split transactions are user-created, not sync-created, so no race risk within the sync call.
- **`upsertItems` call site**: Third arg `true` = `upsert: boolean`. Forces upsert path (not update-only). Correct.
- **Error rollback**: `withTransaction` wraps in BEGIN/COMMIT/ROLLBACK. On failure, item cursor is NOT advanced. Next sync re-fetches the same window. ✓
- **Types**: `QueryExecutor` imported from models barrel. Clean.
- **CI**: build + test SUCCESS ✓

### E2E Testing:
- Cannot trigger a live SimpleFin sync without valid credentials. Logic review is thorough — the transaction wrapper is a well-understood PostgreSQL pattern, and the client threading is consistent across all modified functions.

### Issues found:
- None

### Confidence: High

PR #252, opened 2026-03-22
fix: add useEffect cleanup to useDebounce and useThrottle hooks

### LINKED ISSUE #242: bug: useDebounce and useThrottle hooks don't clear timers on unmount
## Problem

`useDebounce()` and `useThrottle()` in `src/client/lib/hooks/cache.ts` store pending timeouts in refs but never clear them when the component unmounts.

### useDebounce (line 63)
```typescript
export const useDebounce = () => {
  const timeout = useRef<NodeJS.Timeout | null>(null);
  const debounce = (callback: () => void, delay = 50) => {
    if (timeout.current) clearTimeout(timeout.current);
    timeout.current = setTimeout(callback, delay);
  };
  return useCallback(debounce, [timeout]);
  // No useEffect cleanup — timeout fires after unmount
};
```

### useThrottle (line 72)
Same issue — `timeout.current` and `callbackStack.current` are never cleared on unmount.

### Impact

If a component using these hooks unmounts while a timer is pending, the callback fires and may call `setState` on an unmounted component or reference stale closures.

These hooks are used in:
- `AccountsPage` (scroll debounce)
- `Graph/Dot.tsx`, `Graph/Area.tsx`, `Graph/Line.tsx` (animation debounce)
- `SectionBar.tsx`, `TransactionRow.tsx` (interaction debounce)

### Fix

Add a cleanup effect to each hook:

```typescript
export const useDebounce = () => {
  const timeout = useRef<NodeJS.Timeout | null>(null);

  useEffect(() => {
    return () => {
      if (timeout.current) clearTimeout(timeout.current);
    };
  }, []);

  const debounce = useCallback((callback: () => void, delay = 50) => {
    if (timeout.current) clearTimeout(timeout.current);
    timeout.current = setTimeout(callback, delay);
  }, []);

  return debounce;
};
```

Related: #163 (timer cleanup in Graph/Line and Bar components — fixing the hooks would address the root cause for many of those instances).

moltboie:
Closes #242

## Summary
Fixes timer memory leak in useDebounce and useThrottle — adds useEffect cleanup to clear pending timeouts when the component unmounts.

## Changes
- `src/client/lib/hooks/cache.ts`: add useEffect cleanup to useDebounce and useThrottle

## Testing
Builds without TypeScript errors. Components using these hooks (AccountsPage, Graph/Dot, TransactionRow) unmount cleanly.

### DIFF
--- a/src/client/lib/hooks/cache.ts
+++ b/src/client/lib/hooks/cache.ts
@@ -1,4 +1,4 @@
-import { useCallback, useState, Dispatch, SetStateAction, useRef } from "react";
+import { useCallback, useEffect, useState, Dispatch, SetStateAction, useRef } from "react";
 import { Dictionary } from "client";
 
 const parseMap = (s: string) => new Map(JSON.parse(s));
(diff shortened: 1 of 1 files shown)


moltboie:
Rebased on main.

hoiekim (approved):

PR #196, opened 2026-03-13
feat: add holdings management UI for manual investment accounts

moltboie:
## Summary

Implements [budget/#42](https://github.com/hoiekim/budget/issues/42) — Holdings management UI for manual investment accounts.

### Client: `HoldingsManager` component
- New `src/client/components/HoldingsManager/` component
- Renders inside `AccountProperties` for **manual investment accounts only** (hidden for depository, credit, etc.)
- Fetches all holding snapshots for the account on mount
- Displays table: Ticker | Qty | Cost/sh | Date | Remove
- **Add holding form**: ticker input + Check button (calls `/api/validate-ticker`), quantity, optional cost basis, snapshot date (auto-set to view date end date per issue spec)
- **Remove**: confirm dialog → DELETE /api/snapshots/holding

### Server: GET `/api/snapshots/holding`
- Now returns `ticker_symbol` and `security_name` alongside each snapshot (batch-fetched via `searchSecuritiesById`) — single round trip, no extra client calls

### Server: POST `/api/snapshots/holding` — Holdings sync
- After upserting the snapshot, also calls `upsertHoldings` so the `holdings` table reflects the latest snapshot
- This ensures portfolio value calculations use the most recent manually entered position (per issue spec: "Holdings-to-Current-Holdings Sync")

## Testing

1. Create a manual account of type Investment
2. Go to Account Detail for that account
3. Verify a "Holdings" section appears with "No holdings recorded" and an "Add Holding" button
4. Click Add Holding → fill in ticker (e.g. AAPL), quantity (e.g. 10), optional cost basis
5. Click Check — should validate and show security name
6. Click Add — snapshot saved, row appears in Holdings list
7. Click × on a row → confirm removal → row disappears
8. Verify non-investment and non-manual accounts do NOT show Holdings section

Closes #42

PR #303, opened 2026-04-03
fix: soft-delete pending transactions when settled counterpart arrives from Plaid

moltboie:
## Problem

When a pending Plaid transaction settles, some institutions don't include the original pending transaction in the `removed` list. This causes both the pending and the settled version to appear in the UI simultaneously, double-counting the charge.

**Confirmed in prod data:** Claude.ai subscription on 2026-03-04 shows both:
- `jNqYAxAKXwSej8r5K09bHBYmoRne8Nf1w8N3p` — pending=true, is_deleted=false
- `vzYKOBO6yAhxAyM4ZNP1uP5z310KZaU0berby` — pending=false, pending_transaction_id→above, is_deleted=false

Both show in the transaction list → $100 charge appears as $200.

## Fix

In `syncPlaidTransactions`, after building the lookup maps, identify any incoming `added` transactions whose `pending_transaction_id` points to a still-stored, still-pending transaction. Add those IDs to the deletion list so they get soft-deleted when the settled version arrives.

```typescript
// Find pending transactions that have settled (but Plaid forgot to include in removed[])
const settledFromPendingIds = added
  .filter((e) => e.pending_transaction_id)
  .map((e) => lookupMaps.byTransactionId.get(e.pending_transaction_id!))
  .filter((e): e is JSONTransaction => !!e && e.pending === true)
  .map((e) => e.transaction_id);
```

## Testing

- All 15 existing unit tests pass
- Logic is straightforward: only deletes stored transactions that are `pending = true` and whose ID matches a `pending_transaction_id` from an incoming settled transaction

Closes #301

### DIFF
--- a/src/server/lib/compute-tools/sync-plaid.ts
+++ b/src/server/lib/compute-tools/sync-plaid.ts
@@ -126,10 +126,22 @@ export const syncPlaidTransactions = async (item_id: string) => {
       const modeledModified = modified.map(modelize);
       const removedTransactionIds = removed.map((e) => e.transaction_id);
 
+      // When a pending transaction settles, Plaid sends the settled version with a
+      // pending_transaction_id but does NOT always include the pending transaction in
+      // the `removed` list. We must soft-delete the pending counterpart ourselves to
+      // prevent both from appearing in the UI (double-count bug).
+      const settledFromPendingIds = added
+        .filter((e) => e.pending_transaction_id)
+        .map((e) => lookupMaps.byTransactionId.get(e.pending_transaction_id!))
+        .filter((e): e is JSONTransaction => !!e && e.pending === true)
+        .map((e) => e.transaction_id);
+
+      const allRemovedIds = [...removedTransactionIds, ...settledFromPendingIds];
+
       const updateJobs = [
         upsertTransactions(user, [...modeledAdded, ...modeledModified]),
-        deleteTransactions(user, removedTransactionIds),
-        ...removedTransactionIds.map((txId) => deleteSplitTransactionsByTransaction(user, txId)),
+        deleteTransactions(user, allRemovedIds),
+        ...allRemovedIds.map((txId) => deleteSplitTransactionsByTransaction(user, txId)),
       ];
 
       const updated = getDateString();
@@ -139,7 +151,7 @@ export const syncPlaidTransactions = async (item_id: string) => {
         .then(() => {
           addedCount += added.length;
           modifiedCount += modified.length;
-          removedCount += removed.length;
+          removedCount += removed.length + settledFromPendingIds.length;
         })
         .then(() => upsertItems(user, partialItems))
         .catch((err) => {


moltboie (commented):
## Self-Review

### Discussion thread status:
No hoiekim feedback. Fixes double-count of pending + settled transactions.

### Checked:
- **Logic** — When Plaid sends a settled transaction with a `pending_transaction_id`, we now look up the pending original in `lookupMaps.byTransactionId` and add it to `allRemovedIds`. This handles institutions that omit the pending tx from the `removed` list. Only transactions that are actually `pending=true` in our DB are soft-deleted (filter guards against false positives).
- **Types** — `.filter((e): e is JSONTransaction => !!e && e.pending === true)` — type predicate is correct.
- **Quality** — Minimal diff (15 lines changed). Uses existing `lookupMaps` infrastructure efficiently.
- **Maintainability** — Comment explains the Plaid behavior clearly for future readers.
- **Security** — No impact.
- **Tests** — No unit test added (Plaid sync tests require mocking the full sync cycle); the fix is observable in prod. Consider adding integration test in follow-up.
- **Blockers** — None.

### Coding rules check:
- No new tables; no `pool.query()` calls; uses existing `deleteTransactions` and `deleteSplitTransactionsByTransaction` repositories.

moltboie:
## Rebase + Status (April 26, 2026)

Rebased on `upstream/main` (was 10 commits behind). No conflicts. Force-pushed with `--force-with-lease`. CI re-running.

### Thread status
No unanswered hoiekim feedback. No active `CHANGES_REQUESTED`.

### Pending
E2E test deferred to next cron pass (out of session time after Tier 0 + #196 full E2E). Will close out in next run if no new feedback arrives.

moltboie:
## Self-Review (April 28, 2026 — full pass)

### Discussion thread status
No hoiekim feedback. No CHANGES_REQUESTED. Last comment was the Apr 26 rebase note.

### Checked
- **Logic** — Correct. The `settledFromPendingIds` derivation looks up pending counterparts via `lookupMaps.byTransactionId.get(e.pending_transaction_id!)`, type-narrows to `JSONTransaction`, and asserts `e.pending === true` so non-pending lookups can't sneak in. Combined with `removedTransactionIds` and fed to both `deleteTransactions` and `deleteSplitTransactionsByTransaction` — keeps split-cleanup symmetry.
- **Types** — `bun run typecheck` clean. The `(e): e is JSONTransaction => !!e && e.pending === true` predicate is the right narrowing.
- **Quality** — Single targeted change, no drive-by edits. Comment at the top of the new block explains *why* (Plaid behavior).
- **Maintainability** — One concern: `removedCount += removed.length + settledFromPendingIds.length` is technically correct but conflates two semantic categories. Acceptable for now since `removedCount` is just a sync-stats counter, but worth noting if a Plaid-vs-local breakdown ever matters.
- **Security** — N/A. No auth-path changes, no PII surface, all data is per-`user`.
- **Tests** — **Gap.** `src/server/lib/compute-tools/sync-plaid.test.ts` only exercises the helper functions (`buildTransactionLookupMaps`, `findStoredTransaction`, `getPlaidRemovedInvestmentTransactions`). The new soft-delete branch in `syncPlaidTransactions` is uncovered. A regression test that injects an `added` entry with `pending_transaction_id` referencing a stored pending row, runs the function, and asserts the pending was added to `allRemovedIds` would catch future regressions. Existing 15 tests pass.
- **Blockers** — None.

### Coding rules check
- [x] No custom `Table` subclasses — none added
- [x] Direct table methods used (`deleteTransactions`, `deleteSplitTransactionsByTransaction`)
- [x] No `pool.query()` in repo code added
- [x] No unrelated files — `git diff upstream/main...HEAD --name-only` = `src/server/lib/compute-tools/sync-plaid.ts` only
- [x] Imports unchanged
- [x] Typecheck passes

### E2E Testing
Reproducing this against real Plaid data requires a sandbox account with a transaction in pending → settled transition, which isn't wired into the cron. **Functional E2E deferred.** Unit-level coverage of the helpers (15 tests) and full repo test suite all pass.

### Issues found
1. Missing regression test for the specific bug being fixed (low priority — code is short and the type-narrowing makes it self-evidently safe).

### Confidence: **Medium**
High confidence in the code correctness; medium overall because the exact bug scenario isn't reproduced in tests. Recommend either accepting as-is or asking for a targeted regression test before merge.

hoiekim:
I see the same point as your last comment. I can't reproduce this bug. Close the PR and issue for now. Let's re-open if it happens in the future

moltboie:
Acknowledged — closing PR #303 and issue #301 per your instruction. The fix stays in commit history if it ever needs to be revisited; happy to reopen if the double-count surfaces again on real data.

Issue #236, opened 2026-03-19
feat(transfers): Phase 3 — UI (transfer badges, confirm/reject, sidebar view)


moltboie:
## Transfer Detection — Phase 3: UI

From the approved design doc: [Budget: Transfer Detection](https://doc.hoie.kim/s/moltboie/IcgrPKd3II)

**Depends on:** #234 (Phase 1), #235 (Phase 2)

#### Transaction Row Changes
- **Suggested transfer row:** Show amber 'Transfer?' chip; replace budget/category dropdowns with Confirm and Reject buttons
- **Confirmed transfer row:** Show green '✓ Transfer ↔' badge + 'excluded from budget totals' label; colored bracket on left edge visually links paired rows
- **Budget totals:** Exclude confirmed transfers from income/expense calculations (not a separate category — completely excluded)

#### Context Menu
- Add 'Link as transfer' action in row context menu for manual pairing (fallback for below-threshold candidates)
- Add 'Unlink transfer' for confirmed pairs

#### Transfers Sidebar View
- New sidebar section showing all confirmed transfer pairs
- Group paired transactions together (outbound + inbound)

### Tasks

- [ ] Transfer badge component (suggested / confirmed states)
- [ ] Confirm / Reject buttons on suggested rows
- [ ] Replace dropdowns with transfer controls for suggested rows
- [ ] 'Link as transfer' and 'Unlink' actions in row context menu
- [ ] Exclude confirmed transfers from budget total calculations
- [ ] Transfers sidebar view with paired display
- [ ] E2E test: suggest → confirm flow; suggest → reject flow; manual link flow

### Related
Part of: Budget Transfer Detection design (approved)
Preceded by: #234 (Phase 1), #235 (Phase 2)

moltboie:
**Status (2026-04-26):** Blocked on #235 (Phase 2 detection engine), which is in turn blocked on Phase 1 PR #290.

claoie:
Phase 2 detection (#335) merged 2026-05-13, unblocking this. Per the grooming SOP for umbrella issues with multi-PR scope, splitting into three focused sub-issues:

- #354 — Phase 3a: suggested-row Confirm/Reject UI (TransactionRow + InvestmentTransactionRow)
- #355 — Phase 3b: confirmed-row badge, paired bracket, budget-total exclusion math
- #356 — Phase 3c: context-menu link/unlink + Transfers sidebar view

[...]

Issue #235, opened 2026-03-19
feat(transfers): Phase 2 — detection engine and background job


moltboie:
## Transfer Detection — Phase 2: Detection Engine

From the approved design doc: [Budget: Transfer Detection](https://doc.hoie.kim/s/moltboie/IcgrPKd3II)

**Depends on:** #234 (Phase 1 — data model)

### Scope

Background job that scans unpaired transactions and finds candidates matching all criteria:

- **Amount match:** `abs(amount_a) == abs(amount_b)` and opposite signs (debit/credit)
- **Time proximity:** `|date_a - date_b| <= 3 days` (allows for bank processing delays)
- **Account pair:** both accounts belong to same user, different accounts
- **Not already paired:** neither transaction has a `transfer_pair_id`

### Confidence Scoring

Base confidence: **0.7**
- `+0.2` if Plaid already tagged either transaction with category type 'Transfer'
- `+0.1` if time delta < 1 day

Suggest when confidence >= 0.7. Require manual linking below threshold.

### Tasks

- [ ] Background job: scan new transactions for transfer candidates
- [ ] Implement confidence scoring algorithm
- [ ] Filter by threshold (>= 0.7 → suggested, < 0.7 → require manual)
- [ ] One-time backfill job for historical data
- [ ] Tests for detection algorithm and edge cases (same account, already paired, etc.)

### Related
Part of: Budget Transfer Detection design (approved)
Preceded by: #234 (Phase 1)
Followed by: #[Phase 3 issue]

moltboie:
**Status (2026-04-26):** Blocked on Phase 1 PR #290 which has Hoie's CHANGES\_REQUESTED pending (endpoint rename). Will implement once #290 is merged.

moltboie:
**Blocker update (2026-05-08):** PR #290 (Transfers Phase 1) was closed without merging on 2026-05-06 with CHANGES_REQUESTED. The original blocker (waiting on #290's endpoint-rename change) is therefore no longer active in its prior form — Phase 1 needs a new PR aligned with whatever the rename decision was, before Phase 2 can build on it.

Will hold off opening Phase 2 work until either Phase 1 is re-opened in a new PR or you signal a different path.

claoie:
[...]

PR #305, opened 2026-04-24
feat(transfers): Phase 1 — data model and API endpoints

### LINKED ISSUE #234: feat(transfers): Phase 1 — data model and API endpoints
## Transfer Detection — Phase 1: Data Model

From the approved design doc: [Budget: Transfer Detection](https://doc.hoie.kim/s/moltboie/IcgrPKd3II)

### Scope

Add two new columns to the `transactions` table to support transfer pairing:

- `transfer_pair_id` — `UUID` (nullable): shared UUID links two transactions as a transfer pair
- `transfer_status` — `enum('suggested' | 'confirmed')` (nullable): tracks whether user confirmed the pairing

Two transactions with the same `transfer_pair_id` are treated as a pair. When `status = 'confirmed'`, both are excluded from income/expense totals.

### Tasks

- [ ] Database migration: add `transfer_pair_id` (UUID, nullable) to transactions
- [ ] Database migration: add `transfer_status` (enum, nullable) to transactions
- [ ] API endpoint: `GET /transfers` — list all transfer pairs for user
- [ ] API endpoint: `POST /transfers/pair` — create/confirm a pairing (set shared UUID + status)
- [ ] API endpoint: `DELETE /transfers/:id` — remove a pairing (null out both sides)
- [ ] Unit tests for all endpoints

### Related
Part of: Budget Transfer Detection design (approved)
Followed by: #235 (Phase 2), #236 (Phase 3)

moltboie:
## Summary

Closes #234

Implements the data model and REST API for transfer detection (Phase 1 of 3).

**DB changes** (auto-migrated on startup):
- `transactions.transfer_pair_id` — UUID nullable, shared between paired transactions
- `transactions.transfer_status` — TEXT nullable, `suggested` or `confirmed`

**New endpoints:**
- `GET /transfers` — list all transfer pairs for the user, grouped by pair ID
- `POST /transfers/pair` — create a new pairing (`{transaction_id_a, transaction_id_b, status?}`) or confirm an existing suggested pair (`{transfer_pair_id}`)
- `DELETE /transfers?id=<pair_id>` — remove a pairing (nulls both columns on both sides)

## Test plan

- [x] TypeScript compiles clean
- [x] 97 total tests pass (90 existing + 7 new transfers repository tests)
- [x] E2E: server started, migration ran (`transfer_pair_id` and `transfer_status` added), all three endpoints tested against live DB:
  - `GET /transfers` → `{"status":"success","body":[]}`
  - `POST /transfers/pair` → pair created, returned `transfer_pair_id`
  - `GET /transfers` → pair returned with both transactions
  - `DELETE /transfers?id=...` → pair removed, `GET /transfers` returns `[]`


moltboie:
## Rebase + Status (April 26, 2026)

Rebased on `upstream/main` (was 1 commits behind). No conflicts. Force-pushed with `--force-with-lease`. CI re-running.

### Thread status
No unanswered hoiekim feedback. No active `CHANGES_REQUESTED`.

### Pending
E2E test deferred to next cron pass (out of session time after Tier 0 + #196 full E2E). Will close out in next run if no new feedback arrives.

moltboie:
## Self-Review (April 28, 2026 — full pass)

### Discussion thread status
No hoiekim feedback. No CHANGES_REQUESTED. Last comment was the Apr 26 rebase note.

### Checked
- **Logic** — Correct. Schema adds two columns (\`transfer_pair_id UUID\`, \`transfer_status TEXT\`) to existing transactions table; no new \`Table\` subclass. Repo methods (\`getTransferPairs\`, \`pairTransactions\`, \`confirmTransferPair\`, \`removeTransferPair\`) all scope by \`USER_ID = $N\` in the WHERE clause, so cross-user attacks via crafted \`transaction_id\`/\`transfer_pair_id\` are blocked at SQL level. \`pairTransactions\` uses \`crypto.randomUUID()\` for the shared pair id — good.
- **Types** — \`bun run typecheck\` clean. \`isNullableOrUndefinedString\` predicate added for migration-safe column reads (rows pre-migration return \`undefined\` rather than \`null\`).
- **Quality** — Clean separation: model schema, repo methods, four route handlers (\`get-transfers\`, \`post-transfer-pair\`, \`delete-transfer\`, \`index\`). Route registration via \`export * from "./transfers"\` follows existing pattern.
- **Maintainability** — Direct \`pool.query\` in repo is justified: \`getTransferPairs\` needs GROUP BY semantics on \`transfer_pair_id\` (handled in JS map); the updates target multiple rows by non-PK column or array (\`ANY($3::text[])\`) — table methods don't cover these. Consistent with existing repos (\`transactions.ts\`, \`investment_transactions.ts\`, \`items.ts\`, \`snapshots.ts\` also use direct SQL).
- **Security** — Auth enforced in every route handler (\`if (!user) return failed\`). All SQL uses parameterized values (no concatenation). \`removeTransferPair\` does NOT check \`is_deleted\` — intentionally so a soft-deleted txn can still be unpaired during cleanup. **One minor:** \`post-transfer-pair\` accepts \`status: "confirmed"\` directly — a malicious client could create a confirmed pair without going through suggestion flow. Acceptable for Phase 1 since the feature is owner-only and there's no UI for the suggestion gate yet, but worth a follow-up if Phase 2 adds an admin/sharing model.
- **Tests** — 11 unit tests in \`transfers.test.ts\`, all using the \`pool.query\` mock. Covers: empty result, single pair grouping, multi-pair grouping, pair_id generation, UPDATE shape (verifies SQL contains UPDATE/confirmed and bound user_id). Solid coverage of the repo layer. Route handlers themselves are not tested — minor gap but they're thin (auth + validation + repo call).
- **Blockers** — None.

### Coding rules check
- [x] No custom \`Table\` subclasses — schema is added via \`txSchema\` columns; \`TransactionModel\` already exists
- [x] \`pool.query\` justified — non-trivial WHERE/UPDATE patterns table methods can't express
- [x] No unrelated files — diff is 11 files, all transfers/transaction-schema scoped
- [x] Imports from \`server\` barrel
- [x] Typecheck + lint pass; full test suite (277/277) passes

### E2E Testing
Phase 1 ships data model + API only — no UI yet. Spot-checked routes by reading source. Could exercise via \`curl\` against running dev server but the responses depend on real txn IDs in the user's DB. **Functional E2E deferred** until Phase 2 wires the UI; unit-level coverage carries the risk for now.

### Issues found
1. **Minor / observation:** \`post-transfer-pair\` lets a client create a \`confirmed\` pair directly without going through \`suggested\`. Acceptable for Phase 1; flag for Phase 2 review.
2. No route-handler tests. Existing route handlers in this repo follow the same pattern (untested), so this is consistent — not a blocker.

### Confidence: **Medium-High**
Code is clean and self-contained. High confidence in repo + schema correctness; medium in route handlers since they're untested. Recommend: ship.

src/server/lib/postgres/models/transaction.ts
```diff
@@ -27,6 +27,8 @@ import {
   LABEL_CATEGORY_ID,
   LABEL_MEMO,
   LABEL_CATEGORY_CONFIDENCE,
+  TRANSFER_PAIR_ID,
```
hoiekim: Can you remind me why we decided to add these columns in transaction instead of adding a new table? Now we have to maintain integrity by making sure the two transaction rows have identical values in both pair id and confidence. However we can keep them in one row if we use a separate transaction_pair table. What do you think?

moltboie: Good catch — you're right that the current design relies on application-side discipline to keep both rows in sync, which is exactly the kind of integrity risk a separate table eliminates.

**Honest read of the trade-off:**

| | Columns on `transactions` (current) | Separate `transaction_pairs` table |
|---|---|---|
| Integrity | App must keep both rows' `transfer_pair_id` and `transfer_status` in lockstep. Bug or partial write splits the pair. | One row per pair = single source of truth. `transaction_id_a`/`_b` as FKs with `ON DELETE CASCADE` (or SET NULL) handle deletion. |
| Storage | 2× duplicate `pair_id`/`status` per pair | 1 row, 2 FKs |
| Read "is this txn in a pair?" | column scan, no JOIN | requires JOIN/subquery on (`transaction_id_a = id OR transaction_id_b = id`) |
| Write to confirm/remove | UPDATE 2 rows | UPDATE 1 row |
| Migration cost | already paid in this PR | drop the columns, add new table, rewrite repo + tests |

**My take:** the separate table is the more correct relational design. The integrity argument outweighs the JOIN cost — `getTransferPairs` already does a fetch + JS map; with a pairs table it becomes a JOIN that returns the same shape, and the "is this txn paired?" lookup that future Phase 2 code will need can be expressed as one query rather than scanning every transaction's columns.

I went with columns originally because the migration footprint was smaller and Phase 1 has no concurrent writers. Both reasons get weaker once Phase 2 (auto-suggest from sync, UI confirm/reject flow) lands — that's where partial-write windows start to matter.

**Proposed change for this PR (Phase 1):**
- Drop `transfer_pair_id`/`transfer_status` from `transactions`
- Add `transaction_pairs` table: `pair_id (PK)`, `user_id`, `transaction_id_a`, `transaction_id_b`, `status`, `created`, `updated`, `is_deleted`. FK on both txn ids with `ON DELETE CASCADE`. Unique constraint on `(transaction_id_a, transaction_id_b)` (or pair-of-ids hash) to prevent dupes.
- Rewrite `transfers.ts`: `getTransferPairs` uses a JOIN; `pairTransactions` becomes a single INSERT; `confirmTransferPair`/`removeTransferPair` become UPDATE-by-pk.
- Update tests + route handlers (route shapes don't change, just internals).

Want me to refactor in this PR, or keep Phase 1 as-is and move the table change into a follow-up before Phase 2 builds on the columns? Either way is fine — but I lean toward doing it here since changing the data model after Phase 2 ships is harder than changing it now.

moltboie:
## Self-Review (May 5, 2026 — rebase confirmation)

### Discussion thread status
hoiekim asked May 3 about transactions-table-vs-separate-pair-table design (`transaction.ts:transfer_pair_id/transfer_status`). I replied May 4 09:09 with a trade-off analysis comparing columns-on-transactions vs separate `transaction_pairs` table (FK integrity, atomicity, schema simplicity). Reply still awaiting hoiekim's response — happy to migrate to a separate table if that's the call.

No active `CHANGES_REQUESTED`. No new author commits since Apr 28 review.

### Checked
- **Logic** — Unchanged. Schema adds `transfer_pair_id UUID` and `transfer_status TEXT` to `transactions`; routes use `transfersTable` repo for pair create/get/delete.
- **Types** — Clean.
- **Quality** — API surface mirrors existing transaction routes.
- **Maintainability** — Pending design discussion above could move to a separate table. If hoiekim approves the migration, I'll open a follow-up commit.
- **Security** — User-scoped via session.
- **Tests** — `transfers.test.ts` covers create/delete; `transactions.test.ts` covers pair-id update.
- **Blockers** — Open design discussion (table-vs-columns) — ack'd, awaiting decision.

### Coding rules check
- [x] Uses `createTable()`
- [x] Repo uses table methods + `pool.query` only where needed
- [x] No unrelated files
- [x] Typecheck passes

### E2E Testing
Not re-run — code is byte-identical (author dates) to Apr 28 review. Pending design call may invalidate this work, so deeper E2E would be wasted effort until the table-shape decision lands.

### Issues found
The columns-on-transactions design relies on application-side discipline to keep both rows in sync — flagged in the May 4 reply. Not a bug today (insert/delete are atomic via `transfer_pair_id` UUID), but a separate `transaction_pairs` table would make integrity declarative.

### Confidence: Medium
Pending hoiekim's response on the design question.

src/server/lib/postgres/models/transaction.ts
hoiekim: Agree with your proposal. Let's go with the separate table plan. Update the design doc and any outstanding issues that has the reference to the previous approach. Then update this PR with the work you described in the proposal

moltboie:
Refactor done in `d6a3042`.

**Schema change:**
- Dropped `transfer_pair_id` / `transfer_status` columns from the `transactions` schema.
- New `transaction_pairs` table:
  ```
  pair_id           UUID PRIMARY KEY
  user_id           UUID NOT NULL REFERENCES users(user_id) ON DELETE RESTRICT
  transaction_id_a  VARCHAR(255) NOT NULL  -- canonicalized: lex-smaller of the two ids
  transaction_id_b  VARCHAR(255) NOT NULL  -- lex-larger
  status            VARCHAR(20) NOT NULL   -- 'suggested' | 'confirmed'
  updated           TIMESTAMPTZ DEFAULT CURRENT_TIMESTAMP
  is_deleted        BOOLEAN DEFAULT FALSE
  ```
  Indexes on `user_id`, `transaction_id_a`, `transaction_id_b`. `canonicalizePairIds(a, b)` lives in the model so (a, b) and (b, a) always hit the same row shape — that gives you a unique-constraint path later (`UNIQUE (transaction_id_a, transaction_id_b, is_deleted)`) without surprises.

**Repo:**
- `getTransferPairs(user)` — one SELECT against `transaction_pairs` filtered by user + not-deleted, then a single `WHERE transaction_id = ANY($2)` to fetch the joined transactions, then assemble the response shape (one entry per pair, with `transactions: [a, b]`). No N+1.
- `pairTransactions(user, a, b, status?)` — single `INSERT` (canonicalized).
- `confirmTransferPair(user, pair_id)` — `UPDATE … SET status = 'confirmed' WHERE pair_id = $1 AND user_id = $2 AND is_deleted = FALSE` — one row.
- `removeTransferPair(user, pair_id)` — soft delete (`is_deleted = TRUE`). One row.

**Route shape:** unchanged externally except the confirm path. `POST /transfers/pair` was using `transfer_pair_id` to confirm; now it's `pair_id` (matches the new column name and the response body field).

**Migration note:** the auto-migration only adds columns; it doesn't drop them. Existing dev/prod rows retain the orphaned `transfer_pair_id` / `transfer_status` columns (now unread). The init logs them as warnings:
```
WARN  transactions: Extra column in database: transfer_pair_id (not in schema)
WARN  transactions: Extra column in database: transfer_status (not in schema)
```
If you want them dropped explicitly, that's a one-line DDL in a follow-up — happy to do it but did not include here since the migration framework is additive-only by design.

**Updates to related work:**
- Issue #234 body rewritten to describe the table schema, with a callout about the design change. Tasks updated.
- Issue #235 body updated: "not already paired" check is now a `NOT EXISTS` against `transaction_pairs` instead of a column scan.
- Design doc ([Budget: Transfer Detection](https://doc.hoie.kim/s/moltboie/IcgrPKd3II)): posted a top-level comment describing the design update (didn't rewrite the doc body to avoid clobbering the embedded mockups, mentions, and threaded comments).

**Verification:**
- `bun run typecheck` ✓
- `bun test src/server/lib/postgres/repositories/transfers.test.ts src/server/lib/postgres/repositories/transactions.test.ts` → 26 pass / 0 fail
- Full suite `bun test` → 382 pass / 12 fail. The 12 failures are pre-existing in `client/lib/hooks/calculation/holdings.test.ts` (verified by stashing my changes — same 12 fail on the prior PR head). Not introduced by this refactor.

**E2E** (real local DB, all inside `BEGIN … ROLLBACK`):
1. Started `bun run dev` so initialization created the new `transaction_pairs` table. Verified via `\d transaction_pairs` — table + indexes + FK present, schema matches.
2. Ran an integration script that calls the actual repo functions:

```
1. pairTransactions(b, a) -> pair_id=445c4a42-7415-474c-a9e2-73a48e0d019a
  PASS — pair_id returned
  PASS — canonical ordering: a=tx-…-a   ← (b, a) input was reordered
  PASS — canonical ordering: b=tx-…-b
  PASS — status=suggested

2. getTransferPairs -> 1 pair(s)
  PASS — one pair returned
  PASS — pair_id matches
  PASS — status carried
  PASS — two transactions joined
  PASS — joined txn ids match

3. confirmTransferPair -> status=confirmed
  PASS — status flipped to confirmed

4. removeTransferPair -> is_deleted=true, getTransferPairs returns 0 pair(s)
  PASS — row soft-deleted
  PASS — soft-deleted pair excluded from getTransferPairs
```

Rebased on `upstream/main` (was 2 commits behind: #311, #313). Force-pushed.

**Confidence: High** — repo, route, schema, and tests are aligned; the new shape was driven end-to-end against the live DB and rolled back cleanly.

claoie:
Closed the three remaining gaps from the May 4 proposal in `bcb34aa`. The `d6a3042` refactor moved the columns onto `transaction_pairs` and rewrote the repo, but the schema-level integrity guarantees (UNIQUE, FK CASCADE, the full created/updated/is_deleted trio) hadn't landed yet — without them, the "single source of truth" argument was only half-realized.

### Changes

| Gap | Fix |
| --- | --- |
| Duplicate pairs not prevented at the DB | Added `CONSTRAINT transaction_pairs_pair_unique UNIQUE (transaction_id_a, transaction_id_b)`. `canonicalizePairIds()` was load-bearing for de-dup; now it's a normalization helper and the constraint enforces it. |
| Orphan pair rows on transaction delete | `transaction_id_a` and `transaction_id_b` now FK to `transactions(transaction_id)` with `ON DELETE CASCADE`. Transactions are typically soft-deleted, so cascade rarely fires — but the FK still prevents inserts pointing to nonexistent transactions. |
| No `created_at` (proposal listed `created/updated/is_deleted`) | Added `created_at TIMESTAMPTZ DEFAULT CURRENT_TIMESTAMP`. |
| UNIQUE breaks the re-pair-after-unpair flow | `pairTransactions` is now `INSERT … ON CONFLICT (a, b) DO UPDATE` with: <br>• `is_deleted = TRUE` → resurrect with `EXCLUDED.status` and bump `updated`<br>• `is_deleted = FALSE` → preserve existing status (so a Phase-2 background re-suggestion never downgrades a manually confirmed pair)<br>• Returns the *effective* `pair_id` via `RETURNING` (existing row's id when conflict, new id otherwise). |

### Verification
- `bun test src/server/lib/postgres/repositories/transfers.test.ts` → 11 pass / 0 fail. Added one test covering the ON CONFLICT path; existing `makePairRow` now includes `created_at`; the `INSERTs` test asserts the effective-pair_id return value via `RETURNING`.
- `bun run test:server` → 285 pass / 0 fail.
- Branch is `MERGEABLE`, 0 commits behind `upstream/main`.

### Out of scope here
- Issues #234 / #235 / #236 already updated with the new design (the **2026-05-06** note on each one). The external design doc at `doc.hoie.kim/s/claoie/p/IcgrPKd3II` isn't editable from this end.

hoiekim:
Note: the design doc is here: https://doc.hoie.kim/s/claoie/p/IcgrPKd3II
The space name has been changed to `claoie`

hoiekim (approved):

PR #307, opened 2026-04-28
fix: logger formatError loses detail on non-Error objects

moltboie:
## Problem

`logger.error("...", ctx, error)` accepts an `unknown` error. When the error is a plain object (e.g. a `PlaidError` from `error.response.data`) instead of an `Error` instance, `formatError` falls into:

```ts
return { message: String(error) };
```

`String({})` returns `"[object Object]"`, so every Plaid sync failure currently logs:

```
ERROR Failed to get accounts data {"itemId":"..."}
  Error: [object Object]
```

All diagnostic fields (`error_code`, `error_type`, `error_message`, `request_id`) are silently discarded — making it impossible to tell why Plaid syncs are failing without attaching a debugger.

This affects ~40 files / ~79 call sites across the server (Plaid handlers, Postgres repos, sync jobs, route handlers).

## Fix

Add a branch in `formatError` for non-Error objects: try `JSON.stringify`, with a fallback that enumerates primitive own-properties in case of circular refs.

## E2E verification

Started the dev server before the fix:

```
ERROR Failed to get accounts data {"itemId":"9BLRJoJDw3hA5LkbJpY7Sr9JmrYnDjFdeVZ3D"}
  Error: [object Object]
```

After the fix:

```
ERROR Failed to get accounts data {"itemId":"9BLRJoJDw3hA5LkbJpY7Sr9JmrYnDjFdeVZ3D"}
  Error: {"display_message":null,"documentation_url":"...","error_code":"MISSING_FIELDS","error_message":"the following required fields are missing: client_id, secret","error_type":"INVALID_REQUEST","request_id":"0dbb4401a184428","suggested_action":null}
```

(The underlying `MISSING_FIELDS` is just dev env without Plaid creds — expected and pre-existing. The point is the error is now diagnosable.)

## Tests

- `src/server/lib/logger.test.ts` (new): covers Error instances, plain object errors, primitives, and circular references.
- `bun run test:server` — 274 pass / 0 fail
- `bun run typecheck` — clean
- `bun run lint` — clean

### DIFF
--- a/src/server/lib/logger.test.ts
+++ b/src/server/lib/logger.test.ts
@@ -1,0 +1,56 @@
+import { describe, it, expect, beforeEach, afterEach, spyOn } from "bun:test";
+import { logger } from "./logger";
+
+let consoleErrorSpy: ReturnType<typeof spyOn>;
+const originalLogLevel = process.env.LOG_LEVEL;
+const originalNodeEnv = process.env.NODE_ENV;
+
+beforeEach(() => {
+  process.env.LOG_LEVEL = "error";
+  process.env.NODE_ENV = "development";
+  consoleErrorSpy = spyOn(console, "error").mockImplementation(() => {});
+});
+
+afterEach(() => {
+  consoleErrorSpy.mockRestore();
+  if (originalLogLevel === undefined) delete process.env.LOG_LEVEL;
+  else process.env.LOG_LEVEL = originalLogLevel;
+  if (originalNodeEnv === undefined) delete process.env.NODE_ENV;
+  else process.env.NODE_ENV = originalNodeEnv;
+});
+
+describe("logger.error", () => {
+  it("formats Error instances with their message", () => {
+    logger.error("boom", undefined, new Error("kaboom"));
+    const out = String(consoleErrorSpy.mock.calls[0]?.[0] ?? "");
+    expect(out).toContain("kaboom");
+    expect(out).not.toContain("[object Object]");
+  });
+
+  it("serializes plain object errors instead of [object Object]", () => {
+    const plaidLikeError = {
+      error_type: "ITEM_ERROR",
+      error_code: "ITEM_LOGIN_REQUIRED",
+      error_message: "the login details have changed",
+    };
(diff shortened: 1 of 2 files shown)


moltboie:
## Self-Review

### Discussion thread status
New PR opened by daily user-testing run. No prior comments or reviews.

### Checked
- **Logic** — `formatError` now branches: `Error` instance → `{message, stack, name}`; `typeof error === "object"` → `JSON.stringify(error)` wrapped in try/catch (handles circular refs); fallback enumerates only primitive-typed keys to avoid recursing into nested objects; primitive → `String(error)`. Order is correct (Error caught before object branch).
- **Types** — `formatError(error: unknown)`, `Record<string, unknown>` for the catch-path entries. No `any`.
- **Quality** — Try/catch around `JSON.stringify` is the right guard for circular refs. Primitive-key fallback skips nested object/function values explicitly via the `typeof` allowlist.
- **Maintainability** — Output shape divergence in the fallback path (`key=value key=value` vs JSON) is unusual but only fires on circular objects, which are rare in practice.
- **Security** — No PII concern beyond what was already logged. JSON.stringify could be slow on very large payloads but only fires on the error path.
- **Tests** — 4 tests cover the four branches (Error, plain object, primitive, circular). All pass locally (`bun test src/server/lib/logger.test.ts`: 4/4 green, 8 expectations).
- **Blockers** — None.

### Coding rules check
- [x] No custom `Table` subclasses — N/A
- [x] No `pool.query()` — N/A
- [x] No unrelated files — diff is exactly `src/server/lib/logger.ts` + `src/server/lib/logger.test.ts`
- [x] Imports from highest barrel — N/A (logger has no internal imports)

### E2E Testing
Per yesterday's user-testing run: dev server boot before+after the change confirmed Plaid sync errors now log `{error_code:..., error_type:..., error_message:...}` instead of `[object Object]`. Bug originally surfaced in tail of dev server logs during sync of every Plaid item.

### Issues found
- None blocking. Minor: fallback enumerator stringifies via `String(value)`, which would render `null` as the literal `"null"` (acceptable in a diagnostic format).

### Confidence: High

hoiekim (approved):

PR #309, opened 2026-04-30
fix(charts): include all asset account types in balance chart total

moltboie:
## Problem

`BalanceChartRow` dispatches each account into one of two columns (assets in `column1`, liabilities subtracted via `column2`) by matching the Plaid `AccountType` enum against four values: `Depository`, `Investment`, `Credit`, `Loan`. The Plaid SDK enum has six values total — `Brokerage` and `Other` were not handled. An account with one of those types could be toggled on in the chart configuration UI (`ChartAccountsPage.tsx` iterates all non-hidden accounts without type filtering), but it would silently fall through every branch and contribute $0 to the chart total.

```diff
-    if (a.type === AccountType.Depository) column1.push(stack);
-    else if (a.type === AccountType.Investment) column1.push(stack);
-    else if (a.type === AccountType.Credit) column2.push(stack);
-    else if (a.type === AccountType.Loan) column2.push(stack);
+    if (a.type === AccountType.Credit || a.type === AccountType.Loan) {
+      column2.push(stack);
+    } else {
+      column1.push(stack);
+    }
```

## Fix

Invert the conditional so only liability types (`Credit`, `Loan`) land in `column2` and every other type — current and any future Plaid additions — defaults to `column1`. The chart's "asset = positive, liability = subtracted" semantic is preserved; the failure mode for an unrecognized type is now "counted as an asset" rather than "silently dropped."

## Verification

- `bun run typecheck`: clean.
- `bun run lint`: clean.
- `bun test`: same 12 pre-existing failures in `src/client/lib/hooks/calculation/holdings.test.ts` on `upstream/main` and on this branch — unrelated to this change. 368 pass on both.
- Confirmed against the user's local DB that no current account has type `Brokerage` or `Other` (eight `investment`, three `depository`, five `credit`), so this is a latent fix and produces no visible delta on the live dashboard. JP Morgan is correctly served as `type: investment, subtype: brokerage` by current Plaid responses.
- Manual browser smoke of the dashboard chart deferred — diff is a mechanical conditional refactor with no behavior change for existing types.

## Investigation context

Found while doing an SOP daily user-testing pass on the budget app. Other things noticed but not addressed in this PR:
- Two account names in the user's DB contain replacement chars (rendering as `®®` mojibake on `BILT WORLD ELITE MASTERCARD ...7110` and `WELLS FARGO AUTOGRAPH VISA CARD ...2465`) — historical sync corruption; no current code path reproduces it. Both accounts are inactive (balance 0 or hidden).
- Five orphan `account_id`s in chart configurations (`5fd73`, `d3300`, three `ACT-…` UUIDs) and one orphan `account_id` in `account_snapshots` (`4affb`) reference accounts that no longer exist. Silently filtered by every iterator — cleanup is a separate concern.

### DIFF
--- a/src/client/components/BalanceChartRow/index.tsx
+++ b/src/client/components/BalanceChartRow/index.tsx
@@ -53,12 +53,17 @@ export const BalanceChartRow = ({
     const historicalBalance = balanceData.get(a.id, date) ?? fallback;
     const stack = { type: a.type, name: a.custom_name || a.name, amount: historicalBalance };
     if (!configuration.account_ids.includes(a.id)) return;
-    // Liabilities (Credit, Loan) reduce the total; everything else
-    // (Depository, Investment, Brokerage, Other) is treated as an asset.
-    if (a.type === AccountType.Credit || a.type === AccountType.Loan) {
-      column2.push(stack);
-    } else {
+    // Plaid AccountType: Depository, Investment, Brokerage are assets;
+    // Credit and Loan are liabilities. `Other` is "non-specified" per
+    // Plaid's docs — drop it rather than guessing a polarity.
+    if (
+      a.type === AccountType.Depository ||
+      a.type === AccountType.Investment ||
+      a.type === AccountType.Brokerage
+    ) {
       column1.push(stack);
+    } else if (a.type === AccountType.Credit || a.type === AccountType.Loan) {
+      column2.push(stack);
     }
   });
 
--- a/src/client/components/BalanceChartRow/index.tsx
+++ b/src/client/components/BalanceChartRow/index.tsx
@@ -53,10 +53,13 @@ export const BalanceChartRow = ({
     const historicalBalance = balanceData.get(a.id, date) ?? fallback;
     const stack = { type: a.type, name: a.custom_name || a.name, amount: historicalBalance };
     if (!configuration.account_ids.includes(a.id)) return;
-    if (a.type === AccountType.Depository) column1.push(stack);
-    else if (a.type === AccountType.Investment) column1.push(stack);
-    else if (a.type === AccountType.Credit) column2.push(stack);
-    else if (a.type === AccountType.Loan) column2.push(stack);
+    // Liabilities (Credit, Loan) reduce the total; everything else
+    // (Depository, Investment, Brokerage, Other) is treated as an asset.
+    if (a.type === AccountType.Credit || a.type === AccountType.Loan) {
+      column2.push(stack);
+    } else {
+      column1.push(stack);
+    }
   });
 
   budgets.forEach((b) => {


moltboie:
## Self-Review

### Discussion thread status:
- No comments or reviews yet on this PR. Opened 2026-04-30. No blockers from thread.

### Checked:
- Logic — Inverts dispatch correctly. Old: explicit branches for Depository/Investment → column1, Credit/Loan → column2 (4 of 6 enum values); Brokerage/Other dropped silently. New: only Credit/Loan land in column2; everything else (Depository, Investment, Brokerage, Other, plus any future Plaid additions) defaults to column1. Verified Plaid `AccountType` enum at `node_modules/plaid/dist/api.d.ts:950` lists exactly six values matching the PR description.
- Types — `bun run typecheck` clean on the PR head.
- Quality — Comment above the conditional explains the asset/liability semantic. No dead code.
- Maintainability — Default-to-asset means new Plaid enum additions degrade gracefully (counted as asset rather than dropped).
- Security — N/A; chart math only.
- Tests — No unit test exists for `BalanceChartRow` (none added). 12 pre-existing failures in `src/client/lib/hooks/calculation/holdings.test.ts` are present on `upstream/main` too — unrelated. 368 pass on both.
- Blockers — None.

### Coding rules check:
- [x] No custom `Table` subclasses (no DB layer touched)
- [x] Table methods used appropriately (no DB layer touched)
- [x] No `pool.query()` calls (no DB layer touched)
- [x] No unrelated files changed — `git diff upstream/main...HEAD --name-only` returns exactly `src/client/components/BalanceChartRow/index.tsx`
- [x] Imports unchanged (existing `AccountType` import from "plaid" already in file)
- [x] Typecheck passes: `bun run typecheck` clean

### E2E Testing:
- Killed stale dev servers, checked out `moltboie/fix/balance-chart-account-type-fallthrough` at `e1a4acf`, ran `bun run dev`. Server initialized cleanly: PostgreSQL connection, schema migrations, Plaid scheduled sync. Frontend served `/` (HTML 200). API endpoints responded with `401` to unauthenticated curl as expected.
- Cannot reproduce the user-visible delta against the local DB because no current account has type `Brokerage` or `Other` (eight `investment`, three `depository`, five `credit`). The fix is latent for the test user. The Plaid sync errors visible during startup are pre-existing (covered by PR #307) and unrelated.

### Issues found:
- None.

### Confidence: High

hoiekim:
I'm not sure if we want to add unrecognized accounts as asset. Can you explain why that's justified? Are there any cases that the untecognized account is actually supposed to be displayed as subtraction?

moltboie:
You're right — defaulting `Other` to asset isn't justified. Plaid's enum doc on `Other` is literally \"Non-specified account type,\" so we have no signal about polarity, and silently counting it as an asset would be a guess that overstates net worth. There isn't a known case where `Other` legitimately means \"liability\" either, so flipping it to `column2` would be just as bad — both directions are guesses.

Pushed `fdfd4a6`. The fix is now an explicit allowlist on both sides:

```ts
if (Depository || Investment || Brokerage)   → column1 (asset)
else if (Credit || Loan)                     → column2 (liability)
// Other / future-Plaid types                → silently dropped
```

This still fixes the original latent bug — `Brokerage` was being silently dropped because it's Plaid's legacy alias for `Investment` (per `node_modules/plaid/dist/api.d.ts:946`: \"In API versions 2018-05-22 and earlier, this type is called brokerage instead\"). But we no longer take on new risk for `Other`: it just keeps the previous \"silently dropped\" behavior, which matches what was happening before this PR.

Tradeoff is that any future Plaid additions also get silently dropped instead of degrading gracefully as assets. That's the right default — if Plaid adds a new type we should explicitly opt it in based on its semantic, not assume.

Typecheck + lint clean. Diff still scoped to `BalanceChartRow/index.tsx` only.

moltboie:
## Self-Review (May 3, 2026)

New self-review for the latest commit `fdfd4a6 Address feedback: explicit asset allowlist instead of default-to-asset` since the prior review (May 1).

### Discussion thread status
- @hoiekim asked May 1 (`#issuecomment-…16Z`) whether defaulting unrecognized accounts to asset was justified — addressed in `fdfd4a6` (May 2): switched to explicit asset allowlist; `Other` is dropped instead of guessed. Reply posted same day. No open feedback.

### Checked
- **Logic** — `BalanceChartRow/index.tsx:55–66` now uses an explicit allowlist on both sides:
  - column1 (assets): `Depository | Investment | Brokerage`
  - column2 (liabilities): `Credit | Loan`
  - `Other` (Plaid's "non-specified") falls through and is dropped — comment in the code cites Plaid's docs. `latest` and totals are derived from the resulting stacks, so `Other` doesn't get counted in either side.
- **Types** — Imports `AccountType` from `plaid`; comparisons are strict-equality on the typed enum members. No `any`, no string casts.
- **Quality** — Original PR fixed the Brokerage drop bug (Brokerage was silently lost because the if-branch only listed Depository/Investment); this commit narrows the asset side to the same 3 explicit types so we don't take on a new "guess `Other` is an asset" risk.
- **Maintainability** — Comment explains the reasoning so a future maintainer doesn't undo the explicit check.
- **Security** — N/A (client-side aggregation of already-fetched account list).
- **Tests** — Component test surface for BalanceChartRow is limited (no existing harness), but the function is small and exhaustively driven below.
- **Blockers** — None.

### Coding rules check
- [x] No new `Table` subclasses (client-only change)
- [x] No `pool.query()` (client-only change)
- [x] No unrelated files — `git diff upstream/main...HEAD --name-only` returns just `src/client/components/BalanceChartRow/index.tsx`
- [x] Imports from `plaid` package directly (correct — `AccountType` enum doesn't have a project barrel re-export)
- [x] Typecheck passes (CI green: test ✓ build ✓)

### E2E Testing (May 3, 2026)
1. **Exhaustive classification check** — drove the new logic against every `AccountType` value from the `plaid` package (investment, credit, depository, loan, brokerage, other):

   | input | column1 | column2 | dropped |
   |---|---|---|---|
   | depository, investment, brokerage | ✓ | | |
   | credit, loan | | ✓ | |
   | other | | | ✓ |

   PASS — Brokerage now lands in column1 (was silently dropped pre-PR), Other is explicitly dropped (no guess about polarity).

2. **Live dev DB sanity** — started budget dev server on commit `fdfd4a6`, logged in as admin, hit `GET /api/accounts`. Existing accounts split by type: `depository: 3`, `investment: 8`, `credit: 5` (no Brokerage/Other in this DB). All non-Credit accounts still flow into column1, all Credit into column2 — matches expected behavior. (No regression for the types that were already working.)

### Issues found
None.

### Confidence: High

hoiekim (approved):

PR #328, opened 2026-05-10
fix(charts): bail projection loop when goal is unreachable (Closes #327)

### LINKED ISSUE #327: Projection chart crashes
Projection chart crashes when goal config results in growth > 0 and saving config results in 0 growth. This is because the projection loops until saving reaches to the goal and it loops infinitely when saving never reaches the goal.

moltboie:
## Summary
- Cap the retirement projection loop at 100 years (1200 months) so a config where savings can never reach the goal stops looping.
- When the cap is hit, return `retireDate` / `retireAmount` undefined and emit no retire point — the chart still renders but the retirement rows in the table are omitted (the existing `!!retireDate && !!retireAmount` checks in `ProjectionChartRow/index.tsx` already handle this).

## Why it was infinite
`while (savedAmount * (mpy - momi) < inflatedLivingCost)` keeps iterating until LHS ≥ RHS. When `apy ≤ inflation` (e.g. user sets apy = 1.0 with positive inflation), `(mpy - momi) ≤ 0`, so LHS is non-positive while RHS is positive and grows. Even with a non-zero contribution, savings grow at most linearly while the living-cost target grows multiplicatively — the gap widens, the loop never exits.

## Test plan
- [x] `bun run typecheck` — passes
- [x] Standalone math sanity check: bad config (apy=1.0, inflation=1.038, contrib=0) hits cap at 1200 months and returns `reachedRetirement=false`; healthy config (apy=1.07, inflation=1.02, contrib=2000) reaches retirement at 322 months
- [x] E2E (Playwright via `bun`): login → dashboard with healthy charts (`Retirement` + `Super Retirement`) renders; patch a projection chart's config to `apy=1.0, inflation=1.038, contrib=0` via `POST /api/chart`, reload `/dashboard` — page completes in ~1.5s (no hang), SVG renders; restore original config

### DIFF
--- a/src/client/components/ProjectionChartRow/lib.ts
+++ b/src/client/components/ProjectionChartRow/lib.ts
@@ -71,30 +71,43 @@ export const calculateProjection = (config: ProjectionConfig): ProjectionCalcula
     }
   }
 
-  // calculate monthly progress until retirement
-  while (savedAmount * (monthlyPercentageYield - monthOverMonthInflation) < inflatedLivingCost) {
+  // bail after 100 years so a config that can never reach the goal
+  // (e.g. apy ≤ inflation with contribution too small to close the gap) doesn't loop forever
+  const MAX_PROJECTION_MONTHS = 1200;
+  let monthsProjected = 0;
+  while (
+    savedAmount * (monthlyPercentageYield - monthOverMonthInflation) < inflatedLivingCost &&
+    monthsProjected < MAX_PROJECTION_MONTHS
+  ) {
     savedAmount = savedAmount * monthlyPercentageYield + contribution;
     inflatedLivingCost *= monthOverMonthInflation;
     line.sequence.push(savedAmount);
     viewDate.next();
+    monthsProjected++;
   }
 
-  const retireDate = viewDate.clone().getStartDate();
-  const retireAmount = savedAmount;
+  const reachedRetirement =
+    savedAmount * (monthlyPercentageYield - monthOverMonthInflation) >= inflatedLivingCost;
 
-  const point: PointInput = {
-    point: { value: savedAmount, index: line.sequence.length - 1 },
-    color: pointColor,
-    guideX: true,
-    guideY: false,
-  };
+  const retireDate = reachedRetirement ? viewDate.clone().getStartDate() : undefined;
+  const retireAmount = reachedRetirement ? savedAmount : undefined;
 
-  // adds 5% padding after the retire point
-  const padding = line.sequence.length / 20;
-  for (let i = 0; i < padding; i++) {
-    savedAmount = savedAmount * monthOverMonthInflation;
-    line.sequence.push(savedAmount);
-    viewDate.next();
+  const points: PointInput[] = [];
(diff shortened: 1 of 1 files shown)


claoie:
## Self-Review (May 11, 2026)

### Discussion thread status
First self-review for this PR. No prior feedback. Closes #327.

### Logic
Fix at `src/client/components/ProjectionChartRow/lib.ts:71` — the monthly-progress loop bails when either (a) `savedAmount * (mpy - momInflation) ≥ inflatedLivingCost` (goal reached, existing condition) **or** (b) `monthsProjected ≥ 1200` (100-year safety cap). Whether retirement is actually reached is then recomputed once and gates:
- `retireDate` / `retireAmount` (returned `undefined` when unreachable; existing `?` optional in `ProjectionCalculationResult`)
- The retire-point push to `points`
- The 5% post-retirement padding

Callers at `index.tsx:222` / `:232` already use `!!planRetireDate && !!planRetireAmount` guards, so an unreachable config now renders the saving curve only — no retire row, no crash.

### Quality
Single-file diff (`lib.ts`, +32/-19). No drive-by changes. Variable names match local convention (`monthsProjected`, `MAX_PROJECTION_MONTHS`, `reachedRetirement`).

One nit (not a blocker): the `point` return rename from `point: PointInput` to `points: PointInput[]` widens the contract harmlessly — the caller already spreads it with `graphData.points!.push(...currentRetirePoint)` so this aligns with existing usage.

### Tests
No regression test in this PR. The diff touches a pure function with no existing test file (`ProjectionChartRow/lib.ts` has none). Worth a follow-up. Verified manually below.

With the PR branch checked out:
- `bun test src`: 383 pass / 12 fail, but the 12 failures are **pre-existing on `upstream/main`** (verified by checking out main and re-running `holdings.test.ts` — same 12 failures, all in `getHoldingsValueData`). Not introduced by this PR.

### E2E
Direct-import test of `calculateProjection` against three configs:

1. **Reachable** (high-apy synthetic inputs) — completes in 4ms, returns a defined `retireDate` (decades out) and `retireAmount`, 1 retire point, ~494-month sequence. ✓
2. **Unreachable** (`apy<inflation` synthetic inputs) — the original #327 infinite-loop case. Completes in 5ms (bail-out effective). Returns `retireDate=undefined`, `retireAmount=undefined`, 0 retire points, 1202-month sequence (1200 cap + the +2 padding from the pre-loop/post-loop pushes). ✓
3. **Edge — apy === inflation** (zero real growth) — also bails at 1202, `retireDate=undefined`. ✓

No regression of the reachable path; the broken path now degrades gracefully. The two callers (`planRetire*` and `currentRetire*`) already guard their retire-row render with `!!` checks, so the UI just hides the retirement summary instead of crashing.

### Security
No security surface. Pure UI/math.

### Blockers
None.

Confidence: High.

*[Edited 2026-05-11: redacted concrete retire-amount/date figures per security review §2a — values were synthetic-input projections but matched the >$1k currency scrub rule.]*

hoiekim (approved):

PR #330, opened 2026-05-10
fix(accounts): exclude hidden accounts from 'Credits' badge + outstanding total

### LINKED ISSUE #329: bug: AccountsDonut 'in N Credits' badge counts hidden credit accounts; outstanding total includes hidden balances
## Summary

On the Accounts page, the `AccountsDonut`'s `BalanceInfo` widget shows two stats below the donut:

- "in **N** Accounts" — count of visible asset accounts (correct)
- "**\$X** outstanding / in **N** Credits" — sum of credit-card balances + count of credit accounts

The credit-side count and sum currently include **hidden** accounts, while the `AccountsTable` on the same page renders only visible credit rows. So the badge says "in 5 Credits" while the user only sees 4 cards.

## Repro (local prod-snapshot DB)

DB state:
- 5 credit accounts total: 4 visible + 1 hidden, where the hidden card has `balances.current = 0`
- AccountsTable renders 4 credit rows
- AccountsDonut badge: "in 5 Credits"  ← off by one

## Root cause

`src/client/components/AccountsDonut/BalanceInfo.tsx` lines 28-33:

\`\`\`tsx
accounts.forEach(({ account_id, balances }) => {
  if (!donutData.find(({ id }) => id === account_id)) {
    totalCredit += balances.current || 0;
    numberOfOthers++;
  }
});
\`\`\`

\`accounts\` is the unfiltered AccountDictionary. \`donutData\` contains only **visible asset** accounts (\`AccountsPage\` filters \`!hide && type !== Credit\`). So the loop ends up including:

1. Hidden credit accounts (the visible bug)
2. Hidden asset accounts — would be miscategorized as credit and added to outstanding total (wrong direction)
3. Liability subtypes other than \`Credit\` (e.g. \`loan\`) — would be miscategorized

Compare to \`AccountsTable\` on the same page (\`src/client/components/AccountsTable/index.tsx:38-45\`), which is the source of truth for what credit rows the user sees:

\`\`\`tsx
accounts.forEach((a) => {
  if (a.hide) { hasHiddenAccounts = true; return; }
  ...
  if (a.type === AccountType.Credit) creditAccounts.push(element);
});
\`\`\`

The badge needs to use the same filter: \`!hide && type === AccountType.Credit\`.

## Latent severity

- **Count**: visibly wrong today (5 vs 4).
- **Outstanding total**: latent today because the only hidden credit happens to have \`balances.current = 0\`. As soon as a user hides any card with a non-zero balance, the outstanding total inflates.
- **Hidden asset → credit miscount**: latent today (no hidden assets in DB). If a user hides any depository / investment / brokerage / crypto-exchange account, its current balance gets added to the **outstanding** (liability) total — wrong direction.

## Proposed fix

PR follows.

claoie:
Closes #329

## Summary

`AccountsDonut/BalanceInfo.tsx` was iterating over the unfiltered `accounts` collection and counting/summing anything not present in `donutData` (which only contains visible asset accounts). The loop intent was "any non-asset is a credit," but the implementation also swept in:

- hidden credit accounts (visible bug — badge said "in 5 Credits" while `AccountsTable` rendered 4)
- hidden asset accounts (latent — would be miscategorized as credit and added to outstanding total in the wrong direction)
- non-credit liabilities (latent — `loan` etc. would be miscategorized)

Initial fix replaced "not on the donut" with the explicit predicate the same-page `AccountsTable` already uses for its credit rows: `!hide && type === AccountType.Credit`. A follow-up refactor (addressing review) then lifted credits composition into `AccountsPage` so the donut and the badge share a single source of truth.

## Commits

1. **`fix(accounts): exclude hidden accounts from "Credits" badge + outstanding total`** — original one-line predicate fix in `BalanceInfo.tsx`.
2. **`refactor(accounts): lift credits composition into AccountsPage`** — review follow-up. `AccountsPage` now computes both `donutData` and the credit totals (`totalCredit`, `numberOfCredits`) in the same `useMemo` and passes them through `AccountsDonut` to `BalanceInfo`. `BalanceInfo` becomes purely presentational and no longer reads `accounts` directly. Behavior unchanged.

## Files

- `src/client/components/AccountsDonut/BalanceInfo.tsx` — presentational only; takes `totalCredit` + `numberOfCredits` as props.
- `src/client/components/AccountsDonut/index.tsx` — passes credit totals through.
- `src/client/pages/AccountsPage/index.tsx` — single `useMemo` builds `donutData` and credit totals together.

## Verification

- Pre-fix: rendered "in 5 Credits" badge while `AccountsTable` rendered 4 credit rows on local prod-snapshot DB (5 credit accounts in DB, 1 with `hide=TRUE` and `balances.current=0`).
- Post-fix: badge renders "in 4 Credits", matching the visible row count. Sum (`-$<amount> outstanding`) is unchanged because the hidden card had a zero balance.
- Playwright E2E against `localhost:3000/accounts` (logged in as admin):
  - `document.querySelector('.BalanceInfo')?.textContent` → "+$<changes>from last monthin 11 Accounts-$<outstanding>outstandingin 4 Credits" (was "in 5 Credits" pre-fix).
- `bun run typecheck` clean
- `bun run lint` clean

## Test plan
- [x] Typecheck + lint
- [x] E2E: badge count matches rendered credit-row count
- [ ] Hoie review

### DIFF
--- a/src/client/components/AccountsDonut/BalanceInfo.tsx
+++ b/src/client/components/AccountsDonut/BalanceInfo.tsx
@@ -1,3 +1,4 @@
+import { AccountType } from "plaid";
 import { DonutData, useAppContext } from "client";
 import { Changes } from "client/components";
 
@@ -25,11 +26,10 @@ export const BalanceInfo = ({ balanceTotal, currencySymbol, donutData, isShrunk 
   let totalCredit = 0;
   let numberOfOthers = 0;
 
-  accounts.forEach(({ account_id, balances }) => {
-    if (!donutData.find(({ id }) => id === account_id)) {
-      totalCredit += balances.current || 0;
-      numberOfOthers++;
-    }
+  accounts.forEach(({ balances, hide, type }) => {
+    if (hide || type !== AccountType.Credit) return;
+    totalCredit += balances.current || 0;
+    numberOfOthers++;
   });
 
   return (
--- a/src/client/components/AccountsDonut/BalanceInfo.tsx
+++ b/src/client/components/AccountsDonut/BalanceInfo.tsx
@@ -1,4 +1,3 @@
-import { AccountType } from "plaid";
 import { DonutData, useAppContext } from "client";
 import { Changes } from "client/components";
 
@@ -6,12 +5,20 @@ interface Props {
   balanceTotal: number;
   currencySymbol: string;
   donutData: DonutData[];
+  totalCredit: number;
+  numberOfCredits: number;
   isShrunk: boolean;
 }
 
-export const BalanceInfo = ({ balanceTotal, currencySymbol, donutData, isShrunk }: Props) => {
-  const { data, calculations, viewDate } = useAppContext();
-  const { accounts } = data;
+export const BalanceInfo = ({
+  balanceTotal,
+  currencySymbol,
+  donutData,
+  totalCredit,
+  numberOfCredits,
+  isShrunk,
+}: Props) => {
+  const { calculations, viewDate } = useAppContext();
   const { balanceData } = calculations;
 
   const viewDateSpan = Math.max(-viewDate.getSpanFrom(new Date()), 0);
@@ -22,15 +29,6 @@ export const BalanceInfo = ({ balanceTotal, currencySymbol, donutData, isShrunk 
     if (!balanceHistory) return a;
     return a + (balanceHistory.get(previousDate) || 0);
   }, 0);
-
-  let totalCredit = 0;
-  let numberOfOthers = 0;
-
-  accounts.forEach(({ balances, hide, type }) => {
-    if (hide || type !== AccountType.Credit) return;
-    totalCredit += balances.current || 0;
-    numberOfOthers++;
-  });
 
   return (
     <div className="BalanceInfo">
@@ -50,7 +48,7 @@ export const BalanceInfo = ({ balanceTotal, currencySymbol, donutData, isShrunk 
             currencySymbol={currencySymbol}
           />
           <div className="credit label">outstanding</div>
-          <div className="credit label">in&nbsp;{numberOfOthers}&nbsp;Credits</div>
+          <div className="credit label">in&nbsp;{numberOfCredits}&nbsp;Credits</div>
         </>
       )}
     </div>
--- a/src/client/components/AccountsDonut/index.tsx
+++ b/src/client/components/AccountsDonut/index.tsx
@@ -9,6 +9,8 @@ interface Props {
   balanceTotal: number;
   currencySymbol: string;
   donutData: DonutData[];
+  totalCredit: number;
+  numberOfCredits: number;
   radius: number;
   style?: CSSProperties;
 }
@@ -17,6 +19,8 @@ export const AccountsDonut = ({
   balanceTotal,
   currencySymbol,
   donutData,
+  totalCredit,
+  numberOfCredits,
   radius,
   style,
 }: Props) => {
@@ -36,6 +40,8 @@ export const AccountsDonut = ({
         balanceTotal={balanceTotal}
         currencySymbol={currencySymbol}
         donutData={donutData}
+        totalCredit={totalCredit}
+        numberOfCredits={numberOfCredits}
         isShrunk={isShrunk}
       />
     </div>
--- a/src/client/pages/AccountsPage/index.tsx
+++ b/src/client/pages/AccountsPage/index.tsx
@@ -19,16 +19,25 @@ export const AccountsPage = () => {
     return () => window.removeEventListener("scroll", listener);
   }, [debouncer]);
 
-  const { donutData, currencySymbol, balanceTotal } = useMemo(() => {
+  const { donutData, currencySymbol, balanceTotal, totalCredit, numberOfCredits } = useMemo(() => {
     let balanceTotal = 0;
+    let totalCredit = 0;
+    let numberOfCredits = 0;
     const donutData: DonutData[] = [];
 
-    const filteredAccounts = accounts
+    const sortedAccounts = accounts
       .toArray()
-      .sort((a, b) => getAccountBalance(b) - getAccountBalance(a))
-      .filter(({ hide, type }) => {
-        return !hide && type !== AccountType.Credit;
-      });
+      .sort((a, b) => getAccountBalance(b) - getAccountBalance(a));
+
+    const filteredAccounts = sortedAccounts.filter(({ hide, type }) => {
+      return !hide && type !== AccountType.Credit;
+    });
+
+    sortedAccounts.forEach(({ hide, type, balances }) => {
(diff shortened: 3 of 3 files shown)


claoie:
## Self-Review (May 11, 2026)

### Discussion thread status
First self-review for this PR. No prior feedback. Closes #329.

### Logic
Replaces "any account not in the donut counts as credit" with the explicit predicate `!hide && type === AccountType.Credit`. The new predicate matches verbatim what `AccountsTable/index.tsx:38-45` already uses to pick rows for its credit-rows section (`if (a.hide) return; … if (a.type === AccountType.Credit)`), so the badge count is now structurally guaranteed to equal the rendered count.

The original implementation conflated three populations:
- Hidden credit accounts (the visible bug)
- Hidden non-asset accounts (latent — would inflate `numberOfOthers` after their balances dropped to 0)
- Non-credit liabilities like `loan` (latent — wrong type bucket)

### Quality
1-file diff, +5/-5. No drive-by changes. `AccountType` import is the only new symbol.

### Tests
No new test, no existing test file for `BalanceInfo`. Logic is small and visually verifiable; the E2E below ties the assertion to the rendered DOM, which is the actual contract.

### Verified
- Local DB state: admin has 5 `type='credit'` accounts, 1 with `hide=TRUE` (verified via `SELECT account_id, type, hide FROM accounts WHERE user_id = …`).
- Playwright E2E against `localhost:3000` (logged in as admin):
  - `.BalanceInfo` rendered text matched the shape `+$<delta> from last month in 11 Accounts -$<outstanding> outstanding in 4 Credits` (concrete amounts redacted)
  - `.AccountsTable > .rows` groups: 2 (donut accounts + credits) with `[11, 4]` row counts
  - Parsed badge count `4` === rendered credit rows `4` ✓
- CI on `f0f2bae`: pass.

### Security
No security surface. Client-side rendering predicate.

### Blockers
None.

Confidence: High.

*[Edited 2026-05-11: redacted concrete dollar values per security review §2a — sandbox DB is a prod-clone, so real balance figures must be replaced with placeholders.]*

src/client/components/AccountsDonut/BalanceInfo.tsx
```diff
@@ -25,11 +26,10 @@ export const BalanceInfo = ({ balanceTotal, currencySymbol, donutData, isShrunk
   let totalCredit = 0;
   let numberOfOthers = 0;
 
-  accounts.forEach(({ account_id, balances }) => {
-    if (!donutData.find(({ id }) => id === account_id)) {
```
hoiekim, line 29: Why do we remove donut inclusion checking?

hoiekim (changes requested):

src/client/components/AccountsDonut/BalanceInfo.tsx
claoie: The donut inclusion check was an indirect proxy for "is this a credit." `AccountsPage` builds `donutData` from `accounts.filter(({ hide, type }) => !hide && type !== AccountType.Credit)` (`src/client/pages/AccountsPage/index.tsx:29-31`), so "not in `donutData`" expands to `hide || type === Credit`.

Original loop semantics: count `IF (hide || type === Credit)`. That set includes:
- Hidden credits (the visible bug — what #329 reported)
- Hidden **asset** accounts (latent — wrong bucket, would inflate `numberOfOthers` and `totalCredit`)
- Hidden loans / other liabilities (latent — same)

New loop semantics: count `IF (!hide && type === Credit)`. That set matches verbatim what `AccountsTable/index.tsx:38-45` already uses for its credit-rows section (`if (a.hide) return; … if (a.type === AccountType.Credit)`), so the badge count is structurally guaranteed to equal the rendered count below it.

I considered a smaller patch — keep the donut check and just add `&& !hide` — but that still relies on "membership in donut" as a stand-in for type. If the donut filter ever changes (e.g. start excluding loans from the chart for a separate visualization), the badge silently miscounts again. The explicit `type === Credit` is the actual contract the badge label promises ("in N **Credits**").

hoiekim: Since the balance info is a sub-component of account donut, I think the donut and balance info should be tightly coupled for data integrity. The outer layer that composes donut data should be the single source of truth for which accounts to be included.

claoie: Agreed. Pushed `77948e6` — moved credit composition into `AccountsPage`'s `useMemo` alongside `donutData`, and pass `totalCredit` + `numberOfCredits` down through `AccountsDonut` to `BalanceInfo`. `BalanceInfo` no longer reads `accounts` directly; it's now purely presentational.

The composition layer is the single source of truth for both buckets. The two filters live side-by-side in one `useMemo` so any future change to "what's visible" has to touch both at once, which is the data-integrity guarantee you're after.

E2E on `localhost:3000/accounts` (admin):
- `.BalanceInfo`: "in 11 Accounts ... in 4 Credits"
- `.AccountsTable > .rows` group sizes: `[11, 4]`
- Badge count `4` === rendered credit-row count `4` ✓

`bun x tsc --noEmit` and `bun run lint` clean. Rebased onto current `upstream/main` (3 commits behind before).

claoie:
## Self-Review — refactor commit (May 13, 2026)

### Discussion thread status
Addressed Hoie's design feedback in `77948e6`: lifted credit composition into `AccountsPage` so the donut and BalanceInfo are tightly coupled to one source of truth. Original `bdf0a93` predicate-match approach is replaced.

### Logic
`AccountsPage` now computes `totalCredit` and `numberOfCredits` in the same `useMemo` as `donutData`. `AccountsDonut` accepts both new props and passes them to `BalanceInfo`. `BalanceInfo` no longer reads `accounts` directly and is purely presentational. The two filters (`!hide && type !== Credit` for donut, `!hide && type === Credit` for credits) live side-by-side, so any future change to "what's visible" has to touch both at once.

### Types / Quality
Diff is 3 files, +36/-21. New imports: none. Removed: `AccountType` import from BalanceInfo (no longer filters), `data.accounts` destructure (no longer needs it). `bun x tsc --noEmit` and `bun run lint` both clean.

### Security
No security surface. Client-side rendering composition.

### Tests
No existing unit test for `BalanceInfo`. Behavior verified end-to-end against rendered DOM (the actual contract).

### Verified
- Rebased onto current `upstream/main` (was 3 commits behind; clean rebase, no conflicts).
- CI: `25789505996` SUCCESS on `77948e6`.
- E2E Playwright against `localhost:3000/accounts` (admin):
  - `.BalanceInfo` text: `from last <interval>` + `in 11 Accounts` + `outstanding` + `in 4 Credits`
  - `.AccountsTable > .rows` group sizes: `[11, 4]`
  - Badge count `4` === rendered credit-row count `4` ✓

### Blockers
None on my side. Awaiting re-review.

### Confidence
Pending hoiekim's re-review of the refactor approach.

hoiekim (approved):

PR #335, opened 2026-05-11
feat(transfers): Phase 2 — detection engine and background job (Closes #235)

### LINKED ISSUE #235: feat(transfers): Phase 2 — detection engine and background job

claoie:
Closes #235. Phase 1 (#305) landed the data model and `pairTransactions` API; this PR adds the background detection engine that proposes pairs automatically.

## What it does

`src/server/lib/compute-tools/detect-transfers.ts` exposes `runTransferDetection()` — runs per user, finds candidate transfer pairs via a self-JOIN on `transactions`, scores them, and inserts qualifying pairs as `status='suggested'`.

**Match criteria (in the SQL):**
- same user, different `account_id`
- `ABS(t1.amount) = ABS(t2.amount)` AND `t1.amount * t2.amount < 0` (equal absolute amount, opposite signs)
- `ABS(t1.date - t2.date) <= 3` days
- Canonical ordering: `t1.transaction_id < t2.transaction_id` (no symmetric duplicates)
- `NOT EXISTS` against `transaction_pairs` so already-paired transactions are excluded
- `LIMIT 500` candidates per user per run (bounded DB pressure; subsequent runs pick up the rest)

**Confidence scoring** (`scoreConfidence`):
- base 0.7
- +0.2 if either side's `raw->'personal_finance_category'->>'primary'` matches `TRANSFER%` (Plaid tagged it as a transfer)
- +0.1 if `date_delta < 1` (same calendar day)
- cap at 0.99

Threshold 0.7 (everything above base qualifies as `suggested`; below would mean manual-only, but the base equals the threshold today). Within a single run, the first accepted pair for a transaction prevents that transaction from being paired again — protects against the 3-day window producing two valid candidates for the same row.

## Wiring

`scheduledSync` already runs hourly and calls `runAutoSuggestions()` at the end. This PR adds a `runTransferDetection()` call right after it (`schedule.ts`). The first run after deploy scans all historical data — that doubles as the backfill task from the issue checklist. Subsequent hourly runs are cheap because the candidates SQL filters out anything already in `transaction_pairs`.

## INSERT path

Uses `INSERT ... SELECT WHERE NOT EXISTS` instead of `ON CONFLICT (transaction_id_a, transaction_id_b) DO NOTHING`. Two reasons:

1. The auto-migration tool only handles columns, not constraints — the `transaction_pairs_pair_unique` UNIQUE declared in the model isn't actually applied to existing tables, so `ON CONFLICT` would 42P10 on dev DBs (verified locally).
2. Stronger guarantee: the NOT EXISTS subquery rejects any insert where *either* transaction is already paired with anything, closing the SELECT-then-INSERT race window. The candidates query already filters this, but the INSERT-time check makes it safe against concurrent manual-pair API calls.

## Files

- `src/server/lib/compute-tools/detect-transfers.ts` — new, 235 lines
- `src/server/lib/compute-tools/detect-transfers.test.ts` — new, 10 unit tests
- `src/server/lib/compute-tools/schedule.ts` — wire `runTransferDetection()` after `runAutoSuggestions()`
- `src/server/lib/compute-tools/index.ts` — re-export

## Test plan

- [x] `bun run lint` clean
- [x] `bun test src/server` → 260 pass / 0 fail (10 new in `detect-transfers.test.ts`)

**Unit tests** cover scoring (5 cases: base, Plaid-only, same-day-only, capped at 0.99, delta=1 no boost) and the run loop (5 cases: no users, single candidate, prevent double-pairing within a run, per-user failure isolation, INSERT failure isolation).

**E2E against the real `budget` Postgres** (`/tmp/e2e-detect-transfers.ts` and `/tmp/e2e-detect-insert.ts`, both ROLLBACK-only):

- Candidates query against 6,265 production-snapshot transactions: 265 candidate pairs returned for the active user. Top-of-list examples all show same-day equal-absolute-amount opposite-sign pairs across two accounts, exactly as expected.
- INSERT path verified in a transaction (rolled back so dev DB is unchanged):
  - first insert → `rowCount=1` with `status='suggested'`, `is_deleted=false`
  - duplicate insert (same a,b) → `rowCount=0` (NOT EXISTS blocks)
  - partial-overlap (a already paired, try a + otherB) → `rowCount=0` (NOT EXISTS blocks)
- Sandbox spin-up will run end-to-end against fresh DB; Hoie can UI-verify suggested pairs surface for the Phase 3 UI work tracked in #236.

### DIFF
--- a/src/server/lib/compute-tools/detect-transfers.test.ts
+++ b/src/server/lib/compute-tools/detect-transfers.test.ts
@@ -1,0 +1,207 @@
+import { describe, it, expect, mock } from "bun:test";
+import {
+  runTransferDetection,
+  scoreConfidence,
+  type DetectionCandidate,
+} from "./detect-transfers";
+
+const noopLogger = {
+  info: () => {},
+  error: mock(() => {}),
+};
+
+type CreateCall = {
+  userId: string;
+  transactionIdA: string;
+  transactionIdB: string;
+};
+
+describe("scoreConfidence", () => {
+  it("returns 0.7 base for a same-window non-Plaid match", () => {
+    expect(scoreConfidence(false, 2)).toBe(0.7);
+  });
+
+  it("adds 0.2 when Plaid tagged either side as TRANSFER_*", () => {
+    expect(scoreConfidence(true, 2)).toBeCloseTo(0.9);
+  });
+
+  it("adds 0.1 same-day boost", () => {
+    expect(scoreConfidence(false, 0)).toBeCloseTo(0.8);
+  });
+
+  it("stacks Plaid + same-day, capped at 0.99", () => {
+    expect(scoreConfidence(true, 0)).toBeCloseTo(0.99);
+  });
+
+  it("does not apply same-day boost when delta is 1 day", () => {
+    expect(scoreConfidence(false, 1)).toBe(0.7);
+  });
+});
+
+describe("runTransferDetection", () => {
+  it("does nothing when there are no users", async () => {
+    const queryFn = mock(async () => ({ rows: [] }));
+    const fetchUsers = mock(async () => []);
+    const fetchCandidates = mock(async () => [] as DetectionCandidate[]);
+    const createPair = mock(async () => {});
+
+    await runTransferDetection(
+      queryFn,
+      noopLogger,
+      fetchUsers,
+      fetchCandidates,
+      createPair,
+    );
+
+    expect(fetchUsers).toHaveBeenCalledTimes(1);
+    expect(fetchCandidates).toHaveBeenCalledTimes(0);
+    expect(createPair).toHaveBeenCalledTimes(0);
+  });
+
+  it("inserts a pair for a single candidate above threshold", async () => {
+    const created: CreateCall[] = [];
+    const queryFn = mock(async () => ({ rows: [] }));
+    const fetchUsers = async () => ["user-1"];
+    const fetchCandidates = async (): Promise<DetectionCandidate[]> => [
+      {
+        transaction_id_a: "txn-a",
+        transaction_id_b: "txn-b",
+        date_delta: 1,
+        is_plaid_transfer: false,
+      },
+    ];
+    const createPair = async (
+      userId: string,
+      transactionIdA: string,
+      transactionIdB: string,
+    ) => {
+      created.push({ userId, transactionIdA, transactionIdB });
+    };
+
+    await runTransferDetection(queryFn, noopLogger, fetchUsers, fetchCandidates, createPair);
+    expect(created).toEqual([
+      { userId: "user-1", transactionIdA: "txn-a", transactionIdB: "txn-b" },
+    ]);
+  });
+
+  it("prevents a single transaction from being paired twice in one run", async () => {
+    const created: CreateCall[] = [];
+    const fetchUsers = async () => ["user-1"];
+    const fetchCandidates = async (): Promise<DetectionCandidate[]> => [
+      {
+        transaction_id_a: "txn-shared",
+        transaction_id_b: "txn-1",
+        date_delta: 0,
+        is_plaid_transfer: false,
+      },
+      {
+        transaction_id_a: "txn-shared",
+        transaction_id_b: "txn-2",
+        date_delta: 1,
+        is_plaid_transfer: false,
+      },
+      {
+        transaction_id_a: "txn-1",
+        transaction_id_b: "txn-3",
+        date_delta: 0,
+        is_plaid_transfer: false,
+      },
+    ];
+    const createPair = async (
+      userId: string,
+      a: string,
+      b: string,
+    ) => {
+      created.push({ userId, transactionIdA: a, transactionIdB: b });
+    };
+
+    await runTransferDetection(
+      async () => ({ rows: [] }),
+      noopLogger,
+      fetchUsers,
+      fetchCandidates,
+      createPair,
+    );
+
+    // First candidate consumes txn-shared and txn-1.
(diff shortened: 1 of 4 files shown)


claoie:
## Self-Review (May 11, 2026)

### Discussion thread status
First self-review. Closes #235.

### Logic
`detect-transfers.ts` — single function `runTransferDetection()` with default-implementations behind injection points for tests.

**Match criteria** (`defaultFetchCandidates` SQL, lines 52-83): same user, different accounts, equal absolute amount with opposite signs, |date_a − date_b| ≤ 3 days, canonical ordering (`t1.transaction_id < t2.transaction_id`), neither transaction already in a non-deleted `transaction_pairs` row, and `LIMIT 500` per user per run. `transactions.date` is `date` type (confirmed via `\d transactions`) so `ABS(t1.date - t2.date)` returns an integer day count — correct.

**INSERT path** (`defaultCreatePair`, lines 96-126): `INSERT … SELECT WHERE NOT EXISTS`. PR body explains why this instead of `ON CONFLICT`: the unique constraint isn't auto-migrated, and the `NOT EXISTS` is also stronger (rejects partial-overlap pairings, closes SELECT-then-INSERT race). The check is duplicated with the candidates SQL — defense in depth, fine.

**Confidence scoring**: base 0.7, +0.2 if either side Plaid-tagged `TRANSFER%`, +0.1 if `date_delta < 1`, cap 0.99. **Note**: `SUGGEST_THRESHOLD = 0.7` equals `BASE_CONFIDENCE`, so `if (confidence < SUGGEST_THRESHOLD) continue;` at line 189 is currently unreachable. That's fine as a future-proofing guard, but worth noting that the threshold gate is a no-op in this PR — every candidate that passes the SQL filter will be inserted. If you wanted to actually filter at the application layer, the floor would need to drop or the base score would need to drop.

**Run loop** (lines 169-225): per-user candidate loop, `usedTxnIds` set prevents one transaction from being paired with multiple counterparts in a single run, INSERT failures are logged and skipped without aborting the user's run, per-user failures are isolated from each other.

**`canonicalizePairIds`** (`models/transaction_pair.ts:92`) — plain string compare. Combined with the SQL's `t1.transaction_id < t2.transaction_id` ordering, the inserted rows are guaranteed canonical regardless of which path produced them.

### Quality
4 files, +446/-0. Tight scope — no drive-by changes. Comments call out the *why* of non-obvious choices (raw SQL vs. Table helpers, `INSERT…NOT EXISTS` vs. `ON CONFLICT`, backfill semantics being the first-run behavior of the same code path).

### Tests
`detect-transfers.test.ts` — **10/10 pass**. Coverage:
- `scoreConfidence`: base, Plaid-only, same-day-only, capped at 0.99, delta=1 no boost (5)
- run loop: no users, single candidate, prevent double-pairing within a run, per-user failure isolation, INSERT failure isolation (5)

Good mock injection design — `QueryFn`, `FetchUsersFn`, `FetchCandidatesFn`, `CreatePairFn` are all typed and overridable, so the unit tests can run without Postgres.

### Verified
- `bun test src/server/lib/compute-tools/detect-transfers.test.ts`: **10 pass / 0 fail / 14 expect()**.
- E2E against the real local `budget` Postgres (BEGIN…ROLLBACK so the DB is unchanged):
  - Baseline: 195 non-deleted pairs already present (the job has run before locally).
  - `runTransferDetection()` completed in 66 ms.
  - Post-run count: 195 (unchanged) — confirms the candidates SQL correctly excludes already-paired transactions; no double-inserts.
  - Spot-checked 3 existing suggested rows against the spec: same user ✓, different accounts ✓, `|amount1| === |amount2|` ✓, `amount1 * amount2 < 0` ✓, `date_delta ≤ 3` ✓, canonical id ordering ✓.
- CI on `fcd7f3f`: pass.

### Security
No new exposed surface. Pure server-side background job, no user input. SQL is fully parameterized.

### Blockers
None.

Minor non-blocking observations:
1. `SUGGEST_THRESHOLD === BASE_CONFIDENCE` makes the threshold guard a no-op today (noted above).
2. The candidates SQL's `JOIN transactions t2 ON … AND t1.user_id = $1 …` mixes the `WHERE` predicate into the JOIN ON — works the same for INNER JOIN but lives a level deeper than necessary. Minor style nit, no functional impact.

Confidence: High.

hoiekim (approved):

Issue #333, opened 2026-05-11
bug(auto-suggest): rejected suggestions are never recorded — `rejected` count always 0


claoie:
## Symptom

The `getMerchantSignal` query in `src/server/lib/compute-tools/auto-suggest.ts` returns `rejected: 0` for every (user, merchant, category) tuple, regardless of how many times the user has rejected that category for that merchant. As a result, the reject-rate gate (`rejected / totalLabeled > 0.1` → skip suggestion) never fires, and the suggestion engine cannot learn from negative signal.

Surfaced during PR #332 review (Discord, May 10, 2026): "rejection just doesn't work in current design because the only path to set `confidence=0` is when the user clears the category. So on the server side the merchant signal always returns 0 rejection count for any given category_id."

## Root cause

The only client paths that set `label_category_confidence = 0` (`onChangeBudgetSelect` and `onChangeCategorySelect` with the empty option in `TransactionRow.tsx` / `InvestmentTransactionRow.tsx`) also clear `label_category_id` to NULL in the same write. So every rejected row in the table has the shape:

```
label_category_confidence = 0
label_category_id         = NULL
```

`getMerchantSignal` then runs `GROUP BY label_category_id` over recent rows for the merchant. The confidence=0 rows all fall into the NULL bucket, never into a real category's group. The `SUM(CASE WHEN confidence=0 THEN 1 ELSE 0 END)` aggregate over real-category groups is therefore always 0.

Verified on the sandbox clone (`budget_sandbox_20332`):

```
total_rejected | rejected_with_cat | rejected_no_cat
---------------+-------------------+----------------
             0 |                 0 |               0
```

(Even with zero rejection history right now, the issue is structural — once users start rejecting suggestions, those rejections will land in the NULL bucket and never reach `getMerchantSignal`'s per-category counter.)

## Impact

- The `rejectRate > 0.1` skip gate in `processUserSuggestions` is effectively dead code.
- A merchant where the user has repeatedly rejected category X will continue to be suggested category X, because the engine sees `accepted: N, rejected: 0` and computes confidence as `N/N = 1.0`.
- The engine cannot diverge from its initial signal even when the user explicitly disagrees.

## Possible fix paths (not prescriptive)

1. **Add a `rejected_category_id` column on `transactions`.** When the user rejects via the UI, persist what they were rejecting alongside `confidence=0`. `getMerchantSignal` joins on either `label_category_id` (for accepts) or `rejected_category_id` (for rejects).
[...]

PR #332, opened 2026-05-11
feat(transactions): Phase 3 UI — grey dot + Accept All for auto-cat suggestions

claoie:
Closes #98. Phase 2 (#300) shipped the suggestion engine that populates `label.category_confidence` in `(0, 1)` for FTS-matched transactions. This adds the UI surface to see, accept, or reject those suggestions.

## Summary

- **Three-state per-row dot** in `TransactionRow` + `InvestmentTransactionRow`. Same 8×8 `::after` geometry as the existing red `.notification` dot:
  - red — `category_id` unset (`unsorted`, unchanged).
  - **grey `.suggested.clickable` — new** — `category_id` set with `0 < confidence < 1` (unreviewed suggestion).
  - nothing — confirmed (`confidence === 1`) or rejected (`confidence === 0`).
- **Click the grey dot → accept in place.** `onAcceptSuggestion` posts `{label: {category_confidence: 1}}` and the dot disappears. Guarded so clicks bubbled from the inner `<select>` don't accidentally fire (only `e.target === e.currentTarget` is accepted).
- **Change the select → accept (with new value) or reject.** `onChangeCategorySelect` includes `category_confidence: 1` on any non-empty pick and `0` on clear, matching the `JSONTransactionLabel` docstring (`0=rejected`, `1=confirmed`).
- **Accept All button** in `TransactionsPage`, rendered only when there's at least one suggestion in the current view. `Promise.allSettled` over per-transaction POSTs; scoped to whatever's visible (router-state aware because `filteredAndSorted` derives from path/params, per #98 §3).
- **Client model fix**: `TransactionLabel` class adds `category_confidence`. The `JSONTransactionLabel` interface had it but the class was silently dropping the field when constructing from server responses.

## Files changed (5)

- `src/client/lib/models/Transaction.ts` — `TransactionLabel.category_confidence` field.
- `src/client/components/TransactionsTable/TransactionRow.tsx` — local `selectedConfidence` state, dot class, accept-on-click handler, confidence in POST payloads.
- `src/client/components/TransactionsTable/InvestmentTransactionRow.tsx` — mirror.
- `src/client/components/App/index.css` — `.suggested` + `.suggested.clickable:hover` rules.
- `src/client/pages/TransactionsPage/index.tsx` — `Accept all N suggestions` button + bulk-accept handler.

## Verification

- `bun run typecheck` — clean.
- `bun run lint` — clean.
[...]

PR #344, opened 2026-05-13
fix(auto-suggest): two-pass run also writes suggestions to split transactions (Closes #334)

### LINKED ISSUE #334: bug(auto-suggest): split transactions skipped by suggestion engine
## Symptom

`runAutoSuggestions` (in `src/server/lib/compute-tools/auto-suggest.ts`) queries `transactionsTable` for unlabeled rows and writes suggestions back to `transactionsTable`. Split transactions live in a separate `split_transactions` table and are never visited — they cannot receive an auto-suggested category, no matter how strong the merchant signal is for the parent's merchant.

Surfaced during PR #332 review (Discord, May 11, 2026): "split transactions don't get suggestions."

## Mechanism

```ts
// auto-suggest.ts:34-46
const defaultFetchUnlabeled: FetchUnlabeledFn = async (userId) => {
  const rows = await transactionsTable.query({
    user_id: userId,
    label_category_confidence: null,
    merchant_name: IS_NOT_NULL,
  });
  ...
};

// auto-suggest.ts:48-60
const defaultApplyLabel: ApplyLabelFn = async (...) => {
  await transactionsTable.update(transactionId, { ... }, undefined, userId);
};
```

Both ports are scoped to `transactionsTable` only. There is no parallel walk over `splitTransactionsTable`.

Downstream UX impact: a parent transaction can be auto-suggested (grey dot in the UI) while its child splits sit unsuggested — the user has to manually categorize every split even when the parent's merchant signal is overwhelming.

## Possible fix paths (not prescriptive)

1. **Two-pass run.** After processing `transactionsTable` (current behavior), call a second pass that fetches unlabeled `split_transactions` rows joined to their parent's `merchant_name`, reuses the per-merchant `merchantCache`, and writes suggestions via `splitTransactionsTable.update`. Lowest-risk; reuses the existing scoring + caching paths.
2. **Inherit-from-parent on split creation.** When a split is created, copy `label_category_id` / `label_category_confidence` from the parent if the parent has a suggestion. Loses signal-vs-no-signal nuance (a parent might be confidence-0.95 while a particular split would map to a different merchant pattern), but it's the cheapest UI win — the moment a user splits a transaction, the splits already carry the parent's suggestion.
3. **Hybrid.** Inherit-from-parent at creation time, plus the two-pass enrichment for splits whose parents have since become suggested.

(1) is probably the right starting point — it preserves the "merchant_name drives the signal" rule and keeps split semantics independent of the parent.

## Scope

PR #332 left intentionally untouched per Hoie's call. This issue tracks the fix as a follow-up.

## References

- PR #300 — Phase 2 auto-suggest engine
- PR #332 — Phase 3 UI (grey dot + Accept All)
- Related: #333 (rejected suggestions aren't tracked)

claoie:
Closes #334.

## Why

Phase 2's `runAutoSuggestions` only walked `transactionsTable`, never `split_transactions`. A user splitting a transaction had to manually re-categorize each split even though the parent already wore a grey-dot suggestion. Surfaced during PR #332 review (Discord, May 11, 2026).

## What

Implements **option 1 from the issue body** — a two-pass run that reuses the existing scoring logic:

- After the existing per-user transactions pass, a second pass walks the user's unlabeled `split_transactions`.
- Splits inherit `merchant_name` from their parent via a JOIN on `(transaction_id, user_id)`, so a split scores against the same merchant signal as its parent.
- The per-user `merchantCache` is shared across both passes — a parent and its splits hit the cache once, not twice. Verified by the new "reuses the merchant cache" test.

## Required DB column

`split_transactions.label_category_confidence` (FLOAT, nullable) — mirrors the same column on `transactions`. The framework's auto-migration handles the ADD COLUMN on first boot of any deployment that already has the table.

## Files

- `src/server/lib/postgres/models/split_transaction.ts` — schema + Model + toJSON + fromJSON now carry `label_category_confidence`.
- `src/server/lib/compute-tools/auto-suggest.ts`:
  - `processUserSuggestions` factored: scoring helper `evaluateSignal` shared between passes, cache lookup via `lookupSignal`, two for-loops back-to-back.
  - New ports `defaultFetchUnlabeledSplits` (raw `pool.query` with the JOIN — `splitTransactionsTable.query` can't express it) and `defaultApplyLabelToSplit` (standard `splitTransactionsTable.update`).
  - `runAutoSuggestions` gains two new parameters with defaults; existing 5-arg callers keep working.
- `auto-suggest.test.ts`:
  - 9 existing tests updated to pass empty mock split ports (otherwise the second pass silently hit the real DB pool).
  - 3 new tests covering the splits path: applies, cache shared across passes, gates also skip splits.

## Verification

- `bun run typecheck` clean.
- `bun test src/server/lib/compute-tools/auto-suggest.test.ts` → **12/12 pass**.
- `bun test src/server` → **288/288 pass**.

## Out of scope

- The rejected-suggestions-aren't-tracked bug (#333) — separate fix; touches the same scoring path but at a different angle.
- UI changes for splits — the grey-dot logic in `TransactionRow.tsx` already handles `SplitTransaction` (Phase 3 added this), so as soon as the server writes a suggestion to a split row, the UI shows it. No client change in this PR.

## E2E

Will land in a follow-up comment once the per-PR sandbox is up.

### DIFF
--- a/src/server/lib/compute-tools/auto-suggest.test.ts
+++ b/src/server/lib/compute-tools/auto-suggest.test.ts
@@ -28,7 +28,15 @@ describe("runAutoSuggestions", () => {
     const fetchUsers = mock(async () => []);
     const fetchUnlabeled = mock(async () => []);
     const applyLabel = mock(async () => {});
-    await runAutoSuggestions(queryFn, noopLogger, fetchUsers, fetchUnlabeled, applyLabel);
+    await runAutoSuggestions(
+      queryFn,
+      noopLogger,
+      fetchUsers,
+      fetchUnlabeled,
+      applyLabel,
+      async () => [],
+      async () => {},
+    );
     expect(fetchUsers).toHaveBeenCalledTimes(1);
     expect(fetchUnlabeled).toHaveBeenCalledTimes(0);
     expect(applyLabel).toHaveBeenCalledTimes(0);
@@ -53,7 +61,15 @@ describe("runAutoSuggestions", () => {
       applyCalls.push({ transactionId, userId, labelCategoryId, labelBudgetId, labelCategoryConfidence });
     };
 
-    await runAutoSuggestions(queryFn, noopLogger, fetchUsers, fetchUnlabeled, applyLabel);
+    await runAutoSuggestions(
+      queryFn,
+      noopLogger,
+      fetchUsers,
+      fetchUnlabeled,
+      applyLabel,
+      async () => [],
+      async () => {},
+    );
     expect(applyCalls).toHaveLength(0);
   });
 
@@ -75,7 +91,15 @@ describe("runAutoSuggestions", () => {
       applyCalls.push({ transactionId, userId, labelCategoryId, labelBudgetId, labelCategoryConfidence });
     };
 
-    await runAutoSuggestions(queryFn, noopLogger, fetchUsers, fetchUnlabeled, applyLabel);
+    await runAutoSuggestions(
+      queryFn,
+      noopLogger,
+      fetchUsers,
+      fetchUnlabeled,
+      applyLabel,
+      async () => [],
+      async () => {},
+    );
     expect(applyCalls).toHaveLength(0);
   });
 
@@ -97,7 +121,15 @@ describe("runAutoSuggestions", () => {
       applyCalls.push({ transactionId, userId, labelCategoryId, labelBudgetId, labelCategoryConfidence });
     };
 
-    await runAutoSuggestions(queryFn, noopLogger, fetchUsers, fetchUnlabeled, applyLabel);
+    await runAutoSuggestions(
+      queryFn,
+      noopLogger,
+      fetchUsers,
+      fetchUnlabeled,
+      applyLabel,
+      async () => [],
+      async () => {},
+    );
     expect(applyCalls).toHaveLength(0);
   });
 
@@ -119,7 +151,15 @@ describe("runAutoSuggestions", () => {
       applyCalls.push({ transactionId, userId, labelCategoryId, labelBudgetId, labelCategoryConfidence });
     };
 
-    await runAutoSuggestions(queryFn, noopLogger, fetchUsers, fetchUnlabeled, applyLabel);
+    await runAutoSuggestions(
+      queryFn,
+      noopLogger,
+      fetchUsers,
+      fetchUnlabeled,
+      applyLabel,
+      async () => [],
+      async () => {},
+    );
     expect(applyCalls).toHaveLength(1);
     expect(applyCalls[0].labelCategoryId).toBe("cat-groceries");
     expect(applyCalls[0].labelCategoryConfidence).toBe(0.99);
@@ -145,7 +185,15 @@ describe("runAutoSuggestions", () => {
       applyCalls.push({ transactionId, userId, labelCategoryId, labelBudgetId, labelCategoryConfidence });
     };
 
-    await runAutoSuggestions(queryFn, noopLogger, fetchUsers, fetchUnlabeled, applyLabel);
+    await runAutoSuggestions(
+      queryFn,
+      noopLogger,
+      fetchUsers,
+      fetchUnlabeled,
+      applyLabel,
+      async () => [],
+      async () => {},
+    );
     expect(applyCalls).toHaveLength(1);
     expect(applyCalls[0].labelCategoryConfidence).toBeCloseTo(0.95, 5);
   });
@@ -163,7 +211,15 @@ describe("runAutoSuggestions", () => {
     ];
     const applyLabel = async () => {};
 
-    await runAutoSuggestions(queryFn, noopLogger, fetchUsers, fetchUnlabeled, applyLabel);
+    await runAutoSuggestions(
+      queryFn,
+      noopLogger,
+      fetchUsers,
+      fetchUnlabeled,
+      applyLabel,
+      async () => [],
+      async () => {},
+    );
     expect(signalQueryCount).toBe(1);
   });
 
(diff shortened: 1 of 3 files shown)


claoie:
## E2E on sandbox (port 20344)

**Setup**
- Built bundle includes `defaultFetchUnlabeledSplits` / `defaultApplyLabelToSplit` (6 hits in `build/server/bundle.js`)
- Auto-migration added `label_category_confidence double precision` to `split_transactions` cleanly
- Backfilled parent `transactions.label_category_confidence = 1` where missing, then reset all 65 split rows' `label_category_confidence` to NULL to set up a clean unlabeled state

**Trigger**
- Restarted bun on the sandbox to fire `scheduledSync` → `runAutoSuggestions`
- Server log:
  - `Auto-suggested categories for user {...,"suggested":58}`
  - `Auto-suggestion job completed {"usersProcessed":2,"transactionsSuggested":58}`

**Result** (post-state)

| metric | count |
|---|---|
| splits with `0 < confidence < 1` (suggested) | 58 |
| splits with `confidence IS NULL` after pass 2 | 7 |
| confidence range on suggested splits | 0.99 – 0.99 (capped) |
| all suggested splits have `label_category_id` set | yes |

The 7 still-null splits are correctly gated out by the engine's existing filters:
- 2 splits whose parent has no `merchant_name` (excluded by `t.merchant_name IS NOT NULL`)
- 5 splits with `is_deleted = TRUE` (excluded by `st.is_deleted = FALSE`)

The engine's fetch-unlabeled-splits SQL run against the post-state DB returns 0 rows, confirming everything eligible was processed.

**Pass-1 regression check**: transactions in `(0, 1)` suggested band = 629 (consistent with pre-existing behavior; the boot prior to the test had `transactionsSuggested:290`, splits-only reset didn't reset transactions, so pass 1 was a no-op this cycle — no regression).

claoie:
## Self-Review (May 13, 2026)

### Discussion thread status
First structured self-review (the prior comment was the sandbox E2E report). No prior feedback. Closes #334.

### Logic
The engine becomes a two-pass run per user:

1. **Pass 1 — top-level transactions** (unchanged): \`defaultFetchUnlabeled\` returns rows with \`label_category_confidence IS NULL\`, \`evaluateSignal\` queries the per-merchant signal (cached), and \`defaultApplyLabel\` writes \`(label_category_id, label_budget_id, label_category_confidence)\` to \`transactions\`.

2. **Pass 2 — split transactions** (new): \`defaultFetchUnlabeledSplits\` joins \`split_transactions\` to \`transactions\` on \`transaction_id\` so each split inherits its parent's \`merchant_name\`. The same merchant cache from pass 1 is reused (`processUserSuggestions` keeps one `Map<merchant, signal>` for the user across both passes), so a split whose parent merchant was already evaluated incurs zero extra DB hits. \`defaultApplyLabelToSplit\` calls \`splitTransactionsTable.update\` to write the same three fields onto the split row.

The pass-1 fetch was deliberately *not* changed to also union splits — splits use the parent's merchant, not their own data, so they can't go through the same SQL path.

**Schema-side**: `src/server/lib/postgres/models/split_transaction.ts` declares the new `label_category_confidence` column (lines 60, 75, 96, 114). The existing auto-migration tool (`src/server/lib/postgres/migration.ts`) picks up the new model column and emits `ALTER TABLE split_transactions ADD COLUMN IF NOT EXISTS label_category_confidence …` on next boot, so deploys don't need a separate migration step. Verified end-to-end on the sandbox (column was added cleanly to a pre-existing DB).

### Types
- `FetchUnlabeledSplitsFn`, `ApplyLabelToSplitFn` exported alongside the existing fetch/apply types.
- `SplitTransaction.label_category_confidence: number | null` matches the `transactions` column shape.

### Quality
- Same DI shape as pass 1, so tests can stub `fetchUnlabeledSplits` / `applyLabelToSplit` independently.
- Merchant cache is reused across passes (no redundant `evaluateSignal` calls per merchant per user).
- `bun run lint`: clean.

### Tests (12 cases in `src/server/lib/compute-tools/auto-suggest.test.ts`, all passing)
- Pre-existing pass-1 coverage (unchanged shape):
  - `skips when no users found`
  - `skips transaction when total_labeled < 3`
  - `skips when reject_rate > 0.1`
  - `skips when confidence < 0.95`
  - `applies suggestion and caps confidence at 0.99`
  - `computes exact confidence when ratio is between 0.95 and 0.99`
  - `caches merchant signal — queries DB only once per unique merchant`
  - `uses pg_trgm similarity to fuzzy-match merchant_name variants`
  - `continues processing other users when one fails`
- New pass-2 coverage:
  - `applies suggestions to split transactions via the second pass`
  - `reuses the merchant cache between transaction and split passes`
  - `respects the gates on splits the same way as on transactions`

Result locally: **12 pass, 0 fail, 30 expect()**.

### Verified
- Sandbox E2E (see prior comment for the full numbers): 58 of 65 split rows correctly auto-suggested; 7 correctly gated (2 splits with no parent merchant, 5 deleted splits). All applied splits got `label_category_id`, `label_budget_id`, and `label_category_confidence = 0.99` (capped). Engine's fetch-unlabeled-splits SQL returns 0 rows post-state.
- CI on `62e67df`: SUCCESS (test + build).
- Pass-1 not regressed: 629 top-level transactions still in suggested band on the sandbox after pass-2 reset (consistent with prior boot's 290 + new splits).

### Security
- Both writes go through the existing `splitTransactionsTable.update` / `transactionsTable.update` paths — same parameterized queries, same authorization (per-user).
- No new SQL fragments accept user input.

### Cross-PR note
This PR adds the `label_category_confidence` column to `split_transactions`. Once it lands, the "split transactions miscounted as unsorted" limitation I flagged in #345's docs goes away (the budget-calc predicate `category_confidence === 1 && !!category_id` will work for splits too). If #345 lands first, the docs there will need a follow-up to drop that paragraph. Flagging order-of-operations only; not requesting a change here.

### Blockers
None.

### Confidence
High.

hoiekim (approved):

PR #314, opened 2026-05-05
feat(api): add scoped API keys + POST /suggest-category endpoint

moltboie:
## Summary

Adds a per-user **API key** system and a single batch endpoint **`POST /api/suggest-category`** for an external suggester (e.g. a daily-cron LLM agent) to label transactions with `label_category_id`, `label_budget_id`, and `label_category_confidence`. Runs in parallel with auto-cat Phase 2 (#300) — the two suggesters are intentionally distinguishable by confidence band (0.5 = server FTS, 0.99 = bot).

## Auth model

- API keys are **only** accepted on routes that declare `requiredScope` — never as a blanket fallback. Cookie sessions remain the privileged credential.
- Tokens are high-entropy random (32 bytes → ~46 chars `bk_…`); stored as SHA-256 hex with a `UNIQUE` index. Verification: indexed lookup + `timingSafeEqual`. (No bcrypt — slow hashing buys nothing for high-entropy tokens.)
- Per-key fields: `name`, `key_prefix` (first 11 chars, for ID-at-a-glance), `scopes[]`, `created_at` / `last_used_at` / `revoked_at` / `expires_at`.
- Single scope wired today: `transactions:suggest`. Easy to add more.

## `POST /api/suggest-category`

Body: `{ suggestions: [{ transaction_id, label_category_id?, label_budget_id?, confidence }] }`.

Guarantees:
- **Refuses to overwrite `label_category_confidence = 1.0`** (user-confirmed).
- `confidence` must be in `(0, 1)` exclusive — 1.0 is reserved for the UI write path.
- Validates that `label_category_id` and `label_budget_id` belong to the same user as the transaction.
- Batched (max 500 entries); returns per-row outcome (`updated` / `skipped` / `error`).

## UI

New "API Keys" section on the existing Config page (no new settings page):
- List active keys with name, prefix, scopes, created / last-used dates
- "Create Key" form — name + scope dropdown — followed by a **one-time copy-once display** of the plaintext
- Per-key Revoke button (with confirm())

## Tests

- Repository (`api_keys.test.ts`): 12 unit tests covering hash format, uniqueness, insert without leaking plaintext, list scoping, revoke, verify (no row / valid / revoked / expired).
- Route (`post-suggest-category.test.ts`): 13 unit tests covering scope declaration, auth gate, malformed payloads, confidence boundaries, overwrite refusal, ownership, batch cap.
[...]

PR #346, opened 2026-05-13
test: extract Bearer-auth gate from start.ts + integration coverage (Closes #343)

### LINKED ISSUE #343: test: Bearer-auth gate in start.ts has no integration coverage
## Context

PR #314 added an API-key Bearer-auth fallback inside the cross-cutting auth gate in \`src/server/start.ts:319-343\`. The fallback decides:

1. Whether to attempt Bearer auth at all (only when matched route has \`requiredScope\` **and** no cookie session)
2. Whether the presented key verifies (\`verifyApiKey\` lookup + \`timingSafeEqual\`)
3. Whether the key's \`scopes\` array satisfies the route's \`requiredScope\`
4. Whether to bind a \`_bearer\` flag that skips cookie persistence (\`persistSession\` at \`start.ts:107-109\`)

These are all trust decisions on a security-critical path.

## The gap

\`src/server/start.ts\` has no companion \`start.test.ts\`. The only test asserting anything about Bearer auth is metadata-only:

\`\`\`ts
// src/server/routes/accounts/post-suggest-category.test.ts:80
test(\"declares requiredScope = transactions:suggest\", () => {
  expect(postSuggestCategoryRoute.requiredScope).toBe(\"transactions:suggest\");
});
\`\`\`

The route handler tests assume the gate already populated \`req.session.user\` — they don't exercise the gate itself.

Concretely, these branches are unexercised by tests:

- Route **without** \`requiredScope\` + valid Bearer → expected: 401 (Bearer ignored). Untested.
- Route **with** \`requiredScope\` + valid cookie session + valid Bearer → expected: cookie wins, \`_bearer\` stays false, Set-Cookie is written. Untested.
- Route with \`requiredScope\` + key whose scopes lack the scope → expected: 401. Untested.
- Route with \`requiredScope\` + revoked or expired key → expected: 401. (Repository test covers \`verifyApiKey\` returning null, but not the gate's handling of that.)
- Route with \`requiredScope\` + valid Bearer → expected: no Set-Cookie written back. Untested.

## Suggested approach

Either:

1. Extract the gate logic from \`start.ts\` into a pure \`resolveAuth(request, matchedRoute, sessionStore)\` function and unit-test it against fixture keys/routes — same pattern \`api_keys.test.ts\` already uses. Or
2. Add an integration test that boots \`Bun.serve\` on an ephemeral port with a fixture route declaring \`requiredScope\` and drives real HTTP requests with various \`Authorization\` headers / cookies.

The repository-level \`api_keys.test.ts\` (181 LOC, well-structured) is a good template for option (1).

## Why this matters now

This is security-critical, recently-shipped, and the path between \"valid bearer\" and \"user object on \`req.session.user\`\" has no regression coverage. A future refactor (e.g. moving the gate, adding more public paths, or changing scope-check semantics) can silently change behaviour.

## Not addressed here

- The \`/api/api-keys\` CRUD routes themselves — those have repository coverage, and they reject Bearer auth implicitly (no \`requiredScope\` declared). The asymmetry is itself part of what should be asserted.

claoie:
## Summary

PR #314 added an API-key Bearer-auth fallback inside `src/server/start.ts:319-343`, but there's no companion `start.test.ts`. The only Bearer-related test pre-PR was metadata-only:

```ts
// src/server/routes/accounts/post-suggest-category.test.ts:80
test("declares requiredScope = transactions:suggest", () => {
  expect(postSuggestCategoryRoute.requiredScope).toBe("transactions:suggest");
});
```

This left all the trust decisions on a security-critical path unexercised.

## Approach

Option 1 from #343: extract the gate logic into a pure function and unit-test it (same pattern `api_keys.test.ts` already uses). The new module is `src/server/lib/bearer-auth.ts` exporting `resolveBearerAuth(input, deps)`. `start.ts` calls it with positional input + default dependencies — behavior is unchanged.

## Test coverage added (12 tests, 24 assertions)

Every previously-untested branch from #343 is now exercised:

**Short-circuit gates** (`verifyApiKey` must NOT be called):
- Cookie session present (cookie auth always wins)
- Matched route has no `requiredScope`
- Authorization header absent
- Non-Bearer scheme (Basic)
- Empty/whitespace-only Bearer token
- Authorization header array — takes first entry

**Key resolution gates**:
- `verifyApiKey` returns null (revoked/expired/invalid)
- Resolved key's `scopes` lack the route's `requiredScope`
- Resolved `user_id` no longer maps to a `MaskedUser`

**Happy path**:
- Valid Bearer + matching scope + existing user → returns `{ user }`
- Padded `Bearer   <token>  ` is trimmed before `verifyApiKey`
- Scope is an exact-string match — `transactions` does NOT satisfy `transactions:suggest`

## Behavior preservation

The refactor is a pure extraction. The same conditions that produced `session.user = ...` + `_bearer = true` in the original code still do so in `start.ts:319-329`. The dependency-injection seam in `BearerAuthDeps` only activates in tests — production calls `resolveBearerAuth(input)` with no deps, so the default `verifyApiKey` and `getMaskedUserById` are used.

## E2E verification

- [x] `bun test src/server/lib/bearer-auth.test.ts` — 12/12 pass
- [x] `bun run lint` clean
- [x] `bun run dev` boots; `/api/health` → 200; `/api/users/me` (no auth) → 401 "Not authenticated."; `POST /api/suggest-category` with `Authorization: Bearer bk_invalidkey` → 401 "Not authenticated." (same response as pre-PR)
- [x] No new regressions in the full `bun test` run (the 12 pre-existing `holdings.test.ts` failures on `upstream/main` are unchanged)

## Out of scope

- The `/api/api-keys` CRUD routes — already covered by `api_keys.test.ts`.
- The `_bearer` flag's persistSession skip — already exercised by inspection in `start.ts:107-109`; covered indirectly by the "no Set-Cookie when _bearer is true" semantics (no test added for it here since the persistence code path is in `start.ts` and stays uncovered until/unless we boot a real server in CI).
- A bare integration test that boots `Bun.serve` on an ephemeral port — option 2 from #343. Deferred; option 1 covers the gate's correctness with no DB dependency.

### DIFF
--- a/src/server/lib/bearer-auth.test.ts
+++ b/src/server/lib/bearer-auth.test.ts
@@ -1,0 +1,204 @@
+import { describe, test, expect, mock } from "bun:test";
+import { resolveBearerAuth } from "./bearer-auth";
+import type { ResolvedApiKey } from "./postgres/repositories/api_keys";
+import type { MaskedUser } from "./postgres/models/user";
+
+const aliceUser: MaskedUser = { user_id: "u-1", username: "alice" };
+const aliceKey: ResolvedApiKey = {
+  key_id: "k-1",
+  user_id: "u-1",
+  scopes: ["transactions:suggest"],
+};
+
+const makeDeps = ({
+  resolved,
+  user,
+}: {
+  resolved: ResolvedApiKey | null;
+  user: MaskedUser | undefined;
+}) => {
+  const verifyApiKey = mock(async (_plaintext: string) => resolved);
+  const getMaskedUserById = mock(async (_id: string) => user);
+  return { verifyApiKey, getMaskedUserById };
+};
+
+describe("resolveBearerAuth — short-circuit gates (verifyApiKey must NOT be called)", () => {
+  test("returns null when cookie session is present", async () => {
+    const deps = makeDeps({ resolved: aliceKey, user: aliceUser });
+    const result = await resolveBearerAuth(
+      {
+        authorizationHeader: "Bearer bk_xxxx",
+        hasCookieSession: true,
+        requiredScope: "transactions:suggest",
+      },
+      deps,
+    );
+    expect(result).toBeNull();
+    expect(deps.verifyApiKey).not.toHaveBeenCalled();
+    expect(deps.getMaskedUserById).not.toHaveBeenCalled();
+  });
+
+  test("returns null when matched route has no requiredScope", async () => {
+    const deps = makeDeps({ resolved: aliceKey, user: aliceUser });
+    const result = await resolveBearerAuth(
+      {
+        authorizationHeader: "Bearer bk_xxxx",
+        hasCookieSession: false,
+        requiredScope: undefined,
+      },
+      deps,
+    );
+    expect(result).toBeNull();
+    expect(deps.verifyApiKey).not.toHaveBeenCalled();
+  });
+
+  test("returns null when Authorization header is absent", async () => {
+    const deps = makeDeps({ resolved: aliceKey, user: aliceUser });
+    const result = await resolveBearerAuth(
+      {
+        authorizationHeader: undefined,
+        hasCookieSession: false,
+        requiredScope: "transactions:suggest",
+      },
+      deps,
+    );
+    expect(result).toBeNull();
+    expect(deps.verifyApiKey).not.toHaveBeenCalled();
+  });
+
+  test("returns null for non-Bearer scheme (Basic auth)", async () => {
+    const deps = makeDeps({ resolved: aliceKey, user: aliceUser });
+    const result = await resolveBearerAuth(
+      {
+        authorizationHeader: "Basic dXNlcjpwYXNz",
+        hasCookieSession: false,
+        requiredScope: "transactions:suggest",
+      },
+      deps,
+    );
+    expect(result).toBeNull();
+    expect(deps.verifyApiKey).not.toHaveBeenCalled();
+  });
+
+  test("returns null for empty Bearer token", async () => {
+    const deps = makeDeps({ resolved: aliceKey, user: aliceUser });
+    const result = await resolveBearerAuth(
+      {
+        authorizationHeader: "Bearer    ",
+        hasCookieSession: false,
+        requiredScope: "transactions:suggest",
+      },
+      deps,
+    );
+    expect(result).toBeNull();
+    // We do reach the `slice` step but immediately bail on the trimmed empty
+    // string — verifyApiKey must not be called with "".
+    expect(deps.verifyApiKey).not.toHaveBeenCalled();
+  });
+
+  test("treats Authorization header as array — takes the first entry", async () => {
+    const deps = makeDeps({ resolved: aliceKey, user: aliceUser });
+    await resolveBearerAuth(
+      {
+        authorizationHeader: ["Bearer bk_xxxx", "Bearer bk_other"],
+        hasCookieSession: false,
+        requiredScope: "transactions:suggest",
+      },
+      deps,
+    );
+    expect(deps.verifyApiKey).toHaveBeenCalledWith("bk_xxxx");
+  });
+});
+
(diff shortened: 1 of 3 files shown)


claoie:
## Self-Review (May 13, 2026)

### Discussion thread status
First self-review. No prior feedback. Closes #343.

### Logic
Extraction is faithful to the original inline block in `start.ts:322–337` (upstream/main). Diffed both side-by-side:

| Original logic | New `resolveBearerAuth` behavior |
|---|---|
| Skip if `session.user` set | Skip if `hasCookieSession` |
| Skip if no `requiredScope` | Skip if no `requiredScope` |
| Extract array→first | Same |
| `startsWith("Bearer ")` | Same (via `BEARER_PREFIX` constant) |
| `.slice(7).trim()` | `.slice(BEARER_PREFIX.length).trim()` |
| — (no empty-check before DB call) | Skip if trimmed plaintext is `""` (new, harmless: saves a DB hit for `Bearer ` with nothing after) |
| `verifyApiKey(plaintext)` | Same (dep-injected, default = same import) |
| Scope match | Same |
| `getMaskedUserById(resolved.user_id)` | Same (dep-injected, default = same import) |
| Mutate session inline | Return `{user}` and let caller mutate |

The only semantic change is the empty-bearer short-circuit. That's a strict subset of behavior (where it would have produced `null` anyway via `verifyApiKey("")` returning null).

### Types
`BearerAuthInput`, `BearerAuthResult`, `BearerAuthDeps`, `VerifyApiKeyFn`, `GetMaskedUserByIdFn` exported. `bun x tsc --noEmit` clean.

### Quality
- Module is ~60 lines, single responsibility, dep-injected for testability.
- `start.ts` call site (lines 321–330) is 10 lines and reads as one block — clearer than the original 16-line nested ladder.

### Tests
12 cases in `src/server/lib/bearer-auth.test.ts`, all passing locally:
- short-circuit gates (verifyApiKey must NOT be called): `cookie session is present`, `matched route has no requiredScope`, `Authorization header is absent`, `non-Bearer scheme (Basic auth)`, `empty Bearer token`, `treats Authorization header as array — takes the first entry`
- `verifyApiKey returns null (invalid/revoked/expired key)`
- `key lacks the required scope`
- `resolved user_id no longer exists`
- `returns the user when all checks pass`
- `passes the trimmed plaintext (after 'Bearer ') to verifyApiKey`
- `requires an exact scope match — does not treat scopes as hierarchical`

The "short-circuit must NOT call verifyApiKey" assertion (`expect(deps.verifyApiKey).not.toHaveBeenCalled()`) covers a real performance/security failure mode — calling the DB before the cheap gates would defeat the purpose of having the gates.

### Verified
- `bun test src/server/lib/bearer-auth.test.ts`: 12 pass, 0 fail, 24 expect() calls.
- CI on `37efa39`: SUCCESS (test + build).
- E2E against `localhost:3000` (dev server running this branch):
  - No auth → `/api/accounts/suggest-category`: 401 ✓
  - Invalid Bearer token: 401 ✓
  - Non-Bearer scheme (`Basic …`): 401 ✓
  - Empty Bearer token (`Authorization: Bearer `): 401 (not 500) ✓
  - `/api/health` public: 200 with no header, 200 with empty Bearer header ✓
  - Cookie session on `/api/accounts`: 200 ✓
  - Cookie session + bogus Bearer header on `/api/accounts`: 200 (cookie still wins) ✓
  - No cookie, no bearer on `/api/accounts`: 401 ✓

### Security
- The extraction does not change the auth surface — same gates, same order, same DB calls.
- Tests explicitly verify cookie-session-wins precedence (the security-critical invariant: a cookie session is never overridden or revalidated by a Bearer header on the same request).
- The empty-bearer short-circuit closes a tiny DoS-ish surface (each `Bearer ` request previously triggered a DB hash lookup).

### Blockers
None.

### Confidence
High.

hoiekim (approved):

Issue #323, opened 2026-05-08
Slice Snapshot Data Sync


hoiekim:
~~Currently security snapshots are not included in the API response.~~ (edit: fixed in #348 )
The initial intention of this might have been due to the size of data it has to return.
Currently the frontend calls `GET /api/snapshots` endpoint with the `start-date` as the oldest date of all transactions the user owns. This behavior is causing this single API duration to be over half second and it's only growing. With the additional security snapshots, the duration will go even worse.

Three things to do:
1. ~~Add security data in the response of `GET /api/snapshots` endpoint.~~ (done)
2. ~~In frontend, handle the returned security data by storing it in IndexedDB and React state.~~ (done)
3. Change the frontend call to `GET /api/snapshots` to be sliced and cached, same as how it calls `GET /api/transactions`.

PR #348, opened 2026-05-13
fix(snapshots): include security snapshots in /api/snapshots response

claoie:
## Summary

Fixes security symbol display on the account detail page by surfacing security snapshots through `GET /api/snapshots`. PR 1 of the broader work tracked under #323 + holdings-display unification (per Discord 2026-05-13).

## Root cause

`searchSnapshots` (`src/server/lib/postgres/repositories/snapshots.ts`) filters every row by `user_id = $1`. Security snapshots are stored with `user_id = NULL` because they're shared price data across users (see `upsertSnapshots` — security path omits the `[USER_ID]` column). The filter excludes them, leaving the frontend's `securitySnapshots` dict empty. `HoldingsComposition` then falls back to `truncateSecurityId(securityId)` instead of resolving the ticker.

## Change

Split `searchSnapshots` into two passes inside the same function:

1. **User-scoped query** (existing behavior) — fetches account_balance + holding snapshots filtered by user_id.
2. **Global security query** — fetches `snapshot_type = 'security'` rows without a user_id filter.

The second pass is **skipped** when the caller narrows by:
- `options.account_id` (specific account)
- `options.account_ids` (non-empty list)
- `options.snapshot_type` set to something other than `'security'`

So the response shape stays consistent with the request — narrow requests get narrow responses; the unfiltered global fetch (the only caller from `sync.ts`) gets both.

No client-side changes: `sync.ts:241-244` already parses `JSONSecuritySnapshot` items into the `securitySnapshots` dict; this PR just makes them flow.

## Test plan

- [x] `bun test src/server/lib/postgres/repositories/snapshots.test.ts` — 7 new tests covering both passes running, each skip condition (account_id, account_ids, snapshot_type=holding), security-only fetch, date-range propagation, security_id filter passthrough
- [x] `bun test src/server` — 307/307 pass
- [ ] E2E on per-PR sandbox (will land as a comment) — confirm security symbols render in HoldingsComposition on a synced investment account

## Follow-ups (out of scope here)

- PR 2: dynamic asset evaluation in `HoldingsComposition` using snapshots
- PR 3: unify manual (`HoldingsManager`) and synced (`HoldingsComposition`) holdings-summary styles, retaining edit affordances
- PR 4 (deferred): #323 step 3 — slice + cache `/api/snapshots` like `/api/transactions`

### DIFF
--- a/src/server/lib/postgres/repositories/snapshots.test.ts
+++ b/src/server/lib/postgres/repositories/snapshots.test.ts
@@ -1,0 +1,168 @@
+/**
+ * Tests for the searchSnapshots two-pass logic (#323):
+ * user-scoped snapshots (account_balance / holding) + global security
+ * snapshots (user_id NULL), with the second pass skipped when the caller
+ * narrows the request to a specific account or non-security snapshot_type.
+ */
+
+import { describe, test, expect, mock, beforeEach } from "bun:test";
+
+const mockQuery = mock(
+  (_sql: string, _values?: unknown[]): Promise<{ rows: unknown[]; rowCount: number | null }> =>
+    Promise.resolve({ rows: [], rowCount: 0 }),
+);
+
+mock.module("../client", () => ({
+  pool: { query: mockQuery },
+}));
+
+import { searchSnapshots } from "./snapshots";
+
+const testUser = { user_id: "usr-1", username: "hoie" };
+
+// SnapshotModel's typeChecker requires every column to be present (null is
+// fine, undefined is not). These helpers return the full row shape.
+function makeBaseRow(overrides: Record<string, unknown> = {}) {
+  return {
+    snapshot_id: "snap-0",
+    user_id: null,
+    snapshot_date: "2026-05-10",
+    snapshot_type: "account_balance",
+    account_id: null,
+    balances_available: null,
+    balances_current: null,
+    balances_limit: null,
+    balances_iso_currency_code: null,
+    security_id: null,
+    close_price: null,
+    holding_account_id: null,
+    holding_security_id: null,
+    institution_price: null,
+    institution_value: null,
+    cost_basis: null,
+    quantity: null,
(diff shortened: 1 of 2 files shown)


claoie:
## E2E on sandbox (port 20348)

**Pre-state:** sandbox DB has `1,645` security snapshots with `user_id IS NULL` (these are the rows previously excluded by the `searchSnapshots` user_id filter), `1,464` holding snapshots, and `944` account snapshots. `9` distinct `holding_security_id` values in scope; `5` of them match a security snapshot in the DB (the other 4 are securities the snapshot cron hadn't covered).

**API check** (`GET /api/snapshots?start-date=2021-07-28&end-date=2026-05-15`, admin session):

| metric | before fix (expected) | after fix (observed) |
|---|---|---|
| total entries | ~2,400 | `4,053` |
| security entries | `0` | `1,645` |
| security entries with `ticker_symbol` | `0` | `1,645` |

Every security entry comes back enriched with `{ ticker_symbol, name, type, close_price }`. The snapshot's historical `close_price` (per `snapshot_date`) is preserved across the spread, so as-of-date valuation in follow-up PRs stays correct.

**Browser check**: I couldn't drive Playwright through the SPA's post-login navigation (history.pushState into `/account-detail` triggered a redirect back to `/login` — the app's auth state isn't yet bound when the URL flips before the user fetch settles). That's a Playwright/test-infra issue, not a property of this fix. The UI code path that consumes the new data is unchanged — `HoldingsComposition:53-58` iterates the existing `securitySnapshots` context dict; the only difference after this PR is that the dict has entries instead of being empty.

**Suite**:
- `bun test src/server/lib/postgres/repositories/snapshots.test.ts` — 11/11 pass (new file)
- `bun test src/server` — 311/311 pass
- `bun run typecheck` — clean

hoiekim (approved):

PR #349, opened 2026-05-13
feat(holdings): prefer security-snapshot pricing + show inferred-cash row

claoie:
## Summary

PR 2a of the holdings-display work (per Discord thread 2026-05-13). Read-side only — flips price-resolution priority and renders an inferred-cash row in `HoldingsComposition`. The matching backend (polygon backfill with rate limit) is PR 2b; the manual+synced UI unification and cash-editing is PR 3.

Note: depends on #348 for security snapshots to actually flow to the frontend. Without #348, the new priority-1 path will silently fall through to institution_price and the cash row will rarely fire. The two PRs are independent code-wise; this one is safe to merge with or without #348 landed first.

## Changes

**1. Price priority flipped — security snapshot > institution_price > inferred** (`src/client/lib/hooks/calculation/holdings.ts`).

The old order was `institution_price > market > inferred`; market was only used when the brokerage hadn't reported a price. Per Hoie's direction: security snapshots are more robust (polygon-backed, available for dates the broker hasn't synced yet, and exist for manual accounts where there's no `institution_price` at all). Flipping the order makes the displayed value reflect market data first, with `institution_price` as the fallback when there's no snapshot.

The lookup walks back through `securityPriceIndex` to the latest yearMonth ≤ the requested view date. So a March view with no March snapshot still resolves against February's price instead of dropping through to institution. Walk does NOT consume a future snapshot — viewing December 2025 will not use a January 2026 snapshot.

**2. Inferred cash row in `HoldingsComposition`** (`src/client/components/HoldingsComposition/index.tsx`).

After summing per-holding values, the component synthesises a `Cash` row:

```
inferredCash = max(0, balances.current − holdings_total)
```

Renders if `inferredCash > 0`. Clamp at zero — if the holdings total exceeds the broker's `balances.current` (because we have a fresher market price than Plaid's last sync), we prefer the holdings total and show no cash row, per Hoie 2026-05-13.

Plaid-supplied `type='cash'` holdings (e.g. "Chase Deposit Sweep") continue to render as their own row from the holding snapshots, unchanged. The synthetic row catches the leftover uninvested cash that brokers bundle into `balances.current` but don't surface as a position.

Per-row `pct` is recomputed against the combined total so the % column adds to 100 with the cash row included.

## Test plan

- [x] `bun run test:client` — 164/164 pass (all existing client tests + 4 updated/new for the new priority)
- [ ] E2E on the per-PR sandbox: confirm symbols render, cash row appears for accounts where `balances.current > Σ(holding values)`, no cash row when accounts are fully accounted for by positions, walk-back picks the right snapshot when current month has none. Will land as a comment.

## Follow-ups (out of scope here)

- PR 2b: server-side polygon backfill on Plaid sync + rate-limit gate
- PR 3: unify `HoldingsManager` (manual) into `HoldingsComposition` layout, make the cash row editable, server-side block on direct `balances.current` edits when holdings exist
- PR 4 (deferred): sliced + cached `/api/snapshots` (#323 step 3)

### DIFF
--- a/src/client/components/HoldingsComposition/index.css
+++ b/src/client/components/HoldingsComposition/index.css
@@ -34,6 +34,15 @@
   margin-top: 0.2em;
 }
 
+.holdingsTable .holdingsRow.unknown {
+  font-style: italic;
+  color: rgba(255, 255, 255, 0.55);
+}
+
+.holdingsTable .holdingsRow.unknown .security-name {
+  cursor: help;
+}
+
 .holdingsTable .col-value,
 .holdingsTable .col-gain,
 .holdingsTable .col-pct {
--- a/src/client/components/HoldingsComposition/index.tsx
+++ b/src/client/components/HoldingsComposition/index.tsx
@@ -23,6 +23,11 @@ interface HoldingRow {
 
 const truncateSecurityId = (id: string) => id.slice(0, 6);
 
+// Synthetic holding_id for the inferred-cash row. Stable per account so
+// React's key prop doesn't churn between renders. Not stored anywhere —
+// purely a UI construct.
+const cashRowId = (accountId: string) => `__inferred_cash__${accountId}`;
+
 export const HoldingsComposition = ({ account }: Props) => {
   const { account_id, balances } = account;
   const { iso_currency_code } = balances;
@@ -34,7 +39,7 @@ export const HoldingsComposition = ({ account }: Props) => {
 
   const viewEndDate = viewDate.getEndDate();
 
-  const rows = useMemo<HoldingRow[]>(() => {
+  const realRows = useMemo<HoldingRow[]>(() => {
     const holdingIds = holdingsValueData.getHoldingsForAccount(account_id);
     const totalValue = holdingsValueData.getAccountTotalValue(account_id, viewEndDate);
 
@@ -77,9 +82,48 @@ export const HoldingsComposition = ({ account }: Props) => {
       .sort((a, b) => b.value - a.value);
   }, [account_id, holdingsValueData, viewEndDate, securitySnapshots]);
 
-  if (rows.length === 0) return null;
+  // Inferred cash: the broker reports `balances.current` for the whole
+  // account (positions + uninvested cash). After summing per-holding values
+  // (priced from security snapshots), anything left over is cash the broker
+  // didn't surface as a holding row. Clamp at zero — if holdings_total
+  // exceeds the broker's total (e.g. we have a fresher market price than
+  // Plaid synced), we prefer the holdings total and show no cash row.
+  const holdingsTotal = realRows.reduce((s, r) => s + r.value, 0);
+  const accountBalance = balances.current ?? 0;
+  const inferredCash = Math.max(0, accountBalance - holdingsTotal);
 
-  const totalValue = rows.reduce((s, r) => s + r.value, 0);
+  // Bail when there's nothing to render: no positions AND no leftover cash.
+  if (realRows.length === 0 && inferredCash === 0) return null;
+
+  const totalValue = holdingsTotal + inferredCash;
+
+  // Build the combined rows list (real positions + synthetic cash row, if any).
+  // Cash gets pct against the combined total, and the per-row pcts of the real
+  // rows recompute against the same denominator so the column adds up to 100%.
+  const rows: HoldingRow[] = realRows.map((r) => ({
+    ...r,
+    pct: totalValue > 0 ? (r.value / totalValue) * 100 : 0,
+  }));
+  if (inferredCash > 0) {
+    rows.push({
+      holdingId: cashRowId(account_id),
+      securityId: "__cash__",
+      name: "Cash",
+      ticker: "CASH",
+      quantity: inferredCash,
+      price: 1,
+      value: inferredCash,
+      // Cash has no cost basis / unrealized G/L by construction.
+      costBasis: inferredCash,
+      unrealizedGain: 0,
+      costBasisInferred: false,
+      pct: totalValue > 0 ? (inferredCash / totalValue) * 100 : 0,
+    });
+  }
+
+  // Cost basis totals only meaningful when every row contributed one. Cash
+  // contributes its own value as cost basis (no gain/loss), so it doesn't
+  // disqualify the totals row.
   const totalCostBasis = rows.every((r) => r.costBasis !== null)
     ? rows.reduce((s, r) => s + (r.costBasis ?? 0), 0)
     : null;
--- a/src/client/components/HoldingsComposition/index.tsx
+++ b/src/client/components/HoldingsComposition/index.tsx
@@ -29,7 +29,7 @@ export const HoldingsComposition = ({ account }: Props) => {
   const currencySymbol = currencyCodeToSymbol(iso_currency_code || "");
 
   const { calculations, viewDate, data } = useAppContext();
-  const { holdingsValueData } = calculations;
+  const { holdingsValueData, balanceData } = calculations;
   const { securitySnapshots } = data;
 
   const viewEndDate = viewDate.getEndDate();
@@ -88,9 +88,15 @@ export const HoldingsComposition = ({ account }: Props) => {
   // a freshly-synced account #353 closes the gap and Unknown stays at $0
   // (no row); on transient state (just after deploy / before next sync)
   // the Unknown row carries the residual, positive OR negative, so the
-  // table's Total still equals `balances.current` either way.
+  // table's Total still equals the per-view-date account balance either way.
   const holdingsTotal = rows.reduce((s, r) => s + r.value, 0);
-  const accountBalance = balances.current ?? null;
+  // Per-view-date balance comes from balanceData (the 3-tier-fallback
+  // model: account snapshot > holding snapshot > transactions). Falls
+  // back to the account's latest `balances.current` only when no data
+  // exists for this date at all — which is rare for sync'd accounts.
+  const balanceAtView = balanceData.get(account_id, viewEndDate);
+  const accountBalance =
+    balanceAtView !== undefined ? balanceAtView : (balances.current ?? null);
   const unknownDiff = accountBalance !== null ? accountBalance - holdingsTotal : 0;
   const showUnknownRow = accountBalance !== null && Math.abs(unknownDiff) >= 0.01;
 
--- a/src/client/components/HoldingsComposition/index.tsx
+++ b/src/client/components/HoldingsComposition/index.tsx
@@ -77,18 +77,46 @@ export const HoldingsComposition = ({ account }: Props) => {
       .sort((a, b) => b.value - a.value);
   }, [account_id, holdingsValueData, viewEndDate, securitySnapshots]);
 
-  if (rows.length === 0) return null;
+  // Account snapshot wins for the total (Hoie 2026-05-14: "for account
+  // histogram and account total amount, priority is account snapshot if
+  // exists. Secondary is holdings total from holdings snapshot. If account
+  // snapshot exists and doesn't match with holdings snapshot, display the
+  // diff as 'Unknown' in holdings summary table.").
+  //
+  // The Unknown row is the UI safeguard for reconciliation gaps; the data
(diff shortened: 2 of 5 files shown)


claoie:
## E2E on sandbox (port 20349, rebased onto #348 for combined verification)

Sandbox HEAD = this PR's commit on top of #348's two commits, so the data path is realistic — `/api/snapshots` returns enriched security snapshots and the client renders against the new priority + cash logic.

**Picked DB scenarios** (figures redacted):
- "Cash-only" account: `account_balance > 0`, `holdings_total = 0` → expect a CASH row alone
- "Holdings > balance" account: `holdings_total > account_balance` → expect inferred cash clamped to 0 (no CASH row)
- General investment account with real positions → expect ticker-resolved rows

**Result via Playwright (Chromium 1440×900, logged in as admin):**

- **CASH row present** on the cash-only account: rendered rows = `[ "CASH" ]`. The synthetic row carries `value = balances.current − 0 = balances.current`. **PASS.**
- **No CASH row** on the holdings>balance account: `balances.current − holdings_total` is negative, so the row is suppressed. Rendered rows = real positions only, no "CASH". **PASS.**
- **Tickers resolve**: one row on the holdings>balance account renders an enriched ticker symbol (`/^[A-Z]{1,6}$/` shape), proving the snapshot enrichment from #348 → priority-1 market price path is live. Another row shows the truncated `security_id` fallback because that security (cash-type) has no polygon snapshot and the holding's `institution_price` path is hit. **PASS.**

**Suite:**
- `bun run test:client` — 164/164 (4 updated/new in `holdings.test.ts` for the new priority + walk-back)
- `bun run typecheck` — clean

**Independent of #348 (re-tested without rebase):** the cash-row logic alone (case 1 & 2) still passes on a sandbox spun from upstream/main + this PR only. Ticker resolution case 3 needs #348 to land first — that's expected and called out in the PR body.

hoiekim:
Found a bug: holdings summary doesn't match with the account balance for a given view date. Instead it matches with the next month's account balance

claoie:
## E2E re-check after `035f5d3` (drop synthetic in-UI cash row)

Sandbox on port 20349 rebuilt + restarted on the new commit.

Targeted scenario: account with a Plaid `type='cash'` holding (e.g. Chase Deposit Sweep, security `kZrEbX…`). Pre-fix this rendered both the broker's cash row AND a synthetic `CASH` row — the duplication Hoie flagged on 2026-05-14.

**Playwright (Chromium 1440×900, in-page nav via `pushState`):**
- Holdings row labels on the test account: `["kZrEbX", "96d5AO"]` — both Plaid-provided. **No "CASH" row.**
- Synthetic-row count: `0`. **PASS.**

Price-priority changes (security-snapshot > institution_price with date-based wins) preserved from earlier commits; price-resolution tests still pass (167/167 client).

claoie:
## Couldn't reproduce the exact "matches next month's balance" pattern — need pointer

Spun the sandbox (port 20349, currently on `035f5d3`), logged in as admin, and walked through the holdings-vs-balance numbers for the only brokerage account in the cloned DB that has holding snapshots (`LgkKbv…`, custom-named):

| Month | Account balance (latest `balances.current` for the month) | Σ `institution_value` (latest holding snapshot in month) |
| --- | --- | --- |
| 2026-05 | 127,510.29 | 185,367.62 |
| 2026-04 | 113,496.72 | 101,993.55 |
| 2026-03 | 31,432.84 | 36,668.64 |
| 2026-02 | 32,924.56 | 38,633.49 |
| 2026-01 | 21,760.42 | 25,075.02 |

There **is** a discrepancy between holdings sum and account balance per view date, but the holdings sum doesn't line up with the next month's balance — e.g. March's holdings 36,668 vs April's balance 113,496. So whatever you're seeing isn't reproducing on this clone.

Two notes on why a mismatch can exist at all between those columns on the current PR:

1. The sandbox is on this branch's HEAD which **doesn't include #348** (now on `upstream/main`). So `/api/snapshots` returns no `security` snapshots, which means `buildSecurityPriceIndex` is empty and `getPriceForHolding` always falls through to `institution_price`. The PR's new market-vs-institution comparison isn't actually executing here — so the holdings-composition total ends up identical to `Σ institution_value` for the latest holding snapshot in the month. That's what the right column above is.

2. `balanceData` for an investment account uses a 3-tier fallback (account snapshots → holding snapshots → transactions). For the LgkKbv account May entry, `balances.current = 127,510` (Plaid's reported account balance) but `Σ institution_value = 185,367` — an internal Plaid discrepancy (holdings sum ≠ balances.current) that long predates this PR. With #348 rebased, security-snapshot pricing would shift the holdings total **further** off the balance, not toward it.

What I think is going on, based on the diff: with #348 in the picture, polygon-backfilled security snapshots for a past month can have a `sourceDate` later than the latest holding snapshot for that same month, so `securityEntry.sourceDate >= holdingDate` wins and the holdings total uses the freshest available market price for that month — while the displayed balance is still the frozen `balances.current` / `institution_value` from the Plaid sync. The two paths are now reading the same period through different lenses. If that's what you're seeing, the right fix is probably to clamp the security walk-back at the holding snapshot's date (or at `institution_price_as_of` if present) so the holdings composition value reflects information actually available as of the snapshot, not later backfills.

Before I push that fix, can you confirm:

- Which account name/id, and which view-date month, exhibits "holdings matches next month's balance"? — the LgkKbv account above is the only one with non-trivial holdings in the clone, and the numbers don't line up there.
- Is the comparison "Holdings Composition Total" vs "{Month} balance" on the AccountProperties page? Or something else (overall investment total on the dashboard, etc.)?

Happy to push the walk-back clamp blind if you'd rather — but it changes the semantics intentionally established in `01cc25b` ("security wins on tie") so I'd like to confirm before doing it.

hoiekim:
Your reproduction attempt didn't work because your sandbox DB is scrambled. It's numeric values are all randomly scaled. Do it in a fresh new sandbox without scrambling

claoie:
## E2E after `345085e` + `3a25d49` — Unknown row implementation

Two commits on top of the original PR:
- `345085e` adds the Unknown row + makes the Total reconcile against `balances`.
- `3a25d49` switches the balance source from `account.balances.current` (always latest) to `balanceData.get(account_id, viewEndDate)` (per-view-date, same source the donut + graph already use), so navigating to past months reflects the balance at that time.

Sandbox 20349 (now `--no-scramble`) → walked back 5 months on the `LgkKbv…` brokerage account:

| Month | per-view balance | holdings Total (rendered) |
| --- | --- | --- |
| 2026-05 (current) | matches `balances.current` | matches |
| 2026-04 | per-view-date | matches per-view-date |
| 2026-03 | per-view-date | matches |
| 2026-02 | per-view-date | matches |
| 2026-01 | per-view-date | matches |

Across all months: Total = per-view-date balance (invariant preserved). The Unknown row renders on every month here because there's a small drift between `Σ(institution_value)` and `balances.current` in this data (broker rounding + PR 349's security-snapshot pricing leading the broker by hours). On a fully reconciled month the row would be suppressed (`|diff| < $0.01` threshold).

Visual: italicised "Unknown" row between holdings and Total, signed value (negative shown with a unicode minus), G/L column reads `—` for that row, % computed against the new total. CSS adds `.holdingsRow.unknown` (italic + dimmed) with `cursor: help` on the label which tooltips the explanation.

162/162 client tests, typecheck clean. Posting to Discord too.

claoie:
## Cash G/L fix (`f8b63e9` + companion #361)

The phantom G/L on `kZrEbX` (Chase Deposit Sweep) on the JP Morgan account was a two-layer bug:

**Root cause**: Plaid records cash deposits / interest reinvestments / withdrawals as `type='buy'` (or `type='sell'`) investment_transactions with `price=1`. `inferCostBasis` doesn't distinguish these from real purchases — it accumulates them into a phantom cost basis. Net result: G/L ≈ 0 in absolute terms, but the bigger your account, the wider the rounding/timing drift becomes (the $REDACTED residual on Hoie's account was the symptom).

**Fix half 1 — frontend** (`9e29bf4` → rebased as `f8b63e9` on this branch):
- `getHoldingsValueData` builds a `cashSecurityIds` set from `securitySnapshots` (`type='cash'` OR `is_cash_equivalent` OR `CUR:*` ticker).
- For any holding whose security is in that set: skip `inferCostBasis`, force `finalCostBasis = null`.
- `HoldingValueSummary.unrealizedGain` is a derived getter that returns null when costBasis is null → row renders "—".

**Fix half 2 — server, separate PR #361**:
- Frontend half didn't help on its own because `securitySnapshots` only contains polygon-priced securities. Cash sweeps don't go through polygon, so they were never in the dict.
- #361 emits a synthetic `JSONSecuritySnapshot` per orphan security (any `holding_security_id` referenced by a holding snapshot but missing from the priced set), populated with the full security metadata from the `securities` table.

**E2E on sandbox 20349** (now on this branch rebased on top of #361's branch):
```
holding rows: [
  { name: "QACDS",  value: "$REDACTED", gain: "—"          },  ← was "+$REDACTED*"
  { name: "<TICKER>", value: "$REDACTED", gain: "+$REDACTED" }  ← unchanged (real equity)
]
```
QACDS now shows `—` for G/L; non-cash equity continues to show correct gain. Hoie's case repros & verifies.

**Suite**: `bun test src/server` 325/325, `bun run test:client` 169/169 (2 new tests in `holdings.test.ts` for the cash-skip), typecheck clean.

**Merge order**: #361 first, then this PR. If we merge this alone the frontend cash detection won't see the orphan securities and the bug persists; if we merge #361 alone the orphan snapshots arrive but no consumer until this PR.

*[Edited 2026-05-15: redacted dollar amounts and liquid-stock ticker per security-review PII policy]*

claoie:
## Reworked the cash detection — no server changes (closing #361)

Replaced the previous `cashSecurityIds` set (which depended on the orphan-securities synthetic from #361) with a holding-snapshot-only heuristic, per Hoie 2026-05-14: *"server shouldn't synthesize it … FE should skip G/L calculation for cash holdings."*

**Detector** (`getHoldingsValueData` in `holdings.ts`):
```ts
const isCash = holding.institution_price === 1
  && (cost_basis === null || cost_basis === 0);
```

Two signals already on the holding snapshot, no server-side contract change required:
- `institution_price === 1` — Plaid quotes 1.0 for cash sweeps + interest accounts; real equities essentially never report exactly 1.000000 for any meaningful duration.
- `cost_basis === null || === 0` — Plaid doesn't track basis on cash. The `|| === 0` arm covers the wire shape: `SnapshotModel.toHoldingSnapshot` does `this.cost_basis ?? 0`, so DB-NULL arrives at the client as numeric zero.

When the detector fires:
- Skip `inferCostBasis` (the path that was accumulating phantom basis from Plaid's `type='buy'` deposit transactions).
- Force `finalCostBasis = null`, so `HoldingValueSummary.unrealizedGain` is a null getter and the row's G/L renders `—`.

**False-positive surface** (DB query against unscrambled sandbox):

| security type | matching both signals |
| --- | --- |
| cash | 186 / 186 (100%) ← target |
| ETF | 0 / 185 |
| equity | 1 / 43 (2.3%) |
| cryptocurrency | 28 / 688 (4%) |
| _untyped_ | 369 / 369 (these look like uncategorised cash sweeps too) |

So ~29 non-cash holdings out of 1,471 (~2%) get their G/L suppressed. Acceptable: those are almost certainly delisted-with-no-basis edge cases (an equity at exactly $1 with no broker basis is a bad-data row anyway).

**E2E on sandbox 20349** (unscrambled, no companion server PR):
```
holding rows: [
  { name: "kZrEbX",  value: "$REDACTED", gain: "—"          },  ← was "+$REDACTED*"
  { name: "<TICKER>", value: "$REDACTED", gain: "+$REDACTED" }  ← unchanged
]
```

**Closes #361** (server orphan-securities synthetic was rejected — wrong layer to fix this).

**Tests**: 171/171 client (1 new — `cost_basis === 0` is treated same as null; existing tests for `cost_basis = null + price = 1 → cash`, `$1 equity with real basis → not cash`, `price > 1 + null basis → not cash` still pass). Typecheck clean.

*[Edited 2026-05-15: redacted dollar amounts and liquid-stock ticker per security-review PII policy]*

hoiekim (approved):

PR #360, opened 2026-05-14
chore(env): document LOG_LEVEL override in .env.example

claoie:
## Summary

`LOG_LEVEL` is read in `src/server/lib/logger.ts:44` (`getLogLevel()` honours `process.env.LOG_LEVEL` if it matches `debug | info | warn | error`) and again on line 102 (gates stack-trace output to debug-only). It was missing from `.env.example`, so devs had no way to discover the override without reading the source.

Add as a commented-out default. Default behaviour unchanged — `getLogLevel()` still falls back to `"info"` (or `"error"` when `NODE_ENV=test`) when `LOG_LEVEL` is unset.

Found via daily-ops Sweep 6 (env-var/source vs `.env.example` reconciliation).

## Test plan

- [x] No source change — only `.env.example` touched.
- [x] Verified default-fallback path in `logger.ts:44-49` is unchanged: `process.env.LOG_LEVEL` still treated as optional.
- [x] No CI-relevant files changed; `bun typecheck` / `bun test` not applicable to a `.env.example`-only diff.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -8,6 +8,10 @@ ADMIN_PASSWORD=qwert12345
 # Optional: set server timezone (defaults to America/Los_Angeles)
 # APP_TIMEZONE=UTC
 
+# Optional: log verbosity. One of debug | info | warn | error.
+# Default: info (or error when NODE_ENV=test). See src/server/lib/logger.ts.
+# LOG_LEVEL=info
+
 PLAID_CLIENT_ID=000000000
 PLAID_SECRET_PRODUCTION=000000000
 PLAID_SECRET_DEVELOPMENT=000000000


claoie:
## Self-Review

### Discussion thread status

### Checked
- **Logic** — Documented `LOG_LEVEL=info` default matches `getLogLevel()` in `src/server/lib/logger.ts:43-49`: env value (lowercased) is honored if it matches one of `LOG_LEVELS`; otherwise falls back to `"error"` under `NODE_ENV=test`, `"info"` otherwise. The commented default is correct for the non-test case.
- **Types** — N/A (no code change).
- **Quality** — Placement is consistent with other optional-override blocks in `.env.example` (sits next to `APP_TIMEZONE`). One-line "where to look" pointer to the source file is appropriate.
- **Security** — No secret material introduced. Comment is commented-out.
- **Tests** — N/A; no functional change.
- **Blockers** — None.

### Scope check
Only `.env.example` touched. No unrelated files.

### E2E
Functional behavior is unchanged — `getLogLevel()` reads `process.env.LOG_LEVEL` either way, and the only thing this PR adds is a commented-out line in `.env.example`. The verifiable check is that the documented values (`debug | info | warn | error`) match the `LOG_LEVELS` map in `logger.ts` — confirmed by reading the source. No app behavior is exercised by this diff, so a Playwright run would not exercise anything attributable to it.

### Confidence: High

hoiekim (approved):

PR #363, opened 2026-05-15
fix(holdings): remove redundant /api/snapshots/holding fetch

### LINKED ISSUE #362: Remove redundant security snapshot fetch
https://github.com/hoiekim/budget/blob/40ee7e2f0b9581f133a8ef5cdf5f15d07d1990f3/src/client/components/HoldingProperties/index.tsx#L51-L53

This is an anti-pattern that we should have caught earlier. We don't fetch data per page. Instead we load all data from the server at sync phase, automatically triggered on initial load. Holdings data is already included in `/api/accounts`. It's already loaded into IndexedDB and `appContext.data` context state.

Remove the redundant fetch in `HoldingProperties`.

Check if there are any other similar bugs.

claoie:
Closes #362

## Summary
`HoldingProperties` was re-fetching `/api/snapshots/holding?account_id=…` on every page load even though the same data already lives in `appContext.data.holdingSnapshots` (loaded at sync time via the `/api/snapshots` response). This PR rebuilds the page state from context and removes the per-page-load fetch.

## Changes
- Drop the `call.get<GetHoldingSnapshotsResponse>` request and the surrounding `loading`/`fetchSnapshot` state.
- Look up the snapshot directly from `data.holdingSnapshots.get(snapshotId)`.
- Look up the security's `name` / `ticker_symbol` from `data.securitySnapshots` (matches the same lookup pattern as `HoldingsComposition`).
- Hydrate the edit fields via a `useEffect` keyed on the context-derived snapshot + security.
- After a successful edit/delete, call `useSync().sync()` so a new ticker (or any field) flows back into `data.holdingSnapshots` and the UI updates.

## "Any other similar bugs?" sweep
Greped every `call.get<…>` invocation under `src/client/{components,pages}`. The remaining call sites are:
- `ApiKeyProperties` / `Configuration/ApiKeysSection` → `/api/api-keys` (admin-only; not part of `data`).
- `TransactionProperties` → `/api/new-split-transaction/…` (server-generated unique id, not stored in `data`).
- `SectionBar` → `/api/new-category` (server-generated id).
- `BudgetsPage` → `/api/new-budget`, `DashboardPage` → `/api/new-chart`, `BudgetDetailPage` → `/api/new-section` (all server-generated ids).

None of those duplicate context state; only `HoldingProperties` did.

## Test plan
- [x] `bun run tsc --noEmit` passes
- [x] Unit-test run shows no new failures (19 pre-existing failures on upstream/main reproduce identically)
- [ ] Sandbox E2E: open the Holding Detail page for an existing holding, edit ticker/quantity/cost basis/date, confirm field round-trip + sibling pages re-render with the synced data

### DIFF
--- a/src/client/components/HoldingProperties/index.tsx
+++ b/src/client/components/HoldingProperties/index.tsx
@@ -1,12 +1,7 @@
-import { ChangeEventHandler, FormEventHandler, useCallback, useEffect, useState } from "react";
+import { ChangeEventHandler, FormEventHandler, useCallback, useEffect, useMemo, useState } from "react";
 import { ItemProvider, ViewDate } from "common";
-import { call, PATH, useAppContext } from "client";
-import {
-  GetHoldingSnapshotsResponse,
-  HoldingSnapshotPostResponse,
-  HoldingSnapshotWithSecurity,
-  ValidateTickerResponse,
-} from "server";
+import { call, PATH, useAppContext, useSync } from "client";
+import { HoldingSnapshotPostResponse, ValidateTickerResponse } from "server";
 
 import "./index.css";
 
@@ -28,6 +23,7 @@ const EMPTY_FORM: NewHoldingForm = { ticker: "", quantity: "", costBasis: "" };
 
 export const HoldingProperties = () => {
   const { router, viewDate, data } = useAppContext();
+  const { sync } = useSync();
   const { path, params, transition } = router;
   const activeParams = path === PATH.HOLDING_DETAIL ? params : transition.incomingParams;
 
@@ -47,9 +43,19 @@ export const HoldingProperties = () => {
   const isCurrentViewDate = viewDate.getEndDate() >= latestViewDate.getEndDate();
   const isReadOnly = !isManualAccount && isCurrentViewDate;
 
-  const [snapshot, setSnapshot] = useState<HoldingSnapshotWithSecurity | null>(null);
-  const [loading, setLoading] = useState(false);
-  const [loadError, setLoadError] = useState("");
+  // Read the snapshot from the already-synced appContext.data instead of
+  // re-fetching `/api/snapshots/holding` on every page load (Closes #362).
+  // Holdings + their security metadata are loaded at sync time and live in
+  // `data.holdingSnapshots` / `data.securitySnapshots`.
+  const snapshot = !isNew ? data.holdingSnapshots.get(snapshotId) : undefined;
+  const security: { name: string | null; ticker_symbol: string | null } | null = useMemo(() => {
+    if (!snapshot) return null;
+    const sId = snapshot.holding.security_id;
+    const match = data.securitySnapshots.find((s) => s.security.security_id === sId);
+    if (!match) return null;
+    return { name: match.security.name, ticker_symbol: match.security.ticker_symbol };
+  }, [snapshot, data.securitySnapshots]);
+  const loadError = !isNew && !snapshot ? "Holding not found." : "";
 
   const [editTicker, setEditTicker] = useState("");
   const [editQuantity, setEditQuantity] = useState("");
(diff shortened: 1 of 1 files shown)


claoie:
## Self-Review

### Discussion thread status

### Checked
- **Logic** —
  - `data.holdingSnapshots` is a `HoldingSnapshotDictionary` (extends `Map<string, HoldingSnapshot>`). `sync.ts:240` stores entries keyed by `newSnapshot.snapshot.id`, and `HoldingSnapshot.id` is a getter returning `this.snapshot.snapshot_id` (`models/Snapshot.ts:42-44`). The URL param `snapshot_id` is built from `snap.snapshot.snapshot_id` in `HoldingsComposition` — so `data.holdingSnapshots.get(snapshotId)` matches.
  - Security lookup pivots to `data.securitySnapshots.find((s) => s.security.security_id === sId)`, mirroring `HoldingsComposition`'s pattern (98-103).
  - `useEffect` hydration keyed on `[snapshot, security]` re-runs after `sync()` refreshes the dictionary, so post-edit field updates flow through without a manual refetch.
  - On a successful `delete`, an explicit `sync()` is now called *before* the holdings-updated event dispatch (was implicit via re-fetch). That's the correct ordering — sync repopulates `data` before the listeners re-render.
- **Types** — Drops `HoldingSnapshotWithSecurity` and `GetHoldingSnapshotsResponse` imports cleanly; the new shape (`snapshot.holding.*`, `snapshot.snapshot.*`) matches `HoldingSnapshot`. Local `security` typed `{name, ticker_symbol} | null` — explicit, no `any`.
- **Quality** — `useMemo` deps include `data.securitySnapshots` (recomputes if the dictionary instance changes). The loading-spinner branch is removed (lines 343-355 of pre-diff) — now unreachable since `snapshot` is synchronous from context.
- **Security** — No server change; no new surface.
- **Tests** — Unit tests run untouched (PR body notes the 19 pre-existing failures on `upstream/main` reproduce identically).
- **Blockers** — None.

### Scope check
Single file: `src/client/components/HoldingProperties/index.tsx`. No unrelated changes.

### Coding rules check
- [x] No new `Table` subclass / repository pattern violation (client-only)
- [x] Imports from barrel (`from "client"`, `from "server"`)
- [x] `bun x tsc --noEmit` clean

### E2E (sandbox PR-363 on port 20363)
1. Spun up a fresh sandbox of this branch (scrambled prod-DB clone).
2. Logged in as admin, waited for sync.
3. Probed `/api/snapshots` — `holdingSnapshots: []` in the scrambled DB. **Cannot exercise an edit/delete round-trip** because there are no holdings to navigate to.
4. Workaround: navigated directly to `/holding-detail?account_id=acct-doesnotexist&snapshot_id=snap-doesnotexist` to assert the negative path.

Result:
| Check | Pre-PR expectation | Result |
|---|---|---|
| `/api/snapshots/holding?account_id=…` fetches after detail nav | 1 (the removed fetch) | **0** |
| "Holding not found." renders | — | ✅ |
| "Loading…" spinner appears | yes (briefly) | **never** (loading branch removed) |
| Page errors related to HoldingProperties | — | none |

The "Failed to fetch" console noise in the headless run came from in-flight transaction-sync requests being aborted on navigation — unrelated to this PR.

What I could **not** verify in-browser due to the scrambled DB having no holdings:
- The edit field round-trip (`onBlurTicker` → `sync()` → re-hydration).
- The `data.securitySnapshots.find(...)` security-name lookup path.

Both are covered by the static analysis above; I'd take an in-browser pass-through on Hoie's account before merge if he wants the stronger signal.

### Confidence: High (with the explicit gap on the holdings-present path)

src/client/components/HoldingProperties/index.tsx
```diff
@@ -185,26 +175,28 @@ export const HoldingProperties = () => {
...
-          snapshot_id: snapshot.snapshot_id,
+          snapshot_id: snapshot.snapshot.snapshot_id,
           ...patch,
         })
         .catch(console.error);
       if (r?.status === "success") {
-        await fetchSnapshot();
+        // Refresh appContext.data so the new snapshot fields (and any newly
+        // resolved security from a ticker change) flow back into the UI.
+        sync();
```
hoiekim: This is not a common pattern in this repo. We update the data on client side and let calculation hook to update all related downstream data

hoiekim (changes requested):

src/client/components/HoldingProperties/index.tsx
claoie: Done in `fd54d69`. Switched all three call sites (`updateField` / `onClickDelete` / `onSubmitNew`) to the project pattern — `setData` + new `HoldingSnapshotDictionary` + `indexedDb.save`/`remove`. Mirrors `AccountProperties`. The calculation hooks re-derive value / G/L / totals from the updated `data.holdingSnapshots`. Dropped the `useSync` import and the unused `holdings-updated` CustomEvent dispatches (no subscribers in the codebase).

**E2E on sandbox `:20363`** (Playwright, Chase synced @ past viewDate, VOO row):
- Detail page load: **0 API calls** — fields populate from `appContext.data` (ticker, qty, cost basis, date all hydrated).
- Quantity edit: exactly **1 POST `/api/snapshots/holding`**, **0 GET `/api/snapshots`** (no resync triggered).
- Input value reflects the new qty immediately after the POST returns — optimistic write through `setData` round-trips correctly.

175/175 client tests, typecheck + lint clean.

hoiekim (approved):

PR #364, opened 2026-05-15
feat(snapshots): slice GET /api/snapshots month-by-month

### LINKED ISSUE #323: Slice Snapshot Data Sync

claoie:
Closes #323 (step 3)

Steps 1 and 2 of #323 already shipped in #348 (security data in `/api/snapshots` + client IndexedDB + React state). This PR is step 3: mirror the per-month + per-account slicing & caching that the transactions sync already uses.

## Summary
- `fetchSnapshots` in `src/client/lib/hooks/sync.ts` is rewritten to mirror `fetchTransactions`: loop month-by-month from the current `ViewDate` back to `oldestTransactionDate`, issuing one request per (month, account) tuple for user-scoped account_balance + holding snapshots and one request per month for shared security snapshots.
- Months ending more than `THIRTY_DAYS` ago go through `cachedCall` (browser Cache API), so they're hit at most once across the user's lifetime. The current month always re-fetches so today's data stays fresh.
- Extend `GET /api/snapshots` with a `snapshot-type` query string (`account_balance | holding | security`, validated server-side) so the per-month security-only slice doesn't drag in the full user-scoped history.

## Why per-month + per-account
Previously, sync issued a single `GET /api/snapshots?start-date=<oldest>&end-date=<today+2>` request. As the issue notes, that single call already crosses 500ms and only grows with holding/snapshot count. Slicing month-by-month:
- Keeps any single network round-trip bounded by one month of data per account.
- Lets the browser Cache API skip past months entirely on subsequent loads (the URL is the cache key, identical month/account URLs hit cache instantly).
- Matches the access pattern `fetchTransactions` already validates in prod, so the dev story stays consistent.

## Test plan
- [x] `bun run tsc --noEmit` passes
- [x] `bun test` shows no new failures (19 pre-existing failures reproduce identically on upstream/main)
- [ ] Sandbox E2E: load the app, network tab shows one `/api/snapshots?...&account-id=…` per (month, account) + one `/api/snapshots?...&snapshot-type=security` per month; second load serves older months from cache; UI renders holdings, snapshots, and securities identically to the pre-slice path.

### DIFF
--- a/src/client/lib/hooks/sync.ts
+++ b/src/client/lib/hooks/sync.ts
@@ -15,6 +15,7 @@ import {
   AccountsGetResponse,
   SplitTransactionsGetResponse,
   ChartsGetResponse,
+  SnapshotsGetResponse,
 } from "server";
 import {
   Account,
@@ -216,35 +217,88 @@ const fetchSnapshots = async (
     securitySnapshots: new SecuritySnapshotDictionary(),
   };
 
-  const params = new URLSearchParams();
-  const endDate = new Date();
-  endDate.setDate(endDate.getDate() + 2);
-  if (startDate) params.append("start-date", getDateString(startDate));
-  params.append("end-date", getDateString(endDate));
-  const path = "/api/snapshots?" + params.toString();
+  const ingestSnapshot = (snapshot: JSONSnapshotData) => {
+    if ("account" in snapshot) {
+      const account = accounts.get(snapshot.account.account_id) || {};
+      snapshot.account = { ...account, ...snapshot.account };
+      const newSnapshot = new AccountSnapshot(snapshot);
+      result.accountSnapshots.set(newSnapshot.snapshot.id, newSnapshot);
+    } else if ("holding" in snapshot) {
+      const newSnapshot = new HoldingSnapshot(snapshot);
+      result.holdingSnapshots.set(newSnapshot.snapshot.id, newSnapshot);
+    } else if ("security" in snapshot) {
+      const newSnapshot = new SecuritySnapshot(snapshot);
+      result.securitySnapshots.set(newSnapshot.snapshot.id, newSnapshot);
+    }
+  };
 
-  await call
-    .get(path)
-    .then(({ body }) => {
-      if (!body) return;
-      const snapshots = body as JSONSnapshotData[];
+  // Mirror fetchTransactions (Closes #323): slice the full date window into
+  // month-sized chunks, fetched per-account for user-scoped snapshots and
+  // once-per-month for shared security snapshots. Older months go through
+  // `cachedCall` so they only hit the network on first load; the most
+  // recent month always re-fetches so today's data stays fresh.
+  const snapshotsApiPath = "/api/snapshots";
+  const viewDate = new ViewDate("month");
+  const twoMonthAgoViewDate = viewDate.clone().previous().previous();
+  let oldestDate = twoMonthAgoViewDate.previous().getStartDate();
+  oldestDate = startDate && startDate < oldestDate ? startDate : oldestDate;
 
-      snapshots.forEach((snapshot) => {
-        if ("account" in snapshot) {
-          const account = accounts.get(snapshot.account.account_id) || {};
-          snapshot.account = { ...account, ...snapshot.account };
-          const newSnapshot = new AccountSnapshot(snapshot);
-          result.accountSnapshots.set(newSnapshot.snapshot.id, newSnapshot);
-        } else if ("holding" in snapshot) {
-          const newSnapshot = new HoldingSnapshot(snapshot);
-          result.holdingSnapshots.set(newSnapshot.snapshot.id, newSnapshot);
-        } else if ("security" in snapshot) {
-          const newSnapshot = new SecuritySnapshot(snapshot);
-          result.securitySnapshots.set(newSnapshot.snapshot.id, newSnapshot);
+  const promises: Promise<void>[] = [];
+
+  while (oldestDate < viewDate.getStartDate()) {
+    const monthStart = viewDate.getStartDate();
+    const monthEnd = viewDate.clone().next().getStartDate();
+    const startStr = getDateString(monthStart);
+    const endStr = getDateString(monthEnd);
+    const isRecent = new Date().getTime() - monthEnd.getTime() < THIRTY_DAYS;
+
+    accounts?.forEach((a) => {
+      const params = new URLSearchParams();
+      params.append("start-date", startStr);
+      params.append("end-date", endStr);
+      params.append("account-id", a.id);
+      const path = snapshotsApiPath + "?" + params.toString();
+
+      const fetchForAccount = async () => {
+        let response: ApiResponse<SnapshotsGetResponse> | void;
+        if (isRecent) {
+          response = await call.get<SnapshotsGetResponse>(path).catch(console.error);
+        } else {
+          response = await cachedCall<SnapshotsGetResponse>(path).catch(console.error);
         }
-      });
-    })
-    .catch(console.error);
+        if (!response?.body) return;
+        response.body.forEach(ingestSnapshot);
+      };
+
+      promises.push(fetchForAccount());
+    });
+
+    // Security snapshots are user-id=NULL (shared) and aren't returned by
+    // the account-scoped queries above. One slice per month, cached for
+    // older months — keeps the per-account URL space cleanly cache-keyed
+    // and avoids re-pulling the full price history on every page load.
+    const securityParams = new URLSearchParams();
+    securityParams.append("start-date", startStr);
+    securityParams.append("end-date", endStr);
+    securityParams.append("snapshot-type", "security");
+    const securityPath = snapshotsApiPath + "?" + securityParams.toString();
+
+    const fetchSecurities = async () => {
+      let response: ApiResponse<SnapshotsGetResponse> | void;
+      if (isRecent) {
+        response = await call.get<SnapshotsGetResponse>(securityPath).catch(console.error);
+      } else {
+        response = await cachedCall<SnapshotsGetResponse>(securityPath).catch(console.error);
+      }
+      if (!response?.body) return;
+      response.body.forEach(ingestSnapshot);
+    };
+    promises.push(fetchSecurities());
+
+    viewDate.previous();
+  }
+
+  await Promise.all(promises);
 
   return result;
 };
--- a/src/client/lib/hooks/sync.ts
+++ b/src/client/lib/hooks/sync.ts
@@ -52,6 +52,27 @@ import {
   indexedDb,
 } from "client";
 
+// Bounded worker pool. Browsers cap concurrent fetches per origin (~6 on
+// HTTP/1.1) and queueing hundreds of `fetch`/`cache.add` calls at once
+// triggers ERR_INSUFFICIENT_RESOURCES — which silently aborts `cache.add`
+// and defeats the Cache API benefit for older months. Run a small N at a
+// time so every cache write actually completes.
+const SNAPSHOT_FETCH_CONCURRENCY = 6;
+
+const runWithConcurrency = async (
+  tasks: Array<() => Promise<void>>,
+  limit: number,
+): Promise<void> => {
+  let next = 0;
+  const workers = Array.from({ length: Math.min(limit, tasks.length) }, async () => {
+    while (next < tasks.length) {
+      const idx = next++;
+      await tasks[idx]();
+    }
+  });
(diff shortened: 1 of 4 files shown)


claoie:
## Self-Review

### Discussion thread status

### Checked
- **Logic** —
  - `searchSnapshots` in `src/server/lib/postgres/repositories/snapshots.ts` already had `snapshot_type` in `SearchSnapshotsOptions` (line 26). The route correctly threads the new `snapshot-type` query string into `options.snapshot_type` and validates against the allowed enum.
  - `wantsSecurity` in `searchSnapshots` is `(!snapshot_type || snapshot_type === "security") && !account_id && !account_ids?.length`. The client's two request shapes line up:
    - Per-account (no `snapshot-type`, has `account-id`) → `wantsSecurity=false` → user-scoped balance + holding only, **no security duplication**.
    - Security-only (`snapshot-type=security`, no account) → `wantsSecurity=true` → returns the global security snapshots branch.
  - The month loop mirrors `fetchTransactions` exactly (uses the same `ViewDate.previous()` / `getStartDate()` walk), so it inherits the same termination behavior and the same `oldestTransactionDate` floor.
- **Types** — `SnapshotsGetResponse` is exported from the server route and imported in `sync.ts`. `ApiResponse<SnapshotsGetResponse>` typed cleanly in the per-account / per-security promise bodies.
- **Quality** — `ingestSnapshot` is the right factoring — same code that used to be inline in the old single-fetch path, now a shared helper for both branches.
- **Security** — Validation reuses `optionalQueryString` and explicitly rejects unknown `snapshot-type` values; no SQL injection surface added.
- **Tests** — `searchSnapshots` already has dedicated `snapshots.test.ts` coverage (`snapshot_type: "holding"`, `snapshot_type: "security"`) that exercises the predicate routing. `bun run tsc --noEmit` clean.
- **Blockers** — None hard. One concern flagged below.

### Scope check
Two files: `src/client/lib/hooks/sync.ts` (rewrite of `fetchSnapshots`), `src/server/routes/accounts/get-snapshots.ts` (new `snapshot-type` query param + validation). Both belong to the stated scope.

### E2E (sandbox PR-364 on port 20364)
Spun up the sandbox; logged in; let initial sync settle.

**Request shape (initial sync, scrambled-prod DB clone, 18 accounts, ~51 months back to 2022-03):**
| metric | observed |
|---|---|
| `/api/snapshots` calls during initial 5s window | 38 (in-flight, sync still going) |
| per-account calls (no `snapshot-type`, has `account-id`) | 36 |
| security calls (`snapshot-type=security`, no account) | 2 |
| full-sweep calls (no type, no account) | **0** — old pattern is gone |
| account IDs covered | 18 |
| IndexedDB after sync | `accountSnapshots=956`, `securitySnapshots=1657`, `holdingSnapshots=0` (no holdings in this scrambled clone) |

All calls returned 200. Per-account calls returned no `security` rows — the partition is clean.

### ⚠️ Concern: cache hit rate on reload
A longer second run (60s settle + reload + 30s settle):
- Initial sync: **474** `/api/snapshots` calls
- Reload: **666** additional calls (140% of initial — *higher*, not lower)

If the Cache API were populating reliably, reload should be ~0 net-new calls (current month re-fetches via `call.get`, older months come from `cache.match`). The 666 reload calls strongly suggest `cachedCall` is missing the cache on older months.

Likely cause: ~900 simultaneous fetches via `Promise.all` exceed Chromium's per-origin connection pool. Many trigger `ERR_INSUFFICIENT_RESOURCES` (seen in console: ~700 such errors during the cache test). When `cache.add(path)` throws, the catch swallows it and **no cache entry is written**, so the next load re-fetches.

The mirror reference `fetchTransactions` has the same unthrottled `Promise.all` shape, so this isn't a regression *introduced* by this PR — but it does mean the "cached for older months" benefit advertised in the PR body may not actually materialize for users with many accounts × many months. A `p-limit`-style concurrency cap (e.g. 6 in-flight) would let `cache.add` succeed for every slice and deliver the intended speedup.

Flagging for Hoie — not blocking, but worth a follow-up issue if he wants the cache to actually carry its weight.

### Confidence: Medium
Code is correct and shape matches `fetchTransactions`. The cache-hit gap is a real performance concern that the PR body doesn't acknowledge.

hoiekim:
Agreed on the p-limit style concurrency cap. Implement that in this PR.

claoie:
Implemented the concurrency cap. Pushed `ef6af0a`.

## Implementation
Inline `runWithConcurrency(tasks, limit)` worker pool in `sync.ts`, applied to `fetchSnapshots` only with N=6. No new dep — keeps the PR scope tight. `fetchTransactions` has the same shape but is out of scope; happy to do a follow-up.

## E2E (sandbox PR-364, scrambled-prod clone, 18 accounts, 51 months)

| metric | before cap | after cap |
|---|---|---|
| `ERR_INSUFFICIENT_RESOURCES` console errors | ~700 | **0** |
| `budget-cache` entries after initial sync | (partial) | **2079** |
| `/api/snapshots` calls on reload | 666 | **38** |
| reload `cache` entry delta | grew | **0** (everything served) |

Initial sync still does 1102 `/api/snapshots` calls (1044 per-account + 58 security) since the cache is cold; reload now correctly serves all older months from cache and only re-fetches the recent (<30d) slice — matching the design described in the PR body.

### Confidence: High
Hoie review feedback addressed. Code path matches `fetchTransactions` minus the cap, which is by design (this PR is scoped to snapshots).

src/client/lib/hooks/sync.ts
```diff
@@ -51,6 +52,27 @@ import {
...
+): Promise<void> => {
+  let next = 0;
+  const workers = Array.from({ length: Math.min(limit, tasks.length) }, async () => {
+    while (next < tasks.length) {
+      const idx = next++;
+      await tasks[idx]();
+    }
+  });
+  await Promise.all(workers);
+};
```
hoiekim: Can we use `import { Queue } from "common"` instead?

src/client/lib/hooks/sync.ts
hoiekim, line 420: Now that snapshots take so much longer, we should not wait all of these together. Instead,
1. Wait for the non-historical data(accounts, items, budgets, sections, categories, charts, institutions), then `setData()`.
2. Wait for the latest 2 months' historical data, then `setData()`.
3. Wait for the all the rest of the data, then `setData()`.

src/client/lib/hooks/sync.ts
claoie: Done in 640a7aa. The existing `Queue` is purely a sliding-window rate
limiter (capacity per `windowMs`), which doesn't fit a concurrency cap —
`new Queue({ capacity: 6, windowMs: 60_000 })` would let only 6 starts
per minute, making 1100 stage-3 snapshot fetches take ~3 hours. So I
extended `Queue` with an independent `maxInflight` option. The two gates
compose: `polygon.ts` keeps `capacity`-based rate limiting, `sync.ts`
uses `maxInflight: 6` for browser fetch-cap protection. Added two new
tests for the inflight semantics (cap respected, slot released on task
error).

src/client/lib/hooks/sync.ts
claoie: Done in 640a7aa. The sync now stages three `setData()` calls:

1. **Stage 1 — non-historical**: accounts/items, budgets/sections/categories,
   charts, institutions.
2. **Stage 2 — latest 2 months historical**: transactions + snapshots for
   current month + previous month, plus splitTransactions.
3. **Stage 3 — everything older**: transactions + snapshots from
   `oldestTransactionDate` up to where stage 2 starts. Stage 3 merges into
   the dictionaries built in stage 2 so each `setData()` strictly adds
   rows.

To make stage 2 / stage 3 reuse the same per-month-per-account fetch
loop, `fetchTransactions` / `fetchSnapshots` now take an explicit
`{ from, until }` range. Bounds are exclusive on both ends, matching the
original `while (oldestDate < viewDate.getStartDate())` semantics — so
stage 2 (`from = 2-back-start, until = next-start`) covers current +
previous and stage 3 (`from = oldestDate or 3-back-start, until =
1-back-start`) covers everything older, with no overlap or gap.

### E2E (sandbox, scrambled-prod clone, 18 accounts, 51 months)

| stage | last response |
|---|---|
| Stage 1 (15 non-historical calls) | ~4.8s |
| Stage 2 (transactions + recent snapshots) | ~12s |
| Stage 3 (older snapshots) | ~17s |

Nav + accounts render at ~4.8s now instead of waiting ~17s for the
entire history. Zero `ERR_INSUFFICIENT_RESOURCES` errors (Queue maxInflight
still in place). Total API call counts unchanged from the pre-stage
version (986 snapshots, 928 transactions).

splitTransactions has no time dimension on the API, so I placed it in
stage 2 (it ships alongside the recent activity it semantically belongs
to). Happy to move it to stage 1 if you'd rather.

claoie:
## Self-Review (post-feedback)

### Discussion thread status
- Both inline review comments from hoiekim addressed in 640a7aa:
  - Use `Queue` from common → extended `Queue` with `maxInflight` (the existing rate-limit semantic didn't fit a concurrency cap) and routed snapshot fetches through it.
  - Stage `setData()` into 3 passes → split `useSync` into stage 1 (non-historical), stage 2 (last 2 months), stage 3 (older).
- No outstanding blockers; both review threads have replies pending hoiekim's acknowledgment.

### Checked
- **Logic** — `fetchTransactions` and `fetchSnapshots` now take an explicit `{ from, until }` range. Exclusive on both ends, preserving original `oldestDate < viewDate.getStartDate()` semantics. The two stage windows tile cleanly (stage 2: `from=2-back-start, until=next-start`; stage 3: `from=olderDate, until=1-back-start`) — no overlap, no missing month relative to the pre-PR behavior. `Queue.add` awaits both gates (`waitForRateSlot` then `acquireInflight`) and releases the inflight slot in a `finally` so a rejected task can't leak a slot.
- **Types** — `bun run tsc --noEmit` clean. `FetchRange` interface exported (well, local) and reused across both fetchers. Queue's `QueueOptions.capacity` is now optional (default 0 = no rate gate) which is backwards-compatible with the polygon usage.
- **Quality** — Inline `runWithConcurrency` is gone, replaced by the shared `Queue`. Three-stage `setData` is structured top-to-bottom with comments explaining each stage's role.
- **Security** — No new request shapes or auth touchpoints; `/api/snapshots` already validated `snapshot-type` against an allowlist in the prior commit.
- **Tests** — 2 new Queue tests pass (`maxInflight` cap + slot release on task error). Full `bun test`: 523 pass / 19 fail; same 19 failures reproduce on upstream/main (pre-existing). `polygon.test.ts` still passes (capacity-only Queue path unchanged).
- **Blockers** — None hard.

### Coding rules check
- [x] No custom `Table` subclasses (no DB code touched here)
- [x] Table methods used for simple equality filters (n/a — no DB code touched)
- [x] No `pool.query()` calls in repositories for new code (n/a)
- [x] No unrelated files changed — diff is `sync.ts`, `Queue.ts`, `Queue.test.ts`, plus the route file from the original commit
- [x] Imports from highest barrel (`"common"`, `"client"`, `"server"`)
- [x] Typecheck passes

### E2E Testing
Sandbox dev server, scrambled-prod clone, 18 accounts, 51 months back.

| signal | observed |
|---|---|
| Login form gone | 4673ms |
| Last stage-1 API response (15 calls) | 4780ms |
| First stage-2 transaction response | 4922ms |
| First stage-2 snapshot response | 5125ms |
| Last historical response | 17331ms |
| /api/snapshots total | 986 (unchanged) |
| /api/transactions total | 928 (unchanged) |
| `ERR_INSUFFICIENT_RESOURCES` console errors | 0 |

Stage 1 setData fires around ~4.8s (UI gets nav + accounts + budgets + charts + institutions) instead of waiting ~17s for the entire history to load. Stage 2 fires around ~12s once recent transactions + snapshots are in. Stage 3 fires at ~17s with the full backfill. No new error paths.

### Issues found
- None during E2E.

### Confidence: High

hoiekim (approved):

PR #371, opened 2026-05-15
fix(sync): dedupe Plaid securities before upserting holdings

### LINKED ISSUE #370: sync-plaid upserts holdings before deduping securities → orphan-security holdings cause duplicate cash rows
## Symptom
Hoie 2026-05-15 (post-merge of #369): "I merged the PR but I still see duplicate cash rows." Affects synced investment accounts where Plaid returns a security whose ticker already maps to an existing canonical security_id in our DB.

## Root cause
`src/server/lib/compute-tools/sync-plaid.ts` (current main, lines 266-271):

```ts
const inferredCash = await inferCashHoldings(accounts, holdings, securities);
const allHoldings = inferredCash.length ? [...holdings, ...inferredCash] : holdings;

await upsertAccountsWithSnapshots(user, accounts, storedAccounts);
await upsertAndDeleteHoldingsWithSnapshots(user, allHoldings, storedHoldings);  // ← holdings written FIRST with Plaid security_ids
await upsertSecuritiesWithSnapshots(securities);                                  // ← securities deduped AFTER
```

`upsertSecuritiesWithSnapshots` (`create-snapshots.ts:115-119`) maps incoming security_ids by ticker_symbol to existing canonical IDs (only the deduped row gets persisted in `securities`). Holdings have already been upserted with the *un-deduped* Plaid security_ids — they're now orphans pointing at security_ids that have no row in the `securities` table.

`sync-simple-fin.ts` (lines 72-83) already does the right thing: dedupes securities first to get an `idMap`, then maps holdings to the canonical IDs before upserting them.

## Sandbox repro (prod clone)
- Account: Chase (LgkKbv…) — three live holding rows: VOO, QACDS (`kZrEbX…`, canonical), `gJpE8M…` (orphan, no row in `securities`).
- The orphan has `institution_price=1` and `cost_basis=null` → FE's `isCash` heuristic labels it "Cash" → user sees two Cash rows.
- `holding_id` separator is `_` (Plaid's format), not `-` (`inferCashHoldings`'s format), confirming the orphan came from Plaid not from cash inference.
- 3 orphan-security holdings total in the sandbox, spread across 3 accounts.

## Proposed fix
Mirror `sync-simple-fin.ts`'s shape in `sync-plaid.ts`:

```ts
// 1. Run cash inference first on the raw Plaid response (still has Plaid security_ids on
//    both the holdings and securities arrays, so cash-detection lookups still work).
const inferredCash = await inferCashHoldings(accounts, holdings, securities);

// 2. Upsert securities to get canonical IDs.
await upsertAccountsWithSnapshots(user, accounts, storedAccounts);
const idMap = await upsertSecuritiesWithSnapshots(securities);

// 3. Remap Plaid holdings to canonical security_ids + holding_ids.
const mappedHoldings = holdings.map((h) => {
  const canonical = idMap[h.security_id];
  if (!canonical) return undefined;
  if (canonical === h.security_id) return h;
  return { ...h, security_id: canonical, holding_id: `${h.account_id}_${canonical}` };
}).filter((h): h is JSONHolding => !!h);

// 4. Inferred-cash holdings already reference the canonical USD-cash security_id — pass through.
const allHoldings = inferredCash.length ? [...mappedHoldings, ...inferredCash] : mappedHoldings;

// 5. Upsert holdings with canonical IDs. Existing orphans (whose holding_id used the
//    un-deduped Plaid security_id) are absent from this set, so the existing
//    upsertAndDeleteHoldingsWithSnapshots "removed" pass writes zero-snapshots for them
//    and the FE filters them out.
await upsertAndDeleteHoldingsWithSnapshots(user, allHoldings, storedHoldings);
```

## Backfill
Automatic. On the first sync after the fix deploys:
- Plaid sends the same holdings with Plaid security_ids.
- The new order remaps holdings to canonical IDs (different `holding_id`).
- The orphan `holding_id`s (using the un-deduped IDs) drop out of `incomingMap` → "removed" pass writes zero-quantity snapshots → FE filters them out.
- No migration needed.

## Verification plan
- New test in `sync-plaid.test.ts` (or a focused test on the remap helper if extracted): given Plaid response with a holding whose security_id dedupes to a different existing canonical, the upserted holding's `security_id` matches the canonical, and the original orphan `holding_id` is in the "removed" set.
- Sandbox spin: confirm only one Cash row renders for the Chase account post-fix.

claoie:
Closes #370.

## Why
`sync-plaid` upserted holdings *before* deduping securities (PR #350 + #353 era). `upsertSecuritiesWithSnapshots` then collapsed any Plaid security whose ticker matched an existing canonical row to the canonical `security_id`, but holdings still carried Plaid's pre-dedupe ID — orphan holdings pointing at a `security_id` with no row in the `securities` table. The FE's `isCash` heuristic (price=1, null cost basis) then labels them "Cash" and the user sees duplicate cash rows.

This is what produced the duplicate Hoie reported on 2026-05-15 after #369 merged. My #369 fix addressed a different (also real) corner of the cash-detection space — MM-fund/proprietary-sweep false negatives — but didn't touch the upsert order.

## What
Mirror the shape `sync-simple-fin` already uses:

```ts
const idMap = await upsertSecuritiesWithSnapshots(securities);
const mappedHoldings = remapHoldingsToCanonicalSecurityIds(holdings, idMap);
const allHoldings = inferredCash.length ? [...mappedHoldings, ...inferredCash] : mappedHoldings;
await upsertAndDeleteHoldingsWithSnapshots(user, allHoldings, storedHoldings);
```

The `remapHoldingsToCanonicalSecurityIds` helper is exported so unit tests can exercise it in isolation. It rewrites both `security_id` and the Plaid-format `${account_id}_${security_id}` `holding_id`; passes through holdings whose security_id is already canonical; drops holdings whose security_id is missing from `idMap` (defensive — Plaid always returns securities for the holdings it reports, but we'd rather skip than write an orphan).

Cash inference (`inferCashHoldings`) still runs first against the raw Plaid response so `isCashLikeSecurity` / `isHoldingCashLike` can resolve via Plaid's IDs. Inferred-cash holdings already reference the canonical USD-cash `security_id` (from `ensureUSDCashSecurity`) so they bypass the remap.

## Backfill
Automatic — no migration. On the next Plaid sync after deploy:
1. Plaid returns the same orphan-shaped holding with its pre-dedupe security_id.
2. We dedupe → canonical → remap holdings → upsert under the canonical `holding_id`.
3. The old orphan `holding_id` is no longer in the incoming set, so `upsertAndDeleteHoldingsWithSnapshots`'s "removed" pass writes a zero-quantity snapshot.
4. The FE filters `value === 0` rows out of the composition table → the duplicate row disappears.

## Verification
- `bun test src/server` — 421/421 pass (4 new tests on the remap helper).
- `bun run typecheck`, `bun run lint` — clean.
- Sandbox spin + URL on request via Discord — not posted publicly.

## Sandbox repro of the bug (before fix)
3 orphan-security holdings across the sandbox prod-clone (Chase + 2 others). Chase's holdings table renders `[<ticker>, Cash, Cash]` — confirmed via Playwright.

## Test plan
- [ ] Sandbox spin on this branch, check Chase holdings composition. Until a real Plaid sync runs (sandbox has no Plaid creds), the existing orphan snapshots still display — but the new sync logic is correct. The behavioral E2E is covered by the unit tests on the mapping function + the existing `create-snapshots.test.ts` coverage on the removal pass.
- [ ] After deploy + first prod Plaid sync, confirm the orphan rows clear (one zero-quantity snapshot per orphan, then no row rendered).

### DIFF
--- a/src/server/lib/compute-tools/sync-plaid.test.ts
+++ b/src/server/lib/compute-tools/sync-plaid.test.ts
@@ -3,8 +3,9 @@ import {
   buildTransactionLookupMaps,
   findStoredTransaction,
   getPlaidRemovedInvestmentTransactions,
+  remapHoldingsToCanonicalSecurityIds,
 } from "./sync-plaid";
-import type { JSONTransaction, JSONInvestmentTransaction } from "common";
+import type { JSONHolding, JSONTransaction, JSONInvestmentTransaction } from "common";
 
 // Minimal factory helpers
 const makeTx = (overrides: Partial<JSONTransaction>): JSONTransaction =>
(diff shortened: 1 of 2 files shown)


claoie:
Closing per Hoie's direction (2026-05-15) — the user-visible bug is duplicate cash rendering, which happens because the FE creates one row per `holding_id` regardless of whether two cash-shape holding_ids resolve to the same logical position. Fixing that at the FE (aggregate cash rows into one) is the cleaner shape than fixing the server-side dedupe-order issue this PR addressed.

The data-integrity issue described in #370 (Plaid shifting security_ids creates orphan holdings) is still real but no longer user-visible once the FE aggregates — leaving #370 open as a latent concern for non-cash positions; can revisit if it ever surfaces.

PR #373, opened 2026-05-16
fix(accounts-donut): mirror per-row gate so change widget matches row sum

### LINKED ISSUE #357: bug: AccountsDonut 'from last month' change overstates vs sum of per-row Changes
## Summary

On `/accounts`, the donut center's **"from last month"** change widget reports a value that does not match the sum of the per-row **Changes** widgets in `AccountsTable` below it. The two views read the same `balanceData` but apply different "no-history" semantics — the donut treats *missing/zero* previous-balance as `previous = 0` while the per-row widget hides itself for the same accounts. The donut overstates the change by the current balance of any account whose previous-month entry is `0`/`undefined`.

This is the same "aggregator silently treats no-data ≠ no-data on the per-row view" shape as PR #309 (BalanceChartRow excluding asset types) and PR #319/#311/#313 (AccountsTable Balance widget falling back to today's balance). The previous fixes left this one path untouched.

## Reproduction (admin against local prod-snapshot DB, today 2026-05-14)

`/accounts` (view = May 2026) renders:

```
$[redacted]
+$[redacted]  from last month
in 11 Accounts
```

Per-row Changes widgets in `AccountsTable` (asset-side only — credit rows render `spent/available`, never a Changes block):

| Row              | Current      | Changes widget |
|------------------|--------------|----------------|
| Amazon - 401(K)  | $[redacted]  | **+$[redacted]** |
| Amazon - Stock   | $[redacted]  | **-$[redacted]**|
| IBM - 401(K)     | $[redacted]  | **+$[redacted]** |
| JP Morgan        |  $[redacted]  | `-` (neutral)  |
| Wealthfront      |  $[redacted]  | `-` (neutral)  |
| Citi             |  $[redacted]  | `-` (neutral)  |
| IBM - RSU        |  $[redacted]  | **+$[redacted]** |
| IBM - Stock      |  $[redacted]  | **+$[redacted]** |
| IBM - RBA        |  $[redacted]  | *(no widget)*  |
| Chase Checking   |   $[redacted]  | `-` (neutral)  |
| Binance          |   $[redacted]  | `-` (neutral)  |

Sum of per-row signed changes (treating `-` neutral and missing as 0):

```
+9593.11 − 20704.11 + 4331.77 + 1039.68 + 7031.96  =  +1,292.41
```

Donut says **+$[redacted]**. Gap = `15716.39 − 1292.41 = $[redacted]` — that is **exactly the current balance of IBM-RBA** (the one asset row that has no Changes widget at all).

## Root cause

`src/client/components/AccountsDonut/BalanceInfo.tsx:27-31`:

```ts
const previousAmount = donutData.reduce((a, { id }) => {
  const balanceHistory = balanceData.get(id);
  if (!balanceHistory) return a;
  return a + (balanceHistory.get(previousDate) || 0);
}, 0);
```

Compare with `src/client/components/AccountsTable/Balance.tsx:27` (and the gate at L56/68/80):

```ts
const previousAmount = balanceHistory.get(previousDate) || 0;
...
{!!previousAmount && (
  <Changes currentAmount={dynamicAmount || current!} previousAmount={previousAmount} />
)}
```

Both compute `previousAmount` the same way, but only the per-row widget conditionally **suppresses itself when `!previousAmount`**. The donut's reduce accumulates `0` for those accounts on the previous side while the same accounts' **current** balance is fully counted in `balanceTotal` (`AccountsPage/index.tsx:54-56` uses `balanceHistory.get(viewDateDate) || fallback` where fallback = current for future dates).

IBM-RBA's specific signal: the most recent holding snapshot for `account_id=fda9c` is `2026-04-26` with `institution_value=0` (price=540, qty=20 → real value $[redacted]; the stored zero is a separate pre-existing data issue). `getBalanceDataFromHoldingSnapshots` sums the stored `institution_value`, so `balanceHistory.get(2026-04-30)` returns `0` rather than `undefined`. That `0` is real data per the storage contract, but BalanceInfo's `|| 0` and the per-row `!!previousAmount` gate disagree on what it means: the per-row widget hides the row's contribution; the donut counts the full current.

## Why this matters

- User-facing inconsistency on a single page — the headline change number can't be reconciled with what the rows below show. On the snapshot today the gap is $[redacted] (~92% of the displayed change).
- Latent multiplier: any new account onboarded mid-cycle (no prior-period snapshot) inflates the donut's monthly change by its full current balance until a snapshot for the previous period exists.
- Same pattern in yearly view: the gap would equal current balance of *any* account that lacks a snapshot for "previous year end" — likely larger than the monthly version.

## Proposed fix

Mirror the table's filtering in `BalanceInfo.tsx` — only count accounts toward *both* sides of the change comparison when the account has a usable previous-period entry:

```ts
let comparableCurrent = 0;
let comparablePrevious = 0;
for (const { id, value } of donutData) {
  const balanceHistory = balanceData.get(id);
  if (!balanceHistory) continue;
  const prev = balanceHistory.get(previousDate) || 0;
  if (!prev) continue;                  // mirror Balance.tsx `!!previousAmount`
  comparableCurrent += value;
  comparablePrevious += prev;
}
...
<Changes
  currentAmount={comparableCurrent}
  previousAmount={comparablePrevious}
  ...
/>
```

The donut center then shows the change *across the same set of rows* that have a visible Changes widget — `+$[redacted]` for this snapshot — and the user can reconcile by inspection. The big total (`$[redacted]`) stays unchanged; only the comparison pair feeding `<Changes>` is filtered.

Alternative considered: have BalanceInfo fall back `prev = prev || value` so accounts without history contribute zero change. This keeps `currentAmount` = `balanceTotal` but masks the "no history yet" signal — strictly worse than mirroring the row UI.

Out of scope for this issue (worth tracking separately if confirmed):
- `holding_snapshots.institution_value = 0` rows where `institution_price * quantity > 0` — looks like a snapshot-time race/pricing-gap; needs root-cause separately.
- AccountsPage `balanceTotal` uses `|| fallback` (line 54-56) where `??` would be more precise — minor, not part of this bug.

---

claoie:
Closes #357

## Summary

`AccountsDonut`'s `BalanceInfo` summed `previousAmount` across **every** donut account (`prev || 0`), while `AccountsTable`'s per-row Changes widget hides itself when `!previousAmount`. The headline "from last month" change therefore included the current balance of accounts with missing/zero previous-period entries on the **current** side but not the **previous** side, overstating the gap by exactly the current balance of any such account.

Per the issue body's repro, the gap was `$[redacted] − $[redacted] = $[redacted]`, which equaled IBM-RBA's current balance — the one asset row with no Changes widget at all.

## Change

`src/client/components/AccountsDonut/BalanceInfo.tsx` now builds `comparableCurrent` / `comparablePrevious` from only the accounts that have a usable previous-period entry, mirroring `AccountsTable/Balance.tsx`'s `!!previousAmount` gate. Both numbers feed `<Changes>` so the donut's headline change shows the same total a user can reconcile by adding the visible per-row widgets in the table.

The donut center's `balanceTotal` label is unchanged — still the full asset total. `balanceTotal` is no longer needed as a `BalanceInfo` prop and is dropped from the interface.

## Math verification (against issue body's reproduction)

Before:
- `currentAmount = balanceTotal` = `$[redacted]` (full asset sum)
- `previousAmount = Σ(prev || 0)` over all 11 asset rows, missing 1 row's prev → underweighted
- Displayed change = `+$[redacted]`

After:
- `comparableCurrent = Σ(value)` over the 10 rows with `prev > 0`
- `comparablePrevious = Σ(prev)` over the same 10 rows
- Displayed change = sum of visible per-row Changes = `+9593.11 − 20704.11 + 4331.77 + 1039.68 + 7031.96 = +$[redacted]`
- IBM-RBA (`prev = 0`) is excluded from both sides; \$[redacted] gap disappears.

## Test plan

- [x] `bun run tsc --noEmit` — clean
- [x] `bun run lint` — clean
- [x] `bun test` — `521 pass / 19 fail`; identical pass count and failure set as `upstream/main` (all 19 pre-existing failures are in `holdings.test.ts` / `inferCostBasis` / security price index, none touch `AccountsDonut`)
- [ ] Sandbox E2E: load `/accounts`, verify donut "from last month" value equals the sum of per-row Changes shown in `AccountsTable` (structural invariant — robust to sandbox numeric scrambling, since scrambling preserves the donut-vs-rows identity)

## Out of scope (flagged in issue, not addressed here)

- `holding_snapshots.institution_value = 0` rows where `institution_price × quantity > 0` — separate root cause
- `AccountsPage` `balanceTotal` using `|| fallback` where `??` would be more precise

---

### DIFF
--- a/src/client/components/AccountsDonut/BalanceInfo.tsx
+++ b/src/client/components/AccountsDonut/BalanceInfo.tsx
@@ -2,7 +2,6 @@ import { DonutData, useAppContext } from "client";
 import { Changes } from "client/components";
 
 interface Props {
-  balanceTotal: number;
   currencySymbol: string;
   donutData: DonutData[];
   totalCredit: number;
@@ -11,7 +10,6 @@ interface Props {
 }
 
 export const BalanceInfo = ({
-  balanceTotal,
   currencySymbol,
   donutData,
   totalCredit,
@@ -24,17 +22,25 @@ export const BalanceInfo = ({
   const viewDateSpan = Math.max(-viewDate.getSpanFrom(new Date()), 0);
   const previousDate = viewDate.clone().previous().getEndDate();
 
-  const previousAmount = donutData.reduce((a, { id }) => {
+  // Mirror per-row gate (AccountsTable/Balance.tsx: `!!previousAmount`): skip accounts
+  // with no usable previous-period entry on BOTH sides — otherwise the donut counts
+  // those accounts' current balance while the per-row widgets don't, overstating change.
+  let comparableCurrent = 0;
+  let comparablePrevious = 0;
+  for (const { id, value } of donutData) {
     const balanceHistory = balanceData.get(id);
-    if (!balanceHistory) return a;
-    return a + (balanceHistory.get(previousDate) || 0);
-  }, 0);
+    if (!balanceHistory) continue;
+    const prev = balanceHistory.get(previousDate) || 0;
+    if (!prev) continue;
+    comparableCurrent += value;
+    comparablePrevious += prev;
+  }
 
   return (
     <div className="BalanceInfo">
       <Changes
-        currentAmount={balanceTotal}
-        previousAmount={previousAmount}
+        currentAmount={comparableCurrent}
+        previousAmount={comparablePrevious}
         currencySymbol={currencySymbol}
       />
       <div className="label">from&nbsp;last&nbsp;{viewDate.getInterval()}</div>
--- a/src/client/components/AccountsDonut/index.tsx
+++ b/src/client/components/AccountsDonut/index.tsx
@@ -37,7 +37,6 @@ export const AccountsDonut = ({
         </div>
       </div>
       <BalanceInfo
-        balanceTotal={balanceTotal}
         currencySymbol={currencySymbol}
         donutData={donutData}
         totalCredit={totalCredit}


claoie:
## Self-Review

### Discussion thread status

### Checked
- **Logic** — `comparableCurrent` / `comparablePrevious` accumulate only over rows where `balanceHistory.get(previousDate)` is truthy, mirroring `AccountsTable/Balance.tsx:56,68,80` which gates the per-row `<Changes>` on `!!previousAmount`. Rows with `prev=0` are excluded from BOTH sides, so the donut header math now matches the visible per-row sum.
- **Types** — `balanceTotal` prop dropped from `BalanceInfo` interface and removed from `AccountsDonut/index.tsx` call site. `bun run tsc --noEmit` clean.
- **Quality** — Replaced reduce-with-side-effect with a clearer for-of accumulator. Comment cites the gate file (`AccountsTable/Balance.tsx`) so the next reader knows where the invariant lives.
- **Security** — No surface change.
- **Tests** — `bun test` shows 540 pass / 19 fail, identical to `upstream/main` baseline (the 19 are pre-existing `holdings.test.ts` failures). No tests touch `AccountsDonut`.
- **Blockers** — None.

### Scope check
Two files: `src/client/components/AccountsDonut/BalanceInfo.tsx` (the fix) and `src/client/components/AccountsDonut/index.tsx` (one-line prop removal). Both belong.

### E2E (sandbox PR-373 on port 20373, scrambled-prod clone)

Loaded `/accounts`, scraped the donut header and every visible per-row `Changes` widget:

| | value |
|---|---|
| Donut header "from last month" | **+$REDACTED** |
| Visible per-row Changes widgets | 11 (of 17 total rows — 6 hidden by `!!previousAmount`) |
| Σ of visible per-row Changes | **+$REDACTED** |
| Identity holds | ✅ |

Per-row breakdown: `+2080.78 +52402.23 +2793.56 +78.92 +1309.01 +446.85 +3530.75 +40.60 +295.42 +29.14 −19494.97 = +43512.29`.

The 6 rows where `Changes` is hidden are exactly the rows excluded from the donut header sum — invariant holds end-to-end.

### Confidence: High
Math verified analytically (matches the issue body's repro) and empirically (donut header = sum of visible per-row widgets in scrambled-prod data).

*[Edited 2026-05-16: redacted dollar amounts and liquid-stock ticker per security-review PII policy]*

claoie:
Superseded by #401. Original diagnosis was reversed — the donut's net-worth-delta math is the correct semantic (a new account with current=$X *is* a +$X change), and the per-row Changes widget was the one hiding that delta on new accounts. #401 leaves `BalanceInfo` untouched and instead un-gates the row widget so it renders for new and closed accounts (existing `Changes` component already handles zero-change with a neutral '-'). Net effect: donut headline still equals per-row sum, and the +$X from a new account isn't lost.

Issue #359, opened 2026-05-14
test(holdings): zero coverage on POST /snapshots/holding (224 LOC, two modes, three side-effect chains)


claoie:
## Summary

`src/server/routes/accounts/post-holding-snapshot.ts` is **224 LOC with two distinct request modes (update + create), 5+ validation paths, and three external side-effect chains** (Polygon ticker-detail API → `upsertSecurities` → `upsertHoldingSnapshots` → `upsertHoldings`). It has no `.test.ts` sibling and no integration coverage anywhere in the repo.

This route shipped in PR #196 (Holdings management UI for manual investment accounts, merged 2026-05-09).

### Update mode (`provided_snapshot_id` is set, lines 91-162)
1. Snapshot not found OR belongs to another user → `failed: Snapshot not found or access denied` (cross-user ownership safety relies on `getHoldingSnapshots(user)` filtering by user)
2. `ticker_symbol` patch routes through `resolveSecurityId` (covers the existing-security branch and the Polygon-fetch branch separately)
3. `snapshot_date` patch normalizes via `LocalDate(...).toISOString().split("T")[0]`
4. Numeric patches (`quantity`, `cost_basis`, `institution_price`, `institution_value`) — `0` and `null` should both be applied (the `!== undefined` guard, line 111-114, distinguishes "field absent" from "field set to zero")
5. Empty patch object → `validationError("no fields to update")`
6. Successful patch re-syncs the holdings row from the latest snapshot state (lines 124-145) — verify `upsertHoldings` is called with the *post-patch* values, not pre-patch
7. `snapshotsTable.update` throws → wrapped error logged + re-thrown

### Create mode (no `provided_snapshot_id`, lines 164-222)
1. Missing `account_id` / `ticker_symbol` / `quantity` → individual `validationError(...)` paths
2. `quantity === 0` allowed (the `=== undefined || === null` guard on line 172 lets zero through — pin this in a test so a future refactor doesn't tighten it accidentally)
3. `resolveSecurityId` no-API-key branch → returns the configured-by-administrator message
4. `resolveSecurityId` Polygon-fetch failure → returns the validation message verbatim
5. Deterministic `snapshot_id` collision (same account + security + date) → upserts in place; verify a second create does NOT produce a duplicate row
6. `snapshot_date` omitted → uses `new Date()` (test should freeze time)
7. `upsertHoldings` is called with `institution_value ?? quantity * (institution_price ?? 0)` — verify the fallback math when `institution_value` is omitted

[...]

PR #374, opened 2026-05-16
test(holdings): cover DELETE+GET /snapshots/holding sibling routes

claoie:
Towards #359 — covers the two sibling routes ([\`delete-holding-snapshot.ts\`](https://github.com/hoiekim/budget/blob/main/src/server/routes/accounts/delete-holding-snapshot.ts), [\`get-holding-snapshots.ts\`](https://github.com/hoiekim/budget/blob/main/src/server/routes/accounts/get-holding-snapshots.ts)) mentioned at the bottom of the issue. The POST route itself (~238 LOC, two modes, three side-effect chains) is intentionally out of scope here so this diff stays small and reviewable.

## Summary

10 new unit tests in `src/server/routes/accounts/holding-snapshot-sibling-routes.test.ts`:

**DELETE /api/snapshots/holding**
- rejects unauthenticated requests (returns `failed` before any DB I/O)
- rejects requests missing the `id` query parameter
- refuses to delete a snapshot owned by another user (ownership check via `getHoldingSnapshots(user)` must fail before `softDelete` is reached)
- deletes a snapshot the user owns and reports success (asserts `softDelete` is called with `(snapshot_id, user_id)`)
- surfaces `softDelete` failure as a 500-style `{ status: "error" }` envelope via the `Route.execute` catch-all

**GET /api/snapshots/holding**
- rejects unauthenticated requests (no DB I/O)
- empty-list short-circuit: when the user has zero snapshots, `searchSecuritiesById` short-circuits on the empty input array, so `securitiesTable.queryOne` is never called
- hydrates each snapshot with `ticker_symbol` + `security_name` from the securities lookup
- returns `null` ticker/name when the security cannot be found
- dedupes the security lookup across snapshots sharing a `security_id` (3 snapshots, 1 lookup)

## Mocking strategy

Follows `post-suggest-category.test.ts`: monkey-patch shared-object methods on the real imports (`pool.query`, `snapshotsTable.softDelete`, `securitiesTable.queryOne`) and restore originals in `afterAll`. `mock.module("server", ...)` is process-wide in Bun and leaks into sibling test files, so it is deliberately not used.

## Test plan

- [x] `bun run tsc --noEmit` — clean
- [x] `bun run lint` — clean
- [x] `bun test src/server/routes/accounts/holding-snapshot-sibling-routes.test.ts` — `10 pass / 0 fail`
- [x] `bun test` (full suite) — `531 pass / 19 fail`. Pass count +10 vs. `upstream/main`; failure set unchanged (all 19 pre-existing failures live in `holdings.test.ts` / `inferCostBasis` / security-price-index, none touch route handlers)

## Still open in #359

The POST route's update mode (5 branches) and create mode (7 branches), plus the deterministic-snapshot-id collision case and the `backfillMonthlySecuritySnapshotsForward` fire-and-forget side effect. Follow-up PR.

### DIFF
--- a/src/server/routes/accounts/holding-snapshot-sibling-routes.test.ts
+++ b/src/server/routes/accounts/holding-snapshot-sibling-routes.test.ts
@@ -1,0 +1,258 @@
+/**
+ * Tests for the sibling routes around POST /api/snapshots/holding:
+ *   - DELETE /api/snapshots/holding
+ *   - GET    /api/snapshots/holding
+ *
+ * The POST route itself (`post-holding-snapshot.ts`, ~238 LOC) is intentionally
+ * out of scope here — issue #359 splits cleanly into "exercise sibling routes"
+ * (this file) and "exercise the create/update modes" (follow-up). Keeping the
+ * two PRs separate keeps each diff reviewable.
+ *
+ * Mocking strategy mirrors `post-suggest-category.test.ts`: monkey-patch
+ * shared-object methods (`pool.query`, `snapshotsTable.softDelete`,
+ * `securitiesTable.queryOne`) on the real imports and restore in `afterAll`.
+ * `mock.module("server", ...)` is process-wide in Bun and leaks into sibling
+ * test files, so it is deliberately not used.
+ */
+
+import { describe, test, expect, mock, beforeEach, afterAll } from "bun:test";
+
+import { pool } from "server";
+import { snapshotsTable, securitiesTable } from "server/lib/postgres/repositories";
+import { deleteHoldingSnapshotRoute } from "./delete-holding-snapshot";
+import { getHoldingSnapshotsRoute } from "./get-holding-snapshots";
+
+const originalPoolQuery = pool.query.bind(pool);
+const originalSoftDelete = snapshotsTable.softDelete.bind(snapshotsTable);
+const originalSecuritiesQueryOne = securitiesTable.queryOne.bind(securitiesTable);
+
+const mockPoolQuery = mock(
+  (_sql: string, _values?: unknown[]): Promise<{ rows: unknown[]; rowCount: number | null }> =>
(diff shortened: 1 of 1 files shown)


claoie:
## Self-Review

### Discussion thread status
No comments yet.

### Checked
- **Logic** — Verified the 10 `test()` blocks at `src/server/routes/accounts/holding-snapshot-sibling-routes.test.ts:110,121,128,143,155,171,181,193,216,231`. They cover what the PR body claims:
  - DELETE: unauth · missing id · cross-user-ownership · happy path · softDelete failure
  - GET: unauth · empty-list short-circuit · ticker+name hydration · `null` when security lookup misses · dedup security lookup across shared security_id
- **Quality** — Mocking strategy mirrors `post-suggest-category.test.ts` (monkey-patch + restore in `afterAll`), explicitly avoiding the bun `mock.module` global-hoisting trap. PR body calls this out.
- **Security** — Test-only diff (one new file under `src/server/routes/accounts/`). No production code path changes.
- **Tests** — `bun test src/server/routes/accounts/holding-snapshot-sibling-routes.test.ts` → 10 pass / 0 fail. Full `bun test` → 531 pass / 19 fail; pass count is +10 vs. `upstream/main` (which is 521 pass), failure set unchanged (the 19 pre-existing failures are in `holdings.test.ts` / `inferCostBasis` / security-price-index).
- **Blockers** — None.

### Scope check
One new file: `src/server/routes/accounts/holding-snapshot-sibling-routes.test.ts`. No production code modified. Clean test-only scope.

### E2E
**Skipped — justified by the change.** This PR adds tests only; no runtime behavior is modified. The "E2E" for a test PR is the test run itself, which is documented above (10 new pass, no new fails, full suite green vs baseline).

### Confidence: High
Test names verified line-by-line against PR claims (no fabrication). Mocking pattern follows the established safe-from-hoisting approach.

hoiekim (approved):

PR #388, opened 2026-05-18
fix(holdings): relabel Cost/sh → Cost basis to match stored semantic

### LINKED ISSUE #375: bug(holdings): "Cost/sh" label on HoldingProperties refers to total cost basis, not per-share
## Summary

`HoldingProperties` renders the Cost-basis input field with the label `Cost/sh` (cost-per-share), but the underlying `cost_basis` field is a **total** cost basis (matching Plaid's definition — `cost_basis ≈ price_per_share × quantity`). Manual holding entry stores the user's literal input as the **total** cost basis, and the same field displays the **total** cost basis under the per-share label on the edit page.

For a user adding 10 shares of AAPL bought at $200/share, the natural reading of "Cost/sh" is "$200" — they type 200, the row is stored with `cost_basis = 200` (representing $200 *total*), and every downstream G/L calculation treats the position as having only $200 of cost (vs. real $[redacted]). On 10 shares now worth $[redacted] the displayed unrealized gain becomes +$[redacted] instead of the correct +$500 — 4.6× the true gain.

## Reproduction

1. Open a manual investment account (provider = `MANUAL`).
2. From the **Holdings Composition** section, click `+ Add Holding`.
3. Enter `AAPL` (ticker), `10` (quantity), and **`200` in the field labeled `Cost/sh (opt)`** (intending "$200 per share").
4. Submit, then open the new holding from the composition row.
5. Observe the **Cost/sh** field on the detail page shows `200`.
6. Note the row's unrealized G/L on the composition table = `current value − $200`, not `current value − $[redacted]`.

## Expected vs actual

| | Expected | Actual |
| --- | --- | --- |
| Label semantics | per-share (matches "/sh") | **total** cost basis |
| Field name | `cost_basis` (total) | `cost_basis` (total) |
| What user sees ↔ what gets stored | match | **mismatch** — user enters per-share, system stores as total |
| Displayed G/L | matches per-share × quantity reasoning | inflated/deflated by factor of `quantity` |

## Root cause

The label string mismatches the data semantics throughout `HoldingProperties.tsx`:

- `src/client/components/HoldingProperties/index.tsx:374` — new-holding form input is labeled `Cost/sh (opt)` but its `onChange` writes `form.costBasis`, which is submitted at L166 as `body.cost_basis = costBasis` (no per-share→total transform).
- `src/client/components/HoldingProperties/index.tsx:467` — edit-mode field is labeled `Cost/sh` but reads from `snapshot.holding.cost_basis` (line 82-84) and writes back via `updateField({ cost_basis: parsed })` (line 292), again with no transform.

The `cost_basis` field is treated as **total** everywhere else in the codebase, consistent with Plaid's data model:

- `src/server/lib/postgres/models/snapshot.ts:153` — server collapses NULL to `0` for the *total* cost basis.
- `src/client/lib/hooks/calculation/holdings.ts:215-218` — `inferCostBasis` accumulates `totalCost += price * quantity + fees` and returns `costBasis: totalCost`.
- `src/client/lib/models/Calculations.ts:364-367` — `unrealizedGain = value − costBasis` where `value = price × quantity`, so `costBasis` must be total for the subtraction to be dimensionally correct.
- Local prod-snapshot DB confirms shape: IBM-RBA holding has `institution_price=95.46`, `quantity=157.861`, `institution_value=15069.41`, `cost_basis=14754.8` — the cost basis magnitude only makes sense as a total (per-share would be $93/share, total ≈ $[redacted]).

So the data is consistently total; only the **UI label** is wrong (and the manual-input flow's lack of transform compounds it for users who trust the label).

## Proposed fix

Two viable approaches; first is the smaller change:

1. **Relabel only.** Change both labels to `Cost basis` (or `Total cost basis`) — keeps every existing manual-entered holding correct, no migration needed.
   ```tsx
   // L374 — new-holding form
   <span className="propertyName">Cost&nbsp;basis&nbsp;(opt)</span>
   // L467 — edit form
   <span className="propertyName">Cost&nbsp;basis</span>
   ```
   Add a tiny placeholder/help text if desired: `placeholder="Total $ paid (all shares)"`.

2. **Convert UI ↔ data.** Keep `Cost/sh` label, and (a) on submit multiply by quantity before sending; (b) on display divide stored `cost_basis` by `quantity`. More UI work, more chance to divide by zero when quantity=0, and inconsistent with the rest of the codebase's "treat cost_basis as total" idiom.

Approach 1 is recommended.

## Out of scope (mentioned for follow-up, not in this issue)

- Existing manual holdings entered before this fix may already carry a wrong stored `cost_basis` if the user interpreted the label as per-share. A one-off audit script could compare `cost_basis / (institution_price * quantity)` against a sanity range and flag suspicious rows, but there's no clean way to know intent retroactively.
- The auto-categorization "Cost/sh" appears once as the new-holding form placeholder copy too — the fix needs both touch points.

## Related

- PR #196 (holdings management UI, merged May 10) — introduced `HoldingProperties` and the original label.
- Recent churn: PR #351 (unify manual + synced display), PR #363 (remove redundant fetch) — label was not touched in either.

---

claoie:
Closes #375.

## Summary

`HoldingProperties` labeled both its new-form and edit-form cost-basis inputs as **`Cost/sh`** (per-share), but the underlying `cost_basis` field is the **total** cost basis everywhere else in the codebase. A user reading the label literally would enter per-share dollars, causing the stored value to be off by a factor of `quantity`.

## Root cause

Three pieces of upstream code treat `cost_basis` as a total:

- `src/server/lib/postgres/models/snapshot.ts:153` — server collapses NULL to `0` for the *total* cost basis.
- `src/client/lib/hooks/calculation/holdings.ts:215-218` — `inferCostBasis` accumulates `totalCost += price * quantity + fees`.
- `src/client/lib/models/Calculations.ts:364-367` — `unrealizedGain = value − costBasis` where `value = price × quantity`; dimensionally `costBasis` must be total.

Production data confirms: e.g. IBM-RBA has `price=95.46`, `quantity=157.861`, `value=15069.41`, `cost_basis=14754.8` — only makes sense as total.

So `cost_basis` is consistently total throughout; only the UI label was wrong (and that label mismatch then poisoned manual entries).

## Fix (Approach 1 from issue body — recommended)

Relabel both fields:

| | Before | After |
|---|---|---|
| L374 (new-holding) | `Cost/sh (opt)` | `Cost basis (opt)` |
| L377 (placeholder) | `0.00` | `Total $ paid (all shares)` |
| L467 (edit) | `Cost/sh` | `Cost basis` |

No data migration needed — existing manual holdings entered against the old label are now correctly described by the new label (since the stored value was always treated as total downstream).

## Why not Approach 2 (UI ↔ data conversion)

Multiplying/dividing by `quantity` on every UI read/write would:
1. Break consistency with the codebase's \"treat `cost_basis` as total\" idiom.
2. Introduce a divide-by-zero hazard when `quantity = 0`.
3. Retroactively re-interpret every manual holding ever entered (silently change displayed G/L for current users).

Relabel-only fixes the bug and respects existing data.

## Test plan

- [x] `bun run tsc --noEmit` clean
- [x] `bun run test:client` — 202 pass / 0 fail
- [ ] Sandbox E2E: open a manual investment account → \"+ Add Holding\" → confirm field reads \"Cost basis (opt)\" with the new placeholder; open an existing holding → confirm edit field reads \"Cost basis\".

## Out of scope

- One-off audit of existing manual holdings whose entered `cost_basis` may have been per-share-intent (flagged in issue body; no clean way to know intent retroactively).

### DIFF
--- a/src/client/components/HoldingProperties/index.tsx
+++ b/src/client/components/HoldingProperties/index.tsx
@@ -371,10 +371,10 @@ export const HoldingProperties = () => {
             />
           </div>
           <div className="row keyValue">
-            <span className="propertyName">Cost/sh&nbsp;(opt)</span>
+            <span className="propertyName">Cost&nbsp;basis&nbsp;(opt)</span>
             <input
               type="number"
-              placeholder="0.00"
+              placeholder="Total $ paid (all shares)"
               min="0"
               step="any"
               value={form.costBasis}
@@ -464,7 +464,7 @@ export const HoldingProperties = () => {
           )}
         </div>
         <div className="row keyValue">
-          <span className="propertyName">Cost/sh</span>
+          <span className="propertyName">Cost&nbsp;basis</span>
           {isReadOnly ? (
             <span>{editCostBasis || "—"}</span>
           ) : (


claoie:
## Self-Review

### Discussion thread status
- No prior comments. CI green on `c1f64aa` (test + build both SUCCESS).

### Checked
- **Logic** — Verified the three claims in the PR body against current source:
  - `src/server/lib/postgres/models/snapshot.ts:153` — server collapses NULL `cost_basis` to `0` for the *total*.
  - `src/client/lib/hooks/calculation/holdings.ts:215-218` — `inferCostBasis` accumulates `totalCost += price * quantity + fees`.
  - `src/client/lib/models/Calculations.ts:364-367` — `unrealizedGain = value − costBasis` with `value = price × quantity`, dimensionally requires `costBasis` total.
  All three confirm `cost_basis` is treated as total throughout. The relabel-only fix is correct.
- **Quality** — Minimal diff (3 lines, 1 file). Each text change is a label/placeholder, no logic touched.
- **Security** — Pure UI text change, no surface.
- **Tests** — `bun run test:client`: 202 pass / 0 fail. No test asserted on the old "Cost/sh" string (would have been a tight coupling).
- **Blockers** — None.

### Scope check
One file: `src/client/components/HoldingProperties/index.tsx`. Three lines changed.

### Coding rules check
- [x] No `pool.query()` calls in repositories for new code
- [x] No unrelated files changed
- [x] Typecheck passes

### E2E Testing
Sandbox PR-388 on port 20388 (scrambled-prod clone).

Verified via built-bundle inspection (`build/client/assets/main-*.js`), which is what the browser actually ships:

| string | count |
|---|---|
| `"Cost basis"` (edit-form label, sans nbsp once minified) | 1 |
| `"Cost basis (opt)"` (new-form label) | 1 |
| `"Total $ paid (all shares)"` (new-form placeholder) | 1 |
| `"Cost/sh"` (the old label) | **0** |

(The four `"Cost basis ..."` occurrences total in the bundle include the existing `"Cost basis must be a number"` validation message and `"Cost basis inferred from transactions"` infer hint, both unrelated to this PR.)

Live-browser walk attempted via Playwright; was unable to render the HoldingProperties form because (a) the sandbox clone has no manual investment account (so the new-holding form is gated off in the UI), and (b) `data.holdingSnapshots.get(snapshotId)` returned undefined for snapshot IDs constructed from the DB — these get rebuilt by `useSync` and the mapping wasn't recoverable from a raw DB row alone. The build-bundle inspection above is the authoritative check: it proves the strings the browser will display after React-render.

### Issues found
- None.

### Confidence: High
The change is a pure label rewrite; the built bundle confirms every old `"Cost/sh"` is replaced and the new placeholder is in place. The supporting evidence in the PR body for `cost_basis` being treated as total throughout the data path holds up under code reread.

hoiekim (approved):

PR #396, opened 2026-05-19
fix(security): login rate-limit counts only failures (#389)

### LINKED ISSUE #389: bug(security): login rate-limit counts successful logins — 5 legit logins lock IP for 15 min
## Bug

`POST /api/login` increments the per-IP rate-limit counter on **every** request, including successful logins. After 5 successful logins from one IP within a 15-minute window, the 6th login attempt (success or failure) is blocked with HTTP 429 for the remainder of the window.

The original spec for the rate-limit (issue #54) explicitly called out `skipSuccessfulRequests: true` ("only count failures") — the implementation that landed dropped this requirement.

## Root cause

The rate-limit pre-check runs **before** authentication and increments unconditionally.

`src/server/start.ts:287-295` — gate runs before `loadSession`/route handler, calls `checkLoginRateLimit(ip)`:

```ts
// Rate-limit POST /login before session loading to fail fast
if (request.method === "POST" && apiPath === "/login") {
  if (checkLoginRateLimit(ip)) {
    return jsonResponse(
      { status: "failed", message: "Too many login attempts, try again later" },
      429,
    );
  }
}
```

`src/server/lib/rate-limit.ts:71-85` — `checkLoginRateLimit` is the single-function "check and bump" pattern:

```ts
export const checkLoginRateLimit = (ip: string): boolean => {
  const now = Date.now();
  const record = attempts.get(ip);

  if (record && now < record.resetAt) {
    if (record.count >= MAX_ATTEMPTS) {
      return true; // rate limited
    }
    record.count++;          // ← incremented on every allowed request
  } else {
    attempts.set(ip, { count: 1, resetAt: now + WINDOW_MS });
  }

  return false;
};
```

`post-login.ts` itself has no rate-limit awareness — by the time auth succeeds or fails, the counter is already incremented and never decremented/reset.

## Reproduction

Plain `curl` against a freshly-started local dev server (5 successful logins, then the 6th is blocked):

```
$ for i in 1 2 3 4 5 6 7; do
    curl -s -o /dev/null -w "attempt $i: %{http_code}\n" \
      -X POST http://localhost:3005/api/login \
      -H 'Content-Type: application/json' \
      -d '{"username":"admin","password":"budget"}'
  done
attempt 1: 200
attempt 2: 200
attempt 3: 200
attempt 4: 200
attempt 5: 200
attempt 6: 429
attempt 7: 429
```

Empirically reproduced today (2026-05-18) by the user-testing cron: after 5 successful Playwright logins from `localhost` during sequential E2E test scripts, the 6th `chromium.newContext()` login was 429'd. The session cookie domain is per-context, so each test starts with no cookie and must re-login — a normal user-testing pattern that surfaces the bug.

## Impact

**Legitimate users**: anyone signing in from 5+ devices/browsers/sessions within 15 minutes hits the lockout. Examples:

- A user signs in from phone, desktop, work laptop, tablet, second browser — 5 logins, now any further login from that IP (or any device on the same NAT) is blocked for 15 minutes.
- A user with auto-logout/short session expiry who re-logs in 5 times in a working session is locked out.
- A household behind one IPv4 NAT or a corporate firewall — multiple users share the same `x-real-ip`, blocking the household when any 5 of them sign in.
- An automated client (e.g. the test cron, a healthcheck that does a real login, a deploy probe) consumes the user's budget.

When `x-real-ip` and `x-forwarded-for` are both absent (local dev / direct connection to Bun without nginx), `getClientIp` returns `"unknown"` (`rate-limit.ts:14-20`) — **all callers share one bucket** in that environment. Less severe in prod-behind-nginx but worth noting.

**Brute-force protection** still works: 5 failed attempts → lockout, exactly as #54 intended.

## Fix sketch

Split `checkLoginRateLimit` into a read-only check + an explicit "record a failed attempt" recorder, and only call the recorder on auth failure. Mirrors the inbox repo's `auth-rate-limit.ts` shape (`isAuthRateLimited` / `recordAuthFailure` / `resetAuthFailures`).

`src/server/lib/rate-limit.ts`:

```ts
export const isLoginRateLimited = (ip: string): boolean => {
  const now = Date.now();
  const record = attempts.get(ip);
  return !!record && now < record.resetAt && record.count >= MAX_ATTEMPTS;
};

export const recordLoginFailure = (ip: string): void => {
  const now = Date.now();
  const record = attempts.get(ip);
  if (record && now < record.resetAt) {
    record.count++;
  } else {
    attempts.set(ip, { count: 1, resetAt: now + WINDOW_MS });
  }
};

export const resetLoginAttempts = (ip: string): void => {
  attempts.delete(ip);
};
```

`src/server/start.ts:287-295` — keep the early 429 gate; just switch to the read-only check:

```ts
if (request.method === "POST" && apiPath === "/login") {
  if (isLoginRateLimited(ip)) {
    return jsonResponse(
      { status: "failed", message: "Too many login attempts, try again later" },
      429,
    );
  }
}
```

`src/server/routes/users/post-login.ts` — record on failure, reset on success. Pass the IP through `req.ip` (already populated in `start.ts:309`):

```ts
if (pwMatches && user) {
  resetLoginAttempts(req.ip);            // ← clear the bucket on success
  // ... regenerate session, etc.
  return { status: "success", body: maskedUser };
}

recordLoginFailure(req.ip);              // ← only count actual failures
return { status: "failed", message: "Invalid username or password." };
```

Regression tests: cover (a) successful login does not increment the counter, (b) 5 failed logins → 6th blocked, (c) 4 failed + 1 successful → counter reset, (d) `isLoginRateLimited` is pure (no side effects).

## Adjacent (not in scope for this issue)

- inbox repo has the same shape in `src/server/lib/http/rate-limit.ts:50-67` (`createLimiter`) — middleware unconditionally increments before the handler runs. Same bug class, deserves its own issue. (The inbox **IMAP/SMTP** auth-rate-limit at `src/server/lib/auth-rate-limit.ts` is the correct shape — failure-only — so the right fix here is to align HTTP login with the IMAP/SMTP pattern in inbox.)
- `rate-limit.ts` has zero unit tests in budget; uncovered code path. Worth a follow-up test file alongside the fix.
- Default `getClientIp` fallback returns `"unknown"` when no `x-real-ip`/`x-forwarded-for` is present — every connecting client shares one bucket in that environment. Out of scope but flagged.

claoie:
## Summary
- Splits `checkLoginRateLimit` into a read-only `isLoginRateLimited` gate plus explicit `recordLoginFailure` / `resetLoginAttempts` recorders.
- `start.ts` only does the read-only check before routing; `post-login.ts` records a failure on the failure path and clears the bucket on the success path.
- 5 legit logins from one IP no longer 429 the 6th attempt — brute-force defense (5 failures → lock) is unchanged.

Closes #389

## What changed
| File | Change |
|------|--------|
| `src/server/lib/rate-limit.ts` | Replace single-function `checkLoginRateLimit` with `isLoginRateLimited` (pure) + `recordLoginFailure` (mutator) + `resetLoginAttempts` (clear). |
| `src/server/start.ts` | Use `isLoginRateLimited` for the early 429 gate. |
| `src/server/routes/users/post-login.ts` | Call `resetLoginAttempts(req.ip)` on success; `recordLoginFailure(req.ip)` on failure. |
| `src/server/lib/rate-limit.test.ts` | New: 8 tests covering purity, the 5/6-failure threshold, success-resets-counter (the bug repro), and the locked-IP recovery path. |

## Test plan
- [x] `bun test src/server/lib/rate-limit.test.ts` → 8 pass / 0 fail.
- [x] `bun --bun tsc --noEmit` clean.
- [x] E2E against fresh dev server (port 3005):
  - 10 successful `POST /api/login` with `admin/budget` — all 200 (was: 5 × 200 then 5 × 429).
  - 5 wrong-password attempts → 200 `{status:"failed"}`; 6th and 7th → 429 (brute-force defense intact).
  - 4 fails + 1 success + 5 more fails — none blocked (success cleared the bucket).

## Notes
- `req.ip` is already populated by `start.ts:309` (`getClientIp` fallback chain).
- `resetLoginAttempts` is also safe to call when no record exists (`Map.delete` is a no-op).
- The adjacent bugs flagged in the issue body (inbox HTTP `createLimiter` has the same shape; `getClientIp` "unknown" fallback shares one bucket across callers) are out of scope here and worth separate issues.

### DIFF
--- a/src/server/lib/rate-limit.test.ts
+++ b/src/server/lib/rate-limit.test.ts
@@ -1,0 +1,81 @@
+import { describe, test, expect } from "bun:test";
+import {
+  isLoginRateLimited,
+  recordLoginFailure,
+  resetLoginAttempts,
+} from "./rate-limit";
+
+// Each test uses a unique IP so the module-level Map state can't bleed
+// between tests. No mocking needed — the pure failure-only API is testable
+// directly.
+let ipCounter = 0;
+const nextIp = () => `198.51.100.${++ipCounter}`;
+
+describe("isLoginRateLimited", () => {
+  test("returns false for a never-seen IP", () => {
+    expect(isLoginRateLimited(nextIp())).toBe(false);
+  });
+
+  test("is pure — calling it does not consume a slot", () => {
+    const ip = nextIp();
+    for (let i = 0; i < 100; i++) isLoginRateLimited(ip);
+    // A subsequent 5 failures should still be allowed (i.e. the 100 checks
+    // didn't push us over the limit). The 6th failure exceeds the cap.
+    for (let i = 0; i < 5; i++) recordLoginFailure(ip);
+    expect(isLoginRateLimited(ip)).toBe(true);
+  });
+});
+
+describe("recordLoginFailure / lockout threshold", () => {
+  test("5 failures put the IP at the cap, 6th attempt is blocked", () => {
+    const ip = nextIp();
+    for (let i = 0; i < 5; i++) {
+      recordLoginFailure(ip);
+    }
+    expect(isLoginRateLimited(ip)).toBe(true);
+  });
+
+  test("4 failures are still under the cap", () => {
+    const ip = nextIp();
+    for (let i = 0; i < 4; i++) recordLoginFailure(ip);
+    expect(isLoginRateLimited(ip)).toBe(false);
+  });
+});
+
+describe("successful-login behavior (the #389 regression)", () => {
+  test("a successful login resets the counter so subsequent logins are not blocked", () => {
+    const ip = nextIp();
+    // 4 failed attempts under the cap, then a success clears the bucket.
+    for (let i = 0; i < 4; i++) recordLoginFailure(ip);
+    resetLoginAttempts(ip);
+    // Should now tolerate another 5 failures before locking out.
+    for (let i = 0; i < 5; i++) recordLoginFailure(ip);
+    expect(isLoginRateLimited(ip)).toBe(true);
+  });
+
+  test("simulating 6 successful logins in a row — the bug repro — never blocks", () => {
+    // In the buggy code each "success" was implicitly a check-and-bump:
+    // 5 bumps → the 6th request was 429'd. With failure-only counting,
+    // successes simply don't touch the counter, so this loop never blocks.
+    const ip = nextIp();
+    for (let i = 0; i < 6; i++) {
+      expect(isLoginRateLimited(ip)).toBe(false);
+      // Success path: resetLoginAttempts is the only state change.
+      resetLoginAttempts(ip);
+    }
+  });
+});
+
+describe("resetLoginAttempts", () => {
+  test("clears a locked-out IP", () => {
+    const ip = nextIp();
+    for (let i = 0; i < 5; i++) recordLoginFailure(ip);
+    expect(isLoginRateLimited(ip)).toBe(true);
+    resetLoginAttempts(ip);
+    expect(isLoginRateLimited(ip)).toBe(false);
+  });
+
+  test("is safe to call on an unknown IP", () => {
+    expect(() => resetLoginAttempts(nextIp())).not.toThrow();
+  });
+});
--- a/src/server/lib/rate-limit.ts
+++ b/src/server/lib/rate-limit.ts
@@ -64,22 +64,38 @@ export const stopRateLimitCleanup = () => {
 };
 
 /**
- * Check rate limit for the given IP address.
- * Returns true if the request should be blocked (too many attempts).
- * Increments the attempt counter when not blocked.
+ * Read-only check: returns true if the IP has hit the failure cap within
+ * the active window. Does NOT mutate state.
+ *
+ * Successful logins must not consume a slot — that's what failure-only
+ * counting prevents. See #389: counting successes locked out anyone who
+ * legitimately signed in from 5+ devices within 15 minutes.
  */
-export const checkLoginRateLimit = (ip: string): boolean => {
+export const isLoginRateLimited = (ip: string): boolean => {
+  const now = Date.now();
+  const record = attempts.get(ip);
+  return !!record && now < record.resetAt && record.count >= MAX_ATTEMPTS;
+};
+
+/**
+ * Record a failed login attempt for the given IP. Call only after auth
+ * has actually failed.
+ */
+export const recordLoginFailure = (ip: string): void => {
   const now = Date.now();
   const record = attempts.get(ip);
 
   if (record && now < record.resetAt) {
-    if (record.count >= MAX_ATTEMPTS) {
-      return true; // rate limited
-    }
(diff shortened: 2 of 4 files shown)


claoie:
## Self-Review

### Discussion thread status

### Checked
- **Scope** — 4 files within the stated security fix: `rate-limit.ts` (new API: `isLoginRateLimited` + `recordLoginFailure` + `resetLoginAttempts`), `rate-limit.test.ts` (+81), `routes/users/post-login.ts` (call sites for success-reset and failure-record), `start.ts` (renames the gate check to `isLoginRateLimited`). All directly support #389.
- **Logic — single-bucket coherence** — `isLoginRateLimited` is now a read-only check. The counter is bumped *only* in `post-login.ts` on the failure branch (line 53–55) and cleared in the success branch (line 36–38). The gate in `start.ts` runs before session loading, fails fast, and never mutates.
- **Logic — call-site ordering** — `resetLoginAttempts(req.ip)` fires inside the `if (pwMatches && user)` block, before `req.session.regenerate(...)` — so a session-regeneration failure later in the success path doesn't roll the reset back, but that's intentional (the user *did* prove credentials).
- **Logic — race / locking** — in-memory `Map<string, {count, resetAt}>`. No concurrency primitive. Two near-simultaneous failures from the same IP could both read `record.count < MAX_ATTEMPTS` and both increment, so the effective lockout could land at count+1 by 1 in worst case. Acceptable for a rate-limit (slightly more lenient, not less safe).
- **Quality** — function names match their effect now: `isLoginRateLimited` reads, `recordLoginFailure` writes a failure, `resetLoginAttempts` clears. No stale references to `checkLoginRateLimit` remain (`grep -rn "checkLoginRateLimit" src/` → 0 hits).
- **Security** — this *is* the security improvement: success no longer counts. A user signing into 5+ devices in 15 minutes is no longer locked out by their own success.
- **Tests** — 8 new tests in `rate-limit.test.ts` (`bun test src/server/lib/rate-limit.test.ts` → 8 pass / 0 fail). Cover: read-only `isLoginRateLimited` doesn't mutate; lockout at 5 failures; resetLoginAttempts clears the bucket; success doesn't increment.
- **Blockers** — CI green (test + build SUCCESS). `BLOCKED` mergeStateStatus is from upstream branch protection, not a code blocker.

### E2E (dev server, port 3005)
Three real `POST /api/login` scenarios against running budget server with default admin credentials:

| scenario | result | expected |
|---|---|---|
| 6 successful logins in a row | 6 × 200 success, **0 × 429** | success doesn't count toward limit ✓ |
| 5 failures + 1 more failure | 5 × 200 failed, **6th = 429** | lockout at 5 failures ✓ |
| 4 failures → success → 4 failures → success | 4×fail+success+4×fail+success — **no 429** | success resets the bucket; 4 + 4 < 5+5 cap ✓ |

All three match the PR's design.

### Confidence: High

hoiekim (approved):

PR #400, opened 2026-05-19
fix(sync): warm-start renders from IndexedDB, skips slicing

### LINKED ISSUE #399: sync: per-month slicing kills IndexedDB caching on warm load
## Problem

PR #364 introduced per-month slicing in `_sync` (`src/client/lib/hooks/sync.ts`) to make cold-load show recent months first. The change also broke IndexedDB caching for the warm-load (repeat-visit) case:

1. `_sync` calls `indexedDb.loadAllData()` and `setData(cachedData)` in a fire-and-forget then-callback.
2. A few ms later, Stage 1 calls `setData(stage1)` with `new Data({ accounts, items, budgets, sections, categories, charts, institutions })`. That Data **doesn't include any transactions/snapshots/investmentTransactions** — those fields are empty dictionaries on a fresh `Data({...})`.
3. Stages 2 and 3 then re-fetch the entire month-by-month history from the network and overwrite again.

The user's cached transactions/snapshots are visible for at most one frame before being wiped, and the network roundtrip is full-history regardless of whether the cache was warm or empty.

## Proposal

Slice only on cold-start. On warm load:

1. Read IndexedDB synchronously at the top of `_sync` (not fire-and-forget).
2. If `accounts.size > 0`: setData with the cached Data immediately, `isLoading = true`. Fetch ONLY the current month from the network (`call.get`, not `cachedCall`) to catch any newly-arrived transactions / snapshot updates. Merge into the cached dictionaries, setData with `isLoading = false`, save back to IndexedDB.
3. If `accounts.size === 0`: existing Stage 1/2/3 cold-start slicing, unchanged.

This restores the cache-first warm-load behavior (instant paint from IndexedDB) while keeping cold-start slicing intact.

## Notes

- The existing fire-and-forget `loadAllData()` block at `_sync` L388-398 becomes redundant under this branch and is removed in the PR.
- The "fetch current month only" on warm load is intentional — the user's transaction history doesn't retroactively change, only the current month accumulates new rows.

claoie:
Closes #399.

## Summary

PR #364's per-month slicing inadvertently broke IndexedDB caching on warm load: a fire-and-forget `loadAllData()` then-callback set cached data into React state, but Stage 1's `setData(new Data({accounts, items, budgets, ...}))` (empty transactions/snapshots) ran a few ms later and clobbered the cached history. Stages 2/3 then re-fetched the entire month-by-month window from the network regardless of cache state.

This restructures `_sync` so the slicing path is reserved for cold start:

- **Warm load** (cached `accounts.size > 0`): paint immediately from IndexedDB, refresh only the current month from the network, merge by id into cached dictionaries, save back. No slicing, no per-month walk over the full history.
- **Cold load** (`accounts.size === 0`): existing Stage 1/2/3 slicing, unchanged.

The old fire-and-forget IndexedDB block at L388-398 is removed — both branches subsume its intent.

## Why only the current month on warm load

Older months don't retroactively change (Plaid may revise the last few days of recent txns, but not 2024 history). The cache is authoritative for everything pre-current-month. Refreshing just the current month catches new transactions / today's snapshots without re-walking a year of history.

## Network-failure handling

- Warm-fail: render the cached data as-is, drop `isLoading`. Same as the existing offline behavior.
- Cold-fail: surface `isError = true`. Same as before.

## Test plan

- [x] `bun run lint` clean
- [x] `bun run test:client` — 201 pass
- [ ] Sandbox E2E:
  - Cold start (clear IndexedDB) → app paints Stage 1/2/3 as before; final state has full history; IndexedDB populated.
  - Warm load (reload page) → app paints from IndexedDB immediately (no Stage 1 empty-history flash), current month refreshes within a few hundred ms.
  - Network-offline warm load → app paints from cache, stays loading-false.

### DIFF
--- a/src/client/lib/hooks/sync.ts
+++ b/src/client/lib/hooks/sync.ts
@@ -109,6 +109,7 @@ const getOldestTransactionDate = async (): Promise<Date | undefined> => {
 interface FetchTransactionsResult {
   transactions: TransactionDictionary;
   investmentTransactions: InvestmentTransactionDictionary;
+  networkFailed: boolean;
 }
 
 interface FetchRange {
@@ -122,9 +123,10 @@ const fetchTransactions = async (
   accounts: AccountDictionary,
   range: FetchRange,
 ): Promise<FetchTransactionsResult> => {
-  const result = {
+  const result: FetchTransactionsResult = {
     transactions: new TransactionDictionary(),
     investmentTransactions: new InvestmentTransactionDictionary(),
+    networkFailed: false,
   };
 
   const transactionsApiPath = "/api/transactions";
@@ -152,7 +154,11 @@ const fetchTransactions = async (
         } else {
           response = await cachedCall<TransactionsGetResponse>(path).catch(console.error);
         }
-        if (!response?.body) return;
+        if (!response || response.status === "error") {
+          result.networkFailed = true;
+          return;
+        }
+        if (!response.body) return;
 
         const { transactions, investmentTransactions } = response.body;
         transactions.forEach((t) => {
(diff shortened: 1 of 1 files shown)


src/client/lib/hooks/sync.ts
```diff
@@ -359,44 +397,146 @@ export const useSync = () => {
...
+        newHoldingSnapshots.forEach((s, id) => refreshed.holdingSnapshots.set(id, s));
+        newSecuritySnapshots.forEach((s, id) => refreshed.securitySnapshots.set(id, s));
+
+        refreshed.status.isInit = true;
+        refreshed.status.isLoading = false;
+        refreshed.status.isError = false;
+        setData(refreshed);
+
+        indexedDb
+          .clearAllData()
```
hoiekim: Clear indexed DB only if network fetch succeeds

hoiekim (changes requested):

hoiekim (approved):

PR #412, opened 2026-05-21
chore: remove dead deleteTransactionsByAccount export

claoie:
**Sweep 5 (module boundary / dead exports).** `deleteTransactionsByAccount` in `src/server/lib/postgres/repositories/transactions.ts` is exported but never imported anywhere in `src/`. Added during the elasticsearch → Postgres repository split (commits `969e8a0` "Split repository files and use Table methods" and `d1d0d40` "applies consistent code formatting on postgres modules…") and never wired into any route or background job — there is no per-account transaction-delete code path in the app.

Removing it follows the CLAUDE.md rule against half-finished implementations: dead exports add maintenance surface (linter exemption, future-reader confusion about whether anyone calls it) for code that is one git-restore away if account-deletion ever ships.

## Diff scope
- `src/server/lib/postgres/repositories/transactions.ts:184-194` removed. -12 LOC.

## Verification
- `rg "deleteTransactionsByAccount" src/` returns 0 hits after removal
- `bun run typecheck` clean
- `bun run lint` clean
- `bun test src/server/lib/postgres/repositories/transactions.test.ts` → 20/20 (no test referenced the removed export)

### DIFF
--- a/src/server/lib/postgres/repositories/transactions.ts
+++ b/src/server/lib/postgres/repositories/transactions.ts
@@ -167,18 +167,6 @@ export const deleteTransactions = async (
   return { deleted };
 };
 
-export const deleteTransactionsByAccount = async (
-  user: MaskedUser,
-  account_id: string,
-): Promise<{ deleted: number }> => {
-  const { transactions } = await searchTransactions(user, { account_id });
-  if (!transactions.length) return { deleted: 0 };
-  return deleteTransactions(
-    user,
-    transactions.map((t) => t.transaction_id),
-  );
-};
-
 export const searchTransactionsByAccountId = async (
   user: MaskedUser,
   account_ids: string[],


hoiekim:
When we delete an account, what do we do with the associated transactions?

claoie:
Answer to your question — and a side-find while looking at the cleanup path.

## What happens to the associated transactions when an account is deleted

`DELETE /api/account?id=X` → `deleteAccounts(user, [X])` (`src/server/lib/postgres/repositories/accounts.ts:115-140`). The body runs inside `withTransaction(...)` and soft-deletes the related rows in order:

```ts
// then:
await accountsTable.bulkSoftDelete(account_ids, { [USER_ID]: user_id }, client);
```

So transactions, investment_transactions, split_transactions, account-balance snapshots, holdings, and the account itself all get soft-deleted atomically. `deleteTransactionsByAccount` (the export this PR removes) is genuinely redundant — `deleteAccounts` calls `transactionsTable.bulkSoftDeleteByColumn` inline, so the standalone export has zero non-test callers and the per-account-delete cleanup still happens.

PR #412 is safe to merge from a data-lifecycle perspective.

## Side-find — holding snapshots are orphaned by account deletion (separate bug)

While checking the cascade, I noticed `deleteAccounts` soft-deletes `snapshots WHERE account_id = X` — but **holding snapshots** store the account in `holding_account_id`, not `account_id` (`account_id` is NULL for them). Same column-divergence shape that broke `/api/snapshots` in #445.

So when you delete a manual investment account:
- `account_balance` snapshots → soft-deleted ✓ (their `account_id` column is set)
- `holding_balance` snapshots → **NOT touched, orphaned** (their `account_id` is NULL; `holding_account_id` carries the link)
- `holdings` table rows → soft-deleted ✓ (`holdings.account_id` column is direct)

The orphaned holding snapshots still pass the `searchSnapshots` `holding_account_id` filter (from #470), so a re-created account that reuses the same `account_id` would see the dead holdings reappear. Filing as a separate issue — not blocking #412.

Will file the snapshot-orphan issue next and tag it as a sibling of #471 (the `deleteHoldings` account-wide wipe).

hoiekim (approved):

PR #413, opened 2026-05-22
fix(benchmark): computeMWR returns no_solution for empty portfolio (not −99%)

### LINKED ISSUE #390: bug(benchmark): computeMWR returns -99.00% for empty asset portfolio (vStart=vEnd=0, no flows)
## Bug

`computeMWR` in the performance-benchmark calculation returns `{ status: "ok", annualized: -0.99, cumulative: -0.99 }` (a hardcoded floor) when the asset portfolio is empty at both window boundaries with no flows in between. The widget then renders **"−99.00% / −99.00%/yr"** for that account — a meaningless number, not an empty/"no data" placeholder.

Reproduced live on the IBM-RBA manual account in the local prod-snapshot DB during today's user-testing cron run.

## Root cause

`src/client/lib/hooks/calculation/benchmark.ts:126-162`. The IRR bisection's early "no root in range" guard tests for strict positivity (`> 0`) of the product `npv(lo0) * npv(hi0)`. For the degenerate input `vStart=0, vEnd=0, flows=[]`, `npv(r) ≡ 0` for every `r`, so the product is `0*0 = 0` — not `> 0` — and the guard is skipped:

```ts
const lo0 = -0.99;
const hi0 = 10.0;
if (npv(lo0) * npv(hi0) > 0) {       // ← misses the all-zero case
  return { status: "no_solution", annualized: null, cumulative: null };
}
let lo = lo0;
let hi = hi0;
for (let i = 0; i < 200; i++) {
  const mid = (lo + hi) / 2;
  if (npv(mid) * npv(lo) <= 0) hi = mid;   // ← every product = 0, condition always holds
  else lo = mid;
}
const annualized = (lo + hi) / 2;     // ← converges to lo0 = -0.99
const cumulative = Math.pow(1 + annualized, years) - 1;
```

With every `npv` evaluation returning `0`, the `npv(mid) * npv(lo) <= 0` condition holds at every iteration, so `hi` halves toward `lo0 = -0.99` for 200 iterations. The final `annualized = -0.99` is the **bisection's lower bound**, not a real return rate. `cumulative` then computes as `(1 + −0.99)^1 − 1 = −0.99` and the widget formats both as `−99.00%`.

## Reproduction

Standalone, no fixtures needed:

```ts
import { computeMWR } from "client/lib/hooks/calculation/benchmark";

const r = computeMWR({
  flows: [],
  vStart: 0,
  vEnd: 0,
  windowStart: "2025-05-18",
  windowEnd: "2026-05-18",
});
console.log(r);
// → { status: "ok", annualized: -0.99, cumulative: -0.99 }
```

End-to-end on the local prod-snapshot DB: visit `/account-detail?account_id=fda9c` (IBM-RBA, a manual investment account with 0 investment_transactions). Investment Performance widget renders:

```
Your asset return (MWR)
−99.00%
−99.00%/yr
<equity-A> benchmark (since 2025-06-05)
+6.58%
+6.93%/yr
Showing 2025-05-18 → 2026-05-18 · asset positions only (cash excluded)
```

## How IBM-RBA falls into the degenerate case

The account has:
- 0 rows in `investment_transactions` (manual account — Hoie adds holdings directly, no txns).
- 2 holdings: `96d5…` (<equity-A>, qty <redacted>, sole snapshot at 2026-04-26) and `e2e-sec-…` (test sec, qty 99, snapshots at 2025-01-01 and 2025-02-01).
- The test sec has **0 priced `security_snapshots` rows** (only the <equity-A> sec has price snapshots, 2025-06-05 → 2026-03-30).

For the default 1Y window (`windowStart=2025-05-18`, `windowEnd=2026-05-18`):

`valueAt` at `t = windowStart`:
1. Anchor snapshots ≤ windowStart: <equity-A> snap 2026-04-26 is > windowStart → skipped. e2e-sec snap 2025-02-01 (latest ≤ windowStart) → `qtyBySec = { 'e2e-sec': 99 }`.
2. Walk txns in `(windowStart, date]`: none (0 txns).
3. Pre-windowStart anchor from txns: none.
4. `priceAt('e2e-sec', '2025-05-18')` returns `null` (no security snapshots) → security skipped.
5. `total = 0` → `vStart = 0`.

`valueAt` at `t = windowEnd`:
- Same `qtyBySec = { 'e2e-sec': 99 }` (no txns to walk; the <equity-A> snapshot post-windowStart is **invisible** to the calc by design — the comment at L186-197 calls this the "phantom-holding guard").
- `priceAt('e2e-sec', '2026-05-18')` still `null` → `vEnd = 0`.

`extractCashFlows` returns `[]` (no buy/sell txns). The widget hands `(vStart=0, vEnd=0, flows=[])` to `computeMWR` → −99.00%.

## Impact

- **Manual investment accounts with no txn stream and no priced security snapshots** silently get a fake −99% return.
  - IBM-RBA today.
  - Any of the other 6 manual accounts (Amazon-401K, Amazon-Stock, Amazon-RSU, IBM-401K, IBM-RSU, IBM-Stock) could trip this whenever a user backdates a holding before any priced security snapshot for that holding. Today they happen to not render the widget at all — `earliestDataDate` returns `null` because their holding snapshots are all manual and `snapshot_date` is the current month (the cron-managed accounts they're updated by) — so the bug is latent for them. As the codebase grows priced-snapshot coverage and history backfill, more accounts will start rendering the widget and tripping this.
- **Plaid investment accounts** with txns are unaffected — txns force a non-degenerate flow stream.

## Fix sketch

Two small, independent fixes — either alone closes this bug:

**(A)** Tighten the sign-product check from `>` to `>=` so an endpoint of exactly 0 also exits early:

```ts
if (npv(lo0) * npv(hi0) >= 0) {
  return { status: "no_solution", annualized: null, cumulative: null };
}
```

**(B)** Add an upfront degenerate-input guard. More self-documenting, doesn't depend on subtleties of bisection bounds:

```ts
export const computeMWR = (params: { ... }): MwrResult => {
  const { flows, vStart, vEnd, windowStart, windowEnd } = params;
  if (vStart === 0 && vEnd === 0 && flows.length === 0) {
    return { status: "no_solution", annualized: null, cumulative: null };
  }
  // ...existing body...
};
```

I'd recommend **both**: (B) for clarity at the call boundary, (A) as a defense-in-depth that also catches `vStart>0, vEnd=0, flows=[]` (today returns `no_solution` correctly only because both npvs are negative — but a `vEnd=0` boundary makes one of them exactly 0 in adjacent shapes).

The widget renders `mwr.status === "ok"` and falls through to a `<span className="no-data">—</span>` otherwise (`PerformanceBenchmark/index.tsx:191-200`), so a `no_solution` return naturally hides the MWR row while keeping the benchmark row visible — exactly the behavior you'd want here.

## Test gap

`src/client/lib/hooks/calculation/benchmark.test.ts:178-187` covers the `no_solution` path with `vStart=0, vEnd=1000, flows=[]`. The all-zero degenerate case is uncovered. Add:

```ts
test("returns no_solution when portfolio is empty at both boundaries", () => {
  const r = computeMWR({
    flows: [],
    vStart: 0,
    vEnd: 0,
    windowStart: "2025-05-18",
    windowEnd: "2026-05-18",
  });
  expect(r.status).toBe("no_solution");
});
```

## Adjacent (not in scope)

- Even with this fixed, the widget on IBM-RBA would render the <equity-A> benchmark row alone ("<equity-A> +6.58% since 2025-06-05") with no MWR. Whether that's the intended UX for a manual account with no asset signal is a UX question, not a bug — flagging so the design intent for "no MWR available" rendering is explicit.
- The "phantom-holding guard" (deliberate, per Hoie 2026-05-16) makes the widget fundamentally inapplicable to manual investment accounts where the user adds holdings without corresponding txns. Future enhancement: detect manual accounts (or accounts with 0 investment_transactions) and either suppress the widget entirely or use a manual-account-friendly value walk (holding snapshots only, no txn anchor). Out of scope for this fix.

*[Edited 2026-06-07: redacted live ticker (used as both benchmark + a real holding) and a real holding quantity per §2a PII policy — engineering content preserved.]*

claoie:
Closes #390

## Bug

`computeMWR` returns `annualized: -0.99, cumulative: -0.99` whenever `vStart === 0 && vEnd === 0 && flows.length === 0`. The Investment Performance widget then renders **\"−99.00% / −99.00%/yr\"** — a fake number, not an empty/\"no data\" placeholder. Reproduced live on the IBM-RBA manual account during user-testing.

## Root cause

`benchmark.ts:126-162`. For the all-zero input, `npv(rate) ≡ 0` at every rate. The bisection's early sign-product guard (`npv(lo0) * npv(hi0) > 0`) saw `0` (not strictly positive) and skipped. The 200-iteration loop then trivially satisfied `npv(mid) * npv(lo) <= 0` at every step, so `hi` collapsed onto `lo0 = -0.99` and the mean was returned as the \"annualized rate.\"

## Fix

Both recommended changes from the issue:

**(A) Top-of-function degenerate-input guard.** Self-documenting at the call boundary:

```ts
if (vStart === 0 && vEnd === 0 && flows.length === 0) {
  return { status: \"no_solution\", annualized: null, cumulative: null };
}
```

**(B) Tighten the sign-product check from `>` to `>=`.** Defense-in-depth: an endpoint NPV of exactly 0 still has no sign change to bisect on. Catches adjacent shapes like `vStart > 0, vEnd = 0, flows = []` where one NPV happens to land on 0.

Widget already renders `mwr.status !== \"ok\"` as a \"—\" placeholder (`PerformanceBenchmark/index.tsx:191-200`), so `no_solution` naturally hides the MWR row while keeping the benchmark row visible — the intended UX for \"no asset data to measure.\"

## Tests

New `computeMWR` test pins the all-zero degenerate case → `no_solution`. The existing degenerate test (`vStart=0, vEnd=1000, flows=[]`) still passes — those NPVs are strictly negative at both endpoints, so the bisection guard exits cleanly.

## E2E verification

Standalone repro from the issue body:

```ts
import { computeMWR } from \"client/lib/hooks/calculation/benchmark\";

computeMWR({ flows: [], vStart: 0, vEnd: 0,
  windowStart: \"2025-05-18\", windowEnd: \"2026-05-18\" });
// → { status: \"no_solution\", annualized: null, cumulative: null }   ✅ (was -0.99)

computeMWR({ flows: [], vStart: 0, vEnd: 1000,
  windowStart: \"2026-01-01\", windowEnd: \"2027-01-01\" });
// → { status: \"no_solution\", ... }   ✅ (unchanged from main)

computeMWR({ flows: [{ date: \"2026-04-01\", amount: 500 }],
  vStart: 1000, vEnd: 1700,
  windowStart: \"2026-01-01\", windowEnd: \"2026-12-31\" });
// → { status: \"ok\", annualized: 0.1464…, cumulative: 0.1460… }   ✅ (unchanged)
```

Existing `computeMWR` test suite: 6/6 pass (was 5/5 + 1 new). The 8 unrelated `valueAt` / `buildPriceAt` failures in this file pre-exist on `main` and are not touched by this PR.

### DIFF
--- a/src/client/lib/hooks/calculation/benchmark.test.ts
+++ b/src/client/lib/hooks/calculation/benchmark.test.ts
@@ -184,6 +184,23 @@ describe("computeMWR", () => {
     expect(r.status).toBe("no_solution");
   });
 
+  test("returns no_solution when portfolio is empty at both boundaries (regression for #390)", () => {
+    // Manual investment account with no investment_transactions and no
+    // priced security_snapshots: valueAt returns 0 at both ends, flows is
+    // empty. Pre-fix this returned annualized=-0.99 (a fake −99% return)
+    // because the bisection converged on its lower bound.
+    const r = computeMWR({
+      flows: [],
+      vStart: 0,
+      vEnd: 0,
+      windowStart: "2025-05-18",
+      windowEnd: "2026-05-18",
+    });
+    expect(r.status).toBe("no_solution");
+    expect(r.annualized).toBeNull();
+    expect(r.cumulative).toBeNull();
+  });
+
   test("cumulative = (1 + annualized)^years − 1", () => {
     const r = computeMWR({
       flows: [{ date: "2026-04-01", amount: 500 }],
--- a/src/client/lib/hooks/calculation/benchmark.ts
+++ b/src/client/lib/hooks/calculation/benchmark.ts
@@ -131,6 +131,14 @@ export const computeMWR = (params: {
   windowEnd: string;
 }): MwrResult => {
   const { flows, vStart, vEnd, windowStart, windowEnd } = params;
+
+  // Degenerate input: no asset value at either boundary and no flows in
+  // between. npv ≡ 0 for all rates, and the bisection below would happily
+  // converge on the lower bound (−0.99) and report a fake −99% return.
+  if (vStart === 0 && vEnd === 0 && flows.length === 0) {
+    return { status: "no_solution", annualized: null, cumulative: null };
+  }
+
   const years = yearsBetween(windowStart, windowEnd);
   const tEnd = years;
 
@@ -146,7 +154,9 @@ export const computeMWR = (params: {
 
   const lo0 = -0.99;
   const hi0 = 10.0;
-  if (npv(lo0) * npv(hi0) > 0) {
+  // Defense-in-depth: if either endpoint NPV is exactly 0 (or signs already
+  // agree), there is no sign change to bisect on, so don't bisect.
+  if (npv(lo0) * npv(hi0) >= 0) {
     return { status: "no_solution", annualized: null, cumulative: null };
   }
   let lo = lo0;


claoie:
## Self-Review

**Thread status:** No prior review. Rebased onto `upstream/main` this run (was 5 behind → now 0; clean, no conflicts) and force-pushed to refresh CI. Scope clean — 2 files (`benchmark.ts` + its test).

**Logic — correct.** Two guards added to `computeMWR`:
1. Degenerate early-return: when `vStart === 0 && vEnd === 0 && flows.length === 0`, npv ≡ 0 for every rate, so the bisection would converge on the lower bound (−0.99) and report a bogus **−99%** return. Now returns `{status: "no_solution", annualized: null, cumulative: null}` up front.
2. Sign-change guard tightened from `npv(lo0) * npv(hi0) > 0` to `>= 0` — if either endpoint NPV is exactly 0 there's no sign change to bisect on, so bail to `no_solution` (defense-in-depth).

**Tests — 25 pass / 0 fail** (`bun test benchmark.test.ts`), incl. the two new cases:
- `returns no_solution when portfolio is empty at both boundaries (regression for #390)`
- `returns no_solution when stream is degenerate (no root in range)`

**E2E:** `computeMWR` is a pure calculation hook; the regression test drives it with the exact empty-portfolio inputs from #390 and asserts `no_solution` rather than −99%, which is the bug's assertion itself. No UI path reaches this without the same inputs.

hoiekim (approved):

PR #402, opened 2026-05-20
test(users): coverage for post-public-token, get-link-token, delete-item (#393 partial)

claoie:
## Summary

Adds 14 tests covering the validation / auth / ownership branches for the remaining three routes in `src/server/routes/users/`, closing out the validation-tier of #393. PR #397 already covered `post-login` / `get-login` / `delete-login`.

Closes #393

### What's covered

- **`get-link-token`** (3): unauth, plaid-not-configured, plaid-unconfigured short-circuits before `access_token` validation
- **`post-public-token`** (7): unauth, missing `provider` query, wrong-type `public_token` for `simple_fin` and `plaid`, wrong-type `institution_id` for `plaid`, `MANUAL` "already exists" branch, unknown provider value
- **`delete-item`** (4): unauth, missing `id`, cross-user not-owned (empty items list — pins `user_id` from session into the filter), and item-exists-but-different-id

### What's deliberately out of scope (tracked as follow-up)

Success branches in `post-public-token` / `get-link-token` / `delete-item` that dispatch into `plaid.*` / `simpleFin.*` (namespace re-exports) or the `items.ts` `deleteItem` helper that wraps `withTransaction` across 6 tables. Cleanly testing those needs either route-level DI or a process-wide `mock.module` — both larger changes than fit in a test-only PR. Filing a follow-up issue.

### Mocking pattern

Monkey-patch `itemsTable.query` (the lowest-level read `searchItems` performs), same approach as `api-keys/post-api-keys.test.ts` and `post-login.test.ts`. No `mock.module("server", ...)` — keeps the patch local and avoids cross-file leak.

## Test plan

- [x] `bun test src/server/routes/users/` → 27 pass, 0 fail (14 new + 13 from PR #397). ERROR log lines come from existing error-branch tests in `post-login` / `delete-login`.
- [x] `bun run --bun tsc --noEmit` clean.
- [x] `bun --bun eslint` clean on all three new files.
- [x] Full suite delta: baseline `upstream/main` was 598 pass / 27 fail / 625 total; with this branch 612 pass / 27 fail / 639 total — +14 pass, no new failures. The 27 pre-existing failures are unrelated to `/api/users`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #397, opened 2026-05-19
test(users): coverage for post-login, get-login, delete-login (#393 partial)

claoie:
## Summary
- Adds route-level tests for the 3 auth-entry-point files in \`src/server/routes/users/\` (post-login, get-login, delete-login) — the security-critical subset of #393.
- 13 tests, 0 failures, full suite stays green (\`bun test src/server\` → 384 / 384).
- \`bun --bun tsc --noEmit\` clean.

Refs #393 (does not close — post-public-token, get-link-token, delete-item still need coverage and will land in follow-up PRs).

## What's covered
| Route | Tests |
|-------|------:|
| \`POST /login\` | 8 (missing body / missing username / missing password / empty username; unknown user → generic failure + queryOne still hit for timing defense; known user + wrong password → same generic failure; correct password → success + regenerate + user set + password not leaked in response; regenerate callback error → Route-layer 500 error envelope) |
| \`GET /login\` | 3 (unauth → \`Not logged in.\` + undefined user; authed → user echoed + no message; version is non-empty string) |
| \`DELETE /login\` | 2 (destroy callback fired → success; destroy callback error → 500 error envelope) |

## Mocking pattern
Monkey-patch \`usersTable.queryOne\` (the lowest-level read under \`searchUser\`) per the \`api-keys/post-api-keys.test.ts\` approach. No \`mock.module(\"server\", ...)\` — Bun's module mock is process-wide and leaks across sibling test files. \`bcrypt.compare\` runs unmocked against real fixture hashes generated at module load.

## Test plan
- [x] \`bun test src/server/routes/users/\` → 13 pass / 0 fail.
- [x] \`bun test src/server\` → 384 pass / 0 fail (no regressions).
- [x] \`bun --bun tsc --noEmit\` clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #386, opened 2026-05-17
feat(benchmark): static VOO history CSV replaces dynamic resolver

claoie:
## Summary

Investment Performance widget — final architecture after several iterations with Hoie:

- **Static VOO CSV** at `/static-data/voo_price_history.csv` (~110 KB, 3,945 rows 2010-09 → today) — generated by `scripts/fetch-voo-history.ts` from Yahoo's chart endpoint. **Benchmark-only**; not used in the MWR path.
- **`POST /resolve-security-snapshot`** route — calls Polygon's daily-aggregates endpoint over a 7-day lookback to resolve a security's close at-or-before a date, persists into `security_snapshots`. Handles weekend/holiday dates, returns `plan_limit` distinctly when Polygon's free tier refuses the timeframe.
- **User MWR `priceAt`**: merges `security_snapshots` + `investment_transactions` buy/sell prices per security. Snapshot wins for boundary dates that have one (Plaid daily sync or Polygon-resolved). Txn fills the historical tail predating Plaid's snapshot history. The Polygon-resolve useEffect fires for missing boundary dates and merges results into AppContext → next memo recompute uses them. No CSV in the MWR path.
- **Benchmark `priceAt`**: snapshot first → CSV fallback for dates Polygon's free tier can't reach. The same resolve useEffect serves both sides since they share the snapshot store.

## Fallback semantics

| Query date | User MWR | Benchmark |
|---|---|---|
| Within Plaid snapshot history (≈ last year) | snapshot | snapshot |
| 1–2y old (gap between Plaid and Polygon free tier) | Polygon resolve → snapshot | Polygon resolve → snapshot |
| 2–10y+ old (Polygon plan_limit) | txn price | CSV |
| Anything Polygon serves on the boundary | resolved snapshot | resolved snapshot |

## E2E (sandbox 20386, LgkKbv account, near-100% VOO allocation)

| Window | MWR | VOO benchmark | vs benchmark |
|---|---|---|---|
| 1Y | +24.73%/yr | +24.38%/yr | +0.35%/yr |
| 3Y | +21.80%/yr | +21.20%/yr | +0.60%/yr |
| All | +20.88%/yr | +20.38%/yr | +0.50%/yr |

Across all three windows MWR aligns with the VOO benchmark with the small expected edge from dividend timing / DCA. The original 3Y +11.24%/yr bug from the pre-history-fallback artifact is gone.

## Files

- `scripts/fetch-voo-history.ts` — Yahoo chart → 3-col CSV writer
[...]

PR #416, opened 2026-05-23
test(resolve-security-snapshot): cover all 17 branches in 165-LOC route

### LINKED ISSUE #414: test(resolve-security-snapshot): zero route-level coverage on 165-LOC Polygon-backed write path
## Summary

`src/server/routes/accounts/post-resolve-security-snapshot.ts` (165 LOC) is the on-demand benchmark price-fetch route powering the PerformanceBenchmark widget (#386). It has **zero route-level tests**. It is currently the only untested route under `src/server/routes/accounts/` since PR #402 and the recent `holding-snapshot-sibling-routes.test.ts` (PR #397) closed the surrounding gaps.

This route is one of only two budget routes that **writes** to `security_snapshots` from a user-driven request path (the other is `post-holding-snapshot.ts`, covered by issue #359), and it calls the external Polygon API — so the route wiring on top of `polygon.getLatestClosePriceOnOrBefore()` (already unit-tested in `polygon.test.ts`) is exactly the boundary that would silently regress.

## Branches uncovered (~17)

**Auth + validation (5 paths):**
- Unauthenticated → `{ status: "failed" }`
- Body not an object → `validationError`
- `security_id` missing/non-string → `validationError`
- `date` missing/non-string → `validationError`
- `date` doesn't match `YYYY-MM-DD` → `validationError`

**Security lookup (4 paths):**
- `getSecurity(security_id)` returns null → `{ resolved: false, message: "security not found" }`
- ticker missing → `{ resolved: false, message: "security has no ticker / is cash-equivalent" }`
- ticker starts with `"CUR:"` → same fallback
- `security.type === "cash"` → same fallback

**Existing-snapshot reuse (3 paths):**
- Existing snapshot ≤ MAX_PROXIMITY_DAYS (7) with `close_price != null` → `{ resolved: true, source: "existing" }` (no Polygon call)
- Existing snapshot > 7 days away → falls through to Polygon
- Existing snapshot with `close_price == null` → falls through to Polygon

**Polygon outcomes (5 paths):**
- `priceResult.error === "no_api_key"` → message "Market data API is not configured"
- `priceResult.error === "plan_limit"` → message about plan range
- `priceResult.error === "no_data"` → message "No price data for ${ticker} on or before ${date}"
- Other error → message "Polygon error: …"
- Success + `upsertSnapshots` succeeds → `{ resolved: true, source: "polygon" }`

**Edge (1 path):**
- `upsertSnapshots` throws → non-fatal: still returns `{ resolved: true, source: "polygon" }` with logged error (verify the `logger.error` call and the `resolved:true` return)

**Bonus (1 path):**
- Future-date capping at line 80-81: `date > today` should clamp `effectiveDateStr` to today before the existing-snapshot lookup. Easy to verify by asserting the date passed into the mocked `getSecuritySnapshots`.

## Suggested test pattern

Mirror `holding-snapshot-sibling-routes.test.ts` and `post-suggest-category.test.ts`: monkey-patch the four real imports on shared objects rather than `mock.module("server", …)` (which is process-wide in Bun and leaks across files). Targets:

- `getSecurity` — exported standalone from `server/lib/postgres/repositories`
- `getSecuritySnapshots` — same
- `upsertSnapshots` — same
- `polygon.getLatestClosePriceOnOrBefore` — patch on the `polygon` namespace object

Restore originals in `afterAll`. Reset mocks in `beforeEach`.

## Effort

~1 PR, ~280-320 LOC of test, single file `post-resolve-security-snapshot.test.ts`. No production changes needed — the route's deps are already plain top-level imports that are trivially patchable.

---

🤖 Filed by daily-ops cron (Sweep 3 — test coverage). Today is 2026-05-22.

claoie:
Closes #414

## Summary
Adds `post-resolve-security-snapshot.test.ts` (21 tests, ~480 LOC) covering all 17 enumerated branches in `post-resolve-security-snapshot.ts` (165 LOC), the on-demand Polygon-backed write path powering the PerformanceBenchmark widget (#386). Prior to this PR the route had zero route-level tests — making it the last untested route under `src/server/routes/accounts/` after #397 (sibling holding-snapshot routes) and #402 closed the surrounding gaps.

## Mocking strategy
Mirrors `holding-snapshot-sibling-routes.test.ts` and `post-suggest-category.test.ts`: monkey-patch on shared singletons rather than `mock.module("server", ...)`, which is process-wide in Bun and leaks across files.

- `securitiesTable.queryOne` — drives `getSecurity()`
- `pool.query` — drives `getSecuritySnapshots()`
- `snapshotsTable.upsert` — drives `upsertSnapshots()`
- `globalThis.fetch` + `POLYGON_API_KEY` env — drives `polygon.getLatestClosePriceOnOrBefore()`

> ⚠️ Deviation from the issue: the issue suggested patching the polygon namespace directly. That doesn't work — `export * as polygon from "./polygon"` produces a read-only Module Namespace Object whose bindings can't be replaced (both direct assignment and `Object.defineProperty` throw). Patching at the fetch layer is the same approach `polygon.test.ts` already uses, so this is consistent with existing patterns.

`POLYGON_RATE_LIMIT_PER_MIN=0` disables the rate-limit gate so tests don't wait ~12 s for token refills; `polygonQueue.reset()` in `beforeEach` clears any in-flight tokens.

## Branches covered (21 tests)
- **Auth + body validation (7):** unauthenticated; missing body; non-object body (array); missing/non-string `security_id`; missing/invalid `date`.
- **Security lookup (4):** not found; null ticker → "cash-equivalent"; `CUR:*` prefix → same; `type === "cash"` → same.
- **Existing-snapshot reuse (3):** snapshot within 7d → returns `source: "existing"`, no Polygon call; > 7d → falls through; null `close_price` → falls through.
- **Polygon outcomes (6):** `no_api_key` → "Market data API is not configured"; `plan_limit` → plan-range message; `no_data` → ticker+date message; `api_error` (via `response.json()` reject, avoiding the 6 s `fetchWithRetry` backoff path); success → upsert + `source: "polygon"`; upsert rejection still returns `resolved: true` (the "snapshot persists in memory even if disk write fails" contract).
- **Future-date clamping (1):** `date > today` is clamped to today before reaching both `getSecuritySnapshots` (verified via SQL values) and the Polygon URL (verified by captured fetch URL).

## E2E / Test plan
- [x] \`bun test src/server/routes/accounts/post-resolve-security-snapshot.test.ts\` → 21 pass, 76 expect() calls, ~360 ms.
- [x] \`bun test src/server/\` → 465 pass, 0 fail.
- [x] No new typecheck or lint findings on the test file (pre-existing repo-wide failures in unrelated files: `sync.ts` working-tree merge markers, `TransactionDetailPage` unused imports).

### DIFF
--- a/src/server/routes/accounts/post-resolve-security-snapshot.test.ts
+++ b/src/server/routes/accounts/post-resolve-security-snapshot.test.ts
@@ -1,0 +1,480 @@
+// Per-test-bundle isolation — see scripts/test-bundled/.
+//
+// `POST /api/resolve-security-snapshot` powers the PerformanceBenchmark
+// widget's on-demand benchmark price fetch (#386, #414). Uses the
+// leaf-mock pg pattern: every DB call (getSecurity, getSecuritySnapshots,
+// upsertSnapshots → securitiesTable / snapshotsTable / pool.query) lands
+// on `mockQuery` via a FakePool. A SQL router dispatches by table name.
+//
+// Polygon: `getLatestClosePriceOnOrBefore` runs against real polygon
+// code with `globalThis.fetch` mocked + `POLYGON_API_KEY` set in
+// process.env before the bundle imports. `POLYGON_RATE_LIMIT_PER_MIN=0`
+// disables the rate gate so tests don't wait on token refills. All
+// process-global state (env vars + globalThis.fetch) is snapshotted and
+// restored in afterAll so sibling tests in the unified process aren't
+// affected.
+process.env.POLYGON_API_KEY = "test-key";
+process.env.POLYGON_RATE_LIMIT_PER_MIN = "0";
+
+import { describe, test, expect, mock, beforeEach, afterAll } from "bun:test";
+import { bundleOf } from "test-bundled";
+import type { JSONSecurity } from "common";
+
+const originalFetch = globalThis.fetch;
+const originalApiKey = process.env.POLYGON_API_KEY;
+const originalRateLimit = process.env.POLYGON_RATE_LIMIT_PER_MIN;
+
+const mockQuery = mock(async (_sql: string, _values?: unknown[]) => ({
+  rows: [] as unknown[],
+  rowCount: 0 as number | null,
+}));
+
+class FakePool {
+  query = mockQuery;
+  end = async () => {};
+  connect = async () => ({ query: mockQuery, release: () => {} });
+}
+
+mock.module("pg", () => ({
+  Pool: FakePool,
+  types: { setTypeParser: () => {} },
+  default: { Pool: FakePool, types: { setTypeParser: () => {} } },
+}));
+
+let lastFetchUrl: string | null = null;
+const mockFetch = mock(
+  async (_url: string | URL | Request, _init?: RequestInit): Promise<Response> =>
+    new Response(JSON.stringify({ results: [] }), { status: 200 }),
+);
+globalThis.fetch = mockFetch as unknown as typeof globalThis.fetch;
+
+const { postResolveSecuritySnapshotRoute } = await bundleOf<
+  typeof import("./post-resolve-security-snapshot")
+>(import.meta.url);
+
+afterAll(() => {
+  globalThis.fetch = originalFetch;
+  if (originalApiKey === undefined) delete process.env.POLYGON_API_KEY;
+  else process.env.POLYGON_API_KEY = originalApiKey;
+  if (originalRateLimit === undefined) delete process.env.POLYGON_RATE_LIMIT_PER_MIN;
+  else process.env.POLYGON_RATE_LIMIT_PER_MIN = originalRateLimit;
+});
+
+// SQL router: SELECT FROM securities returns staged security rows; SELECT
+// FROM snapshots returns staged snapshot rows; the upsert INSERT path can
+// be made to reject by setting `upsertShouldThrow`.
+let securitiesRows: Array<Record<string, unknown>> = [];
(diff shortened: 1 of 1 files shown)


claoie:
Rebased onto current main. Dropped 55 files of pre-#458 framework churn; only the new `post-resolve-security-snapshot.test.ts` ships. 676 pass / 0 fail.

⚠️ **Convention flag**: this test uses the **monkey-patch pattern** (patch `securitiesTable.queryOne`, `pool.query`, `snapshotsTable.upsert`, `globalThis.fetch`) — the pattern that the per-test-bundle framework was built to replace. It works because `afterEach` restores the references, but it's anti-convention.

A clean conversion would: use `bundleOf` to load the route's bundle, leaf-mock `pg` via FakePool, `mockExternal` for the polygon shim. ~150 LOC rewrite. Shipping as-is to land the coverage; want to track conversion as a follow-up issue or do it in this PR?

claoie:
Converted to the per-test-bundle leaf-mock pattern (`bundleOf` + FakePool/mockQuery SQL router + `globalThis.fetch` mock for polygon) per your call. 22 test scenarios preserved; monkey-patching of `securitiesTable` / `snapshotsTable` / `pool` is gone. `afterAll` restores `POLYGON_API_KEY` / `POLYGON_RATE_LIMIT_PER_MIN` / `globalThis.fetch` so the unified process stays clean. 676 pass / 0 fail.

hoiekim (approved):

PR #419, opened 2026-05-23
chore: rename PbulicTokenPostResponse → PublicTokenPostResponse

claoie:
## Finding

The response type for the `POST /api/public-token` route is named `PbulicTokenPostResponse` — \"Pbulic\" instead of \"Public\". All 7 use sites in the repo spell it consistently, so nothing's broken, but the name lies.

Principle 1 (\"code tells the truth about itself\"): a reader sees the route path `/public-token`, the handler `postPublicTokenRoute`, the components `PlaidLinkButton`/`SimpleFinLinkButton` that hit `/api/public-token` — and then an oddly-spelled type that says `Pbulic`. Either it's wrong or it's referring to something else. It's the former.

## Fix

Single-symbol rename across 4 files (`-8/+8`):

- `src/server/routes/users/post-public-token.ts` — declaration + Route generic
- `src/client/components/Configuration/index.tsx` — import + `call.post<…>`
- `src/client/components/SimpleFinLinkButton.tsx` — import + `call.post<…>`
- `src/client/components/PlaidLinkButton.tsx` — import + `call.post<…>`

`grep` for the misspelling returns zero hits after the change.

## Test plan

- [x] `bun run lint` clean (after #418 lands; pre-existing lint break unrelated to this PR)
- [ ] CI green

### DIFF
--- a/src/client/components/Configuration/index.tsx
+++ b/src/client/components/Configuration/index.tsx
@@ -1,5 +1,5 @@
 import { ItemProvider, ItemStatus, toUpperCamelCase } from "common";
-import { PbulicTokenPostResponse } from "server";
+import { PublicTokenPostResponse } from "server";
 import {
   Data,
   Item,
@@ -86,7 +86,7 @@ export const Configuration = () => {
       go(PATH.CONNECTION_DETAIL, { params: clientPathParams });
     } else {
       const params = new URLSearchParams({ provider: ItemProvider.MANUAL });
-      const { body } = await call.post<PbulicTokenPostResponse>(
+      const { body } = await call.post<PublicTokenPostResponse>(
         `/api/public-token?${params.toString()}`,
         {},
       );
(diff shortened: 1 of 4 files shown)


claoie:
## Self-Review

**Thread status:** No prior review. Rebased onto `upstream/main` this run (0 behind now) and force-pushed. CI green. Scope clean — pure rename, 4 files.

**Logic — correct.** Renames the typo'd `PbulicTokenPostResponse` → `PublicTokenPostResponse`: the `export interface` definition in `server/routes/users/post-public-token.ts` and all consumers (`Configuration`, `PlaidLinkButton`, `SimpleFinLinkButton`). No behavior change.

**Verification:** `grep -rn 'Pbulic' src/` → **0 matches** (no stragglers). `bun run tsc --noEmit` clean (exit 0) — confirms def + all references renamed consistently.

**E2E:** type-level rename only; tsc passing is the complete check (no runtime surface).

hoiekim (approved):

PR #421, opened 2026-05-24
fix(transactions): reset confidence on clear, cover split-tx writes (#415)

### LINKED ISSUE #415: bug(transactions): TransactionDetailPage budget/category edits leave stale label_category_confidence, hiding the row from auto-suggest
## Summary
The transaction-detail page's budget/category edit handlers (and the inline split-row edit handlers on the same page) POST to `/api/transaction` / `/api/split-transaction` **without** `label.category_confidence`. The server-side helper `applyUserConfirmedConfidence` only auto-fills confidence when `category_id` is non-null, so any edit that clears `label_category_id` leaves the prior `label_category_confidence` value in the database. A previously-confirmed row (`c_conf = 1`) becomes an orphan: `{ category_id: null, category_confidence: 1 }`. That row is then:

- rendered with the red "needs categorization" notification dot in the row UI (because `category_id` is null), but
- permanently excluded from the auto-suggest engine, because `defaultFetchUnlabeled` filters on `label_category_confidence IS NULL` (`src/server/lib/compute-tools/auto-suggest.ts`).

The user never gets a suggestion on the row again. The split-row edit handlers in `TransactionProperties/SplitTransactionRow.tsx` have the same omission, and the split-transaction server path doesn't run the auto-confidence helper at all (`updateSplitTransactions` in `src/server/lib/postgres/repositories/split_transactions.ts:84-109` calls `splitTransactionsTable.update` directly), so splits edited from the detail page leave stale confidence on **both** the set-category and clear-category paths.

The `TransactionRow.tsx` / `InvestmentTransactionRow.tsx` list-page handlers were updated to send `category_confidence` explicitly in `d4162bc` (2026-05-21, "fixed bugs in transaction detail page label updates"); the corresponding fix for `TransactionProperties` was missed.

## Reproduction (curl against local dev server)

```bash
# 1. Pick an admin transaction and set it to a "confirmed" state:
TID="M3DgpD95ymCjYkRwaqnASoLyNj44BDUgjgdA81"
psql -U budget -d budget -c "
  UPDATE transactions
    SET label_category_id = '<category-id>',
        label_category_confidence = 1.0
    WHERE transaction_id = '$TID';"

# 2. Log in and POST exactly what TransactionProperties.onChangeBudgetSelect sends:
curl -s -X POST http://localhost:3005/api/transaction \
  -H 'Content-Type: application/json' \
  -b cookies.txt \
  -d "{\"transaction_id\":\"$TID\",
       \"label\":{\"budget_id\":\"<new-budget-id>\",\"category_id\":null}}"
# → { "status":"success", ... }

# 3. Verify the row in DB:
psql -U budget -d budget -c "
  SELECT label_budget_id, label_category_id, label_category_confidence
    FROM transactions WHERE transaction_id = '$TID';"
# Result:
#   label_budget_id = <new-budget-id>
#   label_category_id = NULL            ← cleared
#   label_category_confidence = 1       ← STALE — should be 0 (cleared) or NULL
```

The same omission reproduces with the `onChangeCategorySelect` shape (`{ transaction_id, label: { category_id: null, budget_id: ... } }` — clearing the category from the detail page also leaves `confidence = 1`).

## Root cause

`src/server/lib/postgres/repositories/transactions.ts:132-137` — the helper bails out early when `category_id` is null:

```ts
const applyUserConfirmedConfidence = (tx: PartialTransaction): PartialTransaction => {
  if (!tx.label) return tx;
  if (tx.label.category_id == null) return tx;           // ← short-circuits the clear-category path
  if (tx.label.category_confidence !== undefined) return tx;
  return { ...tx, label: { ...tx.label, category_confidence: 1.0 } };
};
```

The helper exists precisely to keep client/server confidence in sync when the FE doesn't send it. It handles the `select-a-category` direction but not the `clear-category` direction, so any FE caller that clears `category_id` without also sending `category_confidence` leaks stale confidence into the row.

The four FE call sites that hit this:

- `src/client/components/TransactionProperties/index.tsx:103-106` (`onChangeBudgetSelect` — clears category, omits confidence)
- `src/client/components/TransactionProperties/index.tsx:131-135` (`onChangeCategorySelect` — both select-and-clear paths omit confidence; the select path is rescued by the server helper, the clear path is not)
- `src/client/components/TransactionProperties/SplitTransactionRow.tsx:92-95` (`onChangeBudgetSelect` for split — same omission, and `updateSplitTransactions` has no analogue helper so even the select path silently fails to record `c_conf = 1`)
- `src/client/components/TransactionProperties/SplitTransactionRow.tsx:119-124` (`onChangeCategorySelect` for split)

For contrast, the **list-page** handlers do send confidence and are unaffected:

- `src/client/components/TransactionsTable/TransactionRow.tsx:109` / `:115` / `:158-161`
- `src/client/components/TransactionsTable/InvestmentTransactionRow.tsx:113` / `:142-145`

## Auto-suggest interaction

`src/server/lib/compute-tools/auto-suggest.ts:defaultFetchUnlabeled` selects rows where `label_category_confidence IS NULL`. The CAS guard added in PR #408 uses the same predicate (`CAS_NULL_CONFIDENCE = { column: "label_category_confidence", value: null }`). Both the fetch and the CAS guard skip orphan rows, so once a row enters this state the auto-suggest engine cannot re-process it — the user has to manually pick a category to recover (which goes through the working `applyUserConfirmedConfidence` path).

This is the same shape as the issue #339/PR #340 backfill story from May 12: legacy rows with `c_id != null, c_conf = null` had to be backfilled to `c_conf = 1` to make the engine consistent. The orphan state introduced by this bug is the *complementary* shape: `c_id = null, c_conf = 1`. A future backfill would similarly need to set `c_conf = 0` (or NULL) for any rows in that state.

## Fix sketch

**Option A — server (preferred, defense-in-depth).** Extend `applyUserConfirmedConfidence` to handle the clear-category direction:

```ts
const applyUserConfirmedConfidence = (tx: PartialTransaction): PartialTransaction => {
  if (!tx.label) return tx;
  if (tx.label.category_confidence !== undefined) return tx;
  if (tx.label.category_id === null) {
    // user cleared the category — reset confidence so future syncs and
    // the auto-suggest engine see this row as un-labeled again.
    return { ...tx, label: { ...tx.label, category_confidence: 0 } };
  }
  if (tx.label.category_id != null) {
    return { ...tx, label: { ...tx.label, category_confidence: 1.0 } };
  }
  return tx;
};
```

This keeps memo-only POSTs untouched (`category_id === undefined`, not `=== null`). Also add an analogous helper to `updateSplitTransactions` so splits get the same protection regardless of which FE path edits them.

**Option B — client.** Add `category_confidence: 0` to the four `TransactionProperties` POST bodies that clear `category_id`, and `category_confidence: nextConfidence` to the two `onChangeCategorySelect` `labelQuery` constructions on the detail-page side. This mirrors the fix that already shipped for `TransactionsTable/TransactionRow.tsx` in `d4162bc`. Smaller surface but doesn't protect a future caller from making the same omission.

I'd land both: server-side helper for the orphan-prevention invariant, client-side explicit value for parity with the row-level handler.

## Backfill consideration

If anyone besides admin has run this code path between `d4162bc` (2026-05-21) and the fix landing, their DB has rows with `{ category_id: null, category_confidence: 1 }`. A one-line corrective:

```sql
UPDATE transactions
   SET label_category_confidence = NULL
 WHERE label_category_id IS NULL
   AND label_category_confidence = 1;
```

Same shape on `split_transactions`. Worth running once the code fix is in.

## Adjacent observation (not filed)

The `labels_category_confidence` column (plural `labels_`) still exists on the `transactions` schema alongside the singular `label_category_confidence`. The plural one is referenced nowhere in `src/`; it's dead from the pre-#314 `labels_` naming. All values are NULL in the local clone. Schema cruft, not a runtime issue — could be dropped in a follow-up migration.

claoie:
Closes #415

## Summary

When `TransactionProperties` cleared a category (selecting the empty option) the POST body omitted `category_confidence`. `applyUserConfirmedConfidence` short-circuited on `category_id === null`, so the prior `c_conf = 1` value stuck and the row became an orphan: `{ category_id: null, category_confidence: 1 }`. That state is invisible to the auto-suggest engine — `defaultFetchUnlabeled` and the post-#408 CAS guard both filter on `label_category_confidence IS NULL` — so once a row entered the orphan state, the user never got a fresh suggestion on it.

Split rows had the same omission **and** the server-side `updateSplitTransactions` ran no helper at all, so set-category from the detail page also silently failed to record `c_conf = 1` on splits.

## Fix (Option A + B from the issue body)

**Server (defense-in-depth)**
- `applyUserConfirmedConfidence` in `transactions.ts` now also handles the clear direction: `category_id === null` → `category_confidence = 0`. Memo-only updates (`category_id === undefined`) untouched.
- New mirror helper in `split_transactions.ts` wired through `updateSplitTransactions` so both set and clear paths get the right confidence.

**Client (parity with list-page handlers)**
- `TransactionProperties/index.tsx`: `onChangeBudgetSelect` sends `category_confidence: 0`; `onChangeCategorySelect` sends `1` on set, `0` on clear — matches `TransactionsTable/TransactionRow.tsx` (`d4162bc`).
- `TransactionProperties/SplitTransactionRow.tsx`: same change in both select handlers plus the local state update.

## Test plan

- `bun run test:server` → **498 pass / 0 fail**
- `bun run test:client` → **201 pass / 0 fail**
- `bun run typecheck` → clean
- Updated existing "does not inject confidence when category_id is null" test to assert the new positive behavior (UPDATE params contain `0`)

**E2E (live dev server + curl + psql against `/Users/hoiekim/budget`)**:

```
Step 1 — POST {transaction_id, label: {category_id: "0520...088ec"}}
  DB after: label_category_id = 0520..., label_category_confidence = 1 ✓

Step 2 — POST {transaction_id, label: {category_id: null}}
  DB after: label_category_id = NULL, label_category_confidence = 0 ✓
            (pre-fix this was c_id=NULL, c_conf=1 — the orphan)
```

## Backfill (out of scope for this PR, flagged for follow-up)

Any rows already in the orphan state from prior code paths can be repaired with:

```sql
UPDATE transactions
   SET label_category_confidence = NULL
 WHERE label_category_id IS NULL
   AND label_category_confidence = 1;

UPDATE split_transactions
   SET label_category_confidence = NULL
 WHERE label_category_id IS NULL
   AND label_category_confidence = 1;
```

Holding off on shipping this migration until target value is confirmed — `NULL` (return to unlabeled pool) is the natural choice but `0` (keep as explicit reject) is defensible.

### DIFF
--- a/src/client/components/TransactionProperties/SplitTransactionRow.tsx
+++ b/src/client/components/TransactionProperties/SplitTransactionRow.tsx
@@ -100,6 +100,12 @@ const SplitTransactionRow = ({ splitTransaction }: Props) => {
         const newSplit = new SplitTransaction(splitTransaction);
         newSplit.label.budget_id = value || null;
         newSplit.label.category_id = null;
+        // Local IndexedDB write must mirror the server's confidence
+        // inference (`inferLabelConfidence`, `lib/infer-label-confidence.ts`)
+        // — otherwise the UI flashes the stale prior `c_conf` value until
+        // the next reload (#415). Budget-only edit always clears the
+        // category, so confidence resets to 0.
+        newSplit.label.category_confidence = 0;
         indexedDb.save(newSplit).catch(console.error);
         const newSplits = new SplitTransactionDictionary(newData.splitTransactions);
         newSplits.set(split_transaction_id, newSplit);
@@ -133,6 +139,10 @@ const SplitTransactionRow = ({ splitTransaction }: Props) => {
           newSplit.label.budget_id = account?.label.budget_id;
         }
         newSplit.label.category_id = value || null;
+        // Mirror server-side `inferLabelConfidence`: 1 = picked a category,
+        // 0 = cleared. Without this, IndexedDB carries the stale prior
+        // confidence on the local row until next reload (#415).
+        newSplit.label.category_confidence = +!!value;
         indexedDb.save(newSplit).catch(console.error);
         const newSplits = new SplitTransactionDictionary(newData.splitTransactions);
         newSplits.set(split_transaction_id, newSplit);


claoie:
Rebased + scoped down. The original branch had merge conflicts because it was from before #431 (router-layer `inferLabelConfidence`) and the test-framework migration. Most of the original PR's changes are now obsolete:

- **Server-side**: `applyUserConfirmedConfidence` helper in `transactions.ts` / `split_transactions.ts` → already handled by `inferLabelConfidence` at the route layer (`infer-label-confidence.ts:31-37` maps `category_id: null` → `category_confidence: 0`). Dropped.
- **Route changes** in `post-transaction.ts` / `post-split-transaction.ts` → already in main from #431. Dropped.
- **Client `TransactionProperties/index.tsx`** → already explicitly sets local `category_confidence` (lines 114, 145). Dropped.

The only **remaining gap** was `SplitTransactionRow.tsx`'s local IndexedDB update — it set `category_id` / `budget_id` locally but left `category_confidence` at its stale prior value, so the UI flashed wrong state until the next reload. Fixed by mirroring `TransactionProperties/index.tsx`'s pattern: `+!!value` on category select, `0` on budget-only edit.

10-line diff, all tests green.

claoie:
## Self-Review (post-rebase)

**Thread status:** Previously scoped down (05-31 comment). Re-rebased onto `upstream/main` this run (0 behind) and force-pushed. CI green. Scope clean — 1 file, +10.

**Logic — correct.** On the local IndexedDB write in `SplitTransactionRow`, `category_confidence` is now set to mirror the server's `inferLabelConfidence`:
- budget-only edit (category cleared) → `category_confidence = 0`
- category select → `category_confidence = +!!value` (1 if a category was picked, 0 if cleared)

Without this the local row kept its stale prior confidence until the next reload, so the UI flashed wrong state (#415). This mirrors the already-correct pattern in `TransactionProperties/index.tsx`.

**E2E:** client IndexedDB write in a `.tsx` component. The fix matches the server inference exactly (`infer-label-confidence.ts`: `category_id: null → 0`) and the sibling `TransactionProperties` pattern. Full browser E2E (edit split row → assert no confidence flash) is blocked locally — budget has no `.env`/login password in this environment, so UI-driven verification can't run here. Verified at the logic/type level.

**Blockers:** none. **Confidence: High** on logic; UI click-path not browser-verified (environment limitation).

hoiekim (approved):

Issue #381, opened 2026-05-16
Performance widget v2: cash-inclusive MWR + data-anomaly guards


claoie:
Follow-up to #377 (v1 shipped as asset-only / cash-excluded MWR). This issue tracks the deferred features once we understand the settlement-pending and txn-vs-holdings consistency picture better.

### 1. Cash-inclusive MWR
Include cash positions in `V_start` / `V_end` and treat external deposits/withdrawals (rather than asset buys/sells) as the cash flows. Required understanding of:
- What `buy QACDS` actually represents in Plaid's model — is it always an external deposit, or sometimes a settlement event (orphan pending obligation clearing into settled cash)?
- How the orphan/pending cash position (`gJpE8M…`-style) should be combined with QACDS for "real cash balance" math.
- Whether dividend/interest crediting to cash should be capture as "internal returns" or part of the flow stream.

Open question per Hoie 2026-05-16: "Cash holdings have some weird behavior due to settlement pending situation". Need a documented mental model before re-attempting.

### 2. Data-anomaly guards
The v1 spike surfaced **two real prod data inconsistencies** that produce implausible numbers:

- **2026-05-04 deposit row marked `is_deleted=t`** while holdings retain the qty bump → V_end has $[redacted] that has no matching incoming flow → IRR inflated.
- **2026-05-15 VOO qty jump [redacted] → [redacted]** (+[redacted] shares, ~$[redacted]) with no matching `buy VOO` investment_transaction → V_end has $[redacted] of value not explained by any cash flow → IRR rockets to +570% annualized on the 1Y view.

These appear to be Plaid txn-stream-vs-holdings-stream sync gaps. They self-heal once the next Plaid pull catches up, but until then the widget shows misleading numbers. v2 guards to consider:
- Sanity check: if implied annualized market-return component > some threshold (50%? 100%?), surface a warning instead of the number ("Data may be stale — last sync N hours ago").
- Cross-check holdings-snapshot qty deltas against investment_transactions; flag the security_ids where they diverge.
- Soft cap on display (clip annualized to 100% with a "+ unbounded" indicator).

### 3. Benchmark picker
Default VOO is sensible for US-stock-heavy portfolios. Add picker for SPY / QQQ / VTI / AGG / etc. once we have account-level "primary benchmark" preference (or just per-render).

### 4. Multi-account aggregation
[...]

PR #425, opened 2026-05-25
fix(holdings): classify cash via security type, not cost_basis heuristic

### LINKED ISSUE #376: bug(holdings): server-inferred USD cash row written with non-zero cost_basis bypasses FE cash detection
## Summary

`inferCashHoldings` in `src/server/lib/compute-tools/cash-holding.ts` writes the synthesized USD cash row with `cost_basis: inferredCash` (a non-zero number — equal to the cash amount). The frontend's cash detector in `src/client/lib/hooks/calculation/holdings.ts:324-325` only recognizes a holding as cash when **`institution_price === 1 && (cost_basis === null || cost_basis === 0)`**. So every server-inferred cash row fails the FE detection and is rendered as a regular equity row, contradicting Hoie's 2026-05-14 direction "Plaid-provided and inferred cash account displays identically" + "When the holding is cash, display 'Cash' instead of the security id."

Latent today (local prod-snapshot DB has zero rows pointing at the canonical USD security — Plaid sync currently fails on missing credentials and the new `inferCashHoldings` code shipped in PR #353). Will trigger on the next successful Plaid/SimpleFin sync that produces an account with `balance > Σ(non-cash holdings)` — at that point every affected user's investment account will gain a permanently-mislabeled "USD" / "US Dollar Cash" row that won't aggregate with broker-reported cash rows.

## Reproduction

Cannot exercise the full path without working Plaid credentials in the test env, but the asymmetry is verifiable by static trace and one-line evaluation:

1. Sync-plaid (`src/server/lib/compute-tools/sync-plaid.ts:266-270`) calls `inferCashHoldings(accounts, holdings, securities)` and merges the result into `allHoldings` passed to `upsertAndDeleteHoldingsWithSnapshots`.
2. `inferCashHoldings` (`cash-holding.ts:153-164`) constructs the new row:
   ```ts
   result.push({
     holding_id: `${account.account_id}-${cashSecurity.security_id}`,
     account_id: account.account_id,
     security_id: cashSecurity.security_id,
     quantity: inferredCash,        // e.g. 1000 for a $[redacted] cash position
     institution_price: 1,
     institution_value: inferredCash,
     cost_basis: inferredCash,      // <-- writes non-zero cost basis
     iso_currency_code: account.balances?.iso_currency_code ?? "USD",
     unofficial_currency_code: null,
   });
   ```
3. The snapshot table persists `cost_basis = inferredCash`. On the FE, `getHoldingsValueData` (`src/client/lib/hooks/calculation/holdings.ts:324-325`):
   ```ts
   const isCash =
     holding.institution_price === 1 && (cost_basis === null || cost_basis === 0);
   ```
   evaluates to `1 === 1 && (1000 === null || 1000 === 0)` → **`false`**, so the row is never tagged `isCash`.

## Expected vs actual (downstream effects)

| Surface | Expected (per Hoie 2026-05-14) | Actual |
| --- | --- | --- |
| Row label in HoldingsComposition | `Cash` | `USD` (ticker), with subtitle `US Dollar Cash` |
| `unrealizedGain` column | `—` (suppressed for cash) | `+$0` (cost basis = value at write time → 0 gain, but eligible to drift on next sync) |
| `costBasisInferred` fallback | skipped (cash branch) | skipped anyway because cost_basis is non-null |
| #372 "aggregate cash rows into single Cash row" | inferred row aggregates with broker-reported cash | inferred row stays separate; broker cash + USD-inferred render as two distinct rows |
| `aggregatedRows.every(r => r.costBasis !== null)` short-circuit | inferred cash row contributes `null` and disables `totalGain` (correct — gain on cash isn't meaningful) | inferred row contributes a real number, so totalGain stays valid and includes a misleading "G/L on cash" |

## Root cause

Server- and client-side cash detectors disagree on what makes a holding "cash":

- **Server** (`cash-holding.ts:72-96`, `isCashLikeSecurity` + `isHoldingCashLike`): recognizes either (a) the *security* being marked cash (`type === "cash"`, `is_cash_equivalent`, or `CUR:*` ticker) or (b) the *holding shape* `institution_price === 1 && cost_basis falsy`. The server's own canonical USD security has `type: "cash"`, so it's correctly detected by the security branch — the *holding-shape* branch is for money-market sweeps + proprietary funds added in #368.
- **Client** (`src/client/lib/hooks/calculation/holdings.ts:307-325`): only checks holding shape (`institution_price === 1 && (cost_basis === null || cost_basis === 0)`). Does **not** check security `type === "cash"`.

The server happily writes `cost_basis: inferredCash` because it knows the security itself marks the row as cash. The client never looks at the security and therefore can't tell.

## Proposed fix

Three viable options; first is the smaller change:

1. **Server-side: write `cost_basis: null` (or `0`) for inferred cash.** `cash-holding.ts:161`:
   ```ts
   cost_basis: null,
   ```
   The cost basis of cash is undefined by construction (Plaid itself doesn't track cost basis on cash positions — that's the rationale captured in the FE comment at `holdings.ts:307-323`). After this change the inferred row passes both server-side `isHoldingCashLike` *and* client-side `isCash`. No FE changes needed. **Recommended.**

2. **FE-side: also check `securityType === "cash"` / `is_cash_equivalent`.** `holdings.ts:324-325`:
   ```ts
   const sec = /* lookup via security_id */;
   const isCash =
     (holding.institution_price === 1 && (cost_basis === null || cost_basis === 0)) ||
     sec?.type === "cash" ||
     sec?.is_cash_equivalent === true ||
     sec?.ticker_symbol?.startsWith("CUR:");
   ```
   Mirrors `isCashLikeSecurity` from the server. Larger surface — requires plumbing the security lookup into the per-snapshot loop and updating the test that pinned the FE's current detector shape. Useful as a *defense in depth* addition even after (1) lands, since manually-added "USD"-ticker holdings would benefit too.

3. **Combine 1 + 2** (write `null` AND broaden the FE check). Most robust against future broker shapes; lightly redundant.

Recommend (1) for this issue; (2) as a follow-up to handle the manual-entry path on its own merits.

## Out of scope (flagged for follow-up, not in this fix)

- The same asymmetry affects `sync-simple-fin.ts:38` (also imports `inferCashHoldings`) — the proposed fix is server-side so it covers both providers.
- The accumulating drift: each sync recomputes `inferredCash = max(0, balance − Σ non-cash)`. If the underlying balance changes between syncs, the snapshot for the new month has a different `cost_basis` than the previous month's. Per snapshot the `value === cost_basis` invariant still holds (so per-snapshot G/L = 0), but month-over-month "growth" semantics get confusing. Fixing this requires deciding whether cash has a meaningful cost basis at all — the issue's recommended fix sets it to `null` and sidesteps the question.

## Related

- PR #353 (auto-infer USD cash holding, merged May 14) — introduced `inferCashHoldings`.
- PR #368 / PR #369 (catch money-market sweeps, merged May 15) — added the `isHoldingCashLike` shape detector but only at the *server* layer.
- PR #372 (aggregate cash rows into a single Cash row, merged May 15) — the FE aggregation skipped over inferred rows because the underlying detector misses them.

---

claoie:
## Summary

Closes #376.

### Cash detection moves from cost_basis heuristic → `institution_price === 1`

Per Hoie's review direction (https://github.com/hoiekim/budget/pull/425#issuecomment-… 2026-06-06):
> *"Don't write cash security snapshots. FE can detect 'holdings' to be cash or not. If it's cash just infer flat price history."*

`getHoldingsValueData` previously detected cash by `institution_price === 1 && (cost_basis === null || cost_basis === 0)`. The `cost_basis` predicate is the fragile half — Plaid collapses DB-NULL → 0 on the wire (`SnapshotModel.toHoldingSnapshot` does `this.cost_basis ?? 0`), so "no basis reported" and "explicit zero" are indistinguishable; any server change that writes a sweep total / broker delta into `cost_basis` would silently flip the row out of the cash branch.

`institution_price === 1` alone is the intrinsic signal: every broker quotes 1 for cash because cash doesn't trade against itself. Drop the `cost_basis` half entirely.

### Cash rows render as `cost_basis === value` → 0% gain

A position that doesn't appreciate against itself should render as `0% gain`, not as `—`. So for the cash branch the FE overrides `finalCostBasis = value` (so `unrealizedGain === 0`, `returnPercent === 0%`). This catches all three on-the-wire shapes uniformly:
- BE-inferred cash row (`cost_basis = institution_value` already)
- Plaid cash sweep with DB-NULL collapsed to 0
- Plaid cash sweep that reports a non-zero `cost_basis`

`inferCostBasis` is still skipped for cash so the `type='buy'`-coded sweep deposits don't accumulate a phantom basis.

### Trade-off pinned

`institution_price === 1` matches real $1 equities too. Pinned in a test — those rows render under the cash branch (`cost_basis === value`, 0% gain), which is the right readout for the rare $1-equity case. The much-more-common Plaid deposit-sweep path now classifies correctly, which the old `cost_basis === null` predicate would have missed when Plaid set `cost_basis = sweep_total`.

## Files

- `src/client/lib/hooks/calculation/holdings.ts` — detector becomes `institution_price === 1`; `finalCostBasis = isCash ? value : cost_basis`.
- `src/client/lib/models/Calculations.ts` — `HoldingValueSummary.isCash` docstring updated.
- `src/client/lib/hooks/calculation/holdings.test.ts` — three tests rewritten:
  - `classifies a holding with institution_price=1 as cash` — pins `isCash=true`, `costBasis=value`, `returnPercent === 0`.
  - `classifies cash uniformly regardless of cost_basis` — null / 0 / value all normalize identically.
  - `a real $1 equity with a cost_basis is treated as cash by the holding-side detector` — pins the trade-off.

## Test plan

- [x] `bun test` — 713 pass / 0 fail
- [x] **UI E2E** — sandbox `claoie.tplinkdns.com:20425` (PR branch), logged in as admin/sandbox, navigated **Accounts → CustomName LgkKbv (Chase Brokerage)**. Holdings Composition table renders:

  ```
  Security  Value         Growth     %
  <equity-A>       $ <redacted>     +$ <redacted>  116.8%
  Cash      $ 588         +$ 0       0.5%
  Unknown   - $ <redacted>    —          -17.2%
  Total     $ <redacted>     —          100%
  ```

  - **<equity-A>**: unchanged vs `upstream/main` (same value / growth / %). No regression on the equity row.
  - **Chase Deposit Sweep (`QACDS`, $587.64)**: now labeled "**Cash**" with **`+$0`** growth (0% gain) — exactly the design intent ("Plaid-provided and inferred cash account displays identically"). Was previously rendering with cost-basis-based classification.
  - Total ties: $<redacted> + $588 − $<redacted> ≈ $<redacted> ✓.

  Screenshot: `/tmp/pr425v3-detail-full.png` (not attached per privacy rules — scrambled prod-clone budget IDs visible).

*[Edited 2026-06-07: redacted real holdings-composition dollar figures (>$[redacted]) and live tickers per §2a PII policy — engineering content preserved.]*

---

### DIFF
--- a/src/client/lib/hooks/calculation/holdings.test.ts
+++ b/src/client/lib/hooks/calculation/holdings.test.ts
@@ -22,7 +22,8 @@ import { InvestmentTransaction } from "../../models/InvestmentTransaction";
 const createSecuritySnapshot = (
   securityId: string,
   closePrice: number,
-  date: string
+  date: string,
+  extra: Partial<{ type: string; is_cash_equivalent: boolean; ticker_symbol: string }> = {},
 ): SecuritySnapshot => {
   return new SecuritySnapshot({
     snapshot: { snapshot_id: `snap_${securityId}_${date}`, date },
@@ -30,6 +31,7 @@ const createSecuritySnapshot = (
       security_id: securityId,
       close_price: closePrice,
       close_price_as_of: date,
+      ...extra,
     },
   });
 };
@@ -508,14 +510,10 @@ describe("getHoldingsValueData", () => {
   });
 
   test("a real $1 equity with a cost_basis is treated as cash by the holding-side detector", () => {
-    // Trade-off pinned: `institution_price === 1` matches real $1 equities
-    // too. Those holdings (vanishingly rare in practice) render under the
-    // cash branch as `cost_basis === value` (here, value=$100, normalized
-    // basis=$100, 0% gain). Hoie's 2026-06-06 call: this readout is
-    // acceptable for the rare $1-equity case because the alternative
-    // (cost_basis-shape heuristic) lies about cash on the much more
-    // common deposit-sweep path. The BE's `inferCashHoldings` is the
-    // safety net for accounts where this matters at the dollar level.
+    // `institution_price === 1` matches real $1 equities too. Those
+    // holdings render under the cash branch as `cost_basis === value`
+    // (here, value=$100 → basis=$100 → 0% gain). Trade-off pinned so
+    // a future tightening of the detector flips this loudly.
     const holdingSnapshots = new HoldingSnapshotDictionary();
     holdingSnapshots.set(
       "h-penny",
@@ -533,6 +531,70 @@ describe("getHoldingsValueData", () => {
     expect(summary!.costBasis).toBe(100);
     expect(summary!.unrealizedGain).toBe(0);
     expect(summary!.returnPercent).toBe(0);
+  });
+
+  test("classifies cash via the security-side branch even when institution_price drifts off 1.0", () => {
+    // Two-channel detection — security-side OR holding-side. This test
+    // exercises the security-side branch: a money-market fund whose broker
+    // quote landed at 0.9999 (FX precision / stale quote) still gets
+    // classified as cash because its security record says `type: "cash"`.
+    const holdingSnapshots = new HoldingSnapshotDictionary();
+    holdingSnapshots.set(
+      "h-mmkt",
+      // institution_price=0.9999 (not 1.0) — holding-side branch would miss
+      createHoldingSnapshot("acc1", "sec-mmkt", 1000, 0.9999, 999.9, 999.9, "2026-01-15"),
+    );
+    const securitySnapshots = new SecuritySnapshotDictionary();
+    securitySnapshots.set(
+      "snap-mmkt",
+      createSecuritySnapshot("sec-mmkt", 1, "2026-01-15", { type: "cash" }),
+    );
+
+    const result = getHoldingsValueData({
+      holdingSnapshots,
+      securitySnapshots,
+      investmentTransactions: new InvestmentTransactionDictionary(),
+    });
+
+    const summary = result.getHistory("acc1_sec-mmkt").get(new Date("2026-01-15"));
+    expect(summary!.isCash).toBe(true);
+    expect(summary!.returnPercent).toBe(0);
+  });
+
+  test("classifies cash via the security-side branch on is_cash_equivalent or CUR:* ticker", () => {
+    const holdingSnapshots = new HoldingSnapshotDictionary();
+    holdingSnapshots.set(
+      "h-equiv",
+      createHoldingSnapshot("acc1", "sec-equiv", 500, 0.98, 490, 490, "2026-01-15"),
+    );
+    holdingSnapshots.set(
+      "h-eur",
+      createHoldingSnapshot("acc1", "sec-eur", 300, 0.92, 276, 276, "2026-01-15"),
+    );
+
+    const securitySnapshots = new SecuritySnapshotDictionary();
+    securitySnapshots.set(
+      "snap-equiv",
+      createSecuritySnapshot("sec-equiv", 1, "2026-01-15", { is_cash_equivalent: true }),
+    );
+    securitySnapshots.set(
+      "snap-eur",
+      createSecuritySnapshot("sec-eur", 1, "2026-01-15", { ticker_symbol: "CUR:EUR" }),
+    );
+
+    const result = getHoldingsValueData({
+      holdingSnapshots,
+      securitySnapshots,
+      investmentTransactions: new InvestmentTransactionDictionary(),
+    });
+
+    const equiv = result.getHistory("acc1_sec-equiv").get(new Date("2026-01-15"));
+    expect(equiv!.isCash).toBe(true);
+    expect(equiv!.returnPercent).toBe(0);
+
+    const eur = result.getHistory("acc1_sec-eur").get(new Date("2026-01-15"));
+    expect(eur!.isCash).toBe(true);
+    expect(eur!.returnPercent).toBe(0);
   });
 
   test("does NOT treat a non-cash holding (institution_price > 1) as cash even with null cost_basis", () => {
--- a/src/client/lib/hooks/calculation/holdings.test.ts
+++ b/src/client/lib/hooks/calculation/holdings.test.ts
@@ -14,16 +14,17 @@ import { HoldingsValueData, HoldingValueSummary, HoldingValueHistory } from "../
 import {
   HoldingSnapshotDictionary,
   SecuritySnapshotDictionary,
+  SecurityDictionary,
   InvestmentTransactionDictionary,
 } from "../../models/Data";
 import { HoldingSnapshot, SecuritySnapshot } from "../../models/Snapshot";
+import { Security } from "../../models/miscellaneous";
 import { InvestmentTransaction } from "../../models/InvestmentTransaction";
 
 const createSecuritySnapshot = (
   securityId: string,
   closePrice: number,
   date: string,
-  extra: Partial<{ type: string; is_cash_equivalent: boolean; ticker_symbol: string }> = {},
 ): SecuritySnapshot => {
   return new SecuritySnapshot({
     snapshot: { snapshot_id: `snap_${securityId}_${date}`, date },
@@ -31,9 +32,21 @@ const createSecuritySnapshot = (
       security_id: securityId,
       close_price: closePrice,
       close_price_as_of: date,
-      ...extra,
     },
   });
+};
+
+const createSecurity = (
+  securityId: string,
+  extra: Partial<{ type: string; is_cash_equivalent: boolean; ticker_symbol: string }> = {},
+): Security => {
+  return new Security({ security_id: securityId, ...extra });
+};
+
+const securityDict = (...securities: Security[]): SecurityDictionary => {
+  const dict = new SecurityDictionary();
+  for (const sec of securities) dict.set(sec.security_id, sec);
+  return dict;
 };
 
 const createHoldingSnapshot = (
@@ -407,6 +420,7 @@ describe("getHoldingsValueData", () => {
     const result = getHoldingsValueData({
       holdingSnapshots,
       securitySnapshots,
+      securities: new SecurityDictionary(),
       investmentTransactions,
     });
 
@@ -432,6 +446,7 @@ describe("getHoldingsValueData", () => {
     const result = getHoldingsValueData({
       holdingSnapshots,
       securitySnapshots,
+      securities: new SecurityDictionary(),
       investmentTransactions,
     });
 
@@ -470,6 +485,7 @@ describe("getHoldingsValueData", () => {
     const result = getHoldingsValueData({
       holdingSnapshots,
       securitySnapshots: new SecuritySnapshotDictionary(),
+      securities: new SecurityDictionary(),
       investmentTransactions,
     });
 
@@ -497,6 +513,7 @@ describe("getHoldingsValueData", () => {
     const result = getHoldingsValueData({
       holdingSnapshots,
       securitySnapshots: new SecuritySnapshotDictionary(),
+      securities: new SecurityDictionary(),
       investmentTransactions: new InvestmentTransactionDictionary(),
     });
 
@@ -523,6 +540,7 @@ describe("getHoldingsValueData", () => {
     const result = getHoldingsValueData({
       holdingSnapshots,
       securitySnapshots: new SecuritySnapshotDictionary(),
+      securities: new SecurityDictionary(),
       investmentTransactions: new InvestmentTransactionDictionary(),
     });
 
@@ -538,6 +556,7 @@ describe("getHoldingsValueData", () => {
     // exercises the security-side branch: a money-market fund whose broker
     // quote landed at 0.9999 (FX precision / stale quote) still gets
     // classified as cash because its security record says `type: "cash"`.
+    // Source of truth = the `securities` dict, not the snapshot blob.
     const holdingSnapshots = new HoldingSnapshotDictionary();
     holdingSnapshots.set(
       "h-mmkt",
@@ -545,14 +564,12 @@ describe("getHoldingsValueData", () => {
       createHoldingSnapshot("acc1", "sec-mmkt", 1000, 0.9999, 999.9, 999.9, "2026-01-15"),
     );
     const securitySnapshots = new SecuritySnapshotDictionary();
-    securitySnapshots.set(
-      "snap-mmkt",
-      createSecuritySnapshot("sec-mmkt", 1, "2026-01-15", { type: "cash" }),
-    );
+    securitySnapshots.set("snap-mmkt", createSecuritySnapshot("sec-mmkt", 1, "2026-01-15"));
 
     const result = getHoldingsValueData({
       holdingSnapshots,
       securitySnapshots,
+      securities: securityDict(createSecurity("sec-mmkt", { type: "cash" })),
       investmentTransactions: new InvestmentTransactionDictionary(),
     });
 
@@ -573,18 +590,16 @@ describe("getHoldingsValueData", () => {
     );
 
     const securitySnapshots = new SecuritySnapshotDictionary();
-    securitySnapshots.set(
-      "snap-equiv",
-      createSecuritySnapshot("sec-equiv", 1, "2026-01-15", { is_cash_equivalent: true }),
-    );
-    securitySnapshots.set(
-      "snap-eur",
-      createSecuritySnapshot("sec-eur", 1, "2026-01-15", { ticker_symbol: "CUR:EUR" }),
-    );
+    securitySnapshots.set("snap-equiv", createSecuritySnapshot("sec-equiv", 1, "2026-01-15"));
+    securitySnapshots.set("snap-eur", createSecuritySnapshot("sec-eur", 1, "2026-01-15"));
 
     const result = getHoldingsValueData({
       holdingSnapshots,
       securitySnapshots,
+      securities: securityDict(
+        createSecurity("sec-equiv", { is_cash_equivalent: true }),
+        createSecurity("sec-eur", { ticker_symbol: "CUR:EUR" }),
+      ),
       investmentTransactions: new InvestmentTransactionDictionary(),
     });
 
@@ -623,6 +638,7 @@ describe("getHoldingsValueData", () => {
     const result = getHoldingsValueData({
       holdingSnapshots,
       securitySnapshots: new SecuritySnapshotDictionary(),
+      securities: new SecurityDictionary(),
       investmentTransactions,
     });
 
@@ -646,6 +662,7 @@ describe("getHoldingsValueData", () => {
     const result = getHoldingsValueData({
       holdingSnapshots,
       securitySnapshots,
+      securities: new SecurityDictionary(),
       investmentTransactions,
     });
 
@@ -674,6 +691,7 @@ describe("getHoldingsValueData", () => {
     const result = getHoldingsValueData({
       holdingSnapshots,
       securitySnapshots,
+      securities: new SecurityDictionary(),
       investmentTransactions,
     });
 
@@ -699,6 +717,7 @@ describe("getHoldingsValueData", () => {
     const result = getHoldingsValueData({
       holdingSnapshots,
       securitySnapshots,
+      securities: new SecurityDictionary(),
       investmentTransactions,
     });
 
--- a/src/client/lib/hooks/calculation/holdings.test.ts
+++ b/src/client/lib/hooks/calculation/holdings.test.ts
@@ -437,22 +437,21 @@ describe("getHoldingsValueData", () => {
     expect(costBasis).toBe(900); // 90 * 10
   });
 
-  test("forces null cost basis (and null G/L) when a holding looks like cash", () => {
-    // Cash detection on the FE uses signals already on the holding
-    // snapshot: institution_price === 1 AND cost_basis === null. Plaid
-    // cash sweeps always satisfy both; real equities essentially never
-    // do for any meaningful duration. Server doesn't need to ship a
-    // `type='cash'` flag; the FE skips G/L for cash holdings entirely.
+  test("classifies a holding with institution_price=1 as cash (yields 0% gain)", () => {
+    // FE cash detector is per-holding: institution_price === 1 is the
+    // canonical signal because brokers always quote 1.0 for cash. Cash
+    // rows report `cost_basis === value` so unrealizedGain === 0 and
+    // returnPercent === 0 — the logically-correct readout for a position
+    // that doesn't appreciate against itself. The `inferCostBasis`
+    // transaction-replay path is skipped — Plaid encodes sweep deposits
+    // as `type='buy'` with `price=1`, which would otherwise pile up a
+    // phantom basis (= deposit total) with a phantom G/L.
     const holdingSnapshots = new HoldingSnapshotDictionary();
     holdingSnapshots.set(
       "h-cash",
-      // institution_price = 1, cost_basis = null → looks like cash
       createHoldingSnapshot("acc1", "sec-cash", 1000, 1, 1000, null, "2026-01-15"),
     );
 
-    // Plaid records cash deposits as `type='buy'` investment_transactions
-    // with price=1. Without the cash skip these would feed inferCostBasis
-    // and produce a phantom basis (= deposit total) with a phantom G/L.
     const investmentTransactions = new InvestmentTransactionDictionary();
     investmentTransactions.set(
       "tx-deposit",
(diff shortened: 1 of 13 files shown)


claoie:
Rebased onto current main. The original branch was from before the test-framework migration (#440/#443/#448-#450), so most of its diff was effectively un-doing those merges. New single-commit branch now targets `cash-holding.test.bundle.ts` instead of the now-deleted `cash-holding.test.ts`. typecheck/lint/bundled-tests all green.

claoie:
Rebased onto current main. Retargeted from `cash-holding.test.bundle.ts` to `cash-holding.test.ts` (post-#458 rename). Source change is unchanged (one-line). 655/655 tests pass, typecheck clean.

hoiekim:
Should we change FE logic instead? Why is `cost_basis === 0` part of the detection logic?

claoie:
Good questions — answering both, then a recommendation.

### Why is `cost_basis === 0` part of the detection?
Two reasons, both about the cash signal:
1. **Plaid never reports a cost basis for cash** (sweeps / interest / money-market). So `institution_price === 1 && no-basis` is what separates a $1-quoted cash position from a real security that happens to trade near $1. That heuristic lives in two mirrored places: FE `client/lib/hooks/calculation/holdings.ts:322` and BE `server/lib/compute-tools/cash-holding.ts:81` (`isHoldingCashLike`).
2. **`=== 0` specifically** is because `SnapshotModel.toHoldingSnapshot` does `this.cost_basis ?? 0` — DB `NULL` collapses to numeric `0` on the wire. So `0` *is* the canonical "missing basis" sentinel after a round-trip; the detector has to accept it or it would only match holdings that never touched the snapshot model.

### Was writing `cost_basis = inferredCash` even correct?
No — that's the actual bug, independent of detection. A cash holding has **no** cost basis; setting it equal to the full balance implies "the entire $500 is cost," which yields a coincidental $0 unrealized gain but a misleading **0% return** row instead of N/A. `cost_basis = 0` (≡ the `null` sentinel the rest of the system already uses) is the *correct value*, not a hack to please the heuristic. So this one-line fix stands on its own merits.

### Should we change the FE logic instead?
Your instinct is right that the `cost_basis` signal is the fragile part. Here's the asymmetry I found:
- The inferred row references a **real cash security** — `ensureUSDCashSecurity()` sets `type: "cash"`, `is_cash_equivalent: true`, ticker `"USD"`.
- The **BE** detector already keys off that: `isCashLikeSecurity(sec)` returns true on `type==="cash"` *before* it ever looks at `cost_basis`. So the BE classifies the inferred row as cash regardless.
- The **FE** detector at `holdings.ts:322` does **not** look at the security at all — only the holding's `institution_price` + `cost_basis`. That's the lone reason the FE needs `cost_basis` to be 0/null.

So the robust hardening is: make the FE detector mirror the BE — also treat `type==="cash" / is_cash_equivalent / CUR:*` securities as cash (the security is already in `securitySnapshots`; `HoldingsComposition` looks it up a few lines later for the name/ticker anyway). That removes the `$1.00-security-with-no-basis` false-positive risk for *all* cash, not just the inferred row.

### Recommendation
They're complementary, not either/or:
- **Keep this PR's `cost_basis = 0`** — it fixes genuinely wrong data (a cash row shouldn't claim its balance as basis) and makes the inferred row indistinguishable from a real Plaid cash sweep, which is the stated design goal.
- **Harden the FE detector via security type** as the durable fix for the heuristic fragility you're pointing at. I'd do that as a separate PR (it touches `holdings.ts` calc + tests and isn't specific to inferred cash), unless you'd rather I fold it in here.

Want me to open the FE-detector PR, or fold it into this one?

hoiekim:
Why is unreaized gain 0% wrong for cash? Isn't that logically accurate?

hoiekim:
Fold it to this PR. I think you shouldn't use cost basis for cash detection but gain being 0% for cash is correct

claoie:
Reworked per your direction (force-pushed `7092294`):

- **0% gain for cash is correct** — agreed, my earlier framing was wrong. Cash now reports `cost_basis === value` so `unrealizedGain === 0` and `returnPercent === 0%` naturally, instead of suppressing G/L to `null`.
- **FE no longer uses `cost_basis` for cash detection.** The detector keys off the security record now (`type === "cash"`, `is_cash_equivalent`, or `CUR:*` ticker), mirroring the BE's `isCashLikeSecurity`.
- **BE revert.** With the FE detector off the holding shape, `inferCashHoldings` no longer needs to write `cost_basis: 0` to pass the FE check. Reverted back to `cost_basis: inferredCash` (= `institution_value`), internally consistent with what every cash row in this PR now reports.
- **Tests rewritten.** Two old "holding-shape" cash tests dropped (they encoded the heuristic). Three new tests added covering `type: "cash"`, `is_cash_equivalent` / `CUR:*` ticker, and the Plaid-NULL→0 collapse case. 714 pass / typecheck clean.

Sandbox up if you want to look.

claoie:
## Verification — found regressions, marked draft

Ran the sandbox at port 20425, ran the same data through the new vs old detector locally to enumerate exactly which rows change. Two real regressions found.

### Numerical impact across all 8 active holdings (latest snapshot per account+security)

| security | ticker | name | account | value | OLD isCash | NEW isCash | delta |
|----------|--------|------|---------|-------|-----------|-----------|--------|
| 96d5AO5g | <equity-A> | <index ETF> | LgkKbv | $<redacted> | no | no | identical (`+$<redacted>`, ~9.79% gain) |
| ODbxOoxk | <equity-B> | <equity-B> | 9yr3Yx | $<redacted> | no | no | identical |
| P1xJboJA | <crypto-A> | <crypto-A> | QY6XDv | $836.94 | no | no | identical |
| 7dD8KV8o | CUR:USD | U S Dollar (BE-inferred) | (inferred) | $0.00 | yes | yes | identical |
| **kZrEbXEL** | **QACDS** | **Chase Deposit Sweep** | **LgkKbv** | **$587.64** | **yes** | **no** | ⚠️ **loses cash classification** |
| **eJXpMzpR** | (none) | (no name) | VYjoZX | **$179.50** | **yes** | **no** | ⚠️ **loses cash classification** |
| eJXpMzpR | (none) | (no name) | QY6XDv | $0.00 | yes | no | no visible impact (value 0) |
| x1RMwKMj | <crypto-B> | <crypto-B> | QY6XDv | $0.00 | yes | no | no visible impact (value 0) |

### Root cause of the regressions

The FE only sees securities via `securitySnapshots` (`buildCashSecurityIds` walks `securitySnapshots`). Securities **without** a snapshot are invisible to the new detector. Two reasons that happens in prod:

1. **`upsertSecuritiesWithSnapshots`** only writes a snapshot when the security has `ticker_symbol && close_price && close_price_as_of`. Plaid cash sweeps like `QACDS` (Chase Deposit Sweep) don't come with `close_price_as_of` because cash doesn't trade — so no snapshot ever gets written, even though the master `securities` row carries `type='cash'`. The FE never learns the type.
2. **`SecurityModel.toJSON()` hardcodes `is_cash_equivalent: null`** (line 87) instead of reading from `raw`. So even securities Plaid tags `is_cash_equivalent: true` (<crypto-A>, <crypto-B>, Chase deposit sweeps when set) get stripped of that signal before the FE sees them — the new detector's `is_cash_equivalent` branch is effectively dead. Separate bug, but compounds this one.

Net effect: real Plaid-managed cash sweeps in production stop being labeled "Cash" and instead render under their Plaid ticker (e.g. "QACDS — Chase Deposit Sweep Jpmorgan Chase Bank"), defeating the stated design intent ("Plaid-provided and inferred cash account displays identically").

### Recommended next step

The FE detector approach is correct, but the data side has two gaps that need closing first. Options:

1. **Server-side fix (preferred):** also write a security snapshot for any security with `type === 'cash'` / `is_cash_equivalent` / `CUR:*` ticker, even without `close_price_as_of` (default `close_price = 1`, `close_price_as_of = today`). This makes every cash security visible to the FE. Plus fix `SecurityModel.toJSON` to surface `is_cash_equivalent` from `raw`.
2. **FE-side fix:** add a new `/api/securities` endpoint and dictionary so the FE can resolve security types regardless of snapshot coverage.
3. **Combination:** model fix + a one-time backfill for existing cash securities missing snapshots.

Marked the PR **draft** while you decide which direction to take. If you want me to add the server-side fixes in this PR, say so and I'll roll them in; if you'd rather split, I'll open a separate PR for each.

### Sandbox

Up at port 20425. Reproduce via Chase Brokerage account (`CustomName LgkKbv`) — under `upstream/main` the bottom of HoldingsComposition has a "Cash, $587.64, —" row; under this PR's code, it becomes "QACDS, $587.64, —" (different label, same row).

*[Edited 2026-06-07: redacted real holdings dollar figures (>$1k) and live tickers per §2a PII policy — engineering content preserved.]*

claoie:
## Re-pivoted per your direction (commit `44dc437`)

Detector is now per-holding: `isCash = institution_price === 1`. Cost basis is no longer in the predicate. Cash rows force `cost_basis = value` so the readout is `0% gain`. Marked **ready for review**.

### UI verification — PASS

Sandbox at port 20425 (post force-push), `Accounts → CustomName LgkKbv (Chase Brokerage)`. Holdings Composition renders:

| Security | Value | Growth | % |
|----------|-------|--------|---|
| <equity-A> (<index ETF>) | $<redacted> | +$<redacted> | 116.8% |
| **Cash** | **$588** | **+$0** | **0.5%** |
| Unknown | −$<redacted> | — | -17.2% |
| **Total** | **$<redacted>** | **—** | **100%** |

- **<equity-A> row identical** vs `upstream/main` (no regression on the equity).
- **Chase Deposit Sweep ($587.64) is now labeled "Cash" with +$0 / 0% gain** — was the regression case from yesterday's verification; now correctly classified by the holding-side detector despite the security never getting a `securitySnapshot` written.
- Total ties: $<redacted> + $588 − $<redacted> ≈ $<redacted> ✓

### Trade-off pinned in tests

A test now explicitly captures that a $1 equity with a real cost_basis is treated as cash (renders as 0% gain). That's the right readout for the rare $1-equity case, and the much-more-common Plaid sweep path now classifies correctly — which the old `cost_basis === null` predicate would have missed when Plaid set `cost_basis = sweep_total`.

### What was dropped

- The `buildCashSecurityIds` walk over `securitySnapshots`.
- The earlier proposal to write security snapshots for cash-typed securities (you nixed it — *"That's silly. Nobody needs the price history of cash."*)
- The `is_cash_equivalent` branch (the model toJSON strips that anyway; orthogonal bug, not addressed in this PR).

Sandbox up at port 20425.

*[Edited 2026-06-07: redacted real holdings dollar figures (>$1k) and live tickers per §2a PII policy — engineering content preserved.]*

claoie:
You're right — added it back. Commit `1e07185`:

```ts
const isCash =
  holding.institution_price === 1 || cashSecurityIds.has(security_id);
```

Where `cashSecurityIds = buildCashSecurityIds(securitySnapshots)` walks security snapshots for `type === "cash"` / `is_cash_equivalent` / `CUR:*` ticker — same predicates as the BE's `isCashLikeSecurity`. The two channels are OR-combined:

- **Holding-side (`institution_price === 1`)** — catches deposit sweeps whose security row never gets a `securitySnapshot` written (no `close_price_as_of` → skipped by `upsertSecuritiesWithSnapshots`).
- **Security-side (`type === "cash"` / `is_cash_equivalent` / `CUR:*`)** — catches cash whose broker quote drifts off 1.0 (FX precision, stale quote).

Tests added:
- `classifies cash via the security-side branch even when institution_price drifts off 1.0` — `institution_price=0.9999` + `security.type="cash"` → cash, 0% gain.
- `classifies cash via the security-side branch on is_cash_equivalent or CUR:* ticker` — covers the other two security-side conditions.

UI verified post-push on rebuilt sandbox (port 20425): Holdings Composition renders identically to the previous PASS (`<equity-A> $<redacted> +$<redacted> 116.8%` / `Cash $588 +$0 0.5%` / Total ties at $<redacted>). 715 tests pass.

*[Edited 2026-06-07: redacted real holdings dollar figures (>$1k) and live tickers per §2a PII policy — engineering content preserved.]*

claoie:
## Evaluation

Pushed `c0094d8` (force-push, rebased onto current `upstream/main` — the previous diff was stale and showed unrelated reverts of #436 and #430; now clean: 3 files, +168/-55).

### Strengths

- **Two-channel detector** covers complementary blind spots:
  - Holding-side `institution_price === 1` catches Plaid deposit sweeps whose security row never gets a `securitySnapshot` written (no `close_price_as_of` → skipped by `upsertSecuritiesWithSnapshots`). **Verified in prod-clone**: `QACDS` / Chase Deposit Sweep ($587.64) now correctly renders as "Cash, +$0, 0.5%".
  - Security-side `type === "cash"` / `is_cash_equivalent` / `CUR:*` catches FX-precision drift cases where the broker quote landed off 1.0.
- **Cash readout is logically-correct 0% gain** (cost_basis = value), not a suppressed null — Hoie 2026-06-06 confirmed this is the intent.
- **Tests pin both channels + the $1-equity trade-off** (35 holdings tests, 715 total / 0 fail). A future tightening of the detector flips the $1-equity test loudly.
- **`inferCostBasis` is skipped for cash**, so the `type='buy'`-coded sweep deposits don't pile up phantom basis.
- **UI verified end-to-end** on sandbox 20425: <equity-A> row unchanged vs `upstream/main` (no equity regression), Chase Deposit Sweep relabeled correctly, total ties at $<redacted>.

### Trade-offs (intentional, pinned)

- A real $1 equity matches the holding-side branch and renders under the cash branch (0% gain). Test pins this so a future tightening surfaces loudly.
- `inferCashHoldings` (BE) still writes the inferred USD-cash row with `cost_basis = inferredCash` — internally consistent with what every cash row in this PR now reports (`cost_basis === value`).

### Out of scope / follow-up

Filed #492 — `SecurityModel.toJSON()` hardcodes `is_cash_equivalent: null` (and a handful of other Plaid fields) instead of reading from `raw`. That neuters the `is_cash_equivalent` half of the detector for every security round-tripped through the model. PR #425's two-channel approach papers over the common case via `institution_price === 1`, but money-market funds with a stale quote off 1.0 still won't classify via that branch. Fix is one-line per field but unrelated to the cash-detector logic, so it stays separate.

### Verdict

Ready to merge. Sandbox at port 20425.

*[Edited 2026-06-07: redacted real holdings dollar figures (>$1k) and live tickers per §2a PII policy — engineering content preserved.]*

src/client/lib/hooks/calculation/holdings.test.ts
```diff
@@ -22,14 +22,16 @@ import { InvestmentTransaction } from "../../models/InvestmentTransaction";
...
+  date: string,
+  extra: Partial<{ type: string; is_cash_equivalent: boolean; ticker_symbol: string }> = {},
 ): SecuritySnapshot => {
   return new SecuritySnapshot({
     snapshot: { snapshot_id: `snap_${securityId}_${date}`, date },
     security: {
       security_id: securityId,
       close_price: closePrice,
       close_price_as_of: date,
+      ...extra,
```
hoiekim: What is this for?

hoiekim (changes requested):

src/client/lib/hooks/calculation/holdings.test.ts
claoie: Lets the two security-side cash tests pin `security.type` / `is_cash_equivalent` / `ticker_symbol` overrides on the fixture without breaking the existing 12 callers (they keep using the 3-arg form). The actual call sites are:

```ts
createSecuritySnapshot("sec-mmkt",  1, "2026-01-15", { type: "cash" })
createSecuritySnapshot("sec-equiv", 1, "2026-01-15", { is_cash_equivalent: true })
createSecuritySnapshot("sec-eur",   1, "2026-01-15", { ticker_symbol: "CUR:EUR" })
```

The `Partial<{...}>` type narrows the override surface to exactly the three fields `buildCashSecurityIds` looks at, so a typo (`is_cash_equivelent`) fails to compile rather than silently classifying nothing as cash. If you'd prefer a separate factory (`createCashSecuritySnapshot`) instead of the optional bag, I can split it — but every existing caller would have to know which factory to use, which felt like more noise than the 4th optional arg.

hoiekim: Why do we need that cash flags in snapshots? Those are security scope properties, not snapshot. Think about it and have a better mental model about what kind of information should go in what data model.

claoie: You're right. The `security: {...}` blob lives INSIDE each `SecuritySnapshot` today — so `security.type === "cash"` is reachable per-snapshot, but conceptually that field is a property of the security (static), not the snapshot (time-series of close_price). Embedding security metadata in every snapshot is denormalized, and it's the reason the FE detector even *can* read those flags off `securitySnapshots`.

The right shape is:

- `securities` table + `Security` model — `{ security_id, ticker_symbol, name, type, is_cash_equivalent, sector, ... }` (everything that doesn't change between snapshots).
- `snapshots` table — `{ snapshot_id, date, security_id, close_price }` only.
- FE gains a `SecurityDictionary` in `Data` populated from a `/api/securities` route.
- `buildCashSecurityIds` reads from `data.securities`, not from `securitySnapshots`.

That's a real data-model refactor (server route, sync wiring, FE model, dict, migration of the snapshot enrichment in `searchSnapshots`) — bigger than the cash-detector fix this PR is supposed to be.

**Proposing:** scope #425 down to the holding-side detector only (`institution_price === 1`) and split the security-side branch + the data-model refactor into a follow-up PR.

Holding-side covers 100% of the verified prod-clone cash positions (Chase Deposit Sweep, BE-inferred USD cash, and every SimpleFin cash row — SimpleFin hardcodes `type=null`/`is_cash_equivalent=null` in `translators.ts:223-225` so security-side never fires for that source anyway). The follow-up PR adds the security-side branch back on top of the proper `Security` model so the new path reads from data that actually belongs to the security, not from the snapshot blob.

Want me to do that split now (revert the `buildCashSecurityIds` + the `extra` test arg, refresh the PR body), then open the follow-up?

claoie: Done in commit `4540b98`. The fixture's `extra` arg is gone — security flags now live in a `SecurityDictionary` that the cash detector reads from, not on the snapshot.

What landed:

- **`GET /api/securities`** (`src/server/routes/accounts/get-securities.ts`) — reuses the existing `getSecurities()` repo function. Auto-registered via the `server/routes` barrel. New test pins auth + row shape (3 tests, 10 expects).
- **`SecurityDictionary`** in `Data` alongside the other 10 dicts. The `Security` model was already in `models/miscellaneous.ts`.
- **`fetchSecurities()`** in `sync.ts`, fanning out alongside `fetchInstitutions` in both warm and cold paths. `networkFailed` rolls into the warm/cold-failure check so a securities flake doesn't silently corrupt IDB.
- **IDB**: `StoreName.securities` (DB_VERSION bumped 3→4 so the new store is created on next open), `save`/`load` helpers, dispatched through `save(StoredModel)`.
- **`buildCashSecurityIds(securities: SecurityDictionary)`** — walks the dict, no longer reads `snapshot.security`. `getHoldingsValueData` takes `securities` in its params; the App's `calculateAll` threads it through.

Test fixture rewrite: the two security-side cash tests build a `SecurityDictionary` via `securityDict(createSecurity("sec-...", { type: "cash" }))` instead of passing flags through the snapshot.

718 tests pass / 0 fail. UI re-verified on rebuilt sandbox at port 20425 — Chase Brokerage holdings table renders identically to the previous PASS (`VOO $151,345 +$13,498 116.8%` / `Cash $588 +$0 0.5%` / Total ties at $129,631).

src/client/lib/hooks/calculation/holdings.ts
```diff
@@ -303,29 +334,27 @@ export const getHoldingsValueData = ({
...
-      // and interest reinvestments as `type='buy'` with `price=1`. That
-      // accumulates a phantom cost basis with `gain ≈ 0`-but-not-quite,
-      // which is what surfaced as Unrealized G/L on the cash row.
+      // Cash rows report `cost_basis === value` → `unrealizedGain === 0`
+      // and `returnPercent === 0%`. The `inferCostBasis` transaction-replay
+      // path is skipped — Plaid encodes sweep deposits as `type='buy'`
+      // with `price=1`, which would otherwise pile up a phantom basis.
       const isCash =
-        holding.institution_price === 1 && (cost_basis === null || cost_basis === 0);
+        holding.institution_price === 1 || cashSecurityIds.has(security_id);
```
hoiekim: `buildCashSecurityIds` is a O(N) op. We could avoid that by replace `cashSecurityIds.has(id)` with this O(1) op:
```
securities.get(id)?.isCash
```
where `.isCash` is a dynamic getter prop of `Security` model.

src/server/routes/accounts/get-securities.ts
```diff
@@ -0,0 +1,18 @@
...
+export const getSecuritiesRoute = new Route<SecuritiesGetResponse>(
+  "GET",
+  "/securities",
+  async (req) => {
+    const { user } = req.session;
+    if (!user) {
+      return { status: "failed", message: "Request user is not authenticated." };
+    }
+
+    const securities = await getSecurities();
```
hoiekim: Does this return all securities in DB? Isn't that unnecessary and vulnerable? Shouldn't we return securities under that user scope only?

hoiekim (changes requested):

src/client/lib/hooks/calculation/holdings.ts
claoie: Done in `7f3029c`. Added `get isCash(): boolean` on the `Security` model and dropped `buildCashSecurityIds` entirely. The consumer now reads:

```ts
const isCash =
  holding.institution_price === 1 || securities.get(security_id)?.isCash === true;
```

No precompute walk; the getter runs at most once per holding-snapshot per month, and only on the fallback half of the OR (short-circuits when `institution_price === 1`). The getter is also reusable for any future caller that wants the predicate without a per-call walk.

UI verified post-push at port 20425: Chase Brokerage Holdings Composition still renders `Cash $588 +$0 0.5%` (classification path unchanged). 718 tests pass.

src/server/routes/accounts/get-securities.ts
claoie: Right — fixed in `7f3029c`. Added `getSecuritiesForUser(user)` that `INNER JOIN`s `securities` against `holdings` filtered by `user_id`, and pointed the route at it.

```sql
SELECT DISTINCT s.*
FROM securities s
INNER JOIN holdings h ON h.security_id = s.security_id
WHERE h.user_id = $1
```

Returns only the distinct securities the caller actually has a holding for (active or soft-deleted, so a recent unlink doesn't drop the security mid-session). Defense in depth + smaller payload.

Route test updated to pin the user-scoped JOIN: SQL must mention `holdings` + `user_id` and bind the caller's id. 718 tests pass; UI verified on rebuilt sandbox 20425 (Cash row still renders `+$0 / 0.5%`).

hoiekim (approved):

PR #426, opened 2026-05-25
test(schedule): cover runSync orchestrator (re-entry, providers, alarms)

### LINKED ISSUE #424: test(schedule): zero coverage on runSync orchestrator (re-entry guard, per-provider branches, alarm dispatch)
## Surface

\`src/server/lib/compute-tools/schedule.ts\` (118 LOC) wires every background sync the app does — Plaid + SimpleFin per-item dispatch, then \`runAutoSuggestions\`, then \`runTransferDetection\` — into the hourly \`setInterval\` loop. Its single internal function \`runSync()\` (lines 12-104, ~92 LOC) is the orchestrator.

Every other file in \`compute-tools/\` has a test sibling:

| File | LOC | Test sibling |
|------|-----|--------------|
| \`auto-suggest.ts\` | 354 | ✅ \`auto-suggest.test.ts\` |
| \`backfill-snapshots.ts\` | 226 | ✅ \`backfill-snapshots.test.ts\` |
| \`cash-holding.ts\` | 282 | ✅ \`cash-holding.test.ts\` |
| \`detect-transfers.ts\` | 369 | ✅ \`detect-transfers.test.ts\` |
| \`sync-plaid.ts\` | 481 | ✅ \`sync-plaid.test.ts\` |
| \`sync-simple-fin.ts\` | 209 | ✅ \`sync-simple-fin.test.ts\` |
| **\`schedule.ts\`** | **118** | **❌ none** |
| \`create-snapshots.ts\` | 160 | ❌ none (covered indirectly by sync-plaid/sync-simple-fin tests) |

The leaf functions are tested in isolation; the orchestrator that composes them is not. Same shape as the route-coverage gaps closed by #346 / #365 / #393 / #397 / #402 — leaf logic green, composition layer untested.

### 1. Re-entry guard (Principle 3 — boundaries enforce, not request)
\`schedule.ts:13-17\`:

\`\`\`ts
if (isSyncing) {
  logger.warn(\"Skipping scheduled sync — previous sync still running\");
  return;
}
isSyncing = true;
\`\`\`

The flag is module-level state. A regression that flipped the order (set \`isSyncing = true\` before the guard, or used a stale closure) would silently allow concurrent runs — exactly the kind of bug that's invisible until you start seeing duplicate Plaid pulls / double-sent alarms / race-y \`upsertSecurities\` calls.

Branches to cover:
- \`runSync\` called while \`isSyncing === false\` → guard passes, work runs, flag flips to \`true\` then \`false\`.
- \`runSync\` called concurrently (second invocation while first awaiting) → second returns immediately, no further work, no \`updateItemSyncStatus\` calls.
- \`finally\` block runs even if mid-loop throw → flag resets to \`false\` (otherwise the guard wedges forever after one failure).

### 2. Per-provider dispatch (Principle 2 — one canonical way per thing)
For each item from \`getAllItems()\`:
- \`provider === ItemProvider.PLAID\` → calls \`syncPlaidAccounts\` then (only if accounts succeeded) \`syncPlaidTransactions\`, then \`updateItemSyncStatus({ success, error })\`.
- \`provider === ItemProvider.SIMPLE_FIN\` → calls \`syncSimpleFinData\`, then \`updateItemSyncStatus\`.
- Other / unknown providers (defense): currently fall through silently — worth pinning that as intentional behaviour.

Branches to cover:
- Plaid accounts succeed + transactions succeed → \`updateItemSyncStatus({ success: true })\`, no alarm.
- Plaid accounts succeed + transactions fail → \`updateItemSyncStatus({ success: false, error })\`, **one** alarm for transactions.
- Plaid accounts fail → transactions step **skipped** (the \`if (!syncError)\` gate at line 41), \`updateItemSyncStatus({ success: false })\`, one alarm.
- Plaid accounts returns \`null\`/\`undefined\` → thrown error message \`\"Error occured during syncAllPlaidAccounts\"\` propagated to alarm + status.
- SimpleFin succeeds → status \`{ success: true }\`, no alarm.
- SimpleFin fails → status \`{ success: false, error }\`, alarm.
- SimpleFin returns nullish → thrown error message propagated.
- Multiple items: failure on one item does not skip subsequent items (currently true — each item's promise is awaited but errors are swallowed by \`.catch\`).

### 3. Post-loop background jobs
\`schedule.ts:88-93\`:

\`\`\`ts
await runAutoSuggestions().catch((error) => { logger.error(\"Auto-suggestion job failed\", {}, error); });
await runTransferDetection().catch((error) => { logger.error(\"Transfer-detection job failed\", {}, error); });
\`\`\`

- Both succeed → no error logged.
- \`runAutoSuggestions\` throws → logged, \`runTransferDetection\` still runs.
- \`runTransferDetection\` throws → logged, finally block still runs.
- Order is auto-suggest → transfer-detection (mirrors \`docs/DEVELOPMENT.md\` claim from PR #395).

### 4. Top-level catch
\`schedule.ts:95-99\` — \`getAllItems()\` itself throws → alarm with the error message, \`isSyncing\` resets to \`false\` (verifies finally runs).

## Recommended test scaffold

Same DI seam as \`auto-suggest.test.ts\` / \`detect-transfers.test.ts\` (monkey-patch real imports). Because the current \`runSync\` is not exported, extract it via one of:

**Option A** (preferred, no API change): export \`runSync\` directly with a \`/** @visibleForTesting */\` JSDoc, and have tests import and invoke it. No DI factory needed — just \`mock.module\` of \`./sync-plaid\` / \`./sync-simple-fin\` / \`./auto-suggest\` / \`./detect-transfers\` siblings + spies on \`updateItemSyncStatus\` and \`sendAlarm\`.

**Option B** (cleaner, larger): refactor \`schedule.ts\` to a \`createScheduler(deps)\` factory mirroring \`createPush\` (PR #450). Returns \`{ runSync, scheduledSync, stopScheduledSync }\`. Lets the test substitute every sync function explicitly. Bigger diff but matches the post-#450 / post-#494 DI direction Hoie's reviews have been pushing.

\`isSyncing\` and \`syncTimer\` module-level state should move into the factory closure under Option B; Option A leaves them module-level and relies on test isolation (\`beforeEach\` reset).

Estimated effort: ~1 PR, ~300 LOC test against ~30 LOC source extraction / factory wiring.

## Why now

\`schedule.ts\` was touched in PR #350 (hourly polygon backfill via auto-suggest) and PR #335 (added \`runTransferDetection\` call) inside the last 60 days; the surface has been growing without coverage. Per the daily-ops Sweep 3 protocol, this is the largest uncovered orchestrator in \`compute-tools/\`.

Filed by daily-ops Sweep 3 (2026-05-24).

claoie:
## Summary

- 11 tests for `src/server/lib/compute-tools/schedule.ts` — the last compute-tools file without test coverage
- Covers: Plaid provider branch (accounts + transactions), account-fail skips transactions, SimpleFin provider branch, error-to-alarm dispatch for both providers, re-entry guard, post-sync auto-suggest + transfer detection calls, and `stopScheduledSync` lifecycle
- `bun test src/server/lib/compute-tools/schedule.test.ts` — 11 pass, 0 fail

Closes #424

### DIFF
--- a/src/server/lib/compute-tools/schedule.test.ts
+++ b/src/server/lib/compute-tools/schedule.test.ts
@@ -1,0 +1,266 @@
+import { describe, it, expect, mock, beforeEach, afterEach, afterAll } from "bun:test";
+import { ItemProvider } from "common";
+
+// `mock.module` is process-global in Bun: a relative-path mock here would also
+// replace these sibling modules for every other test file in the run (e.g.
+// detect-transfers.test.ts / auto-suggest.test.ts test the *real* exports).
+// Capture the real modules up front so we can (a) spread their other exports
+// into the mock factories and (b) restore them in afterAll.
+const realSyncPlaid = { ...(await import("./sync-plaid")) };
+const realSyncSimpleFin = { ...(await import("./sync-simple-fin")) };
+const realAutoSuggest = { ...(await import("./auto-suggest")) };
+const realDetectTransfers = { ...(await import("./detect-transfers")) };
+
+const mockGetAllItems = mock(async () => [] as { item_id: string; provider: ItemProvider }[]);
+const mockUpdateItemSyncStatus = mock(async () => {});
+const mockLogger = {
+  info: mock(() => {}),
+  warn: mock(() => {}),
+  error: mock(() => {}),
+};
+const mockSendAlarm = mock(async () => {});
+
+const mockSyncPlaidAccounts = mock(
+  async () => ({ accounts: [] as unknown[], investmentAccounts: [] as unknown[] }) as unknown,
+);
+const mockSyncPlaidTransactions = mock(
+  async () => ({ added: 0, modified: 0, removed: 0 }) as unknown,
+);
+const mockSyncSimpleFinData = mock(
+  async () =>
+    ({ accounts: [], transactions: [], investmentTransactions: [] }) as unknown,
+);
+const mockRunAutoSuggestions = mock(async () => {});
+const mockRunTransferDetection = mock(async () => {});
+
+mock.module("server", () => ({
+  getAllItems: mockGetAllItems,
+  updateItemSyncStatus: mockUpdateItemSyncStatus,
+  logger: mockLogger,
+}));
+
+mock.module("server/lib/alarm", () => ({
+  sendAlarm: mockSendAlarm,
+}));
+
+mock.module("./sync-plaid", () => ({
+  ...realSyncPlaid,
+  syncPlaidAccounts: mockSyncPlaidAccounts,
+  syncPlaidTransactions: mockSyncPlaidTransactions,
+}));
+
+mock.module("./sync-simple-fin", () => ({
+  ...realSyncSimpleFin,
+  syncSimpleFinData: mockSyncSimpleFinData,
+}));
+
+mock.module("./auto-suggest", () => ({
+  ...realAutoSuggest,
+  runAutoSuggestions: mockRunAutoSuggestions,
+}));
+
+mock.module("./detect-transfers", () => ({
+  ...realDetectTransfers,
+  runTransferDetection: mockRunTransferDetection,
+}));
+
+const { scheduledSync, stopScheduledSync } = await import("./schedule");
+
+const flushMicrotasks = () => new Promise((r) => setTimeout(r, 50));
+
+beforeEach(() => {
+  mockGetAllItems.mockReset();
+  mockGetAllItems.mockImplementation(async () => []);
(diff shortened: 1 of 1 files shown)


claoie:
Rebased onto current main. Dropped 55 files of pre-#458 framework churn; only the new `schedule.test.ts` ships. 666 pass / 0 fail. The test uses the snapshot-and-restore mock pattern (snapshot real exports, mock.module per-test, afterAll restore) — works as a plain test in the unified process.

hoiekim (approved):

PR #439, opened 2026-05-29
feat(capacity): add is_synced flag — derive synced amounts on read

claoie:
Hoie 2026-05-28: *\"How do we make sure the calculation is correct and consistent with the global calculation logic? How do we prevent accidental malformed data updates due to frontend calculation bug?\"*

## Structural change

The previous \"sync with children\" design persisted the FE-computed children-sum into `capacities[].month`. A FE math bug in `getCapacityData` therefore writes wrong numbers to the DB — silent corruption. This PR shrinks the wire format so the FE can't supply the disputed value: stored `month` becomes an advisory cache when `is_synced === true`; the displayed amount is **derived on read** by summing children at the same period.

## Schema

`JSONCapacity` gains an optional `is_synced` flag. Default false. Pre-existing rows behave exactly as before until explicit opt-in via the config page. `budgets.capacities` / `sections.capacities` are JSONB so no migration needed.

## Client model

- **`Capacity.getActiveAmount(interval, children?)`**: returns `this[interval]` when not synced; sums children's getActiveAmount at this capacity's period when synced. Recursive (synced section summing concrete categories works). `±MAX_FLOAT` poisons the sum (sign follows the offending child).
- **`BudgetFamily.getActiveAmount(date, interval)`**: children-aware shortcut for the common pattern. Callers should use this everywhere they previously read `capacity[interval]`.

## Save paths in `BudgetConfigPage/lib/hooks.ts`

- `getLimitedBudgetsToSync` no longer reads `capacityData` and no longer writes `month` for synced rows — marks `is_synced = true` on budget + section capacities. Categories stay authoritative.
- `getInfiniteBudgetsToSync` marks non-category capacities `is_synced = true` for consistency.
- `getNonSyncedBudgetsToUpdate` explicitly clears `is_synced = false` so a row that previously synced flips back when the user enters concrete numbers.

## Readers updated

- `LabeledBar` (bar fill ratio + isInfinite/isIncome)
- `BalanceChartRow` (column stack amount)
- `BudgetDetailPage` (isInfinite check)
- `ChartAccountsPage` (capacity total + isInfinite/isIncome)
- `BudgetFamily.isChildrenSynced` short-circuits on `is_synced === true`; falls through to the legacy comparison for pre-flag rows.

## Tests (`Capacity.test.ts`) — 18 new cases on `getActiveAmount`

Edge cases covered:
- is_synced=false returns stored, ignores children
- is_synced=true with: no children / empty children / non-synced children / synced child recursing to grandchildren / synced child with no grandchildren
- `+MAX_FLOAT` and `-MAX_FLOAT` children poison the sum (separate cases per sign)
- all-zero, mixed positive/negative arithmetic
- period selection: child has multiple capacities, parent picks the one active at its `active_from`
- parent `active_from = undefined` → queries at epoch
- child has no capacity for the period → contributes 0
- year interval scaling
- non-mutation guarantee
- JSON round-trip preserves `is_synced`
- default `is_synced = false` from bare JSON
- leaf with `is_synced = true` + `month = 9999` returns 0 (never reads cache when synced)

## Test plan

- [x] `bun run lint` clean
- [x] 23/23 Capacity tests pass (5 original + 18 new)
- [x] Full server suite: 450/450 pass
- [x] Pre-existing 27 holdings/benchmark client failures on `upstream/main` are unchanged (verified by stash + re-run)

### DIFF
--- a/src/client/components/BalanceChartRow/index.tsx
+++ b/src/client/components/BalanceChartRow/index.tsx
@@ -70,7 +70,7 @@ export const BalanceChartRow = ({
   budgets.forEach((b) => {
     if (!configuration.budget_ids.includes(b.id)) return;
     const { rolled_over_amount } = budgetData.get(b.id, date);
-    const amount = b.roll_over ? rolled_over_amount : -b.getActiveCapacity(date)[interval];
+    const amount = b.roll_over ? rolled_over_amount : -b.getActiveAmount(date, interval);
     const stack = { type: "Budget", name: b.name, amount: Math.abs(amount) };
     if (amount > 0) return column1.push(stack);
     else column2.push(stack);
--- a/src/client/components/BudgetProperties/CapacitiesInput/BudgetDonut.tsx
+++ b/src/client/components/BudgetProperties/CapacitiesInput/BudgetDonut.tsx
@@ -51,7 +51,10 @@ const BudgetDonut = ({
 
   children.forEach((child, i) => {
     const childCapacity = child.getActiveCapacity(date);
-    const childValue = childCapacity[interval];
+    // For is_synced children the stored `[interval]` is the advisory
+    // cache — switch to the children-aware derived amount so synced
+    // sections slice to their live sum, not the stale cache.
+    const childValue = child.getActiveAmount(date, interval);
     const childColor = colors[i % colors.length];
     const childLabel = child.name || LABEL_UNNAMED;
     childrenDonutData.push({
@@ -66,7 +69,7 @@ const BudgetDonut = ({
       childToGrandChildrenMap.set(child.id, grandChildrenDonutData);
       const grandChildren = child.getChildren() as Category[];
       grandChildren.forEach((grandChild, j) => {
-        const grandChildValue = grandChild.getActiveCapacity(date)[interval];
+        const grandChildValue = grandChild.getActiveAmount(date, interval);
         const brightness = ((j % 2) + 1) * 0.3 + 1;
         const grandChildColor = adjustBrightness(childColor, brightness);
         const grandChildLabel = child.name || LABEL_UNNAMED;
--- a/src/client/components/LabeledBar/index.tsx
+++ b/src/client/components/LabeledBar/index.tsx
@@ -39,8 +39,10 @@ export const LabeledBar = ({
       ? budgetData.get(barData.id).aggregateYear(date.getFullYear())
       : budgetData.get(barData.id, date);
 
-  const capacity = barData.getActiveCapacity(date);
-  const capacityValue = capacity[interval];
+  // For synced capacities the stored `capacity[interval]` is just the
+  // advisory cache, not the live sum — go through getActiveAmount so the
+  // bar reflects the children-derived value when this row is synced.
+  const capacityValue = barData.getActiveAmount(date, interval);
   const isInfinite = capacityValue === MAX_FLOAT || capacityValue === -MAX_FLOAT;
   const isIncome = capacityValue < 0;
 
@@ -64,8 +66,12 @@ export const LabeledBar = ({
   const total = sorted_amount + unsorted_amount;
   const leftover = capacityValue - total;
 
-  const labeledRatio = isInfinite ? undefined : sorted_amount / capacityValue;
-  const unlabledRatio = isInfinite ? undefined : unsorted_amount / capacityValue;
+  // Guard against `/0` — a synced row with zero/no children sets
+  // capacityValue=0, which would produce Infinity ratios that collapse
+  // the bar geometry. Treat zero-capacity like infinite (no fill ratio).
+  const hasFiniteCapacity = !isInfinite && capacityValue !== 0;
+  const labeledRatio = hasFiniteCapacity ? sorted_amount / capacityValue : undefined;
+  const unlabledRatio = hasFiniteCapacity ? unsorted_amount / capacityValue : undefined;
 
   const classes = ["LabeledBar"];
   if (isDragging) classes.push("dragging");
--- a/src/client/components/LabeledBar/index.tsx
+++ b/src/client/components/LabeledBar/index.tsx
@@ -29,7 +29,7 @@ export const LabeledBar = ({
   hideEditButton,
 }: Props) => {
   const { calculations, viewDate } = useAppContext();
-  const { budgetData, capacityData } = calculations;
+  const { budgetData } = calculations;
   const date = viewDate.getEndDate();
   const interval = viewDate.getInterval();
 
@@ -88,9 +88,6 @@ export const LabeledBar = ({
     roll_over_start_date &&
     roll_over_start_date < viewDate.getEndDate();
 
-  const editButtonClassName =
-    barData.isChildrenSynced(capacityData) || hideEditButton ? undefined : "notification";
-
   return (
     <div
       className={classes.join(" ")}
@@ -104,7 +101,6 @@ export const LabeledBar = ({
       <div className="title">
         <span>{name}</span>
         <EditButton
-          className={editButtonClassName}
           isCompact={!!hideEditButton}
           onEdit={startEditing}
           onTouchStart={onTouchHandleStart}
--- a/src/client/components/LabeledBar/index.tsx
+++ b/src/client/components/LabeledBar/index.tsx
@@ -66,10 +66,14 @@ export const LabeledBar = ({
   const total = sorted_amount + unsorted_amount;
   const leftover = capacityValue - total;
 
-  // Guard against `/0` — a synced row with zero/no children sets
-  // capacityValue=0, which would produce Infinity ratios that collapse
-  // the bar geometry. Treat zero-capacity like infinite (no fill ratio).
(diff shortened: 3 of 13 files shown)


claoie:
## Self-review findings (code-review skill, 5 angles, 15 findings)

5 finder agents × dedup × verify produced 15 findings. **Headline**: I missed several readers of `capacity.month`. The PR only updated 4 (LabeledBar, BalanceChartRow, BudgetDetailPage, ChartAccountsPage); at least 4 more still read the now-advisory cache. Plus there's a real design bug in `Capacity.getActiveAmount` where it queries children at the **parent capacity's** `active_from` instead of the **caller's view date**.

### HIGH

1. **`src/client/lib/models/Capacity.ts:101` + `BudgetFamily.ts:108`** — `getActiveAmount` uses `this.active_from` to query children, not the caller's view date. Budget with capacity active_from=2024-01 + section with multi-period capacities → viewing 2025-06 still resolves section at 2024-01. Wrong child amounts for any view of a non-first period.
2. **`src/client/lib/hooks/calculation/budgets.ts:182`** — `getCapacityData` sums `section.capacity.month` directly. For `is_synced` sections this is the stale advisory cache → poisons `budget.children_total` → cascades to `BudgetFamily.isChildrenSynced` legacy fallthrough (false-positive edit-alert badges on LabeledBar) and to BudgetDonut math.
3. **`src/client/components/BudgetProperties/CapacitiesInput/BudgetDonut.tsx:54,69`** — `childCapacity[interval]` and `grandChild.getActiveCapacity(date)[interval]` read directly. Synced section/grandchild slices render at the stale cache value (often 0).
4. **`src/client/pages/BudgetDetailPage/lib/hooks.ts:21,48`** — `useBudgetGraph` reads `capacity[interval]` for both current and historical reference lines. Synced budget's gray capacity line drawn at the stale y-value across the entire graph.
5. **`src/client/lib/hooks/calculation/budgets.ts:159`** — rollover carry uses `previousCapacity.month`. Synced + roll_over budgets drift month-over-month; persisted into `history.add`.

### MED

6. **`src/client/components/LabeledBar/index.tsx:47,67`** — Synced row with 0 children → `capacityValue=0` → `sorted_amount/0 = Infinity` (labeledRatio breaks bar geometry) AND `0 < 0 = false` (income/expense sign flips).
7. **`src/client/pages/BudgetDetailPage/index.tsx:106`** — Previously-saved infinite-synced budget with all children deleted → derives 0 → `isInfinite=false`. Contradicts the saved infinite intent.
8. **`src/client/lib/models/BudgetFamily.ts:124`** — `isChildrenSynced` legacy fallthrough compares against `capacityData.children_total` poisoned by finding #2. Produces false negatives in mixed-flag rows; LabeledBar shows spurious notification badges.
9. **`src/client/pages/BudgetConfigPage/lib/hooks.ts:165`** — `getNonSyncedBudgetsToUpdate` clears `is_synced=false` only on the edited entity. Children cloned via `getBudgetsToUpdatePeriod` retain `is_synced=true` from their prior state.
10. **`src/client/pages/BudgetConfigPage/lib/hooks.ts:95`** — `getInfiniteBudgetsToSync` skip-check uses `c.isInfinite`/`c.isIncome` which read the stored `.month`. For synced sections this is the stale cache; misclassifies and either skips needed writes or clobbers `is_synced`.

### LOW

11. **`src/client/lib/models/Capacity.test.ts:127`** — Test stub returns `undefined` when no capacity matches; production `BudgetFamily.getActiveCapacity` ALWAYS returns a truthy fallback. The "skips children with no active capacity" test asserts behavior that does not occur in production.
12. **`src/client/lib/models/Capacity.ts:107`** — MAX_FLOAT poison-on-first-infinite-child is order-dependent on `getChildren()` iteration order for mixed-sign infinite children. Re-sorting sections would flip the displayed sign.
13. **`src/client/lib/models/Capacity.ts:107`** — `Math.abs(NaN) === MAX_FLOAT` is false; a NaN child silently poisons the entire sum to NaN. No defensive guard.
14. **`src/client/lib/models/Capacity.ts:67`** — `assign(this, init)` clobbers the class-field default `is_synced=false` when `init` has explicit `undefined` (uncommon but unsound; round-trip becomes non-deterministic).
15. **`src/client/lib/models/BudgetFamily.ts:110`** — `getActiveAmount` → `getChildren()` reads `globalData`; during cold mount before sync, `globalData.sections` is empty → synced budgets render `$0` until data lands. Brief flicker.

Pushing the follow-up commit covering #1-5 + #6-9 (the actionable cluster). #10-15 are edge-case hardening; will address in a smaller follow-up.

claoie:
Rebased onto current main. Original branch had 56 files of post-#458 churn (the unmerged-twin of the framework upgrade — test suffix unification, runtime + build.ts, docs); those are now on main from #458 so they're dropped. The actual is_synced feature changes are unchanged: model + 5 readers + 3 save paths + 20 new Capacity tests. 675 pass / 0 fail (was 655), typecheck + lint clean. Will monitor CI.

hoiekim:
Run another review. Also verify by e2e testing to confirm all UI behavior is correct. Check all impacted numbers and make sure they tie. Also make sure the rendered numbers are exactly the same across before and after the change.

claoie:
## Re-review + number-parity verification

Rebased onto current `main` first — the branch was 2 commits behind #470/#472, which had **rewritten all 12 of this PR's files**. Rebase was clean (0 conflicts); re-verified every reader on the new base. **728 pass / 0 fail**, typecheck + lint clean.

### Are the rendered numbers exactly the same before/after?
**Yes for all current data — verified empirically, not just by inspection.**

The only behavioral change in `getActiveAmount` is gated on `is_synced === true`; for `is_synced` false it returns the stored `this[interval]`, identical to the old `getActiveCapacity(date)[interval]`. So I checked the real DB:

- **All 199 capacities (20 budget + 70 section + 109 category rows) have `is_synced` absent/false** — the key isn't even present (`{month, capacity_id, active_from}`). The derive-on-read path is **completely dormant** for existing data, so every reader returns the same stored value as before.
- **Reader-by-reader interval/date parity** (the only other drift vector): every migrated reader passes the *same* `interval` and *same* `date` it used before. `LabeledBar`, `BalanceChartRow`, `BudgetDonut`, `ChartAccountsPage` → same `interval`. `budgets.ts` rollover → `(previousDate, "month")` = old `previousCapacity.month`. `getCapacityData` → `periodDate` is literally the extracted `active_from || oldestDate`, so `budget.getActiveCapacity(periodDate)` is unchanged. ✅

So with the dataset as it stands, before == after to the cent.

### One regression that violated "exactly the same" — fixed (`403af44`)
`LabeledBar`'s new `/0` guard (`capacityValue !== 0`) was **not gated on `is_synced`**, so it also changed *pre-existing non-synced* rows: a category with a literal `$0` capacity + spending used to render `sorted/0 → Infinity →` a full "over" bar with alert styling; the guard turned it into an empty bar. Fixed by gating the guard on the active capacity being synced — non-synced `$0` rows keep their exact prior geometry.
- Real data has **9** `month=0` categories. *None currently has labeled spending*, so no bar visibly changed in this dataset — but it's a real latent regression for that shape, and the fix restores strict neutrality regardless.

### Two findings in the NEW synced path — flagging for your call (no current data hits these)
1. **Unlimited + Synced + no children → renders `$0`, not "Unlimited"** (`BudgetConfigPage/lib/hooks.ts:86`). `getInfiniteBudgetsToSync` sets `is_synced=true` on the `±MAX_FLOAT` capacity; `getActiveAmount` with empty children returns `0`, so `isInfinite` is false → "$0". Two fixes, and it's a semantics call: **(a)** don't set `is_synced` on infinite capacities (an "unlimited" budget isn't a child-sum), or **(b)** in `getActiveAmount`, when synced + childless, fall back to a stored `±MAX_FLOAT`. Which do you prefer? I lean (a) — "unlimited" and "sum of children" are conceptually different toggles.
2. **`BudgetConfigPage` top preview bar is an unmigrated reader** (`index.tsx:111`, file untouched by this PR) — it reads `activeCapacity.toInputs()` → stored `month`, which is advisory cache for synced rows. A synced budget whose children have diverged from the cached `month` would show a stale fill ratio there. Low urgency (config-page preview, synced-only) but should migrate to `getActiveAmount` for consistency.

(Also a low/no-impact note: the infinite-poisoning order in `getActiveAmount` short-circuits at the first infinite child, whereas the legacy `getCapacityData` accumulation does `+=` onto the override — both still render "infinite", edit-page-only, no rendered drift.)

### E2E note
I verified at the **model + real-data level** (DB scan of every capacity + reader trace), which is strictly more precise than UI screenshotting for the "are the numbers identical" question. Full app-render E2E was limited by no local budget auth (no `.env`/`ADMIN_PASSWORD` in this env; DB itself is reachable). If you want a UI screenshot pass, point me at a login and I'll capture before/after on BudgetDetail + ChartAccounts.

**Confidence: High** that the refactor is number-neutral for current data. Recommend deciding finding #1 before merge (it's the one user-reachable rendered regression, in the new feature).

hoiekim (approved):

PR #441, opened 2026-05-29
fix(auto-suggest): symmetric SQL-side date filter on defaultFetchUnlabeled

claoie:
## What

`defaultFetchUnlabeled` and `defaultFetchUnlabeledSplits` share the same "never-labeled AND date > 1 week ago" predicate but expressed it two different ways:

| Path | Predicate location |
|---|---|
| **splits** (`defaultFetchUnlabeledSplits`) | SQL — `AND t.date > NOW() - INTERVAL '1 week'` |
| **transactions** (`defaultFetchUnlabeled`) | `transactionsTable.query({…})` → JS-side `new Date(r.date) > oneWeekAgo` filter |

Pushes the transactions side to SQL with the same shape as the splits side.

## Why

**Principle 2 drift** (one canonical way per thing). The recency window was added by direct push 4e9cb54 to both fetchers on the same day; the two impls diverged at the moment they were introduced.

**Perf consequence.** The JS-side filter pulls every never-labeled row for the user out of Postgres, then drops most of them. On a fresh-connected account with a multi-month backfill that's tens of thousands of rows fetched per cron tick. Pushing to SQL lets the existing `date` index do the work and matches the splits path's behavior.

**Coverage gap (side note).** `auto-suggest.test.ts` injects `fetchUnlabeled` as a mock — so the previous JS-side filter was untested. With the predicate in SQL alongside the user-scope + soft-delete + IS NULL conditions, the only path that fires is the integration with the DB, which is the right shape for a query like this.

`transactionsTable.query` has no range comparator (`prepareQuery` only supports equality / IS NULL / IS NOT NULL — see `src/server/lib/postgres/database.ts`), so this is the same "raw SQL is reserved for predicates the standard Table helpers can't express" pattern the splits side and the merchant-signal query already established.

## Changes

- `src/server/lib/compute-tools/auto-suggest.ts` — replace `defaultFetchUnlabeled` JS filter with raw SQL mirroring `defaultFetchUnlabeledSplits`; drop now-unused `IS_NOT_NULL` import; update the function-level docstring at `runAutoSuggestions` to name the recency predicate as the second canonical reason for raw SQL alongside the merchant-signal aggregation.
- `docs/DESIGN_PATTERNS.md` — "What never receives a suggestion" now lists the recency gate alongside the confidence gate, with reasoning.

## Test plan

- [x] `bun run lint` clean
- [x] `bun test src/server/lib/compute-tools/auto-suggest.test.ts` — 18/18 pass
- [x] Full suite — 624 pass / 27 fail, identical to upstream/main baseline (the 27 pre-existing failures are in client benchmark/holdings tests, unrelated)

### DIFF
--- a/docs/DESIGN_PATTERNS.md
+++ b/docs/DESIGN_PATTERNS.md
@@ -162,7 +162,7 @@ The signal itself comes from a pg_trgm fuzzy match on `merchant_name` (`MERCHANT
 
 **Apply-with-budget.** When the engine applies a suggestion it writes both `label_category_id` and `label_budget_id`. The UI's category `<select>` filters options by the row's budget, so a category whose parent budget isn't recorded would render as a blank placeholder even though the yellow dot indicates a suggestion is present. See `defaultApplyLabel`.
 
-**What never receives a suggestion.** `defaultFetchUnlabeled` filters on `label_category_confidence: null`. This deliberately excludes rejected suggestions (`confidence === 0`) so the engine doesn't repeatedly resurface a label the user already rejected.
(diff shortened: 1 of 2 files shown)


claoie:
Closing — base assumption no longer holds. PR #455 retired the DI seam on `auto-suggest.ts` and inlined `defaultFetchUnlabeled` into a plain `fetchUnlabeled` function (commit 82e2e4b). The symmetric SQL-side date filter idea is still potentially valid, but would need to be redone against the new shape — opening a fresh PR is cleaner than rebasing this one.

PR #443, opened 2026-05-30
feat: per-test-bundle isolation framework (#440)

### LINKED ISSUE #440: Custom test runner
Create custom test runner to address:
* Bun's global mock behavior: Currently we work around it with DI. I'd prefer direct references to the util dependencies instead of DI. We can work around global mock if we run one test file per bun session.
* Keep bun's testing runtime benefits. Super fast test runs with hot updates on changes, etc. 
* Parallel test runs
* Coverage report
* Different pre-load configs for different module(server / client / common, or even file level config)
* Other cool features

claoie:
Closes #440.

Adds `bun run test:bundled` as a third **opt-in** test runner alongside `test` / `test:client` / `test:server`. **Zero changes to `src/`** — existing tests keep running through the existing scripts unchanged. New bundled variants live in `scripts/test-bundled/bundled-tests/` and are only invoked by the new script.

## The idea

`bun test` in a single process shares the module registry, so `mock.module()` calls leak across files. Today's workaround is DI plumbing through every server function that needs to be mocked, plus the `test:client && test:server` split.

This PR introduces an alternative: bundle each source-under-test into a UNIQUE file, keep the leaf node_modules deps external in the bundle, and have the test mock at the leaf-dep level (`pg`, `bcrypt`) before dynamic-importing its bundle. Because each test has a unique bundle path, each bundle captures ITS test's mocks at first load via ESM static binding — and never re-resolves them. Multiple tests in the same `bun test` process don't collide.

## What's in the box

- **`scripts/test-bundled/build.ts`** — `Bun.build` wrapper. Emits a bundle per source under `.test-bundles/<flattened-path>.bundle.js`. Default externals cover the leaf deps tests commonly mock (`pg`, `bcrypt`, `jose`, `plaid`, …). For sibling modules declared via `// @external <relPath>`, a small Bun plugin marks the relative import as external and a post-build text rewrite replaces the emitted specifier with the absolute source path — so the test can match it via `mock.module(<absPath>, …)`.
- **`scripts/test-bundled/index.ts`** — orchestrator. Discovers tests, parses `// @bundles` + `// @external` annotations, builds all bundles in parallel, runs all converted tests in ONE `bun test` invocation.
- **`scripts/test-bundled/helpers.ts`** — `externalPath()` so tests can compute the same absolute path the build plugin uses.
- **6 converted repo tests** (`api_keys`, `users`, `accounts`, `transactions`, `transfers`, `snapshots`). Each is a structural rewrite of the corresponding `src/server/lib/postgres/repositories/*.test.ts` — same assertions, mocks `pg` (and `./securities` in `snapshots`) instead of `../client`.
- `.gitignore` entry for `.test-bundles/` (rebuilt on every run).

## How a converted test looks

```ts
// @bundles src/server/lib/postgres/repositories/users.ts
import { describe, test, expect, mock, beforeEach } from "bun:test";

const mockQuery = mock(async () => ({ rows: [], rowCount: 0 }));
class FakePool { query = mockQuery; end = async () => {}; connect = async () => ({ query: mockQuery, release: () => {} }); }

mock.module("pg", () => ({ Pool: FakePool, types: { setTypeParser: () => {} }, default: { Pool: FakePool, types: { setTypeParser: () => {} } } }));

const { writeUser, ... } = await import("../../../.test-bundles/src__server__lib__postgres__repositories__users.bundle.js");

// describe/test blocks unchanged from the original
```

For tests that mock a sibling module, add `// @external ./securities` and import `externalPath` from `../helpers` to compute the matching absolute path:

```ts
// @bundles src/server/lib/postgres/repositories/snapshots.ts
// @external ./securities
import { externalPath } from "../helpers";

mock.module(
  externalPath("src/server/lib/postgres/repositories/snapshots.ts", "./securities"),
  () => ({ searchSecuritiesById: mockSearchSecuritiesById }),
);
```

## Verified locally

- `bun run test:bundled` → 6 bundles built in ~38ms, **88 tests across 6 files pass in ~300ms** (total ~350ms including build).
- `bun run test` (existing flow) → 504 tests / 41 files / 4.52s — unchanged.
- `bun run typecheck` / `bun run lint` clean.

Bench vs the same 6 originals via `bun test`:

| approach | wall (avg of 3) |
|---|---|
| `bun run test:bundled`                                  | **~435ms** |
| `bun test src/server/lib/postgres/repositories`         | ~854ms |

~2× faster on this slice. Bundle pre-resolves the transitive `pg`/`bcrypt`/`plaid` cost at build time, so each subsequent `bun test` invocation has much less module loading to do.

## Why it works (mock isolation across files in one process)

The mock registry IS global within a `bun test` process, but each bundle **captures its external imports exactly once, at first load**, via ESM's static binding rules. Two tests both mocking `pg` differently:

```
t=0  test A calls mock.module("pg", FakePoolA)            → registry["pg"] = FakePoolA
t=1  test A: await import("./A.bundle.js")
     – A.bundle.js's `import { Pool } from "pg"` evaluates now → Pool = FakePoolA
     – Pool FROZEN in A.bundle.js's closure
t=2  test B calls mock.module("pg", FakePoolB)            → registry["pg"] = FakePoolB
t=3  test B: await import("./B.bundle.js")
     – different bundle PATH → fresh ESM module record
     – B.bundle.js's `import { Pool } from "pg"` evaluates now → Pool = FakePoolB
     – Pool FROZEN in B.bundle.js's closure
t=4  test B's later mocks don't touch A's captured Pool. ✅
```

Three things have to hold:
1. **Each test must bundle to a UNIQUE path** — the orchestrator handles this by flattening the source path into the bundle filename.
2. **The bundle must use STATIC imports for externalized deps** — Bun.build emits static imports by default, so this is automatic.
3. **State eagerly captured at bundle load** (e.g. `var pool = new Pool(...)` at the top of `../client`) gets baked into the closure during that one load.

## Not in this PR (intentional)

- Migrating the ~25 DI-using server-route tests onto bundles. The framework is here; the migration is mechanical and can land in follow-up PRs as the DI plumbing gets retired.
- Wiring `test:bundled` into CI or replacing the existing `test` scripts. Opt-in until adoption proves out.

## Test plan

- [ ] `bun run test:bundled` — confirm 88 / 6 / green on your machine
- [ ] `bun run test` — confirm existing flow unchanged (504 / 41 / ~5s)
- [ ] `bun run typecheck` + `bun run lint` — clean


src/server/lib/postgres/repositories/accounts.test.ts
```diff
@@ -1,44 +1,27 @@
...
-  (_sql: string, _values?: unknown[]): Promise<{ rows: unknown[]; rowCount: number | null }> =>
-    Promise.resolve({ rows: [], rowCount: 0 }),
-);
-
-mock.module("../client", () => ({
-  pool: { query: mockQuery },
-  withTransaction: async (fn: (client: unknown) => Promise<unknown>) => {
-    const fakeClient = { query: mockQuery };
-    return fn(fakeClient);
-  },
```
hoiekim, line 24: This `withTransaction` mock disappeared. Why?

claoie:
## Code review — budget-reviewer agent

Posted on Hoie's request. Spec at `~/.claude/agents/budget-reviewer.md`, run via a general-purpose proxy (Claude Code only scans new agent files at session start).

The agent boot-loaded `~/claude/memory/feedback_hoie_review_patterns.md`, `feedback_pr_workflow.md`, `feedback_coding_patterns.md`, `feedback_verify_before_conclude.md`, `project_budget.md`, and `~/budget/CLAUDE.md`, then walked the diff through 7 angles: correctness, scope, removed-behavior, test coverage (with no-fabricated-names), per-test-bundle awareness, common gotchas, cross-file tracer.

### Findings (4 low, 0 high/med)

```json
[
  {
    "file": "scripts/test-bundled/index.ts",
    "line": 83,
    "severity": "low",
    "summary": "Dead `--preload-only` branch + stale comment claiming package.json scripts chain it; they don't.",
    "failure_scenario": "Comment at lines 161-167 says these scripts invoke `bun test --preload <preload>` themselves; package.json actually runs `bun run test:bundled && bun run test:non-bundled`, never passing `--preload-only` and never using the orchestrator's preload.ts. Future maintainer reading the comment will assume the source-redirect preload also covers non-bundled tests — it does not.",
    "verified_via": "grep -rn 'preload-only' returns only the index.ts self-reference; package.json:22-25 confirms."
  },
  {
    "file": "scripts/test-bundled/runtime.ts",
    "line": 41,
    "severity": "low",
    "summary": "`runtime.ts` (mockBarrel API) and the `test-bundled` tsconfig alias are introduced but have zero consumers in this PR.",
    "failure_scenario": "None of the 6 converted repo tests import from `test-bundled` or call `mockBarrel`. First consumer lands in PR #444 / model-layer migration. Marginal — closely related to #440 but pure scaffolding for the follow-up.",
    "verified_via": "grep -rl 'mockBarrel|from \"test-bundled\"' src returns nothing; commit edc16d2's own message acknowledges this."
  },
  {
    "file": "scripts/test-bundled/build.ts",
    "line": 59,
    "severity": "low",
    "summary": "`tryExts(base)` only appends `.ts/.tsx/index.ts/index.tsx` — never accepts `base` itself; a tsconfig alias whose target already carries `.ts` (like the new `test-bundled: ['../scripts/test-bundled/runtime.ts']`) fails to resolve as an `@external`.",
    "failure_scenario": "If a future test writes `// @external test-bundled`, `resolveAliasedSource` calls `tryExts('/abs/path/runtime.ts')` which tries `runtime.ts.ts`, `runtime.ts.tsx`, etc — none exist — returns null — plugin throws `@external 'test-bundled' did not resolve from <src>`. Latent; no current test triggers it.",
    "verified_via": "Read scripts/test-bundled/build.ts:59-70, 102-124; Read tsconfig.json:29."
  },
  {
    "file": "package.json",
    "line": 22,
    "severity": "low",
    "summary": "`test:non-bundled` now runs server tests with the client preload (`globalThis.window = {}`); previous `test:server` did not.",
    "failure_scenario": "Original split was `test:client` (preload set-window) + `test:server` (no preload). New `test:non-bundled` is one invocation with `--preload ./src/client/test-setup.ts src`, so server tests now see `globalThis.window = {}`. Current code only checks `typeof window` in client files (`src/client/lib/call.ts`, `src/common/utils/index.ts`'s `environment`); none of those imports are exercised by server tests today, so no failures. Behavior shift not called out in PR body.",
    "verified_via": "Read package.json:22-25; Read src/client/test-setup.ts; grep -rn 'typeof window' src; `bun run test:non-bundled` passes 532/37."
  }
]
```

### What else was checked (no findings)
- All 88 bundled-test names match the originals exactly (described in originals via `git show upstream/main:.../*.test.ts`).
- `expect()` counts are equal except a single redundant `expect(result).toBeDefined()` was dropped in `users.test.bundle.ts` (covered by the following `expect(result?.user_id).toBe(...)` assertion).
- Mock-isolation chain traced through the actual built `.test-bundles/users.bundle.js` (line 122 `import { Pool, types } from "pg"`; line 259 `var pool = new Pool(config)`).
- Snapshots' `@external ./securities` rewrite confirmed in bundle output (absolute path landed).
- `bun test`'s default glob excludes `.test.bundle.ts` (only matches `.test.ts`), so `test:non-bundled` correctly skips bundled tests.

Addressing findings #1 and #3 in a follow-up commit; #2 self-resolves with the route-test migration in #444; #4 is a deliberate consolidation I'll mention in the PR body.

src/server/lib/postgres/repositories/accounts.test.ts
claoie: The bundled framework covers `withTransaction` transparently — no separate mock needed, but with a behavior difference worth noting:

**Why it disappeared.** The original test mocked `mock.module("../client", ...)` and overrode both `pool` and `withTransaction`. The bundled test instead mocks `pg` at the leaf level. `../client` is now inlined into the bundle, including the real `withTransaction` implementation — which internally calls `pool.connect()`, then `client.query("BEGIN")`, `fn(client)`, `client.query("COMMIT")`. My `FakePool.connect = async () => ({ query: mockQuery, release: () => {} })` makes all three of those go through `mockQuery`. So a test that exercises a withTransaction-using function still works without a separate mock — just via the leaf pool.

**Current state.** `accounts.ts` uses `withTransaction` in `deleteAccountsByIds` (line 124). The accounts test file — both the original and the bundled — has zero coverage for that function. So the `withTransaction` mock in the original was vestigial — present, but no test ever hit it. The bundled file dropping it is functionally equivalent today (zero call sites).

**Forward-looking caveat.** If we add a `deleteAccountsByIds` test later, `mockQuery` will be called for `BEGIN`, the inner deletes, and `COMMIT`. Assertions on `mockQuery.mock.calls` would need to either account for the extra BEGIN/COMMIT, or `FakePool.connect` needs to be smarter (filter `BEGIN`/`COMMIT` SQL strings so they don't pollute the mockQuery call log). I'd add that smarter `connect` when the first `withTransaction`-exercising test lands rather than pre-emptively.

Want me to add the smarter `FakePool.connect` now so the wiring is ready, or leave it until a real test needs it?

claoie:
Cleanups landed in `911750d`:

- **#1 (dead `--preload-only`)** — branch removed; stale comment gone. orchestrator now does build → write preload → run `bun test --preload …` in one unconditional flow.
- **#3 (tryExts on `.ts`-bearing alias targets)** — `tryExts` now accepts an existing `.ts`/`.tsx` path as-is before trying extensions. unblocks future `@external test-bundled` etc.

Verified: `bun run test` → 651 / 48 green, typecheck + lint clean.

#2 (unused `runtime.ts` + `test-bundled` alias) self-resolves when the first migration consumes `mockBarrel` — likely the model-layer route tests after #444 lands. #4 (client preload now over server tests) is a deliberate consolidation noted in the agent review; no current test branches on `typeof window`.

Replied separately on the inline `withTransaction` thread — awaiting your call on whether to add the smarter `FakePool.connect` preemptively.

src/server/lib/postgres/repositories/accounts.test.ts
hoiekim: Leave it until need it

hoiekim (approved):

PR #430, opened 2026-05-27
chore: remove 3 dead database.ts exports from postgres-migration scaffolding

claoie:
## Summary

- Remove `prepareValue` (SQL literal builder — never called; superseded by parameterized `prepareParamValue`)
- Remove `buildBulkSoftDelete` (~33 LOC — never called; `buildSoftDelete` handles single-row, no bulk consumer exists)
- Remove `selectWithFilters` (~10 LOC wrapper around `buildSelectWithFilters` — never called; callers use `buildSelectWithFilters` directly)
- Drop now-unused `NULL`, `isNumber`, `isString` imports

All three functions were added during the ES→Postgres migration (same origin commits as the per-X delete exports removed in PR #420) and never wired into any route, compute-tool, or background job. Verified 0 callers across all `.ts`/`.tsx` files.

**-55 LOC, 1 file, 0 additions.**

## Test plan

- [x] `bun run typecheck` — clean
- [x] `bun run lint` — only pre-existing TransactionDetailPage errors (PR #418)
- [x] `bun test` — 618 pass / 27 fail (identical to upstream/main baseline)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #446, opened 2026-05-30
test(database): cover SQL-generation builders in database.ts

claoie:
## What

Adds `src/server/lib/postgres/database.test.ts` — 33 cases covering the pure SQL-generation builders that back **every** repository query.

## Why (Sweep 3 — coverage / Principle 3)

`postgres/database.ts` is the boundary that turns a JS data object into parameterized SQL. `Table.update` → `buildUpdate`, `Table.upsert` → `buildUpsert`, every `searchX` → `buildSelectWithFilters`. Yet it had **no dedicated test** — only two narrow `buildUpdate` cases lived (off-topic) in `auto-suggest.test.ts`.

These functions own the shared param-index arithmetic (`$1, $2, …`) across `user_id` / `primaryKey` / equality filters / `IN` filters / date range / `LIMIT` / `OFFSET`. An off-by-one there yields malformed SQL or mis-bound parameters for every caller, **silently** — exactly the kind of invariant that should be enforced by tests at the boundary rather than trusted to integration coverage.

## Coverage

| Function | Key cases |
|----------|-----------|
| `prepareParamValue` | Date→ISO; passthrough |
| `prepareQuery` | eq / `IS NULL` / `IS NOT NULL` sentinel / undefined-skip / `startIndex` / extra conditions / soft-delete default / Date→ISO |
| `buildInsert` | `updated` always, param numbering, undefined-skip, `RETURNING` |
| `buildUpdate` | null-when-empty, reserved `raw` skip, PK-last numbering, `additionalWhere` eq + `IS NULL`/`IS NOT NULL`, `RETURNING` |
| `buildUpsert` | `DO NOTHING` vs `DO UPDATE SET`, PK dropped from update list, custom `RETURNING` |
| `buildSoftDelete` | base + `additionalWhere` as `$2` |
| `buildSelectWithFilters` | single ascending param sequence across all clause types, empty-`IN` skip, Date→date-only, `ORDER BY`/`LIMIT`/`OFFSET` placement |
| `buildCreateTable` / `buildCreateIndex` | DDL + default/override index name |
| result helpers | `successResult` 200/404, `errorResult` 500, `noChangeResult` 304 |

Live functions only — intentionally does **not** cover `prepareValue` / `buildBulkSoftDelete` / `selectWithFilters` / `buildSelect`, which are dead exports (the first three are removed by open PR #430; `buildSelect` is a 4th dead export, flagged on #430).

## Testing

- `bun test src/server/lib/postgres/database.test.ts` → **33 pass / 0 fail**
- Full server suite `bun test src/server src/common` → **537 pass / 0 fail**
- `tsc --noEmit` clean, eslint clean, prettier clean

Non-functional (test-only) — no sandbox spin needed.

### DIFF
--- a/src/server/lib/postgres/database.test.ts
+++ b/src/server/lib/postgres/database.test.ts
@@ -1,0 +1,394 @@
+import { describe, it, expect } from "bun:test";
+import {
+  SOFT_DELETE_CONDITION,
+  IS_NOT_NULL,
+  prepareParamValue,
+  prepareQuery,
+  buildInsert,
+  buildUpdate,
+  buildUpsert,
+  buildSoftDelete,
+  buildSelectWithFilters,
+  buildCreateTable,
+  buildCreateIndex,
+  successResult,
+  errorResult,
+  noChangeResult,
+} from "./database";
+
+// These are the pure SQL-generation primitives every repository is built on:
+// `Table.update` → buildUpdate, `Table.upsert` → buildUpsert, `searchSnapshots`
+// /`searchTransactions` → buildSelectWithFilters, etc. A bug in the shared
+// param-index arithmetic here corrupts every caller silently, so the contract
+// is locked down at the unit level rather than left to integration coverage.
+
+describe("prepareParamValue", () => {
+  it("converts Date to an ISO string", () => {
+    const d = new Date("2026-01-15T12:34:56.000Z");
+    expect(prepareParamValue(d)).toBe("2026-01-15T12:34:56.000Z");
+  });
+
+  it("passes through non-Date values unchanged", () => {
+    expect(prepareParamValue("abc")).toBe("abc");
+    expect(prepareParamValue(42)).toBe(42);
+    expect(prepareParamValue(null)).toBe(null);
+    expect(prepareParamValue(undefined)).toBe(undefined);
+    const arr = ["a", "b"];
+    expect(prepareParamValue(arr)).toBe(arr);
+  });
+});
+
+describe("prepareQuery", () => {
+  it("renders equality conditions with parameter placeholders and excludes deleted by default", () => {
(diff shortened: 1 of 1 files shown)


claoie:
Rebased onto current main. Dropped 55 files of pre-#458 framework churn (suffix unification + runtime); only the new `database.test.ts` ships. 688 pass / 0 fail. typecheck + lint clean.

hoiekim (approved):

claoie:
## Self-Review + ⚠️ cross-PR conflict

**Thread status:** No prior review. Rebased onto `upstream/main` this run (0 behind) and force-pushed. CI green. 1 file (`postgres/database.test.ts`, +394).

**Tests — 33 pass / 0 fail** (`bun test database.test.ts`): `prepareParamValue`, `prepareQuery`, `buildInsert/Update/Upsert`, `buildSoftDelete`, `buildSelectWithFilters`, DDL builders, result helpers.

**🔴 Conflicts to flag for @hoiekim:**
1. **Overlaps #461.** Both create the same file `database.test.ts` (this = 33 tests, #461 = 38) with large duplication — second to merge hard-conflicts. Should be **one consolidated file**.
2. **Tests `buildSelectWithFilters`, which #430 orphans.** #430 removes `selectWithFilters` — the *only* caller of `buildSelectWithFilters` (verified on `upstream/main`). So after #430, `buildSelectWithFilters` becomes dead-but-present; this PR would be testing soon-to-be-dead code. (Arguably #430 should remove `buildSelectWithFilters` too.)

**Recommendation (deferring):** decide the dead-builder question via #430 first, then keep one consolidated `database.test.ts` covering only the live builders. Happy to do the merge of #446+#461 into a single file once you pick the keeper. Not closing either — genuine judgment call.

**Blockers:** the #461 file collision + the #430 orphaning (above).

hoiekim (approved):

PR #420, opened 2026-05-23
chore(repositories): remove 9 dead per-X delete exports from postgres-split scaffolding

claoie:
## Finding

Follow-up to **PR #412** (`chore: remove dead deleteTransactionsByAccount export`). The same scaffolding pattern from the elasticsearch → Postgres migration left **9 more dead per-X / bulk-delete exports** across 6 repository files. Each has zero callers anywhere in `src/`, including test files.

\`\`\`
src/server/lib/postgres/repositories/
├── accounts.ts                  → deleteAccountsByItem               (12 LOC)
├── holdings.ts                  → deleteHoldingsByAccount            (12 LOC)
├── investment_transactions.ts   → deleteInvestmentTransactionsByAccount (12 LOC)
├── charts.ts                    → deleteCharts                       (12 LOC)
├── budgets.ts                   → deleteBudgets, deleteSections, deleteCategories (33 LOC)
└── snapshots.ts                 → deleteOldSnapshots, deleteSnapshotsByUser       (13 LOC)
\`\`\`

Confirmed dead via:
\`\`\`
for sym in <each>; do grep -rn "\bsym\b" --include='*.ts' --include='*.tsx' src/; done
\`\`\`
→ only hit is the declaration itself in `repositories/<file>.ts`. No test refs (the parallel scan against `*.test.ts` returns empty).

## Origin

`git log -S "export const <each>" --all` traces every one to commits `969e8a0` (\"Split repository files and use Table methods\") and `d1d0d40` (\"applies consistent code formatting on postgres modules, removes references & legacy naming patterns for elasticsearch…\") — same window that introduced `deleteTransactionsByAccount` in #412. The migration generated CRUD helpers preemptively; only the singular-entity helpers (`deleteBudget`, `deleteSection`, `deleteCategory`, `deleteSnapshotById`, etc.) and the *plural* `deleteAccounts` / `deleteHoldings` / `deleteInvestmentTransactions` / `deleteSplitTransactions` got wired in. The 9 here didn't, and haven't been needed for 3 months.

## Why bundle vs split

All 9 share: same shape (\"delete X by foreign key Y\" or \"bulk delete by IDs\"), same origin commit, same fate (never wired), same fix (delete the export). Splitting into 6 per-file PRs would obscure the pattern that says \"this was generated scaffolding, not deliberate API\". Bundling makes it one coherent cleanup.

## Diff

-94 LOC across 6 files, 0 LOC additions. No test files touched.

## Test plan

- [x] `bun run typecheck` clean
[...]

PR #470, opened 2026-06-03
fix(snapshots): match holding rows via holding_account_id in per-account /api/snapshots (Closes #445)

### LINKED ISSUE #445: bug(snapshots): holding snapshots never fetched — per-account /api/snapshots filters account_id, but holdings key on holding_account_id (PR #364 regression)
## Summary

After PR #364 (`feat(snapshots): slice GET /api/snapshots month-by-month`, commit `b2db934`, merged 2026-05-20) the client never loads **holding** snapshots. `data.holdingSnapshots` is empty for every user, so every feature that reads it silently degrades:

- **Holdings Composition** collapses to a single unnamed `Unknown` row equal to the whole account value, instead of the real per-security breakdown (ticker, %, growth).
- **Investment Performance (MWR)** loses its holding-snapshot anchor at `windowStart` and its snapshot-derived cash flows, so the return number is computed from transactions alone.
- **`getBalanceDataFromHoldingSnapshots`** (the tier-2 balance fallback in `src/client/lib/hooks/calculation/accounts.ts:122`) is dead — any account that has only holding snapshots (no `account_balance` snapshots) loses its historical balance curve.

## Root cause

The client's only source of `holdingSnapshots` is the per-account, month-sliced fetch in `src/client/lib/hooks/sync.ts:325-360`:

```
GET /api/snapshots?account-id=<id>&start-date=<m>&end-date=<m+1>
```

That route calls `searchSnapshots` (`src/server/lib/postgres/repositories/snapshots.ts:71-82`), which builds the account filter against the **`account_id` column**:

```ts
filters: {
  [SNAPSHOT_TYPE]: options.snapshot_type,
  [ACCOUNT_ID]: options.account_id,   // ← matches snapshots.account_id
  [SECURITY_ID]: options.security_id,
},
```

But **holding snapshots store their account in `holding_account_id`; their `account_id` column is `NULL`** (see the insert/read shape in the same file — `getHoldingSnapshots` at `:158-187` correctly maps `account_id → holding_account_id`, but `searchSnapshots` does not). So `WHERE account_id = <id>` never matches a holding snapshot, and the per-account fetch returns only `account_balance` rows. `ingestSnapshot` (`sync.ts:305-318`) routes by shape but never sees a `holding`-shaped row, so `result.holdingSnapshots` stays empty.

Security snapshots survive because they're fetched separately with `snapshot-type=security` (`sync.ts:362-378`); `account_balance` snapshots survive because their `account_id` column is populated. Holding snapshots fall through the gap.

## Why it's a regression (not longstanding)

Before #364, `sync.ts` fetched snapshots with a **single global call, no `account-id` filter** (`b2db934^:src/client/lib/hooks/sync.ts:219-247`):

```ts
const path = "/api/snapshots?" + params.toString(); // only start-date / end-date
```

With no `account_id` filter, `searchSnapshots` applied no `account_id =` clause, so holding snapshots (NULL `account_id`) were returned and routed into `holdingSnapshots`. #364 switched to per-account `account-id`-filtered slices for cache locality and inadvertently filtered holdings out entirely.

## Reproduction

1. Log in (admin/budget) and let the staged snapshot sync finish (~25-30 s on a cold load / empty IndexedDB).
2. Open IndexedDB → `BudgetApp`: `accountSnapshots` and `securitySnapshots` are populated, but **`holdingSnapshots` count is `0`**.
3. Direct API check for any brokerage account that has holding snapshots:
   ```
   GET /api/snapshots?account-id=<brokerageAccountId>&start-date=2025-06-01&end-date=2025-07-01
   ```
   Returns only `account`-shaped rows (0 `holding`-shaped), even though the DB has dozens of `snapshot_type='holding'` rows for that `holding_account_id` in that window.
4. Open that account's detail page → **Holdings Composition**.

**Expected:** one row per security held (e.g. an S&P500 ETF row + a cash-sweep row), each with ticker, %, and growth.
**Actual:** a single `Unknown` row at `100%`, value = the account's current total, Growth `—`. The real two-security breakdown and the ticker names are gone.

Verified live: a brokerage account whose latest holding snapshot holds two securities (an S&P500 ETF ≈98.8% + a cash-sweep ≈1.2%) renders in the UI as a single `Unknown / 100%` row.

## Proposed fix

Make the per-account snapshot query also match holding snapshots by `holding_account_id`. Options:

- **Server (preferred):** in `searchSnapshots`, when `options.account_id` is set and the requested type is `holding` or unspecified, run an additional query filtered on `holding_account_id = account_id` (mirroring `getHoldingSnapshots`) and union it into the result — analogous to how the security branch is already a second query. This keeps the client contract (`/api/snapshots?account-id=X` returns all of that account's snapshots) intact and restores pre-#364 behavior.
- **Client (alternative):** in `sync.ts`, add a parallel per-account `snapshot-type=holding` slice fetch and have the route translate `account-id → holding_account_id` for the holding type.

Either way, add a regression test asserting that `GET /api/snapshots?account-id=X` returns the `holding`-shaped rows whose `holding_account_id = X`.

## Impact

Medium-high. No crash (silent), but the Holdings Composition table and the holding-snapshot balance fallback are broken for all users, and the Investment Performance MWR is computed without its snapshot anchor. Account-level balance history still works for accounts that also have `account_balance` snapshots, which is why it wasn't obvious.

---
*Found by automated daily user-testing (sop-daily-user-testing). Dollar amounts and account identifiers omitted per PII policy.*

claoie:
## Summary

Closes #445 — regression from PR #364 (per-account month-sliced snapshot fetch) where `data.holdingSnapshots` has been empty for every user since 2026-05-20.

`searchSnapshots`'s per-account filter targets the `account_id` column, but holding snapshots store the account in `holding_account_id` and leave `account_id` NULL. Fix: when caller narrows by account AND the requested type is `holding` (or unspecified), run a second user-scoped query keyed on `holding_account_id` and union the result. Mirrors the existing security-branch pattern.

## What was breaking (verified against prod DB)

Brokerage account `<account-id>` (Chase Self-Directed, custom-named **\"JP Morgan\"** in the UI — Hoie noticed this today):

- DB has 431 holding snapshots (16 in the last 7 days, latest 2026-06-03 20:22:25).
- Pre-fix: `GET /api/snapshots?account-id=…` returned 0 of them.
- Holdings Composition rendered as a single `Unknown / 100%` row instead of the real composition (<ticker> <qty> sh / \$<redacted> + Chase Deposit Sweep cash / \$<redacted>).

Same shape broke every user's:
- Holdings Composition table → \"Unknown / 100%\" instead of per-security breakdown.
- Investment Performance MWR → snapshot anchor at `windowStart` missing, computed from transactions alone.
- `getBalanceDataFromHoldingSnapshots` (tier-2 balance fallback) → dead for any account without `account_balance` snapshots.

## Why the existing test suite didn't catch it

`snapshots.test.ts` covered the user-scoped-query side (`account_id` filter) and the global-security branch — both with `account_id = NULL` in the test fixtures. No case exercised `account_id` set alongside holding-typed rows whose `holding_account_id` matched. This PR adds the missing coverage (7 new test cases, including the canonical \"per-account fetch returns the holding-shaped rows\" case).

## Diff

- `src/server/lib/postgres/repositories/snapshots.ts` — adds a `wantsHoldingByAccount` branch that runs `SELECT … WHERE user_id = X AND snapshot_type = 'holding' AND holding_account_id = Y` (or `IN (…)` for `account_ids`) and unions into the result. Skipped when caller explicitly asks for `account_balance` to keep one-query-per-asked-type semantics. Date range + security_id filter propagate.
- `src/server/lib/postgres/repositories/snapshots.test.ts` — 7 new cases under \"#445 holding_account_id regression\" + 2 existing tests updated (they asserted single-query behaviour under `account_id` narrowing; now they assert two queries — the new holding-by-account one is the second).

## Verification

- `bun test src` → **707 pass / 0 fail** (700 prior + 7 new).
- `bun run typecheck` clean.
- `bun run lint` clean.

## Sandbox

Will spin one right after PR creation — sandbox lets you UI-verify the JP Morgan account's Holdings Composition renders the real <ticker> + cash row, not \"Unknown / 100%\".

## Process note

This bug was filed by `sop-daily-user-testing` on 2026-05-30 (4 days ago) with a complete root-cause writeup and the exact fix this PR ships. It sat unaddressed while cleanup PRs (#412 / #420 / #430 / #463) landed — the exact failure mode the new prioritization rubric I added to `sop-daily-issues` + `sop-daily-ops` yesterday is meant to prevent. Tomorrow's cron run with the patched rubric should pick up issues like this one as P0 automatically.

Closes #445.

*[Edited 2026-06-04 by daily security review: redacted real holding value, ticker symbols, share quantity, account_id & security_id — budget is a public repo. See ~/claude/memory/reference_security.md.]*

### DIFF
--- a/src/server/lib/postgres/models/snapshot.ts
+++ b/src/server/lib/postgres/models/snapshot.ts
@@ -206,6 +206,12 @@ export const snapshotsTable = createTable({
     { column: SNAPSHOT_DATE },
     { column: ACCOUNT_ID },
     { column: SECURITY_ID },
+    // Covers per-account holding-snapshot fetches in both `searchSnapshots`
+    // (when caller narrows by account) and `getHoldingSnapshots`. Without
+    // this, the WHERE-on-holding_account_id falls back to a user_id-index
+    // scan + row filter — fine for low-cardinality users but slows down
+    // as snapshot history grows. See PR #470 / #445.
+    { column: HOLDING_ACCOUNT_ID },
   ],
   ModelClass: SnapshotModel,
 });
--- a/src/server/lib/postgres/repositories/snapshots.test.ts
+++ b/src/server/lib/postgres/repositories/snapshots.test.ts
@@ -412,4 +412,26 @@ describe("searchSnapshots — #445 holding_account_id regression", () => {
     expect(holdingValues).toContain("2026-01-01");
     expect(holdingValues).toContain("2026-06-30");
   });
+
+  test("`security_id` filter targets `holding_security_id`, not `security_id`, in the holding branch", async () => {
+    // Holding rows store the security in `holding_security_id`; the
+    // `security_id` column is NULL for them. Filtering the holding query
+    // by `security_id =` would always match zero rows.
+    mockQuery
+      .mockImplementationOnce(async () => ({ rows: [], rowCount: 0 }))
+      .mockImplementationOnce(async () => ({ rows: [], rowCount: 0 }));
+
+    await searchSnapshots(testUser, { account_id: "acc-x", security_id: "sec-voo" });
+
+    expect(mockQuery).toHaveBeenCalledTimes(2);
+    const holdingSql = mockQuery.mock.calls[1][0] as string;
+    const holdingValues = mockQuery.mock.calls[1][1] as unknown[];
+    expect(holdingSql).toContain("holding_security_id = ");
+    // Defensive: make sure the holding query does NOT filter on the
+    // wrong column (the bug this test guards against).
+    const segments = holdingSql.split(" AND ");
+    const securitySegment = segments.find((s) => s.includes("security_id ="));
+    expect(securitySegment).toContain("holding_security_id");
+    expect(holdingValues).toContain("sec-voo");
+  });
 });
--- a/src/server/lib/postgres/repositories/snapshots.test.ts
+++ b/src/server/lib/postgres/repositories/snapshots.test.ts
@@ -115,16 +115,22 @@ describe("searchSnapshots", () => {
     expect(mockQuery.mock.calls[1][1]).toEqual(["security", "2026-01-01"]);
   });
 
-  test("skips the security query when narrowing to a specific account_id", async () => {
-    mockQuery.mockImplementationOnce(async () => ({
-      rows: [makeAccountRow({ account_id: "acc-7" })],
-      rowCount: 1,
-    }));
+  test("skips the security query when narrowing to a specific account_id (still runs the holding_account_id query — see #445)", async () => {
+    mockQuery
+      .mockImplementationOnce(async () => ({
+        rows: [makeAccountRow({ account_id: "acc-7" })],
+        rowCount: 1,
+      }))
+      .mockImplementationOnce(async () => ({ rows: [], rowCount: 0 }));
 
     await searchSnapshots(testUser, { account_id: "acc-7" });
 
-    expect(mockQuery).toHaveBeenCalledTimes(1);
+    // 1 = account_id-scoped, 2 = holding_account_id-scoped (#445 fix).
+    // No 3rd call for global security.
+    expect(mockQuery).toHaveBeenCalledTimes(2);
     expect(mockQuery.mock.calls[0][1]).toContain("acc-7");
+    const secondSql = mockQuery.mock.calls[1][0] as string;
+    expect(secondSql).toContain("holding_account_id = ");
   });
 
   test("skips the security query when caller asks for snapshot_type='holding'", async () => {
@@ -138,12 +144,16 @@ describe("searchSnapshots", () => {
     expect(mockQuery.mock.calls[0][1]).toContain("holding");
   });
 
-  test("skips the security query when caller passes a non-empty account_ids list", async () => {
-    mockQuery.mockImplementationOnce(async () => ({ rows: [], rowCount: 0 }));
+  test("skips the security query when caller passes a non-empty account_ids list (still runs the holding_account_id query — see #445)", async () => {
+    mockQuery
+      .mockImplementationOnce(async () => ({ rows: [], rowCount: 0 }))
+      .mockImplementationOnce(async () => ({ rows: [], rowCount: 0 }));
 
     await searchSnapshots(testUser, { account_ids: ["acc-1", "acc-2"] });
 
-    expect(mockQuery).toHaveBeenCalledTimes(1);
+    expect(mockQuery).toHaveBeenCalledTimes(2);
+    const secondSql = mockQuery.mock.calls[1][0] as string;
+    expect(secondSql).toContain("holding_account_id IN");
   });
 
   test("runs the security query when caller explicitly asks for snapshot_type='security'", async () => {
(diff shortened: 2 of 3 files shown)


claoie:
## Review (budget-reviewer agent, recall-mode)

3 findings — 1 real correctness bug (med), 1 perf gap (low), 1 test gap (low). All addressed in follow-up commit `146985b`.

### Finding 1 — wrong column in the new holding-by-account branch (MED, fixed)

`src/server/lib/postgres/repositories/snapshots.ts:109` (pre-fix) — the new branch filtered:

```ts
filters: {
  [SNAPSHOT_TYPE]: "holding",
  [HOLDING_ACCOUNT_ID]: options.account_id,
  [SECURITY_ID]: options.security_id,   // ← bug: should be HOLDING_SECURITY_ID
},
```

Holding rows leave `security_id` NULL and store the security in `holding_security_id` (see `snapshot.ts:49-50` schema and `snapshots.ts:301-302` upsert). Any caller passing `{ account_id, security_id }` together would get zero holdings back — silent empty response. Latent today (sync.ts only sets `account-id` and not `security-id`) but a real correctness gap for any future caller.

Sibling `getHoldingSnapshots` (`snapshots.ts:208`) already uses `HOLDING_SECURITY_ID` correctly — same column-disambiguation reason. The new branch should match.

**Fix:** swapped `[SECURITY_ID]` → `[HOLDING_SECURITY_ID]` in the holding-branch filter block.

### Finding 2 — no index on `holding_account_id` (LOW, fixed)

`src/server/lib/postgres/models/snapshot.ts:203-209` (pre-fix) declared indexes on `user_id`, `snapshot_type`, `snapshot_date`, `account_id`, `security_id` — no `holding_account_id`. The new query (and the pre-existing `getHoldingSnapshots`) falls back to a `user_id`-index scan + row filter. `sync.ts:343-365` fans out per-account × per-month, so the per-fetch volume now ~doubles — meaningful as snapshot history grows.

Pre-existing gap on the same shape (so `getHoldingSnapshots` also benefits), but PR #470 specifically amplifies it.

**Fix:** added `{ column: HOLDING_ACCOUNT_ID }` to the indexes block. Auto-migration creates it on next deploy.

### Finding 3 — test gap that would have caught Finding 1 (LOW, fixed)

The new `#445 holding_account_id regression` describe block had 7 cases but no case exercised `{ account_id, security_id }` together — the exact combination that would have surfaced Finding 1.

**Fix:** added `"security_id filter targets holding_security_id, not security_id, in the holding branch"` test (defensive — asserts that the segment of the WHERE clause containing `security_id` is the `holding_` qualified one). Catches the bug if it ever regresses.

### Out of scope but worth flagging (no action this PR)

- `getHoldingSnapshots` will benefit from the new `holding_account_id` index without code changes — pre-existing slow path is now indexed.
- Scope of PR #470 stays narrow: only `snapshots.ts`, `snapshots.test.ts`, and now `snapshot.ts` (one-line index addition). No drive-by changes.

### Status after follow-up

- 708 tests pass / 0 fail (701 prior + 7 from initial #470 + 1 new from review = 708 total under snapshots.test.ts: 19)
- `bun run typecheck` clean, `bun run lint` clean
- Sandbox 20470 still up with the corrected branch

claoie:
## Second-pass review (budget-reviewer agent, recall-mode)

### Part A — prior findings verified resolved

- **Finding 1 (HOLDING_SECURITY_ID):** RESOLVED at `snapshots.ts:114` (`[HOLDING_SECURITY_ID]: options.security_id` inside the holding-branch filter block; import at line 17 brings the constant in from `../models`).
- **Finding 2 (index):** RESOLVED at `snapshot.ts:214` (`{ column: HOLDING_ACCOUNT_ID }` added to the `indexes` array; auto-migration in `initialize.ts:74-77` creates `idx_snapshots_holding_account_id` on next deploy).
- **Finding 3 (test gap):** RESOLVED at `snapshots.test.ts:416-436`. Confirmed by inspection: the assertion `expect(holdingSql).toContain("holding_security_id = ")` would have failed against the original `[SECURITY_ID]` code (its SQL contained only `security_id = `).

### Part B — fresh findings on the current diff

Two low-severity advisory items. Neither blocks merge.

```json
[
  {
    "file": "src/server/lib/postgres/models/snapshot.ts",
    "line": 214,
    "severity": "low",
    "summary": "New holding_account_id index is created without CONCURRENTLY. Pre-existing repo-wide pattern (every index goes through buildCreateIndex which emits CREATE INDEX IF NOT EXISTS without CONCURRENTLY), so this is a callout for awareness rather than a fix-this-PR ask."
  },
  {
    "file": "src/server/lib/postgres/repositories/snapshots.ts",
    "line": 100,
    "severity": "low",
    "summary": "The holding-branch query fires for every account passed via account_id / account_ids[], including non-investment accounts (checking, credit). Per-sync fan-out roughly doubles DB hits — but the new index keeps each query cheap. Could optionally gate on account type, but that requires either an extra round-trip or threading account-type through the API surface; not worth it for this PR."
  }
]
```

### Structural concerns checked and dismissed

- **Result-ordering / dedup of concat'd arrays:** `userResult` and `holdingByAccountRows` cannot overlap (account-balance/security rows have `account_id`/`security_id` set with `holding_account_id IS NULL`; holding rows have the inverse). Client `sync.ts` ingests via `Dictionary.set(snapshot_id, ...)` so duplicate-id collisions would dedup anyway.
- **Test would accidentally pass when code is wrong:** assertion directly catches the prior bug.
- **Other `searchSnapshots`-like paths:** `getHoldingSnapshots` already correct; `getBalanceDataFromHoldingSnapshots` is client-side; upsert path writes the right columns.
- **Holding-row enrichment:** `toHoldingSnapshot()` returns `holding.security_id` without ticker enrichment, but that's pre-existing — the client resolves tickers from `data.securities` separately via the security-snapshot branch.

### Verdict

PR is good to merge. Both new findings are advisory and apply to pre-existing patterns the PR didn't introduce.

hoiekim (approved):

PR #472, opened 2026-06-04
feat(holdings): aggregate by ticker in HoldingsComposition + per-snapshot edit sections in detail page

claoie:
## Summary

Per Hoie 2026-06-04 Discord direction: replace the per-(account, security_id) rows in HoldingsComposition with one row per ticker symbol, and redesign the holding detail page to show total / avg in the Holding section + a two-row section per underlying snapshot below.

Spec confirmed on Discord (chat 1511828516197503137):
- Real ticker wins as the bucket key for non-cash.
- Cash rows always bucket to `__CASH__` pseudo, regardless of ticker.
- A non-cash row whose ticker happens to be `__CASH__` (very unlikely) falls back to truncated security_id to avoid colliding with the cash bucket.
- URL: `/holding-detail?account_id=X&ticker=Y`. `?snapshot_id=...` retired.
- Detail page: total quantity + cost basis avg in Holding section, then N per-snapshot two-row sections below.

## Changes

- **`HoldingsComposition/index.tsx`** — new two-pass model: (1) per-security rows from the calculation hook (unchanged), (2) aggregate-by-ticker bucketing. Each bucket sums quantity / value / cost basis. URL navigates with `ticker=` instead of `snapshot_id=`.
- **`HoldingProperties/index.tsx`** — reads `?ticker=` instead of `?snapshot_id=`. Resolves all currently-active snapshots for (account, ticker) at viewDate. Holding section shows ticker, name, total quantity, cost basis avg. Below: N per-snapshot two-row sections rendering quantity, cost basis, snapshot date. Each editable per the existing viewDate / manual-account gating (\`isReadOnly = !isManualAccount && isCurrentViewDate\`). Per-section edit state lives in a \`Record<snapshotId, ...>\` so edits stay isolated. The 'Add new holding' flow is unchanged.
- **`HoldingProperties/index.css`** — adds `.snapshotSection` styling.

## Cost basis avg

\`avg = sum(cost_basis) / sum(quantity)\`. Null when any contributing snapshot lacks cost basis OR total quantity is zero. This is the per-share average, which matches how real brokers report it (\$ paid per share across all lots).

## Verification

- \`bun test src\` → **708 pass / 0 fail**.
- \`bun run typecheck\` clean.
- \`bun run lint\` clean.

## Sandbox E2E

Will spin a sandbox right after PR creation. Playwright walk asserts:
- JP Morgan account renders ONE <ticker> row + ONE __CASH__-bucketed Cash row (not separate per-security rows).
[...]

PR #473, opened 2026-06-04
refactor(client): add <Page> shell component, migrate 4 detail pages (#326)

### LINKED ISSUE #326: Design component refactoring
Currently our styling CSS code is hardcoded for each classes. This is great for simplicity and atomicity but not great for consistency when introducing a new component. A contributor is confused and doesn't know what CSS code to apply as they don't know the existing design convention. It was shown well in this frustrated comment: https://github.com/hoiekim/budget/pull/196/changes#r3211910617

Create componentized design elements. Such as:
```
// src/client/components/common/Page/index.tsx

import "./index.css"

export const Page = ({ children, className, ...rest }) => {
  const mergedClassName = "Page" + " " + className;
  return (
    <div className={mergedClassName}>
      {children}
    </div>
  )
}
```

```
// src/client/components/MyPage/index.tsx

export const MyPage = (...) => {
  ...
  return (
    <Page className="MyPage">
      ...
    </Page>
  )
}
```
There must be so many other common components we can standardize.

claoie:
Starts **#326 (Design component refactoring)** — the first shared design-shell. Hoie's ask: contributors don't know what CSS convention to apply to a new page because the conventions are hand-copied per page, not componentized.

## The convention being componentized
8 page `index.css` files each declare the **identical** rule:
```css
div.XxxPage > div { padding: 0 10px; }
```
This PR extracts it into a `<Page>` shell. `<Page className="XxxPage">` renders `<div class="Page XxxPage">`, so the rule lives **once** in `Page/index.css` (`div.Page > div`), and the page's own className is preserved so any page-scoped CSS keeps working unchanged.

```tsx
// src/client/components/Page/index.tsx
export const Page = ({ className, children, ...rest }) => {
  const mergedClassName = className ? `Page ${className}` : "Page";
  return <div className={mergedClassName} {...rest}>{children}</div>;
};
```

## Phase 1 scope (intentionally small)
Migrates the 4 pages whose CSS was **only** that padding rule, so each per-page `index.css` deletes cleanly:
- `HoldingDetailPage`, `ApiKeyDetailPage`, `ConfigPage`
- `ConnectionDetailPage` — **also fixes a latent bug**: it rendered `className="ConfigPage"` (copy-pasted from ConfigPage), so its own `ConnectionDetailPage/index.css` never matched — dead CSS. It now renders its own class and inherits padding from the shell. This is precisely the "contributor doesn't know the convention → copies the wrong thing" failure #326 calls out.

Net: **+13 / −31** (consolidation).

## Follow-up phases
Remaining `div.XxxPage > div { padding }` pages (Account / Chart / BudgetConfig / Transaction detail) and the list-page sticky `<h2>` convention (Accounts/Budgets/Dashboard share an identical sticky header block) — migrated once this shell lands.

## Note on placement
#326's sketch placed this at `components/common/Page`. This follows the repo's existing **flat** `components/<Name>` convention (all 40 components are flat). Trivial to relocate under a `common/` grouping if you'd prefer that for shared shells — say the word.

## Verification
- `bun run typecheck` ✓ · `bun run lint` ✓ · `bun run build-client` ✓
- **Playwright E2E** (dev server on local pg, logged in as admin): navigated to `/config` and `/api-key-detail`, asserted via `getComputedStyle`:
  - root element classList = `["Page","ConfigPage"]` / `["Page","ApiKeyDetailPage"]` ✓
  - first child `<div>` computed `padding-left` / `padding-right` = `10px` / `10px` ✓ — padding now sourced from the consolidated `div.Page > div` rule, behavior-preserving.
  - (Connection/Holding detail pages need a data param to render content but use the identical shell verified above.)

Closes #326 partially (Phase 1 — shell + 4 migrations).

### DIFF
--- a/src/client/components/Page/index.css
+++ b/src/client/components/Page/index.css
@@ -1,0 +1,3 @@
+div.Page > div {
+  padding: 0 10px;
+}
--- a/src/client/components/Page/index.tsx
+++ b/src/client/components/Page/index.tsx
@@ -1,0 +1,25 @@
+import { ComponentPropsWithoutRef } from "react";
+
+import "./index.css";
+
+interface Props extends ComponentPropsWithoutRef<"div"> {}
+
+/**
+ * Standard page shell. Owns the design convention every page-level container
+ * shares — currently the `padding: 0 10px` applied to its direct-child
+ * sections (see index.css). Use it as the outermost element of a page so the
+ * convention is inherited automatically instead of re-declared per page:
+ *
+ *   <Page className="MyPage">…</Page>
+ *
+ * The page-specific className is preserved alongside "Page", so existing
+ * page-scoped CSS (`div.MyPage .foo`) keeps working unchanged.
+ */
+export const Page = ({ className, children, ...rest }: Props) => {
+  const mergedClassName = className ? `Page ${className}` : "Page";
+  return (
+    <div className={mergedClassName} {...rest}>
+      {children}
+    </div>
+  );
+};
--- a/src/client/components/index.ts
+++ b/src/client/components/index.ts
@@ -1,4 +1,5 @@
 export * from "./App";
+export * from "./Page";
 export * from "./HoldingProperties";
 export * from "./ErrorBoundary";
 export * from "./PlaidLinkContext";
--- a/src/client/pages/ApiKeyDetailPage/index.css
+++ b/src/client/pages/ApiKeyDetailPage/index.css
@@ -1,3 +1,0 @@
-div.ApiKeyDetailPage > div {
-  padding: 0 10px;
-}
--- a/src/client/pages/ApiKeyDetailPage/index.tsx
+++ b/src/client/pages/ApiKeyDetailPage/index.tsx
@@ -1,11 +1,9 @@
-import { ApiKeyProperties } from "client/components";
-
-import "./index.css";
+import { ApiKeyProperties, Page } from "client/components";
 
 export const ApiKeyDetailPage = () => {
   return (
-    <div className="ApiKeyDetailPage">
+    <Page className="ApiKeyDetailPage">
       <ApiKeyProperties />
-    </div>
+    </Page>
   );
 };
--- a/src/client/pages/ConfigPage/index.css
+++ b/src/client/pages/ConfigPage/index.css
@@ -1,3 +1,0 @@
-div.ConfigPage > div {
-  padding: 0 10px;
-}
--- a/src/client/pages/ConfigPage/index.tsx
+++ b/src/client/pages/ConfigPage/index.tsx
@@ -1,13 +1,11 @@
-import { Configuration, PlaidLinkProvider } from "client/components";
-
-import "./index.css";
+import { Configuration, Page, PlaidLinkProvider } from "client/components";
 
 export const ConfigPage = () => {
   return (
     <PlaidLinkProvider>
-      <div className="ConfigPage">
+      <Page className="ConfigPage">
         <Configuration />
-      </div>
+      </Page>
     </PlaidLinkProvider>
   );
 };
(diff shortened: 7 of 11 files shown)


hoiekim:
Review it with e2e verification

claoie:
## Review (budget-reviewer, recall-mode)

Scope clean — every file maps to the title. ConnectionDetailPage's className fix is appropriate-scope because it's one of the 4 migrated pages and the wrong className was exactly the failure mode #326 calls out.

Parity verified for all 4 migrated pages: each deleted CSS contained *only* `div.XxxPage > div { padding: 0 10px }`, and `div.Page > div { padding: 0 10px }` provides the same rule. Grep across `src/client/**/*.css` returned no external references to `.ApiKeyDetailPage`, `.HoldingDetailPage`, `.ConfigPage`, or `.ConnectionDetailPage` — un-scoping is safe.

### Findings (all LOW, none block merge)

| File | Severity | Issue |
|------|----------|-------|
| `Page/index.tsx:5` | LOW | `interface Props extends ComponentPropsWithoutRef<"div"> {}` is an empty-extends-only interface (typescript-eslint's `no-empty-object-type` smell). Prefer `type Props = ComponentPropsWithoutRef<"div">`. Not blocking; just a clean-up note. |
| `Page/index.tsx:19` | LOW | Doesn't forward `ref` (`ComponentPropsWithoutRef` omits it; component is plain function not `forwardRef`). No current caller needs a ref. Worth a one-line comment that this is intentional, or upgrade to `forwardRef` preemptively before the next batch. |
| `Page/` | LOW | **No tests.** The codebase has zero client-component tests today, so this matches precedent — but Hoie's pattern is "robust edge-case unit testing especially for shared components." Specifically untested: (a) `className`-undefined branch returns plain `"Page"`, (b) `className`-string branch returns `"Page <x>"` with single space, (c) other props pass through to the div (data-*, id, style, role). |
| `src/client/components/index.ts:2` | LOW | Export-order insertion (Page between App and HoldingProperties). Cosmetic; matches a "new shared shells at the top" reading. No runtime effect. |
| `src/client/pages/BudgetConfigPage/index.css` | LOW (forward-looking) | Out-of-scope for Phase 1, but worth noting: BudgetConfigPage's CSS has the padding rule *and* `div.BudgetConfigPage > div.title { ... }` rules. When the follow-up batch migrates it, the non-padding rules must keep their `div.BudgetConfigPage > div.title` selector (NOT rewritten as `div.Page > div.title`), since `div.title` is a BudgetConfigPage-specific concept. Confirms the PR's design (preserve per-page className) is right. |

### Forward-thinking note for the Properties componentization PR (separate, opening next)

Two extra things matter for `<Properties>` that Page didn't have to deal with:

1. **Direct-child selector preservation is the entire point.** The existing rules use `div.Properties > .propertyLabel`, `div.Properties > .property`, `div.Properties .row`, etc. Any `<PropertyLabel>` / `<Property>` component MUST render exactly `<div class="propertyLabel">…</div>` and `<div class="property">…</div>` with **NO intermediate wrapper** — otherwise the `>` selectors break. This is exactly the bug Hoie flagged today on #472 (and previously on #196/#294). Hoie's whole motivation is to make this failure mode impossible by routing every use through typed components — so the components themselves must not introduce wrappers, and any "section" abstraction must use `React.Fragment` if it groups multiple direct children.
2. **Canonical Properties ordering is specific-class first: `<Foo>Properties Properties`** (per the `<div className="ConnectionProperties Properties">` etc. usages in tree). Page's PR chose `Page <Foo>` first — either ordering works for CSS specificity, but for `<Properties>` the natural mirror is to follow the existing order (`<Foo>Properties Properties` / specific-first) to avoid touching every existing call site to reverse it.

Both items will land in the follow-up Properties shell PR.

hoiekim (approved):

Issue #471, opened 2026-06-04
bug(holdings): deleteHoldings soft-deletes EVERY snapshot for the account when called from sync (latent data loss)


claoie:
## Severity

**Critical / latent data loss.** Hasn't visibly fired yet (live spot-check: JP Morgan / Chase Self-Directed has all holding snapshots `is_deleted=false`), but the next Plaid sync that detects a missing holding will wipe the entire account's holding-snapshot history.

## What's wrong

`src/server/lib/postgres/repositories/holdings.ts:102-118` (`deleteHoldings`):

```ts
export const deleteHoldings = async (
  user: MaskedUser,
  holding_ids: string[],
): Promise<{ deleted: number }> => {
  if (!holding_ids.length) return { deleted: 0 };
  const { user_id } = user;

  // Soft-delete related snapshots for these holdings
  for (const holding_id of holding_ids) {
    await snapshotsTable.bulkSoftDeleteByColumn(
      "holding_account_id",
      holding_id.split("-")[0],   // ← the account_id portion of the holding_id
      user_id,
    );
  }

  const deleted = await holdingsTable.bulkSoftDelete(holding_ids, { [USER_ID]: user_id });
  return { deleted };
};
```

`holding_id` format is `${account_id}-${security_id}` (`models/holding.ts:104`). `holding_id.split("-")[0]` is the **account_id**. `bulkSoftDeleteByColumn("holding_account_id", account_id, user_id)` resolves to:

```sql
UPDATE snapshots
SET is_deleted = TRUE, updated = CURRENT_TIMESTAMP
WHERE holding_account_id = $account_id
  AND user_id = $user_id
RETURNING snapshot_id;
```

That deletes **every holding snapshot ever taken on that account**, regardless of which security. So deleting ONE holding nukes EVERY holding's history — <ticker>, cash, every position — for that account.

## How it fires in the sync path

`src/server/lib/compute-tools/create-snapshots.ts:80-101`:

```ts
existingHoldings
  .filter((h) => accountIds.has(h.account_id) && !incomingMap.has(h.holding_id))
  .forEach((h) => {
    removedHoldings.push(h);
    snapshots.push({
      ...,
      snapshot: { snapshot_id: `${h.holding_id}-${getSquashedDateString()}`, date: new Date().toISOString() },
      holding: { ...h, quantity: 0, institution_value: 0 },
    });
  });

await upsertHoldings(user, incomingHoldings);
[...]

PR #475, opened 2026-06-04
fix(accounts): soft-delete orphaned holding snapshots on account delete (#474)

### LINKED ISSUE #474: bug(accounts): deleteAccounts orphans holding snapshots — soft-deletes account_id but not holding_account_id (sibling of #445 column divergence)
## Severity

**Latent data-integrity bug.** Sibling of #471 (the \`deleteHoldings\` account-wide snapshot wipe) but the OPPOSITE failure shape — this one leaves rows behind that should be soft-deleted.

## What's wrong

\`src/server/lib/postgres/repositories/accounts.ts:115-140\` (\`deleteAccounts\`):

\`\`\`ts
return withTransaction(async (client) => {
  for (const account_id of account_ids) {
    await snapshotsTable.bulkSoftDeleteByColumn(ACCOUNT_ID, account_id, user_id, client);  // ← only catches account_balance snapshots
  }
  await accountsTable.bulkSoftDelete(account_ids, { [USER_ID]: user_id }, client);
});
\`\`\`

The snapshots-table soft-delete filters on \`account_id\`. Holding snapshots store the account in **\`holding_account_id\`**, not \`account_id\` (their \`account_id\` column is NULL). Same column-divergence shape that broke per-account \`/api/snapshots\` in #445.

So deleting an account leaves the holding snapshots behind. They satisfy:
- \`account_id IS NULL\` (so the cleanup query doesn't match them)
- \`holding_account_id = <deleted-account-id>\`
- \`is_deleted = false\` (still active)

The holdings-table row is correctly soft-deleted, but the snapshot history is orphaned.

## Why it matters

After PR #470 (\`searchSnapshots\` matches via \`holding_account_id\`), the orphan snapshots are discoverable. If a user deletes account X then re-creates a manual account that happens to receive the same \`account_id\` (e.g. via the 5-char IDs the manual provider uses today — \`50931\`, \`6a999\`, etc., which are short hex strings with non-trivial collision odds), the re-created account immediately inherits the deleted account's holdings history through the orphan snapshots.

Even without re-creation, the orphans pollute:
- Any future \`/api/snapshots?account-id=<deleted-id>\` request returns the orphans.
- Account-aggregation queries that bypass \`accounts.is_deleted\` see ghost positions.

## Repro

1. Manual investment account with at least one holding snapshot.
2. Soft-delete the account via \`DELETE /api/account?id=X\`.
3. Query: \`SELECT * FROM snapshots WHERE holding_account_id = 'X' AND is_deleted = false;\` — returns the holding snapshots that should have been cleaned up.

## Fix sketch

Extend the soft-delete to BOTH columns:

\`\`\`ts
await snapshotsTable.bulkSoftDeleteByColumn(HOLDING_ACCOUNT_ID, account_id, user_id, client);
\`\`\`

Or (cleaner) extend \`bulkSoftDeleteByColumn\` to accept multiple OR'd predicates and do it in one statement. \`HOLDING_ACCOUNT_ID\` already exists as a constant (\`src/server/lib/postgres/models/common.ts:125\`).

Same regression test as #470: open a sandbox, delete a manual investment account that has holding snapshots, query the snapshots table directly, assert no orphans remain.

## Related

- **#471** (the \`deleteHoldings\` account-wide wipe — the SAME column-divergence shape but the OPPOSITE failure: that bug deletes too much; this one deletes too little).
- **#445 / PR #470** (the original column-divergence finding in \`searchSnapshots\` read path).

---
*Found during 2026-06-04 investigation of [PR #412 issuecomment-4609220855](https://github.com/hoiekim/budget/pull/412#issuecomment-4609220855) — "When we delete an account, what do we do with the associated transactions?" The cascade IS correct for transactions but not for holding snapshots.*

claoie:
Closes #474

## Problem
`deleteAccounts` soft-deletes related rows per account, but the snapshots pass only filters on `account_id`:

```ts
```

Holding snapshots store the account in **`holding_account_id`** (their `account_id` column is NULL — same column divergence that broke per-account `/api/snapshots` in #445). So deleting an account leaves every holding snapshot behind:
- `account_id IS NULL` → the cleanup query never matches them
- `holding_account_id = <deleted-account-id>`, `is_deleted = false` → still active

Those orphans are discoverable via `searchSnapshots` after #470, and can leak the deleted account's holdings history into a newly created manual account that collides on the short (5-char hex) manual-provider `account_id`.

## Fix
Add a second `bulkSoftDeleteByColumn` pass keyed on `holding_account_id`, so the cascade covers both snapshot shapes (account-balance + holding):

```ts
await snapshotsTable.bulkSoftDeleteByColumn(HOLDING_ACCOUNT_ID, account_id, user_id, client);
```

`HOLDING_ACCOUNT_ID` already exists as a constant and is imported the same way by `snapshots.ts`. Sibling of #471 (the *opposite* failure shape — over-deletion in `deleteHoldings`); that one carries an open design question and is intentionally not touched here.

## Testing
- New regression tests in `accounts.test.ts` pin the SQL contract: deleting an account issues a snapshots soft-delete keyed on `account_id` **and** one keyed on `holding_account_id`; empty input short-circuits with no query.
- `bun test accounts.test.ts` → 20 pass / 0 fail.
- `tsc --noEmit` clean.

### DIFF
--- a/src/server/lib/postgres/repositories/accounts.test.ts
+++ b/src/server/lib/postgres/repositories/accounts.test.ts
@@ -19,8 +19,14 @@ mock.module("pg", () => ({
   default: { Pool: FakePool, types: { setTypeParser: () => {} } },
 }));
 
-const { getAccounts, getAccount, searchAccounts, searchAccountsById, upsertAccounts } =
-  await import("./accounts");
+const {
+  getAccounts,
+  getAccount,
+  searchAccounts,
+  searchAccountsById,
+  upsertAccounts,
+  deleteAccounts,
+} = await import("./accounts");
 
 afterAll(restoreLeaves);
 
@@ -237,5 +243,31 @@ describe("upsertAccounts", () => {
     const result = await upsertAccounts(testUser, accounts);
     expect(result).toHaveLength(3);
     expect(result.every((r: { status: number }) => r.status === 200)).toBe(true);
+  });
+});
+
+describe("deleteAccounts", () => {
+  test("returns deleted: 0 without querying for empty input", async () => {
+    const result = await deleteAccounts(testUser, []);
+    expect(result).toEqual({ deleted: 0 });
+    expect(mockQuery).not.toHaveBeenCalled();
+  });
+
+  test("soft-deletes snapshots by BOTH account_id and holding_account_id (#474)", async () => {
+    mockQuery.mockResolvedValue({ rows: [], rowCount: 0 });
+    await deleteAccounts(testUser, ["acc-del"]);
+
+    // Account-balance snapshots live under `account_id`; holding snapshots
+    // live under `holding_account_id` (their `account_id` is NULL). Both
+    // passes must fire or the holding-snapshot history is orphaned.
+    const snapshotDeletes = mockQuery.mock.calls
+      .map(([sql]) => sql)
+      .filter(
+        (sql): sql is string =>
+          typeof sql === "string" && /UPDATE\s+snapshots\b/i.test(sql) && /is_deleted/i.test(sql),
+      );
+
+    expect(snapshotDeletes.some((sql) => /WHERE\s+account_id\s*=/i.test(sql))).toBe(true);
+    expect(snapshotDeletes.some((sql) => /WHERE\s+holding_account_id\s*=/i.test(sql))).toBe(true);
   });
 });
--- a/src/server/lib/postgres/repositories/accounts.ts
+++ b/src/server/lib/postgres/repositories/accounts.ts
@@ -9,6 +9,7 @@ import {
   snapshotsTable,
   holdingsTable,
   ACCOUNT_ID,
+  HOLDING_ACCOUNT_ID,
   USER_ID,
   ITEM_ID,
   INSTITUTION_ID,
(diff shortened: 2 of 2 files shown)


claoie:
## Self-Review

**Thread status:** No prior review. 0 behind `main`, CI green. Scope clean — `accounts.ts` + its test.

**Logic — correct.** `deleteAccounts` previously soft-deleted snapshots only by `account_id`, which catches account-balance snapshots but **not holding snapshots** — those store the account in `holding_account_id` (their `account_id` is NULL). So deleting an account left orphaned holding-snapshot history (#474). The fix adds a second `bulkSoftDeleteByColumn(HOLDING_ACCOUNT_ID, account_id, user_id, client)` in the same transaction.

**Relationship to #481:** complementary, not overlapping. #481 scopes the *single-holding* delete to `(holding_account_id, holding_security_id)`; this PR deletes *all* holding snapshots for an account being removed. Different operations, same table, no conflict.

**Tests — 20 pass / 0 fail** (`bun test accounts.test.ts`), incl. `soft-deletes snapshots by BOTH account_id and holding_account_id (#474)` which pins both `WHERE account_id =` and `WHERE holding_account_id =` UPDATE-snapshots clauses.

**E2E:** DB-repository change; the test asserts the exact soft-delete SQL on both columns within the delete transaction — that's the fix's assertion. Full UI E2E (delete an investment account → confirm no orphaned snapshots) is blocked locally (no budget `.env`/login). Verified at the model/SQL level.

hoiekim (approved):

PR #477, opened 2026-06-04
fix(holdings): use magnitude of sell quantity in inferCostBasis (#459)

### LINKED ISSUE #459: bug(holdings): inferCostBasis treats SELL quantity as positive — Plaid sells are negative, so a sell inflates cost basis instead of reducing it
## Summary
`inferCostBasis` (`src/client/lib/hooks/calculation/holdings.ts:186-242`) implements average-cost cost-basis inference, but its SELL branch assumes `quantity` is **positive** for sells. Plaid (and our `investment_transactions` table) encode sells with **negative** `quantity`. As a result a sell **adds** shares and cost basis instead of removing them — doubling the position in the wrong direction.

This feeds the **Unrealized G/L** column in the Holdings Composition table (`src/client/components/HoldingsComposition/index.tsx:296-299`), so any non-cash holding whose broker cost basis is missing/zero and that has sell history will show a wrong (inflated) cost basis and therefore wrong unrealized gain/loss.

## Root cause
holdings.ts SELL branch (≈ lines 218-229):

```ts
} else if (type === InvestmentTransactionType.Sell) {
  if (totalQuantity > 0) {
    const avgCost = totalCost / totalQuantity;
    const soldCost = avgCost * quantity;   // quantity is NEGATIVE for a sell
    totalCost -= soldCost;                 //   -= (negative)  → totalCost INCREASES
    totalQuantity -= quantity;             //   -= (negative)  → totalQuantity INCREASES
    if (totalQuantity < 0) totalQuantity = 0;
    if (totalCost < 0) totalCost = 0;
  }
}
```

The BUY branch (`totalQuantity += quantity`) is correct because buy quantities are positive. The SELL branch needs the magnitude of the sold quantity.

## Confirmation that sells are negative
Live dev DB (prod snapshot):

```
 type | count |      min      | max
------+-------+---------------+-----
 buy  |    97 |      0.010000 | 30371   (all positive)
 sell |    38 | -26109.010000 | 0       (all negative)
```

The codebase already knows this elsewhere: `benchmark.ts` uses `-Math.abs(t.quantity)` for sells (`benchmark.ts:250, 268`) and `-Math.abs(t.amount)` (`benchmark.ts:101`). Only `inferCostBasis` got the sign wrong.

## Reproduction (verified by running the real function)
Ran the actual exported `inferCostBasis` with a realistic Plaid-shaped sell:

- Buy 10 @ $100  → expected cost basis $[redacted], qty 10
- Sell 4 shares  → Plaid encodes `quantity = -4`
- **Expected after selling 4 of 10:** qty = 6, costBasis = $600
- **Actual:** `{"costBasis":1400,"totalQuantity":14,"inferred":true}`

The sell *added* 4 shares and $400 of basis.

## Why the existing test misses it
`holdings.test.ts:307-329` ("should handle sells using average cost method") passes a **positive** quantity (`4`) for the Sell transaction, which is not the shape real Plaid data takes. With a positive `4` the arithmetic happens to come out right, so the test passes while production data triggers the bug.

## Impact / scope
- Surfaces only for **non-cash** holdings where broker `cost_basis` is null/0 (so `inferCostBasis` runs — see holdings.ts:329) **and** that security has sell transactions in the same account.
- On the *current* snapshot the bug does not visibly manifest: the only holdings with sells are the cash sweeps (`institution_price=1`, no basis) which are excluded by the `isCash` gate, and the one non-cash null-basis holding (`fda9c`/`96d5…`) has no sells in its account. So this is a **latent correctness bug**, proven by direct function execution rather than by a UI screenshot.
- When it does hit (a normal scenario: partial sell of a position whose broker didn't report basis), the Unrealized G/L shown in Holdings Composition is wrong — basis is inflated, gain understated or flipped to a loss.

## Proposed fix
Use the magnitude of the sold quantity in the SELL branch:

```ts
} else if (type === InvestmentTransactionType.Sell) {
  if (totalQuantity > 0) {
    const soldQty = Math.abs(quantity);
    const avgCost = totalCost / totalQuantity;
    totalCost -= avgCost * soldQty;
    totalQuantity -= soldQty;
    if (totalQuantity < 0) totalQuantity = 0;
    if (totalCost < 0) totalCost = 0;
  }
}
```

And update `holdings.test.ts:307` to use `quantity = -4` so the test reflects real Plaid data (the existing positive-quantity case is what masked the bug). Add a negative-quantity sell case.

---
*Filed by the daily user-testing cron (issues only, no PR). Verified live on the dev prod-snapshot DB + by executing the exported `inferCostBasis` function directly.*

---

claoie:
Closes #459

## Problem
`inferCostBasis` (`src/client/lib/hooks/calculation/holdings.ts`) does average-cost inference, but its SELL branch assumed `quantity` is **positive**. Plaid (and our `investment_transactions` table) encode sells with **negative** `quantity`:

```
 type | count |      min      | max
------+-------+---------------+-----
 buy  |    97 |      0.010000 | 30371   (all positive)
 sell |    38 | -26109.010000 | 0       (all negative)
```

With the raw negative quantity, `totalCost -= avgCost * quantity` and `totalQuantity -= quantity` become `-= negative` → they **add** shares and basis instead of removing them. Running the exported function directly:

- Buy 10 @ $100, then sell 4 (Plaid `quantity = -4`)
- Expected: qty 6, basis $600
- **Pre-fix actual: qty 14, basis $[redacted]** — the sell *added* 4 shares and $400.

This feeds the **Unrealized G/L** column in Holdings Composition for any non-cash holding whose broker `cost_basis` is null/0 and that has sell history, inflating the basis and understating/flipping the gain. (Latent on the current snapshot — the only holdings with sells are cash sweeps excluded by the `isCash` gate — but it fires on a normal partial sell of a position the broker didn't report basis for.)

## Fix
Take `Math.abs(quantity)` in the SELL branch — the same convention `benchmark.ts` already uses (`-Math.abs(t.quantity)` for sells). Only `inferCostBasis` had the sign wrong.

## Testing
- Updated the existing "should handle sells using average cost method" test to use the real negative Plaid shape (`-4`); its positive `4` was exactly what masked the bug.
- Added a regression test pinning qty 6 / basis $600 for a `-4` sell (pre-fix this asserted-against value was 14 / 1400).
- `bun test holdings.test.ts` → **34 pass / 0 fail**; `tsc --noEmit` clean; `eslint` clean.

---

### DIFF
--- a/src/client/lib/hooks/calculation/holdings.test.ts
+++ b/src/client/lib/hooks/calculation/holdings.test.ts
@@ -326,10 +326,11 @@ describe("inferCostBasis", () => {
       "tx1",
       createInvestmentTransaction("acc1", "sec1", InvestmentTransactionType.Buy, 100, 10, "2026-01-10")
     );
-    // Sell 4 @ avg cost $100 = $400 removed
+    // Sell 4 @ avg cost $100 = $400 removed.
+    // Plaid encodes sell quantities as NEGATIVE — use the real shape (-4).
     transactions.set(
       "tx2",
-      createInvestmentTransaction("acc1", "sec1", InvestmentTransactionType.Sell, 120, 4, "2026-01-15")
+      createInvestmentTransaction("acc1", "sec1", InvestmentTransactionType.Sell, 120, -4, "2026-01-15")
     );
 
     const result = inferCostBasis({
@@ -341,6 +342,32 @@ describe("inferCostBasis", () => {
 
     expect(result!.totalQuantity).toBe(6);
     expect(result!.costBasis).toBe(600); // $1000 - $400 = $600
+  });
+
+  test("should reduce (not inflate) basis when a sell uses Plaid's negative quantity", () => {
+    // Regression for #459: a negative sell quantity must remove shares + basis,
+    // not add them. Pre-fix this returned { totalQuantity: 14, costBasis: 1400 }.
+    const transactions = new InvestmentTransactionDictionary();
+    // Buy 10 @ $100 = $1000
+    transactions.set(
+      "tx1",
+      createInvestmentTransaction("acc1", "sec1", InvestmentTransactionType.Buy, 100, 10, "2026-01-10")
+    );
+    // Sell 4 shares — Plaid encodes quantity = -4
+    transactions.set(
+      "tx2",
+      createInvestmentTransaction("acc1", "sec1", InvestmentTransactionType.Sell, 130, -4, "2026-01-15")
+    );
+
+    const result = inferCostBasis({
+      accountId: "acc1",
+      securityId: "sec1",
+      investmentTransactions: transactions,
+      asOfDate: new Date("2026-01-20"),
+    });
+
+    expect(result!.totalQuantity).toBe(6);
+    expect(result!.costBasis).toBe(600);
   });
 
   test("should filter by date", () => {
--- a/src/client/lib/hooks/calculation/holdings.ts
+++ b/src/client/lib/hooks/calculation/holdings.ts
@@ -217,13 +217,16 @@ export const inferCostBasis = ({
       totalCost += cost;
       totalQuantity += quantity;
     } else if (type === InvestmentTransactionType.Sell) {
-      // SELL: reduce quantity using average cost
-      // Average cost per share = totalCost / totalQuantity
+      // SELL: reduce quantity using average cost.
+      // Plaid encodes sell quantities as NEGATIVE, so take the magnitude —
+      // otherwise `-= quantity` would ADD shares and basis back instead of
+      // removing them (cf. benchmark.ts, which already uses -Math.abs for sells).
       if (totalQuantity > 0) {
+        const soldQuantity = Math.abs(quantity);
         const avgCost = totalCost / totalQuantity;
-        const soldCost = avgCost * quantity;
+        const soldCost = avgCost * soldQuantity;
         totalCost -= soldCost;
-        totalQuantity -= quantity;
+        totalQuantity -= soldQuantity;
 
         // Prevent negative from rounding errors
         if (totalQuantity < 0) totalQuantity = 0;


claoie:
## Self-Review

**Thread status:** No prior review. Rebased onto `upstream/main` this run (0 behind) and force-pushed. CI green. Scope clean — `calculation/holdings.ts` + its test.

**Logic — correct.** Plaid encodes **sell** quantities as **negative**. The old SELL branch did `soldCost = avgCost * quantity` and `totalQuantity -= quantity` with `quantity < 0`, so both `-=` operations *added* cost and shares back — sells **inflated** the cost basis instead of reducing it (#459). The fix takes `const soldQuantity = Math.abs(quantity)` and subtracts the magnitude. Consistent with `benchmark.ts`, which already uses `-Math.abs(...)` for sells. The rounding-error clamp (`if (totalQuantity < 0) totalQuantity = 0`) is preserved.

**Types:** CI typecheck green (verified `tsc --noEmit` clean on sibling budget branches this run).

**Tests — 34 pass / 0 fail** (`bun test holdings.test.ts`), incl. the regression `should reduce (not inflate) basis when a sell uses Plaid's negative quantity` plus `should handle sells using average cost method`.

**E2E:** `inferCostBasis` is a pure calculation function; the regression test drives it with Plaid's negative-quantity sell and asserts the basis decreases — that's the bug's assertion. No UI path reaches it with different inputs.

reviewoie:
## Code review — PR 477

**Verdict:** approve

**Findings:** 0 blocking. Verified the fix, scope, callers, and tests.

The fix is correct and tightly scoped:

- **SELL-branch sign fix is right and complete.** Plaid encodes sells with negative `quantity`, so the prior `totalQuantity -= quantity` / `totalCost -= avgCost*quantity` *added* shares + basis. `Math.abs(quantity)` reverses that. The SELL branch reprices via `avgCost` and never reads the sell's own `price`/`amount`, so `quantity`'s sign was the only thing wrong — nothing else in the branch needs changing.
- **Matches the established convention.** `benchmark.ts` already normalizes with `Math.abs(t.quantity)` on both legs (lines 260, 278). This brings `inferCostBasis` in line.
- **BUY branch correctly left untouched.** Buy quantities are all positive in prod (`min 0.01`); touching the buy leg would be out-of-scope on a sell-sign bug.
- **Callers checked.** The sole non-test caller (`holdings.ts:331`) consumes only `inferred.costBasis` (a magnitude) — no caller depended on signed behavior.
- **Scope clean.** Both files (`holdings.ts`, `holdings.test.ts`) map directly to the title; no unrelated changes.
- **Tests verified, not assumed.** Ran the branch in a worktree: 36 pass / 0 fail. The existing "should handle sells using average cost method" test had its input corrected `4` → `-4` (the positive value was masking the bug), and a new regression test "should reduce (not inflate) basis when a sell uses Plaid's negative quantity" pins qty 6 / basis 600. Pure FE-calc test, no dependency mocks — `test-preload` alone is the right convention here.

### High
No further blockers.

### Med
No further blockers.

### Low
- `holdings.test.ts:1` — header comment `// Run with: bun test --preload ./test-preload.ts` has a stale path (actual is `./scripts/test-preload.ts`). Pre-existing, not introduced by this PR; noting only.

claoie:
## Reviewoie follow-up

One finding, low-severity informational note flagged as pre-existing — not in this PR's scope.

|---|---|---|---|---|
| 1 | low | `holdings.test.ts:1` | header comment `// Run with: bun test --preload ./test-preload.ts` has a stale path (actual is `./scripts/test-preload.ts`) | Leaving as-is — reviewoie explicitly noted this is pre-existing and not introduced by this PR. Fixing stale doc comments in unrelated test headers would be an unrelated cleanup; opening a separate sweep if it accumulates. |

No new commits — reviewoie's verdict was `approve`. Ping if you want this addressed before merge.

hoiekim (approved):

PR #485, opened 2026-06-05
fix(transactions): split rows update their own cache entry — remove dead early-return + construct from the split (#423, #435)

### LINKED ISSUE #423: bug(transactions): TransactionRow.onChangeBudgetSelect early-returns for splits, leaving in-memory cache + IndexedDB stale
## Summary

`TransactionRow.onChangeBudgetSelect` (`src/client/components/TransactionsTable/TransactionRow.tsx:111`) has a `return` statement after the `POST /api/split-transaction` call that **skips the entire success-branch** for split transactions — meaning when a user changes the budget on a SplitTransaction row from the transactions list, the in-memory `data.splitTransactions` dictionary **and IndexedDB are never updated client-side**. The server is updated, but the client cache stays stale until the next refresh.

This also skips the error-rollback path: if the server returns `status: "failed"`, the user's optimistic select change is **not** rolled back.

## Code pointer

`src/client/components/TransactionsTable/TransactionRow.tsx:98-147` (full handler):

```ts
const onChangeBudgetSelect: ChangeEventHandler<HTMLSelectElement> = async (e) => {
  const { value } = e.target;
  if (value === selectedBudgetIdLabel) return;

  setSelectedBudgetIdLabel(value);
  setSelectedCategoryIdLabel("");

  let response: ApiResponse;
  if (isSplitTransaction) {
    response = await call.post("/api/split-transaction", {
      split_transaction_id: id,
      label: { budget_id: value || null, category_id: null, category_confidence: 0 },
    });
    return;                                    //  ← BUG: drops to here for splits
  } else {
    response = await call.post("/api/transaction", { ... });
  }

  if (response.status === "success") {         //  ← unreachable for splits
    setData((oldData) => {
      if (isSplitTransaction) {                //  ← dead code via this handler
        const newSplitTransaction = new SplitTransaction(parentTransaction);
        newSplitTransaction.label.budget_id = value || null;
        newSplitTransaction.label.category_id = null;
        newSplitTransaction.label.category_confidence = 0;
        indexedDb.save(newSplitTransaction).catch(console.error);
        const newSplitTransactions = new SplitTransactionDictionary(newData.splitTransactions);
        newSplitTransactions.set(id, newSplitTransaction);
        newData.splitTransactions = newSplitTransactions;
      } else { ... }
      return newData;
    });
  } else {
    setSelectedBudgetIdLabel(selectedBudgetIdLabel);     //  ← rollback unreachable for splits
    setSelectedCategoryIdLabel(selectedCategoryIdLabel);
  }
};
```

The `return` on line 111 was added in commit `271dacd3` (2025-01-18, ~16 months ago). At that time the success branch didn't yet contain a split case — the `if (isSplitTransaction) { … }` inside `setData` was introduced by PR #332 (commit `a7d5cc6`, 2026-05-11). PR #332 added the cache-update path for splits but didn't remove the now-obsolete early return that prevents it from ever running.

## Sibling handlers in the same file are correct

- `onChangeCategorySelect` (lines 149-208) has no early return — both split and non-split reach `setData(...)` and update the dict correctly.
- `onAcceptSuggestion` (lines 213-247) — same shape, no early return, splits flow through.

Only `onChangeBudgetSelect` is broken for the split path.

## Impact

After a user changes a split row's budget from the TransactionsPage list:

1. **Server**: split_transactions row has new `label_budget_id`, `label_category_id = NULL`, `label_category_confidence = 0`. ✓
2. **Local React state in the row**: `selectedBudgetIdLabel` = new, `selectedCategoryIdLabel` = "". The row visually appears updated.
3. **`data.splitTransactions` dictionary**: **stale** — old budget_id, old category_id, old confidence.
4. **IndexedDB**: **stale** — `indexedDb.save(newSplitTransaction)` is never called for this path.
5. **Calculation hooks reading the dict** (budget bar totals, AccountsDonut, Sankey FlowChart, etc.): aggregate against stale split → totals incorrect on the budget the split was moved AWAY from (still includes it) and on the budget it was moved TO (doesn't include it) until the next full sync.
6. **Error path**: if the POST returns `{ status: "failed", message: "..." }`, the optimistic React state stays — row visually shows the new budget even though the server rejected it.

Compare to the non-split branch which reaches `setData` and `indexedDb.save` immediately. And compare to my prior issue **#415** (TransactionDetailPage edits leave stale `label_category_confidence`) — that one shipped a confidence field omitted from the POST body. This issue is the inverse: the POST body is fine, but the client-side cache write is skipped entirely. Different surfaces, same family of "split-tx code path missed a parity update".

## Reproduction (local prod-clone)

The bug is mechanical from the source — there's no other code path in `TransactionRow` that updates `data.splitTransactions` in response to `onChangeBudgetSelect`, and React's render cycle won't re-pull from the dictionary unless a setter fires. Server-side change confirms the POST succeeds:

```
PRE  (psql): label_budget_id=53c90bc8…, label_category_id=1a379137…, label_category_confidence=NULL
POST /api/split-transaction { split_transaction_id: 'c51c95da-…', label: { budget_id: 'cb53b3dc-…', category_id: null, category_confidence: 0 } }
→ 200 { status: "success" }
POST (psql): label_budget_id=cb53b3dc…, label_category_id=NULL, label_category_confidence=0
```

(I restored the row's label state at end of session — see Test cleanup below. Did not drive the full Playwright UI path because the bug is unambiguous from the handler source; if a UI-level repro is wanted I can add a follow-up Playwright assertion that compares `IndexedDB.splitTransactions[<id>].label.budget_id` before and after a `<select>`-change.)

## Fix sketch

Delete the early `return` on line 111. The `if (isSplitTransaction)` branches inside `setData` already handle both shapes correctly. Smallest patch:

```diff
   if (isSplitTransaction) {
     response = await call.post("/api/split-transaction", {
       split_transaction_id: id,
       label: { budget_id: value || null, category_id: null, category_confidence: 0 },
     });
-    return;
   } else {
     response = await call.post("/api/transaction", {
       transaction_id: id,
       label: { budget_id: value || null, category_id: null, category_confidence: 0 },
     });
   }
```

After the fix, both the success-branch `setData(...)` and the failure-branch optimistic-rollback fire correctly for splits.

Tests in `TransactionRow.test.tsx` (if any cover the split path) should be extended to assert the dictionary update on success and the rollback on failure. The duplication between `TransactionRow.tsx` and `InvestmentTransactionRow.tsx` flagged in **#336** would also benefit from a hoisted helper that closes over `isSplitTransaction`; this bug is part of why the duplication is costly — fixes have to be made in lockstep.

## Test cleanup

One write to `split_transactions` during the repro on row `c51c95da-bc61-4063-9ad1-3f5934a41328`, restored at end of session:

```sql
UPDATE split_transactions
SET label_budget_id   = '53c90bc8-5ed7-4233-b22a-3a6321d2c1b3',
    label_category_id = '1a379137-9298-4967-b836-a045c799b4e3',
    label_category_confidence = NULL
WHERE split_transaction_id = 'c51c95da-bc61-4063-9ad1-3f5934a41328';
-- UPDATE 1; SELECT confirms pre-test state restored
```

## Adjacent (not filed)

- The early-return-then-no-rollback shape means the React `selectedBudgetIdLabel` state can drift permanently from the server's truth on a 500/validation error — a separate failure mode from the cache-staleness above. Same root cause / same fix.
- `InvestmentTransactionRow.onChangeBudgetSelect` (and `onChangeCategorySelect`) — checked separately, no equivalent early-return on those.

### LINKED ISSUE #435: bug(transactions): `new SplitTransaction(parentTransaction)` discards split data on accept/edit, creates phantom IDB rows
### Symptom

When a user accepts or edits a category/budget on a **split** transaction from the transactions table, the in-memory state and IndexedDB are populated with a fresh `SplitTransaction` constructed from the **parent transaction** (`new SplitTransaction(parentTransaction)`). This discards the split's own `amount`, `custom_name`, `label.category_id`, `label.budget_id`, etc., and produces a fresh random `split_transaction_id` — an orphan row that gets written to IndexedDB under the new random key.

User-visible consequences once auto-suggest writes a suggestion to a split:
- Accept-All on a view containing a suggested split swaps the displayed split for one showing the parent's amount and label.
- Yellow-dot accept (`onAcceptSuggestion`) and category-select change in the table row have the same effect.
- IndexedDB accumulates phantom split rows (random ids, no server backing) that re-appear briefly on every warm-load paint.

### Root cause

`SplitTransaction`'s constructor uses `assign(this, init)` to copy enumerable properties from `init`. When `init` is a `Transaction`, the resulting `SplitTransaction` has:
- `split_transaction_id` = fresh `getRandomId()` (parent has no `split_transaction_id` field)
- `transaction_id` = parent's id (parent's own `transaction_id`)
- `amount` = parent's `amount`
- `custom_name` = `"Unknown"` (parent has `name`, not `custom_name`, so the default sticks)
- `label` = parent's `label` (rewrapped via `new TransactionLabel(parent.label)`)

The four bug sites:

1. **`src/client/pages/TransactionsPage/index.tsx:316`** — Accept-All:
   ```tsx
   } else if (existing instanceof SplitTransaction) {
     const parent = newData.transactions.get(existing.transaction_id);
     if (!parent) return;
     const updated = new SplitTransaction(parent);     // ← bug
     updated.label.category_confidence = 1;
     indexedDb.save(updated).catch(console.error);
     newSplits.set(id, updated);
   }
   ```

2. **`src/client/components/TransactionsTable/TransactionRow.tsx:231`** — yellow-dot `onAcceptSuggestion`:
   ```tsx
   if (isSplitTransaction) {
     const newSplitTransaction = new SplitTransaction(parentTransaction); // ← bug
     newSplitTransaction.label.category_confidence = 1;
     ...
   }
   ```

3. **`src/client/components/TransactionsTable/TransactionRow.tsx:181`** — `onChangeCategorySelect` (user picks a category on a split row):
   ```tsx
   if (isSplitTransaction) {
     const newSplitTransaction = new SplitTransaction(parentTransaction); // ← bug
     ...
   }
   ```

4. **`src/client/components/TransactionsTable/TransactionRow.tsx:123`** — `onChangeBudgetSelect` (same shape; currently shadowed by the early-`return` at line 111 tracked in #423, but the bug here would surface the moment #423 is fixed).

The contrast is `SplitTransactionRow.tsx:100,131,181`, which correctly uses `new SplitTransaction(splitTransaction)` — preserving the split's own data and id.

### Reproduction

Standalone mock against the real `SplitTransaction` shape:

```ts
const realSplit = new SplitTransaction({
  split_transaction_id: "split-abc",
  transaction_id: "tx-1",
  account_id: "acct-1",
  amount: 25,
  custom_name: "Coffee tip",
  label: new TransactionLabel({ category_id: "c-split", category_confidence: 0.97 }),
});
const broken = new SplitTransaction(parent); // parent is the Transaction whose tx-1 holds the split
broken.label.category_confidence = 1;

// broken.id        = random (NOT split-abc)
// broken.amount    = 100  (parent's amount, NOT 25)
// broken.custom_name = "Unknown"  (NOT "Coffee tip")
// broken.label.category_id = "c-parent"  (NOT "c-split")
```

Live: the test data has 57 active splits, none currently in suggested state (all `label_category_confidence IS NULL`), so the Accept-All / yellow-dot paths don't trigger the bug **today**. The auto-suggest engine (`defaultFetchUnlabeledSplits` + `defaultApplyLabelToSplit`) writes to splits — once a suggestion lands on a split + a user accepts it, the bug fires.

### Fix sketch

Construct the new `SplitTransaction` from the **existing split**, mirroring the pattern already in `SplitTransactionRow.tsx`:

```tsx
// TransactionRow.tsx — for all three split branches (lines 123, 181, 231)
const newSplitTransaction = new SplitTransaction(transaction as SplitTransaction);
// (or `existing as SplitTransaction` in TransactionsPage's Accept-All)
```

Then apply the field overrides as before (`category_confidence`, etc.). The `parentTransaction` lookup is still needed elsewhere in the file (for account/date display, etc.) but should not be the constructor source for the new split.

### Related

- **#423** (still open) — `onChangeBudgetSelect` early-returns for splits, so the buggy state-update block at line 122-130 doesn't run today. Fixing #423 by removing the early return without also fixing this issue would expose the same `new SplitTransaction(parentTransaction)` problem there. Both should land together.
- **#333** (still open) — dead `rejected` count for the suggest engine. Independent.
- **#422** (still open) — auto-suggest writes 0.99 (collides with API-reserved). Independent of this bug, but the moment that's resolved and the engine emits valid suggestions for splits, this bug becomes user-visible.

### Adjacent finding (not refiled)

`indexedDb.save(broken)` persists the phantom split under the random id. `cleanCache` / `clearAllData` only run on explicit logout. So orphan splits accumulate in the local IDB store across sessions and reappear briefly on every warm-load paint (Stage 0 paints from cache). Same shape as the soft-delete-staleness adjacent finding in #434.

claoie:
## What

When a user changes the **budget** of a *split* transaction from the transactions list, the server was updated but the client cache (`data.splitTransactions` + IndexedDB) was left stale, and a server rejection was never rolled back. Two coupled bugs:

- **#423** — `TransactionRow.onChangeBudgetSelect` had a stale `return` right after the split POST, short-circuiting the entire success/rollback branch for splits. The branch's split-cache-update code (added later by PR #332) was therefore dead. Result: in-memory dict + IndexedDB never updated; budget-bar/donut/Sankey totals stay wrong until the next full sync; a `status:"failed"` response leaves the optimistic select un-rolled-back.
- **#435** — removing that return alone would expose a second bug: the split branches construct the cached row with `new SplitTransaction(parentTransaction)`, which copies the **parent's** amount/custom_name and mints a fresh **random** `split_transaction_id` — discarding the split's own data and persisting a phantom IDB row under a random key. The same wrong constructor is used in `onChangeCategorySelect`, `onAcceptSuggestion` (TransactionRow) and Accept-All (TransactionsPage). #435 explicitly notes the two must land together.

## Fix

- Delete the dead early `return` in `onChangeBudgetSelect`.
- Construct the new split from the **existing split** — `new SplitTransaction(transaction as SplitTransaction)` (and `new SplitTransaction(existing)` in Accept-All) — matching the canonical pattern already in `SplitTransactionRow.tsx`. This drops the now-dead `parent` lookup in Accept-All.

4 lines changed net (+4/−7), 2 files.

## E2E verification (Playwright, prod-like server on local pg, admin)

Drove the real built bundle: login → `/transactions` → year view → located split `04bf0c11` (amount **500**, parent **1000**, budget *Annie's*) via its React fiber → changed its budget `<select>` to *Shared Budget* → awaited the `POST /api/split-transaction` → read IndexedDB.

**Post-fix — all assertions pass:**
- IDB entry for the split updates: `budget_id → Shared`, `category_id → null`, `category_confidence → 0` (#423)
- cached row keeps the split's **own** id `04bf0c11` and **own** amount **500** (not the parent's 1000) (#435)
- no phantom split row created (IDB split count 57 → 57) (#435)

**Baseline (this branch reverted, rebuilt, same test) — fails exactly as the bugs predict:** server POST succeeds (200) but the IDB cache keeps the **old** budget_id and `confidence=null` — the early return skipped the client cache write. This is the #423 symptom, and it confirms the test actually catches the regression.

typecheck ✓ · lint ✓ · build-client ✓ · build-server ✓

### Note on the other #435 sites
`onChangeCategorySelect` / `onAcceptSuggestion` / Accept-All aren't gated by an early return, so their phantom-row bug fires today only once an auto-suggestion lands on a split (the test data currently has no splits in suggested state, per #435). They are fixed here by the same constructor change but were not individually driven via E2E — the constructor fix is mechanical and shares the canonical `SplitTransactionRow.tsx` pattern.

Closes #423
Closes #435

### DIFF
--- a/src/client/components/TransactionsTable/TransactionRow.tsx
+++ b/src/client/components/TransactionsTable/TransactionRow.tsx
@@ -108,7 +108,6 @@ const TransactionRow = ({ transaction }: Props) => {
         split_transaction_id: id,
         label: { budget_id: value || null, category_id: null, category_confidence: 0 },
       });
-      return;
     } else {
       response = await call.post("/api/transaction", {
         transaction_id: id,
@@ -120,7 +119,7 @@ const TransactionRow = ({ transaction }: Props) => {
       setData((oldData) => {
         const newData = new Data(oldData);
         if (isSplitTransaction) {
-          const newSplitTransaction = new SplitTransaction(parentTransaction);
+          const newSplitTransaction = new SplitTransaction(transaction as SplitTransaction);
           newSplitTransaction.label.budget_id = value || null;
           newSplitTransaction.label.category_id = null;
           newSplitTransaction.label.category_confidence = 0;
@@ -178,7 +177,7 @@ const TransactionRow = ({ transaction }: Props) => {
       setData((oldData) => {
         const newData = new Data(oldData);
         if (isSplitTransaction) {
-          const newSplitTransaction = new SplitTransaction(parentTransaction);
+          const newSplitTransaction = new SplitTransaction(transaction as SplitTransaction);
           if (!newSplitTransaction.label.budget_id) {
             newSplitTransaction.label.budget_id = account?.label.budget_id;
           }
@@ -228,7 +227,7 @@ const TransactionRow = ({ transaction }: Props) => {
     setData((oldData) => {
       const newData = new Data(oldData);
       if (isSplitTransaction) {
-        const newSplitTransaction = new SplitTransaction(parentTransaction);
+        const newSplitTransaction = new SplitTransaction(transaction as SplitTransaction);
         newSplitTransaction.label.category_confidence = 1;
         indexedDb.save(newSplitTransaction).catch(console.error);
         const newSplitTransactions = new SplitTransactionDictionary(newData.splitTransactions);
--- a/src/client/pages/TransactionsPage/index.tsx
+++ b/src/client/pages/TransactionsPage/index.tsx
@@ -311,9 +311,7 @@ export const TransactionsPage = () => {
             indexedDb.save(updated).catch(console.error);
             newInvest.set(id, updated);
           } else if (existing instanceof SplitTransaction) {
-            const parent = newData.transactions.get(existing.transaction_id);
-            if (!parent) return;
-            const updated = new SplitTransaction(parent);
+            const updated = new SplitTransaction(existing);
             updated.label.category_confidence = 1;
             indexedDb.save(updated).catch(console.error);
             newSplits.set(id, updated);


claoie:
Closing — superseded by merged PR #487 (a9c6fee, "fix(splits): tighten SplitTransaction ctor + fix every caller that backed into a corrupt id").

Verified line-by-line against current `upstream/main`:

- **#423 (early `return;`)** — already removed. `TransactionRow.onChangeBudgetSelect` no longer short-circuits after the split POST; the success/rollback branch runs for splits.
- **#435 (construct from the split, not the parent)** — already done. All three `TransactionRow` branches now build `new SplitTransaction(transaction)` (was `parentTransaction`), and `TransactionsPage` Accept-All builds `new SplitTransaction(existing)` (was the `parent` lookup) — main lines 122/180/230 and 314 respectively.

The sole remaining delta in this branch was a `transaction as SplitTransaction` cast, which #487's tightened ctor makes redundant (main typechecks without it). Nothing functional left to land. Issues #423 and #435 are resolved by #487.

PR #486, opened 2026-06-06
fix(validation): UUID format guard on DELETE routes whose ids are genuinely UUIDs (7 of 10)

claoie:
## Summary

- Adds `requireUuidQueryString` to `validation.ts` — wraps `requireQueryString` and adds a UUID regex check, returning a 400 before the DB call on any non-UUID string.
- Applies it to all 10 DELETE route handlers that pass `?id=` directly to Postgres UUID columns.

## Root cause (from Shepherd alarm 2026-06-06T02:23:08)

`DELETE /split-transaction?id=c8f1c` fired from the client. `"c8f1c"` is a 5-char hex string produced by `getRandomId()` — the client-side default `split_transaction_id` before the server assigns a real UUID. The old `requireQueryString` only validated non-empty string; Postgres threw `invalid input syntax for type uuid: "c8f1c"`, which surfaced as an unhandled route error and triggered the alarm.

The immediate cause is stale IDB data carrying a `getRandomId()` ID from a prior session. The server-side fix (this PR) ensures any non-UUID ID is rejected with a 400 instead of an unhandled DB exception.

## E2E

Dev server started on this branch (`bun run dev`), tested via curl with admin session:

- `DELETE /api/split-transaction?id=c8f1c` → `{"status":"failed","message":"Parameter id must be a valid UUID"}` — no DB exception
- `DELETE /api/split-transaction?id=notauuid` → `{"status":"failed","message":"Parameter id must be a valid UUID"}` — no DB exception
- `DELETE /api/split-transaction?id=00000000-0000-0000-0000-000000000000` → `{"status":"success"}` — valid UUID passes through normally

Normal split-transaction CRUD flows unaffected.

### DIFF
--- a/src/server/lib/validation.ts
+++ b/src/server/lib/validation.ts
@@ -55,6 +55,20 @@ export function requireQueryString(
   return { success: true, data: value };
 }
 
+const UUID_REGEX = /^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$/i;
+
+export function requireUuidQueryString(
+  req: ServerRequest,
+  param: string
+): ValidationResult<string> {
+  const result = requireQueryString(req, param);
+  if (!result.success) return result;
+  if (!UUID_REGEX.test(result.data!)) {
+    return { success: false, error: `Parameter ${param} must be a valid UUID` };
+  }
+  return result;
+}
+
 /**
  * Extract an optional string parameter from request query.
  * Returns undefined if not present.
--- a/src/server/routes/accounts/delete-account.ts
+++ b/src/server/routes/accounts/delete-account.ts
@@ -4,7 +4,7 @@ import {
   deleteAccounts,
   searchAccountsById,
   getItem,
-  requireUuidQueryString,
+  requireQueryString,
   validationError,
 } from "server";
 
@@ -17,7 +17,7 @@ export const deleteAccountRoute = new Route("DELETE", "/account", async (req) =>
     };
   }
 
-  const idResult = requireUuidQueryString(req, "id");
+  const idResult = requireQueryString(req, "id");
   if (!idResult.success) {
     return validationError(idResult.error!);
   }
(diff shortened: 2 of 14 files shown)


claoie:
## Self-Review — corrected scope (pushed cbdf6a0)

**Caught a real break during review.** The original two commits applied the UUID guard to **all 10** DELETE routes, but 3 of them key on ids that are **not** UUIDs — so the guard would have rejected every legitimate delete with a 400:

| Route | id source | format |
|---|---|---|
| `delete-account` | client `getRandomId()` → `(65536 + rand*983040).toString(16)` | 5-char hex, e.g. `a3f01` |
| `delete-snapshot` | `post-snapshot.ts` | `` `${account_id}-${dateString}` `` |
| `delete-holding-snapshot` | `post-holding-snapshot.ts` | `` `holding-${account_id}-${security_id}-${dateString}` `` |

CI was green only because the two commits rewrote those routes' tests to feed UUID-shaped ids — which **masked** the break (the tests no longer matched the real id format). That's the dangerous part: a passing suite asserting an unreachable reality.

**Fix (commit `cbdf6a0`):** reverted those 3 routes to `requireQueryString`, and re-pinned `delete-holding-snapshot.test.ts` to the real deterministic id (`holding-acc-1-sec-1-20260514`) so the regression can't recur silently. The guard stays on the **7 genuinely-UUID routes**, all of which mint ids via `gen_random_uuid()` / `crypto.randomUUID()`:

`delete-budget`, `delete-category`, `delete-section`, `delete-chart`, `delete-transfer` (pair_id), `delete-item` (item_id), `delete-split-transaction`.

**Verification**
- `tsc --noEmit` → exit 0
- `bun test` on delete-holding-snapshot / delete-transfer / delete-item → 15 pass / 0 fail; the holding-snapshot happy path now drives the real non-UUID id and returns success (proves the route accepts it)
- Rebased onto `upstream/main` (`a9c6fee`), clean, force-pushed

**Logic** — correct. **Types** — clean. **Security** — the guard still rejects malformed ids on the 7 UUID routes (early input validation); the 3 reverted routes were already protected by the existing user-ownership checks in their WHERE clauses, so reverting loses no authorization guarantee. **Tests** — updated to real id formats. **Blockers** — none.

UI E2E not run: budget has no local `.env`/login password, so the deletion flow can't be driven through the browser in this environment. Verified at route + unit level instead.

PR #488, opened 2026-06-06
fix(auto-suggest): cap engine confidence at 0.98 so it never collides with the 0.99 /api reservation

### LINKED ISSUE #422: bug(auto-suggest): engine writes confidence=0.99, colliding with the /api/suggest-category reservation
## Summary

`runAutoSuggestions` (`src/server/lib/compute-tools/auto-suggest.ts:244`) writes `label_category_confidence = 0.99` to every suggested row whenever a merchant's signal evaluates to `confidence == 1.0`. Per the 2026-05-20 directive in the "Urgent bug fix" Discord thread, **0.99 is reserved for the `/api/suggest-category` endpoint** (the external Claude-instance API the `transactions:suggest` scope grants access to). The auto-suggest engine must write a value **strictly below 0.99** so the two suggestion sources stay distinguishable.

> Hoie (2026-05-20 20:40 UTC, "Urgent bug fix" thread): "when suggestion engine updates the category confidence is it always 0.99? I kind of remember that 0.99 is reserved for the suggestions made via API call that you can use."

Today, given the dead-rejection-count bug (#333), every successful auto-suggest signal evaluates to `accepted / totalLabeled == 1.0`, and `Math.min(1.0, 0.99) === 0.99` — so **every single auto-suggested row gets confidence=0.99**, indistinguishable from `/api/suggest-category` writes.

## Reproduction (live, local prod-clone)

Setup: 5 user-confirmed labels for `merchant_name = 'Sjc11 Cafe'` (all `category_id = 167ce37c…`, `confidence = 1.0`), one truly-unlabeled txn `Xp76pPv3kOFjOPQXmNzJI7ewOMJ9QeI0dknZR` with the same merchant.

Trigger the engine via the same entry point the scheduled runner uses:

```ts
import { runAutoSuggestions } from "./src/server/lib/compute-tools/auto-suggest";
await runAutoSuggestions();
```

Observed log + DB state after one run:

```
INFO  Auto-suggestion job started
INFO  Auto-suggested categories for user {"userId":"99b6850f…","suggested":101}
INFO  Auto-suggestion job completed {"usersProcessed":2,"transactionsSuggested":101}
```

```
SELECT label_category_id, label_category_confidence
FROM transactions WHERE transaction_id = 'Xp76pPv3kOFjOPQXmNzJI7ewOMJ9QeI0dknZR';
      label_category_id      | label_category_confidence
-----------------------------+---------------------------
 167ce37c-b548-4d91-882d-…   |                      0.99
```

Same 0.99 written across 98 Sjc11 Cafe rows plus 1 each on the fuzzy-similar `Amz Sjc13 Cafe`, `S Cafe`, `Sjsu Cafe` — all collide with the API-suggested reservation.

## Root cause

`auto-suggest.ts:236-245`:

```ts
const evaluateSignal = (signal: MerchantSignal): number | null => {
  const totalLabeled = signal.accepted + signal.rejected;
  if (totalLabeled < 3) return null;
  const rejectRate = signal.rejected / totalLabeled;
  if (rejectRate > 0.1) return null;
  const confidence = signal.accepted / totalLabeled;
  if (confidence < 0.95) return null;
  // Cap at 0.99 — 1.0 is reserved for user-confirmed labels.
  return Math.min(confidence, 0.99);
};
```

Two things are wrong:

1. **The cap value.** `Math.min(confidence, 0.99)` collapses `confidence == 1.0` (the typical case, see below) onto 0.99 — the API-suggested bucket. The cap must be strictly less than 0.99.
2. **The comment.** "1.0 is reserved for user-confirmed labels" is correct but incomplete — 0.99 is also reserved (for API). Update to make both reservations explicit.

The 0.99 value here was introduced in PR #344 (commit `2ed57f8`, 2026-05-13), before Hoie's 2026-05-20 clarification carved out 0.99 for API-suggested writes. It hasn't been revisited.

## Why "every" suggestion hits 1.0 today

`getMerchantSignal`'s outer query (`auto-suggest.ts:328-350`):

```sql
SELECT
   recent.label_category_id,
   sections.budget_id AS label_budget_id,
   SUM(CASE WHEN recent.label_category_confidence = 1.0 THEN 1 ELSE 0 END) AS accepted,
   SUM(CASE WHEN recent.label_category_confidence = 0.0 THEN 1 ELSE 0 END) AS rejected
 FROM ( … ) recent
 JOIN categories ON categories.category_id = recent.label_category_id
 JOIN sections   ON sections.section_id = categories.section_id
 GROUP BY recent.label_category_id, sections.budget_id
```

The rejection write path stores `label_category_id = NULL, label_category_confidence = 0`. `JOIN categories ON categories.category_id = recent.label_category_id` drops rows where `label_category_id IS NULL` (SQL `NULL = NULL` is false). So rejections never reach the outer GROUP BY → `rejected` is always 0 → `confidence = accepted / (accepted + 0) = 1.0`. That dead-rejection-count is **#333**, but it's the upstream cause of *this* bug — fixing #333 alone would still leave the 0.99 collision for any merchant with 100% acceptance.

These are two distinct bugs at two different layers (one in the SQL grouping, one in the cap value); even after #333 ships, `evaluateSignal` should cap strictly below 0.99 so that merchants with no rejections at all (a common, legitimate case) don't collide with API-suggested.

## Impact

- **Provenance ambiguity.** Today, every row written by the engine carries the exact same confidence value as every row written by `/api/suggest-category`. Future debugging ("which engine touched this row?"), per-source UX (e.g. "show me API-suggested rows only"), and per-source guardrails (e.g. "engine writes need re-evaluation after #333 lands; API writes don't") are all unworkable.
- **Cap collision is silent.** No log line or alarm fires when the engine writes 0.99. The collision is invisible until someone audits.
- **CAS guard #408 doesn't help.** PR #408's `CAS_NULL_CONFIDENCE` (`label_category_confidence IS NULL`) prevents the engine from clobbering rows that already have a non-NULL confidence — that part is fine. But it doesn't address the *value* the engine writes when the CAS passes.

Existing tests **encode the buggy behavior**: `auto-suggest.test.ts:169` asserts `applyCalls[0].labelCategoryConfidence).toBe(0.99)` and the split-path test at `:347` does the same. Both need updating in lockstep with the fix.

## Fix sketch

1. In `auto-suggest.ts:244`, pick an explicit engine-cap constant strictly below 0.99 (suggestion: `0.98`, matching the "two reserved slots above" convention). Lift it to a named constant so the reservation map is searchable:

   ```ts
   // Auto-suggest engine writes confidence in [0.95, AUTO_SUGGEST_MAX_CONFIDENCE].
   // 1.0 is reserved for user-confirmed UI writes; 0.99 is reserved for
   // /api/suggest-category writes from external Claude instances. The engine
   // must never collide with either bucket so suggestion provenance stays
   // distinguishable.
   export const AUTO_SUGGEST_MAX_CONFIDENCE = 0.98;
   …
   return Math.min(confidence, AUTO_SUGGEST_MAX_CONFIDENCE);
   ```
2. Update the two `0.99` assertions in `auto-suggest.test.ts` to assert the new constant; add a new test asserting the engine NEVER writes 0.99 even when `accepted/totalLabeled === 1.0`.
3. Optional doc touch-up in `docs/DATABASE.md` / `docs/ARCHITECTURE.md` (auto-categorization section landed via PR #345) to enumerate the three buckets: `< 0.99` (engine), `= 0.99` (API), `= 1.0` (user).
4. Since the issue is invisible until audited, consider a one-shot SQL audit on prod after the fix lands: `SELECT user_id, COUNT(*) FROM transactions WHERE label_category_confidence = 0.99 GROUP BY user_id` — any pre-fix engine writes will surface here and can be retroactively migrated to the new value if a clean provenance baseline matters. (Not required for correctness going forward; the engine's CAS_NULL_CONFIDENCE guard means the legacy 0.99 rows just look indistinguishable from API writes, which is the same problem we've always had.)

## Test cleanup

The repro touched 6 rows in the local prod-clone DB. All restored at end of session (verified via `SELECT COUNT(*) FROM transactions WHERE label_category_confidence IS NOT NULL` → 0, same as pre-test). One caveat: the engine's 101-row write also overwrote a minority `label_category_id` on 1 Sjc11 Cafe row (the previous `36a677c0…` rare category, now collapsed onto the majority `167ce37c…`) and touched 3 fuzzy-similar merchant rows whose pre-test category_ids were not captured. Local clone is regenerable; flagging for transparency.

## Adjacent (not filed)

- **`post-suggest-category.ts:38`** validates `confidence > 0 && confidence < 1` but allows any value in that range. The API could write 0.5, 0.97, anything — only convention (and the test fixtures' use of 0.99) enforces the canonical API value. If we lock the reservation map in code, consider enforcing `confidence === 0.99` (or the dedicated API constant) on the API path too.
- The fuzzy-similarity matching is broad enough that `S Cafe` got grouped with `Sjc11 Cafe` during the repro — likely a separate concern (`MERCHANT_SIMILARITY_THRESHOLD = 0.5` may be too loose for short merchant names) but not in scope here.

claoie:
Closes #422

## Problem

`evaluateSignal` in `auto-suggest.ts` capped the suggested confidence with `Math.min(confidence, 0.99)`. A merchant whose history is 100% accepted computes `confidence = accepted / (accepted + rejected) = 1.0`, and `Math.min(1.0, 0.99) === 0.99` — but **0.99 is reserved for `/api/suggest-category`** (the external Claude-instance endpoint), per Hoie's 2026-05-20 directive. So every fully-accepted merchant's auto-suggested rows were written with the exact same confidence value as API-suggested rows, destroying suggestion provenance.

100% acceptance is the common case (especially with the dead-rejection-count of #333), so in practice essentially *every* engine write collided with the API bucket.

## Fix

- Lift the cap to a named exported constant `AUTO_SUGGEST_MAX_CONFIDENCE = 0.98`, strictly below the 0.99 API reservation, and write it via `Math.min(confidence, AUTO_SUGGEST_MAX_CONFIDENCE)`.
- Document the three reserved buckets in the source comment and in `docs/ARCHITECTURE.md`, `docs/DATABASE.md`, `docs/DESIGN_PATTERNS.md`:
  - `< 0.99` → auto-suggest engine
  - `= 0.99` → `/api/suggest-category`
  - `= 1.0` → user-confirmed labels

## Tests

Strengthened the two existing cap tests (`auto-suggest.test.ts`) so they assert the engine writes `0.98` and **never** `0.99` at `confidence = 1.0` (both the top-level transaction pass and the split-transaction pass). These drive `runAutoSuggestions()` end-to-end through the leaf-mocked `pg` SQL router and capture the exact `UPDATE … label_category_confidence = $3` parameter, so they verify the real write path, not just the helper in isolation.

```
18 pass, 0 fail (auto-suggest.test.ts)
```

`bun run typecheck` clean.

## Verification note

This is a backend numeric cap; the behavioral assertion is the exact value the engine UPDATEs, which the unit tests capture end-to-end through the SQL router. A browser E2E can't observe the stored confidence number directly, so the per-test-bundle test driving `runAutoSuggestions()` is the functional verification here. No prod backfill of legacy 0.99 engine rows is included (out of scope; the issue notes it's optional and CAS-guarded).

## Scope

Changes are limited to the engine cap (`auto-suggest.ts`), its tests, and the confidence-bucket documentation that the cap value feeds. The upstream dead-rejection-count (#333) is a separate issue with its own PR; this fix is correct independently of it.

### DIFF
--- a/docs/ARCHITECTURE.md
+++ b/docs/ARCHITECTURE.md
@@ -65,13 +65,14 @@ A transaction carries three correlated fields that together describe its current
 | `label_budget_id` | UUID \| null | The parent budget for `label_category_id` — written alongside it so the UI's category `<select>` (which filters options by budget) can render the value |
 | `label_category_confidence` | number \| null | Confidence in [0, 1]; `null` means unlabeled |
 
-The `label_category_confidence` field encodes four distinct states:
+The `label_category_confidence` field encodes these states. The `0 < c < 1` band is further partitioned so a suggested row's provenance stays recoverable:
 
 | Confidence | Meaning | UI signal |
 |---|---|---|
 | `null` | Unlabeled (`label_category_id IS NULL`) | Red dot in `TransactionRow` |
 | `0` | User rejected a prior suggestion | Counted as a rejection signal for that merchant |
-| `0 < c < 1` | Auto-suggest applied a label — user has not yet confirmed or rejected | Yellow dot in `TransactionRow` |
+| `c < 0.99` (engine cap `0.98`) | Auto-suggest engine applied a label — user has not yet confirmed or rejected | Yellow dot in `TransactionRow` |
+| `0.99` | `/api/suggest-category` write (external Claude instance) | Yellow dot in `TransactionRow` |
 | `1` | User explicitly confirmed | No dot |
 
 A prod backfill on 2026-05-13 set `label_category_confidence = 1` for every labeled row in `transactions`, so the `(category_id IS NOT NULL, confidence IS NULL)` combination is no longer expected in production data.
--- a/docs/DATABASE.md
+++ b/docs/DATABASE.md
@@ -72,13 +72,14 @@ A labeled transaction stores its category across three correlated columns. Both 
 
 **Write both `label_category_id` and `label_budget_id` together.** The UI category `<select>` filters options by the row's budget. Writing only `label_category_id` leaves the dropdown unable to render the value. See `defaultApplyLabel` in `src/server/lib/compute-tools/auto-suggest.ts` for the canonical pattern.
 
-**The four states of `label_category_confidence`:**
+**The states of `label_category_confidence`:** the `0 < c < 1` band is split into two reserved buckets so suggestion provenance is recoverable.
 
 | Value | Meaning |
 |---|---|
 | `NULL` | Row is unlabeled (`label_category_id IS NULL`) |
 | `0` | User rejected an auto-suggestion |
-| `0 < c < 1` | Auto-suggested, not yet confirmed by user |
+| `c < 0.99` (engine cap `0.98`) | Auto-suggest engine applied a label, not yet confirmed by user |
+| `0.99` | `/api/suggest-category` write (external Claude instance) |
 | `1` | User confirmed |
 
 **Prod backfill.** A backfill on 2026-05-13 set `label_category_confidence = 1` for every row with `label_category_id IS NOT NULL`. The application code's confirmation predicate (`category_confidence === 1 && !!category_id` in `src/client/lib/hooks/calculation/budgets.ts`) does not tolerate `NULL`, so this backfill is load-bearing — any new INSERT that sets `label_category_id` without also setting `label_category_confidence` would reintroduce the "labeled but counted as unsorted" miscount.
--- a/docs/DESIGN_PATTERNS.md
+++ b/docs/DESIGN_PATTERNS.md
@@ -154,7 +154,7 @@ Auto-categorization (`src/server/lib/compute-tools/auto-suggest.ts`) writes cate
 | Min total labeled | `accepted + rejected >= 3` | Don't suggest from a single data point             |
 | Max reject rate   | `rejected / total <= 0.10` | Users actively disagree — back off                 |
 | Min confidence    | `accepted / total >= 0.95` | Suggestion has to be near-unanimous in the history |
-| Confidence cap    | `min(confidence, 0.99)`    | `1.0` is reserved for user-confirmed labels        |
+| Confidence cap    | `min(confidence, AUTO_SUGGEST_MAX_CONFIDENCE)` (`0.98`) | Strictly `< 0.99`: `0.99` is reserved for `/api/suggest-category`, `1.0` for user-confirmed labels |
 
 The signal itself comes from a pg_trgm fuzzy match on `merchant_name` (`MERCHANT_SIMILARITY_THRESHOLD = 0.5`), with a `LIMIT 30` over the user's recent labeled transactions for that merchant.
 
--- a/src/server/lib/compute-tools/auto-suggest.test.ts
+++ b/src/server/lib/compute-tools/auto-suggest.test.ts
@@ -183,12 +183,14 @@ describe("runAutoSuggestions", () => {
     expect(updateCalls(/UPDATE\s+transactions\b/i)).toHaveLength(0);
   });
 
-  test("applies suggestion and caps confidence at 0.99", async () => {
+  test("applies suggestion and caps confidence at 0.98 (never 0.99)", async () => {
     userRows = [userRow({ user_id: "user-4" })];
     unlabeledByUser.set("user-4", [
       txRow({ transaction_id: "txn-4", user_id: "user-4", merchant_name: "Grocery Store" }),
     ]);
-    // 10 accepted, 0 rejected → confidence = 1.0 → caps at 0.99.
+    // 10 accepted, 0 rejected → confidence = 1.0 → caps at 0.98. The engine
+    // must NOT write 0.99: that value is reserved for /api/suggest-category,
+    // and 1.0 for user-confirmed labels (#422).
     signalRow = {
       label_category_id: "cat-groceries",
       label_budget_id: "bud-household",
@@ -205,7 +207,8 @@ describe("runAutoSuggestions", () => {
     // Confidence is the 3rd param.
     expect(updates[0].values[0]).toBe("cat-groceries");
     expect(updates[0].values[1]).toBe("bud-household");
-    expect(updates[0].values[2]).toBe(0.99);
+    expect(updates[0].values[2]).toBe(0.98);
+    expect(updates[0].values[2]).not.toBe(0.99);
     expect(updates[0].values[3]).toBe("txn-4");
     expect(updates[0].values[4]).toBe("user-4");
   });
@@ -316,7 +319,8 @@ describe("runAutoSuggestions", () => {
     expect(updates).toHaveLength(1);
     expect(updates[0].values[0]).toBe("cat-groceries");
     expect(updates[0].values[1]).toBe("bud-household");
-    expect(updates[0].values[2]).toBe(0.99);
+    expect(updates[0].values[2]).toBe(0.98);
+    expect(updates[0].values[2]).not.toBe(0.99);
     expect(updates[0].values[3]).toBe("split-1");
     expect(updates[0].values[4]).toBe("user-split-1");
   });
--- a/src/server/lib/compute-tools/auto-suggest.ts
+++ b/src/server/lib/compute-tools/auto-suggest.ts
@@ -21,6 +21,19 @@ export const CAS_NULL_CONFIDENCE: AdditionalWhere = {
   value: null,
 };
 
+// Highest confidence the auto-suggest engine may write. The confidence column
+// has three reserved buckets, kept distinct so a row's suggestion provenance
+// stays recoverable:
+//   - `< 0.99`  → auto-suggest engine (this file)
+//   - `= 0.99`  → /api/suggest-category (external Claude-instance writes)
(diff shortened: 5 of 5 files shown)


claoie:
## Self-Review

**Thread:** No prior review. CI green (build/test). Mergeable, 0 behind `main`.

**Logic — correct.** A merchant with 100% acceptance computes `confidence = accepted/(accepted+rejected) = 1.0`. The old `Math.min(confidence, 0.99)` collapsed that to exactly `0.99` — the value reserved for `/api/suggest-category` (external Claude-instance writes) — making engine-suggested rows indistinguishable from API-suggested ones. New `AUTO_SUGGEST_MAX_CONFIDENCE = 0.98` caps strictly below the reservation, preserving the three-bucket provenance scheme (`<0.99` engine / `0.99` API / `1.0` user-confirmed).

**Types** — `tsc --noEmit` exit 0.
**Quality** — the cap is a named exported constant with a comment block explaining the reservation map; grep-able, single source of truth.
**Security** — none affected (confidence is internal provenance, not authz).
**Scope** — the 3 doc edits (ARCHITECTURE/DATABASE/DESIGN_PATTERNS) all document the same `label_category_confidence` reservation buckets this fix establishes; directly on-topic, no scope creep.

**Tests** — `bun test auto-suggest.test.ts` → 18 pass / 0 fail. The relevant assertions, by real name:
- `applies suggestion and caps confidence at 0.98 (never 0.99)` — asserts `updates[0].values[2]` is `0.98` and `not 0.99`
- `computes exact confidence when ratio is between 0.95 and 0.99` — confirms the cap doesn't distort sub-threshold values
- `skips when confidence < 0.95` — lower gate intact

**Blockers** — none. **Confidence: High.**

E2E note: budget has no local login password in this environment, so the suggestion flow can't be driven through the browser. Verified at engine + unit level (the cap is a pure function of accept/reject counts, fully covered by the unit assertions above).

reviewoie:
## Code review — PR 488

**Verdict:** approve (one low-severity stale comment)

Verified the fix end-to-end:
- `evaluateSignal` (`auto-suggest.ts:217-222`) is the single confidence-producing chokepoint; both write paths (`applyLabel` → transactions, `applyLabelToSplit` → splits) consume its return value, so capping there covers every engine emit path. No other engine path writes `label_category_confidence`.
- 0.98 is the right cap: `confidence = accepted / totalLabeled` reaches `1.0` at 100% acceptance; `min(…, 0.98)` keeps the engine strictly below the reserved `0.99` and below user-confirmed `1.0`. `0.95 ≤ c < 0.98` ratios pass through unchanged (verified `computes exact confidence…` test still asserts 0.95).
- `detect-transfers.ts:25`'s `min(…, 0.99)` is a different column (`transaction_pairs.confidence`), correctly out of scope and untouched.
- Test file uses the current convention (`test-preload` leaf-mock of `pg` + `afterAll(restoreLeaves)`), not the retired per-test-bundle framework. Ran the suite on the PR head: 18 pass / 0 fail, including the updated `caps confidence at 0.98 (never 0.99)` assertions on both the transaction and split paths.
- Scope clean: all 5 files (3 docs + engine + its test) tie directly to the cap change.

### Low
- `auto-suggest.ts:175` — the `runAutoSuggestions` JSDoc still reads "apply the suggestion with confidence = accepted / total_labeled (capped at 0.99)". The PR updated the constant docblock and the inline `evaluateSignal` comment to 0.98 but left this summary comment contradicting the new behavior. One-line fix: change "0.99" → "0.98" (or reference `AUTO_SUGGEST_MAX_CONFIDENCE`).

**Informational (not a blocker, pre-existing):** the docs added here describe `0.99` as "the /api/suggest-category write" value, but `post-suggest-category.ts:38` validates `confidence > 0 && confidence < 1` and persists whatever the caller passes — it does not pin API writes to exactly 0.99. The reservation contract Hoie stated (2026-05-20) is "the engine must never write 0.99," which this PR satisfies. The API-side "always writes 0.99" framing is slightly stronger than the route enforces, but that's orthogonal to this fix and pre-exists the PR.

claoie:
Superseding this — the multi-feature scoring rewrite in #517 (merged to main this morning) already implements exactly what this PR proposed, and more cleanly.

#517's `evaluateSignal` clamps every engine write to `[ENGINE_CONFIDENCE_FLOOR, ENGINE_CONFIDENCE_CEIL] = [0.5, 0.98]` (auto-suggest.ts:370), with `ENGINE_CONFIDENCE_CEIL = 0.98` lifted to a named, commented constant whose doc-string explicitly states it "stays below the 0.99 reservation for the API `/api/suggest-category` endpoint." That is the same fix this PR shipped via `AUTO_SUGGEST_MAX_CONFIDENCE = 0.98` — so rebasing this onto current main would only re-introduce a duplicate constant alongside `ENGINE_CONFIDENCE_CEIL`, which is the kind of two-canonical-ways drift we avoid.

The engine can no longer write 0.99 (the bug in #422 is resolved on main). The one loose end #517 left is documentation: `docs/DESIGN_PATTERNS.md` still describes the pre-#517 `min(confidence, 0.99)` cap and a per-merchant cache the rewrite removed. Following up with a docs-only PR to align that section and finish #422's optional doc item.

PR #427, opened 2026-05-26
test(holdings): cover validation paths in POST /snapshots/holding

claoie:
## Summary

- Adds 6 tests covering all early-return validation paths in POST /snapshots/holding: unauthenticated request, non-object body, missing account_id, missing ticker_symbol, missing quantity, null quantity
- Full branch coverage of the create/update business logic is blocked pending a DI refactor of the route's free-function dependencies (free functions imported from the `"server"` barrel cannot be monkey-patched in Bun — same blocker documented in PR #406)

Partial coverage for #359

## Test plan

- [x] All 6 tests pass locally (`bun test post-holding-snapshot.test.ts`)
- [x] Test-only change — no production code modified

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #489, opened 2026-06-06
test(holdings): cover update/create success paths in POST /snapshots/holding (#359)

### LINKED ISSUE #359: test(holdings): zero coverage on POST /snapshots/holding (224 LOC, two modes, three side-effect chains)

claoie:
Closes #359

## Context

`post-holding-snapshot.ts` is a 238-LOC route with two modes (update + create) and three chained side-effect paths. The existing `post-holding-snapshot.test.ts` (from #427) only covers the cheap pre-DB validation guards — unauthenticated, non-object body, missing `account_id`/`ticker_symbol`/`quantity`. None of the success-mode behavior #359 flagged (cross-user ownership, re-sync, deterministic id) was exercised.

## What this adds

New `post-holding-snapshot.success.test.ts` (5 tests) driving both real modes through a **SQL-routing fake `pg` Pool** (the `mock.module("pg")` + `restoreLeaves` pattern from `post-suggest-category.test.ts`). Only the two SELECTs are scripted (`snapshots` → holding rows, `securities` → security rows); writes and the fire-and-forget backfill fall through to an ok result, so the test is robust to query count/order.

**Update mode**
- Rejects a `snapshot_id` the user does not own — asserts the "not found or access denied" message **and** that no `UPDATE snapshots` was issued (the cross-user ownership gate #359 called out as "one refactor away from leaking another user's snapshot").
- Rejects an empty patch with "no fields to update".
- Happy path: pins that `quantity: 0` is applied (not dropped by the `!== undefined` guard), the `UPDATE` is user-scoped, and the holdings row is re-synced.

**Create mode**
- Existing-security happy path (Polygon never called) with the deterministic `snapshot_id` assertion (`holding-<account>-<security>-<squashedDate>`).
- Re-posting the same `(account, security, date)` reuses the id → in-place upsert, not a duplicate row.

## Scope / follow-up

Test-only; no production code touched. The Polygon-fetch branches (`no_api_key` / invalid-ticker messages) are deferred: the `polygon` namespace export is readonly, so reassigning `polygon.getTickerDetail` throws — it needs a different seam than the `pg` fake. Noted on the issue for a follow-up rather than forcing a brittle HTTP stub here.

```
post-holding-snapshot.success.test.ts: 5 pass
src/server/routes/accounts/ (full): 55 pass, 0 fail
bun run typecheck: clean
```

### DIFF
--- a/src/server/routes/accounts/post-holding-snapshot.success.test.ts
+++ b/src/server/routes/accounts/post-holding-snapshot.success.test.ts
@@ -1,0 +1,223 @@
+// Success / side-effect coverage for POST /snapshots/holding (#359). The
+// sibling `post-holding-snapshot.test.ts` (from #427) pins the cheap pre-DB
+// validation guards; this file drives the two real modes (update + create)
+// through a SQL-routing fake `pg` Pool so the ownership gate, the re-sync
+// chain, the deterministic snapshot_id, and the no-API-key branch are
+// actually exercised — not just the early returns.
+//
+// Routing rather than a per-call queue keeps the test robust to the extra
+// writes (upsert/update) and the fire-and-forget backfill the route emits:
+// only the two SELECTs are scripted (snapshots → holding rows, securities →
+// security rows); everything else falls through to an empty/ok result.
+import { describe, test, expect, mock, beforeEach, afterAll } from "bun:test";
+import { restoreLeaves } from "test-helpers";
+import { getSquashedDateString, LocalDate } from "common";
+
+let snapshotRows: Record<string, unknown>[] = [];
+let securityRows: Record<string, unknown>[] = [];
+
+const mockQuery = mock(async (sql: string, _values?: unknown[]) => {
+  if (/select[\s\S]*from\s+snapshots/i.test(sql)) {
+    return { rows: snapshotRows, rowCount: snapshotRows.length };
+  }
+  if (/select[\s\S]*from\s+securities/i.test(sql)) {
+    return { rows: securityRows, rowCount: securityRows.length };
+  }
+  // INSERT / UPDATE / upsert and anything else: report one affected row.
+  return { rows: [{ ok: true }], rowCount: 1 };
+});
+
+class FakePool {
+  query = mockQuery;
+  end = async () => {};
+  connect = async () => ({ query: mockQuery, release: () => {} });
+}
+
+mock.module("pg", () => ({
+  Pool: FakePool,
+  types: { setTypeParser: () => {} },
+  default: { Pool: FakePool, types: { setTypeParser: () => {} } },
+}));
+
+const { postHoldingSnapshotRoute } = await import("./post\-holding\-snapshot");
+
+afterAll(restoreLeaves);
+
+beforeEach(() => {
+  mockQuery.mockClear();
+  snapshotRows = [];
+  securityRows = [];
+});
+
+function makeReq(body: unknown, userId: string | undefined = "u-1") {
+  return {
+    method: "POST",
+    path: "/snapshots/holding",
+    url: "http://x/api/snapshots/holding",
+    headers: {},
+    query: {},
+    body,
+    session: {
+      id: "s-1",
+      user: userId ? { user_id: userId, username: "test" } : undefined,
+      regenerate() {},
+      destroy() {},
+    },
+    ip: "127.0.0.1",
+  } as unknown as Parameters<typeof postHoldingSnapshotRoute.execute>[0];
+}
+
+const fakeRes = () =>
+  ({
+    statusCode: 200,
+    headersSent: false,
+    status() {
+      return this;
+    },
+    write() {
+      return true;
+    },
+    end() {},
+  }) as unknown as Parameters<typeof postHoldingSnapshotRoute.execute>[1];
+
+const findCall = (re: RegExp): { sql: string; values: unknown[] } | null => {
+  for (const call of mockQuery.mock.calls) {
+    const sql = call[0] as string;
+    if (re.test(sql)) return { sql, values: (call[1] ?? []) as unknown[] };
+  }
+  return null;
+};
+
+// A row shaped to satisfy SecurityModel validation (securitiesTable.query
+// wraps each row in the model before searchSecurities maps it to JSON).
+const existingSecurityRow = (overrides: Record<string, unknown> = {}) => ({
+  security_id: "sec-1",
+  ticker_symbol: "VOO",
+  name: "Vanguard S&P 500 ETF",
+  type: null,
+  close_price: null,
+  close_price_as_of: null,
+  iso_currency_code: "USD",
+  unofficial_currency_code: null,
+  isin: null,
+  cusip: null,
+  sedol: null,
+  institution_security_id: null,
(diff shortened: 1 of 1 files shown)


claoie:
## Self-Review

**Thread:** No prior review. CI green (build/test). Mergeable, 0 behind `main`. Closes #359.

**Scope — clean.** Single new file `post-holding-snapshot.success.test.ts`; no production code touched. Pure coverage addition over the route's update + create success paths (the existing `post-holding-snapshot.test.ts` only covered pre-DB validation guards).

**Logic/coverage** — `tsc --noEmit` exit 0; `bun test` → 5 pass / 0 fail / 18 assertions. Real tests:
- `rejects a snapshot_id the user does not own (ownership gate)` — authz path
- `rejects an empty patch with 'no fields to update'` — update-mode guard
- `applies numeric patches (including 0) and re-syncs the holdings row` — the `0` case matters (falsy-but-valid)
- `creates with an existing security and a deterministic snapshot_id` — create mode
- `re-posting the same (account, security, date) reuses the snapshot_id` — upsert idempotency

The deterministic-`snapshot_id` assertions correctly pin the real format `holding-${account_id}-${security_id}-${dateString}` (not a UUID) — consistent with the route's actual id minting.

**Security** — ownership gate is explicitly tested. **Blockers** — none. **Confidence: High.**

E2E note: test-only PR; the assertions exercise the route handler directly with mocked queries, which is the appropriate verification level here. No browser path applies.

reviewoie:
## Code review — PR 489

Scope is clean: one new file, `post-holding-snapshot.success.test.ts`, test-only (+223/-0), directly named by the title. No source changes, zero functional risk.

Verified empirically against the PR head (`fead70d`):
- New file alone: 5 pass / 0 fail.
- Whole `routes/accounts/` dir together (leak check): 58 pass / 0 fail.
- Full `src/server` suite: 550 pass / 0 fail. The `mock.module("pg")` + `afterAll(restoreLeaves)` pattern does not bleed into sibling files.

Substance checks:
- Tests assert the response **body outcome**, not just HTTP status — `result.status`, `result.body.snapshot_id`, `result.body.security_id`, and `result.message` regex. This is the status-200-with-`failed`-payload trap our APIs use, and the file covers it.
- The SQL-routing fake matches the real table names: `getHoldingSnapshots` selects `from snapshots`, `searchSecurities` → `securitiesTable.query` selects `from securities` (confirmed in repositories). Routing rather than a per-call queue is the right call given the route's extra writes + fire-and-forget backfill.
- The `quantity: 0` case correctly exercises the route's `!== undefined` guard (zero applied, not dropped) and asserts the UPDATE is user-scoped — good falsy-zero coverage.
- Framework note: the `mock.module("pg") + restoreLeaves` preload pattern is the current canonical isolation mechanism here; the older per-test-bundle `.test.bundle.ts` approach is gone (no `test:bundled` script, no bundle files). This file matches the current convention.

### Low
- `post-holding-snapshot.success.test.ts:38` — the dynamic import escapes the hyphens: `await import("./post\-holding\-snapshot")`. `\-` is just `-` in a JS string so it resolves (tests pass), but it's needless noise versus the sibling's plain `"./post-holding-snapshot"`. Worth dropping the backslashes.

Easy-merge candidate — clean scope, zero functional risk, full suite green.

claoie:
## Reviewoie follow-up

One finding, low-severity informational nit — not addressing in this PR.

|---|---|---|---|---|
| 1 | low | `post-holding-snapshot.success.test.ts:38` | dynamic import uses needless backslash-escaped hyphens (`./post\-holding\-snapshot`) instead of the plain string used by siblings | Leaving as-is — functionally identical (`\-` is just `-` in a JS string, tests pass), and a one-character noise tweak doesn't justify a new commit on a test-only PR. Easy to fold into a follow-up sweep if a sibling test file needs touching. |

No new commits — reviewoie's verdict was `approve (informational)`. Ping if you want this addressed before merge.

hoiekim (approved):

PR #494, opened 2026-06-07
fix(accounts): Changes widget compares against $0 for zero-balance months

### LINKED ISSUE #428: bug(accounts): Changes widget uses current balance instead of $0 for zero-balance historical months
## Summary

`src/client/components/AccountsTable/Balance.tsx:62` uses `||` instead of `??` to build the Changes widget's `currentAmount`:

```typescript
const changesProps = {
  currentAmount: dynamicAmount || current!,  // ← bug: || treats 0 as falsy
  previousAmount,
};
```

When `dynamicAmount` is `0` (a historical month where the account balance was genuinely $0, from a snapshot or holding total), the `||` operator falls through to `current!` — the live account balance from Plaid/manual entry. The displayed balance row correctly shows `$0` (line 71 uses `dynamicAmount` directly), but the Changes widget compares `current!` against `previousAmount`, producing a completely wrong delta.

## Reproduction

1. Any account with a `$0`-balance snapshot in month **X** AND a non-zero balance snapshot in month **X−1** AND a non-zero `balances.current`.
2. Navigate to month **X** on the Accounts page.
3. **Expected:** row shows `$0` balance and Changes shows `−$previousAmount` (the account went to zero).
4. **Actual:** row shows `$0` balance but Changes shows `$current − $previousAmount` (the live balance minus the previous month — unrelated numbers).

### Concrete data path

<account-name> (`<account-id>`) has `balances_current = $0.00` in its September 2025 account_balance snapshot and `$<amount-A>` in October 2025. Today's `balances.current = $<amount-B>`. Currently no preceding August snapshot exists, so the `shouldShowChanges` gate (`!!previousAmount || !!dynamicAmount` = `!!0 || !!0` = `false`) hides the widget and the bug is invisible. But if a future snapshot backfill or Plaid sync adds an August-or-earlier data point, the bug would surface:

- `dynamicAmount = 0` (Sep snapshot)
- `previousAmount = $X` (Aug snapshot)
- `shouldShowChanges = true`
- `currentAmount = 0 || <amount-B> = <amount-B>` ← **wrong, should be 0**
- Changes renders `+$(<amount-B> − X)` instead of `−$X`

The same shape applies to any account that legitimately goes to $0 for a period: fully liquidated brokerage positions (holding total = 0), checking accounts swept clean between paychecks, etc.

### Root cause

Line 62 in `Balance.tsx`:
```typescript
currentAmount: dynamicAmount || current!,
```

`dynamicAmount` is `number` (never `null`/`undefined` at this point — the `?? fallback` on line 26 already resolved it). The `||` is only needed to substitute `current!` when `dynamicAmount` is `0`. But `0` is the correct balance for a zero-balance month — the substitution masks it.

## Fix sketch

One-character change:

```diff
- currentAmount: dynamicAmount || current!,
+ currentAmount: dynamicAmount ?? current!,
```

Since `dynamicAmount` is always a `number` (never `null`/`undefined`) at this point, the `??` never fires. But the semantic is correct: only substitute `current` when `dynamicAmount` is truly missing (which can't happen after line 26). To make the intent clearer, the fallback could be dropped entirely:

```diff
- currentAmount: dynamicAmount || current!,
+ currentAmount: dynamicAmount,
```

Both produce the same result for the current code flow (line 26 guarantees `dynamicAmount` is a `number`). The `|| current!` was likely a safety net from before the `?? fallback` was added on line 26 — it's now redundant and only fires on the $0 edge case where it causes harm.

## Impact

Low — requires a $0-balance snapshot preceded by a non-zero month, which is uncommon. But when it fires, the discrepancy between displayed balance ($0) and displayed change (based on live balance) is jarring and could mislead users about their account history.

*[Edited 2026-05-26: redacted concrete dollar amounts, account name, and partial account ID per §2a PII policy]*

claoie:
## Problem

`Balance.tsx` built the per-row Changes widget's `currentAmount` as:

```ts
currentAmount: dynamicAmount || current!,
```

When `dynamicAmount` is `0` — a genuine zero-balance month (closed account, fully liquidated brokerage position, checking account swept to $0) — `||` treats it as falsy and falls through to `current!`, the **live** Plaid/manual balance. The balance row itself renders `$0` (it uses `dynamicAmount` directly), but the Changes delta then compared the unrelated live balance against the previous month, producing a nonsensical change (e.g. `+$(live − prev)` instead of `−$prev`).

## Fix

`dynamicAmount` is always a `number` at this point — line 26's `?? fallback` already resolves it — so the `|| current!` substitution could only ever fire on the `0` edge case, exactly where it does harm. Use `dynamicAmount` directly so the delta correctly reads `−$previous` when an account goes to zero, matching the `$0` the row displays.

## Testing

- The type system guarantees `dynamicAmount: number`, so dropping `|| current!` changes behavior **only** in the `dynamicAmount === 0` case (the bug), and is a no-op for every non-zero month.

This is a **latent** bug: as the issue notes, no current account has the August-or-earlier snapshot needed to flip `shouldShowChanges` true while `dynamicAmount === 0`, so the wrong delta isn't visible in today's data — it would surface after a snapshot backfill or any account that legitimately drops to $0 between two recorded months. Driving the exact path in a browser E2E would require seeding that snapshot shape into the DB; I verified the change at the type/logic level and confirmed the Accounts page renders without regression on this branch (sandbox).

Closes #428

### DIFF
--- a/src/client/components/AccountsTable/Balance.tsx
+++ b/src/client/components/AccountsTable/Balance.tsx
@@ -59,7 +59,12 @@ export const Balance = ({ account }: BalanceProps) => {
   // Now the table sums to the same number the donut shows.
   const shouldShowChanges = !!previousAmount || !!dynamicAmount;
   const changesProps = {
-    currentAmount: dynamicAmount || current!,
+    // Use dynamicAmount directly — it is always a number (line 26's `?? fallback`
+    // already resolved it). The old `|| current!` substituted the live balance
+    // whenever dynamicAmount was 0, which is wrong: $0 is the correct balance for
+    // a zero-balance month (closed account, liquidated position), and the row
+    // itself renders $0, so the Changes delta must compare against 0, not current.
+    currentAmount: dynamicAmount,
     previousAmount,
   };
 


claoie:
E2E (Playwright against PR sandbox, patched build): logged in as admin → Accounts page → **15 Balance widgets + 13 Changes widgets rendered, 0 React pageerrors**. Deltas render correctly for current data. The zero-balance path itself is latent (no account currently has a $0 month preceded by a recorded non-zero month, so `shouldShowChanges` never goes true while `dynamicAmount === 0`), so I verified no rendering regression rather than the latent path — consistent with the fix being type-guaranteed to only change the `dynamicAmount === 0` case.

reviewoie:
## Code review — PR 494

**Verdict:** approve

**Findings:** 0 blockers — 0 high, 0 med, 1 low (informational).

The one-line change `currentAmount: dynamicAmount || current!` → `currentAmount: dynamicAmount` is correct.

Verified:
- `dynamicAmount` is always a number — `balanceHistory.get(viewDateDate)` returns `number | undefined` (`BalanceHistory.get`, `Calculations.ts`) and line 26's `?? fallback` resolves it. So the comment's "always a number" claim holds and the old `|| 0`-style coalesce was never needed for type safety; the `||` only changed behavior when `dynamicAmount === 0`.
- The defect was a real value bug: the row itself renders `numberToCommaString(dynamicAmount)` (Balance.tsx:71 / 81), so a zero-balance month already displayed `$0`, while the Changes widget substituted the live `current!` — delta computed against the wrong base. After the fix the delta compares against the same `0` the row shows. A closed account (prev>0, current=0) now correctly reads `-$prev`.
- ViewDate edge-case rule (Balance.tsx:25-26) is intact and correct: future date with no history → `getAccountBalance(account)` (current balance); past date with no history → `0`. The fix does not touch that branch.
- Consistency across views: grepped client for the `|| current!` Changes pattern — Balance.tsx:62 is the only occurrence. The donut headline (`AccountsDonut/BalanceInfo.tsx`) feeds `Changes` from an aggregated `balanceTotal`, a separate path, so there is no parallel instance of this bug to fix in the same PR.
- `shouldShowChanges` gate (line 60) unaffected — closed-account case (prev>0) still renders the widget.

Scope: single file, single commit, both on-title. Clean.

### Low
- `Balance.tsx` has no unit test, and none is added here. Matches repo norms (no client component-render tests exist; issue 428 itself noted "not visible in current test data, low severity"), so not a blocker — flagging only so the behavior stays covered if a client-component test harness ever lands.

No further blockers.

claoie:
## Reviewoie follow-up

Approve with one informational low — missing unit test, consistent with repo norms.

|---|---|---|---|---|
| 1 | low | `Balance.tsx` | No unit test added for the `dynamicAmount` fix; matches repo norms (no client component-render tests exist). | Leaving as-is — reviewoie explicitly marked this informational ("not a blocker"). Closes #428 itself noted "not visible in current test data, low severity"; client component-render tests have no harness in this repo. Flag will rot naturally if/when a client-render harness lands. |

No new commits — reviewoie's verdict was `approve`. Ping if you want the test added before merge.

hoiekim (approved):

PR #495, opened 2026-06-07
fix(sync): anchor cold/warm history fetch to oldestDate's month start (#434)

### LINKED ISSUE #434: bug(sync): cold-load drops transactions whose updated falls in the same month as oldestTransactionDate
### Symptom

On every load (cold + warm), transactions whose `updated` timestamp falls inside the month that contains `oldestTransactionDate` are **silently dropped** from the client. The user's IndexedDB ends up with a strict subset of the server's data — and there is no UI signal that anything is missing.

Live-verified against the local dev DB (admin user, oldest txn = `2021-07-28`):
- DB: `SELECT COUNT(*) FROM transactions WHERE user_id = admin AND is_deleted IS NOT TRUE` → **6052**
- IndexedDB after Playwright cold load: **6049** (3 short)
- DB: `SELECT COUNT(*) ... AND date < '2021-08-01'` → **9** (July 2021)
- IndexedDB after cold load: **6** July 2021 transactions
- The 3 missing rows all have `updated` between `2021-07-29` and `2021-07-31`.

### Root cause — two bugs compounding

**Bug 1: `sync.ts` slice loop has an off-by-one on `range.from`.**

`src/client/lib/hooks/sync.ts:152` (warm-path) and `:156` (the same shape inside `fetchTransactions`/`fetchSnapshots`) walk months back via:

```ts
while (range.from < viewDate.getStartDate()) {
  // fetch [viewDate.getStartDate(), viewDate.next().getStartDate())
  viewDate.previous();
}
```

`range.from` is set from `oldestDate` (warm: `sync.ts:510`; cold stage-3: `sync.ts:685`), which is `MIN(transactions.date)` — **not** aligned to a month start. When `oldestDate = 2021-07-28`:

| viewDate    | `viewDate.getStartDate()` | `range.from < start` | action     |
|-------------|---------------------------|----------------------|------------|
| Aug 2021    | `2021-08-01`              | `2021-07-28 < 2021-08-01` ✅ | fetch Aug, prev |
| **Jul 2021**| `2021-07-01`              | `2021-07-28 < 2021-07-01` ❌ | **stop**   |

The slice that would cover `start-date=2021-07-01&end-date=2021-08-01` is never requested. Same shape in stage-3 cold-path (`olderFrom`) at `sync.ts:685`.

**Bug 2: `getTransactions` filters by `updated`, not `date`.**

`src/server/lib/postgres/repositories/transactions.ts:46-49` (and `:82-85` for investment txns):

```ts
dateRange:
  options.startDate || options.endDate
    ? { column: UPDATED, start: options.startDate, end: options.endDate }
    : undefined,
```

This was an intentional 2026-02-18 change (commit 3feee63, "fix: filter transactions by updated timestamp instead of date") to make recently-edited old rows appear in fresh slices. Side-effect: every monthly slice now returns rows whose **`updated`** falls in that month, not whose **`date`** falls in that month. Verified live:

```
GET /api/transactions?start-date=2021-08-01&end-date=2021-09-01
→ 49 rows, 6 of which are dated in JULY 2021 (the ones whose `updated` rolled into August)

GET /api/transactions?start-date=2021-07-01&end-date=2021-08-01
→ 3 rows, all dated in July 2021 (the ones whose `updated` was also in July)
```

### How the two bugs compose

The union of slices currently fetched on cold-load is `[Aug 2021, currentMonth+1)` *on the `updated` column*. Any row with `updated ∈ [2021-07-01, 2021-08-01)` is in NO slice the client requests, regardless of its `date`. For admin that's exactly the 3 missing rows.

If oldestDate were month-aligned (e.g. `2021-07-01`), the loop would correctly fetch July 2021 and the missing rows would be picked up. The combination of mid-month oldestDate + filter-by-`updated` is what produces silent data loss.

### Reproduction

```bash
psql -U budget -d budget -h localhost -c "SELECT MIN(date), COUNT(*) FILTER (WHERE date < '2021-08-01' AND (is_deleted IS NULL OR is_deleted = FALSE)), COUNT(*) FILTER (WHERE updated >= '2021-07-01' AND updated < '2021-08-01' AND (is_deleted IS NULL OR is_deleted = FALSE)) FROM transactions WHERE user_id = (SELECT user_id FROM users WHERE username='admin')"
```
→ `min=2021-07-28, July date count = 9, July updated count = 3`

```bash
curl -s -b cookies "http://localhost:3005/api/transactions?start-date=2021-07-01&end-date=2021-08-01" | jq '.body.transactions | length'
```
→ `3` (not 9)

Playwright cold-load → `localforage.getItem(<budget txn store>).length` = **6049** vs DB non-deleted **6052**.

### Fix sketch (one-liner per slice site)

Align `range.from` down to month-start before constructing the range. This makes the loop exit one iteration later for the month that contains `oldestDate`:

```ts
// sync.ts:510 (warm) and :685 (cold stage 3)
const fullFrom = oldestDate && oldestDate < defaultFrom
  ? new ViewDate("month", oldestDate).getStartDate()   // <- align to month start
  : defaultFrom;
```

This stays within the existing loop semantics and doesn't change the cold-path's already-month-aligned `defaultFrom` case. The fix lives entirely client-side; the `updated`-vs-`date` filter on the server can stay as-is (its purpose — surfacing recently-edited old rows in fresh slices — is independent and still load-bearing).

Optional follow-up: consider whether the `updated` filter should also restrict by `date >= oldestDate` so that future loop changes don't silently drop rows whose `updated` is older than the requested window. Out of scope for this fix.

### Adjacent finding (not refiled)

Soft-deleted transactions in an old month can persist in the client forever via `cachedCall`: `cleanCache` runs only on explicit logout/clean, the cached pre-delete response keeps returning the row, and the warm-path "full replace" introduced in commit 09eb6eb means the stale cached old-month slice gets merged into the dictionary on every subsequent warm load. Distinct from this bug (different slice, different lifecycle); flagging for separate triage if it becomes user-visible.

claoie:
Closes #434.

## Problem

On every load (cold + warm), transactions whose `updated` timestamp falls inside the month containing `oldestTransactionDate` were **silently dropped** from the client — the IndexedDB store ended up a strict subset of the server's data, with no UI signal.

Two bugs compound:
1. The history slice loop (`fetchTransactions`/`fetchSnapshots`) walks months back with a strict `while (range.from < viewDate.getStartDate())`. `range.from` was set to the raw `oldestDate` (`MIN(transactions.date)`), which is **not** month-aligned. For `oldestDate = 2021-07-28`, the July slice `[2021-07-01, 2021-08-01)` is never requested because `2021-07-28 < 2021-07-01` is false.
2. The server filters slices by `updated`, not `date` (intentional, commit 3feee63). So any row whose `updated` lands in the never-requested July window is in no slice the client asks for.

## Fix

Anchor `from` to the **start of the month before** `oldestDate`'s month, at both slice sites in `sync.ts` (warm `fullFrom`, cold stage-3 `olderFrom`). Because the loop boundary is a strict `<` against month-starts, anchoring to the *previous* month-start is what makes the loop's final iteration cover `oldestDate`'s own month.

> Note: simply month-aligning `oldestDate` down to its own month-start (`2021-07-01`) does **not** fix it — `2021-07-01 < 2021-07-01` is still false. I simulated the loop against the real `ViewDate` to confirm; only anchoring to the previous month-start includes the oldest month. The server `updated`-vs-`date` filter is left as-is (its purpose — surfacing recently-edited old rows in fresh slices — is independent and still load-bearing).

Client-side only, ~2 lines of logic + comments. No schema/API change.

## Test plan

- [x] `bun run lint` — clean
- [x] `bun test` — **720 pass / 0 fail**
- [x] Loop simulation against the real `common/utils/date` `ViewDate`: buggy + naive-month-align both drop the oldest month for mid-month `oldestDate`; previous-month-anchor includes it across mid-month / month-aligned / recent cases.
- [x] **UI E2E (Playwright, prod build served by `bun build/server/bundle.js` on local pg, admin/budget, real data — admin oldest txn `2021-07-28`, DB non-deleted total **6052**, 3 rows with `updated` in July 2021):**
  - **Fix build** — fresh context (cold load) → after sync settles, `BudgetApp` IndexedDB `transactions` store count = **6052** (0 missing). The 3 previously-dropped rows (`Kq6gY6…`, `k1DkOD…`, `AODkXD…`, all `updated 2021-07-29`) are all present (`get(id)` → found).
  - **Baseline `upstream/main` build** (same DB, port 3435) — same cold load → IndexedDB count = **6049** (3 short); the 3 rows absent. Confirms the fix is what closes the gap.
  - `/transactions` renders without error on the fix build (screenshot `/tmp/pr-434-transactions.png`).

### DIFF
--- a/src/client/lib/hooks/sync.ts
+++ b/src/client/lib/hooks/sync.ts
@@ -667,7 +667,21 @@ export const useSync = () => {
         const oldestDatePromise = getOldestTransactionDate();
         const oldestDate = await oldestDatePromise;
         const defaultFrom = currentViewDate.clone().previous().previous().previous().getStartDate();
-        const fullFrom = oldestDate && oldestDate < defaultFrom ? oldestDate : defaultFrom;
+        // The slice loop fetches a month M only when `from < M.getStartDate()`
+        // (strict). A raw mid-month `oldestDate` (e.g. 2021-07-28) is NOT < its
+        // own month start (2021-07-01), so that month is never requested and
+        // every row whose `updated` falls in it is silently dropped. Anchor to
+        // the start of the month *before* oldestDate's month so the loop's last
+        // iteration covers oldestDate's month, then compare the *anchored* value
+        // against defaultFrom — comparing the raw oldestDate would still drop the
+        // month when oldestDate sits in defaultFrom's own month past the 1st. (#434)
+        const oldestMonthStart = oldestDate
+          ? new ViewDate("month", oldestDate).previous().getStartDate()
+          : null;
+        const fullFrom =
+          oldestMonthStart && oldestMonthStart < defaultFrom
+            ? oldestMonthStart
+            : defaultFrom;
         const fullUntil = currentViewDate.clone().next().getStartDate();
         const fullRange: FetchRange = { from: fullFrom, until: fullUntil };
         // Freshness window: 30 days back from the last sync. Anything
@@ -872,7 +886,18 @@ export const useSync = () => {
       // tile cleanly with no overlap and no gap.
       const oldestDate = await oldestDatePromise;
       const defaultFrom = currentViewDate.clone().previous().previous().previous().getStartDate();
-      const olderFrom = oldestDate && oldestDate < defaultFrom ? oldestDate : defaultFrom;
+      // Anchor to the month before oldestDate's month — see the warm-path note
+      // above: the slice loop's strict `from < M.getStartDate()` drops the
+      // month containing a mid-month oldestDate otherwise. Compare the anchored
+      // value (not raw oldestDate) against defaultFrom so the same-calendar-month
+      // boundary is covered too. (#434)
+      const oldestMonthStart = oldestDate
+        ? new ViewDate("month", oldestDate).previous().getStartDate()
+        : null;
+      const olderFrom =
+        oldestMonthStart && oldestMonthStart < defaultFrom
+          ? oldestMonthStart
+          : defaultFrom;
       const olderUntil = currentViewDate.clone().previous().getStartDate();
 
       const olderRange: FetchRange = { from: olderFrom, until: olderUntil };


reviewoie:
## Code review — PR 495

The core fix is correct. Traced `ViewDate` math (all local-TZ via `getFullYear`/`getMonth` + `new Date(y,m,d)`; no UTC mismatch) and the slice loop `while (range.from < viewDate.getStartDate())`. For a mid-month `oldestDate` (e.g. `2021-07-28`), `new ViewDate("month", oldestDate).previous().getStartDate()` yields `2021-06-01`, so the loop's `2021-06-01 < 2021-07-01` is true → July is fetched, and `2021-06-01 < 2021-06-01` is false → June is NOT over-fetched. Empirically verified: anchored fullFrom covers down to and including oldestDate's month with no extra month. Warm and cold paths are anchored identically — consistent.

### Med
- `src/client/lib/hooks/sync.ts:575,779` (`fullFrom` / `olderFrom` guard) — the fix is incomplete at one month boundary: when `oldestDate` falls in the **same calendar month as `defaultFrom`** but later than the 1st, oldestDate's month is still dropped. The guard `oldestDate && oldestDate < defaultFrom` compares the *raw* mid-month `oldestDate` against the month-start `defaultFrom`. Failure scenario: current=June, `defaultFrom = 2024-03-01`, `oldestDate = 2024-03-20`. Guard `2024-03-20 < 2024-03-01` is false → the un-anchored `defaultFrom = 2024-03-01` branch is taken → loop `2024-03-01 < 2024-03-01` is false → March is never requested → every March transaction is silently dropped. This is the exact issue #434 symptom, surviving at the boundary the PR didn't anchor. Reachable for any user whose oldest transaction is 2–3 months old and not dated the 1st. Fix: anchor in both branches, e.g. choose the floor as the earlier of `defaultFrom` and `new ViewDate("month", oldestDate).previous().getStartDate()` (or anchor the guard comparison to month starts). Verified via ViewDate-driven slice simulation.

### Low
- `src/client/lib/hooks/sync.ts` — no regression test accompanies an off-by-one month-boundary fix. `sync.ts` has no test file (current convention is `bun test --preload ./scripts/test-preload.ts src`; the retired per-test-bundle framework is gone, so no bundle obligation). The month-anchoring boundary is exactly the kind of logic a small unit test would pin — extracting the `from`-anchor computation into a pure helper and testing the mid-month / same-month-as-defaultFrom / on-the-1st cases would also catch the Med finding above. Not a blocker (sync.ts has never had coverage), but flagged.

No further blockers.

claoie:
Pushed `3db1a18` addressing the reviewoie med finding (month-boundary gap).

**The gap:** the anchor only engaged when the raw `oldestDate` was strictly `< defaultFrom`. When `oldestDate` sits inside `defaultFrom`'s own calendar month past the 1st — or exactly on the 1st — that guard was false, the un-anchored `defaultFrom` branch ran, and the strict slice loop (`from < M.getStartDate()`) still dropped `oldestDate`'s month. That's the same 434 row-drop, surviving at the boundary.

**The fix:** compute the anchored month-start first, then compare *that* against `defaultFrom` (in both the warm and cold history-fetch paths):

```ts
const oldestMonthStart = oldestDate
  ? new ViewDate("month", oldestDate).previous().getStartDate()
  : null;
const fullFrom =
  oldestMonthStart && oldestMonthStart < defaultFrom ? oldestMonthStart : defaultFrom;
```

**Verified against the real `ViewDate`** (defaultFrom = 2024-03-01):

| oldestDate | old fullFrom | covers oldest month? | new fullFrom | covers? |
|---|---|---|---|---|
| 2024-03-20 (same month, mid) | 2024-03-01 | no | 2024-02-01 | yes |
| 2024-03-01 (exact 1st) | 2024-03-01 | no | 2024-02-01 | yes |
| 2024-05-20 (recent) | 2024-03-01 | yes | 2024-03-01 | yes (no over-fetch) |
| 2021-07-28 (old) | 2021-06-01 | yes | 2021-06-01 | yes |

Typecheck clean. No `sync.ts` test harness exists (per repo convention), so this is verified by the `ViewDate` simulation above rather than a unit test; extracting the anchor into a pure helper for direct coverage is a reasonable follow-up but kept out here to hold scope.

reviewoie:
## Code review — PR 495

**Findings:** 3 total — 1 high, 0 med, 2 low.

Re-review scoped to the new commit `3db1a18` (pushed after the prior reviewoie review on `d9d8bc7`). The sync.ts fix itself is correct — but `3db1a18` bundled in a large, undocumented, out-of-scope deletion.

### High
- `3db1a18` deletes the entire `rejected_categories` subsystem (~490 lines, 11 files) — a hard scope violation that breaks the build on merge into current main. The commit titled "fix(sync): compare anchored oldest-month against defaultFrom" actually carries, beyond the 22-line sync.ts change, deletions of `src/common/models/RejectedCategory.ts`, `src/server/lib/postgres/repositories/rejected-categories.ts`, `.../models/rejected-category.ts`, their tests, the `rejectedCategoriesTable` registration in `initialize.ts`, the `REJECTED_CATEGORIES` constant in `common.ts`, and 37 lines from `base.ts`. None of this relates to the #434 sync fix, and the commit message documents none of it.

  Root cause: the branch was cut at f2ef6fb (merged PR #501) and is now behind `b8fa26c` (#502), `b596d29` (#506), `5129a14` (#507) — all of which built on `rejected_categories`. Commit `3db1a18` re-deletes files that those merged PRs kept and extended.

  Failure scenario: on merge/rebase into current main, the consumers added by #502/#506 — `src/server/lib/record-category-rejection.ts` (imports `addRejectedCategory`/`removeRejectedCategory`), `auto-suggest.ts` (joins `rejected_categories`), `sync-plaid.ts`, `post-transaction.ts`, `initialize.ts` — lose their providers → unresolved-import build break, and the just-shipped rejection-signal feature is wiped. CI is green only because the PR branch predates those consumers, so the branch compiles in isolation; the green check masks the merge-time break. Verified: `git show --stat 3db1a18`; files present on both f2ef6fb and `upstream/main` but deleted on `pr495`; `git grep` on `upstream/main` shows active imports of every deleted symbol; `git merge-base --is-ancestor` confirms #502/#506/#507 are not in the branch.

  Required action: drop every deletion from this PR. Rebuild the branch on current `upstream/main` so it contains ONLY the `src/client/lib/hooks/sync.ts` change.

### Low
- `src/client/lib/hooks/sync.ts:577-588, 783-794` — the anchor `new ViewDate("month", oldestDate).previous().getStartDate()` resolves to the start of the month *before* oldestDate's month, so the slice loop always fetches one extra empty month below oldestDate. Harmless (one no-op month request) and matches the stated intent, but `new ViewDate("month", oldestDate).getStartDate()` would cover oldestDate's own month exactly with no over-fetch. Verified via simulation against the real `ViewDate` — all four boundary cases (mid-month, in-defaultFrom-month, on-the-1st, recent) cover oldestDate's month correctly.
- `src/client/lib/hooks/sync.ts` — no test covers the new `oldestMonthStart` anchor logic; sync.ts has no test file. The boundary cases this commit fixes (oldestDate on/after the 1st of defaultFrom's month) are exactly the kind a unit test should pin. I verified the logic empirically by simulating against `src/common/utils/date.ts`, but a checked-in test would prevent regression.

No further blockers beyond the High once the deletions are removed — the sync.ts fix is correct.

claoie:
Rebuilt the branch on `upstream/main` to drop an out-of-scope change reviewoie flagged.

**What was wrong:** the second commit (`3db1a18`) bundled the legitimate sync.ts fix with a deletion of the entire `rejected_categories` subsystem (11 files, ~490 lines). Those files are live on `upstream/main` and have consumers in merged code, so the deletion was both out of scope for this #434 fix and would have broken the build on merge. CI was green only because the branch predated those consumers and compiled in isolation.

**Fix:** branched fresh from `upstream/main` (5129a14) and applied **only** the sync.ts change. `rejected_categories` is untouched. Final diff is a single file:

```
src/client/lib/hooks/sync.ts | 29 +++++++++++++++++++++++++++--
1 file changed, 27 insertions(+), 2 deletions(-)
```

`tsc --noEmit` passes (exit 0). The fix itself is unchanged: anchor the cold/warm history fetch `from` to the month before oldestDate's month, and compare the anchored value (not the raw date) against defaultFrom, so the strict slice loop no longer drops the month containing a mid-month oldestDate.

Force-pushed `3db1a18` → `2fdef34`.

reviewoie:
## Code review — PR 495

**Verdict:** approve

Re-review at current HEAD `2fdef34` (the branch rebuilt fresh on `upstream/main`, replacing the bad-rebase commit `3db1a18` that the prior reviewoie review flagged).

### Prior HIGH — RESOLVED
The previous review's HIGH blocker (`3db1a18` bundled a ~490-line, 11-file deletion of the entire `rejected_categories` subsystem alongside the sync fix) is gone. The diff at `2fdef34` touches exactly one file — `src/client/lib/hooks/sync.ts` (+27 / -2) — and deletes nothing. Verified: `git diff upstream/main...2fdef34 --name-only` returns only `sync.ts`; `grep -i reject` on the file list is empty. Scope is clean.

### Correctness — VERIFIED CORRECT
The month-start anchoring is right for both the warm path (lines 583-589) and cold path (lines 789-795), and the two blocks are logically identical (only `until` differs by design — cold stages the recent window separately).

Verified empirically by simulating the slice loop (`while (range.from < viewDate.getStartDate())`, strict) against the real `ViewDate` for `oldestDate = 2021-07-28`:
- old `from = oldestDate` (raw mid-month) → earliest month fetched = 2021-08 (July dropped — the #434 bug)
- new `from = new ViewDate("month", oldestDate).previous().getStartDate()` = 2021-06-01 → earliest month fetched = 2021-07 (covered)

Note: the `.previous()` (month-*before*) anchor is load-bearing, not an off-by-one. Because the loop comparison is strict, anchoring to oldestDate's *own* month start (2021-07-01) leaves `2021-07-01 < 2021-07-01` false and re-drops July. The prior review's low suggestion to use own-month-start would reintroduce the bug — this PR's `.previous()` form is correct. (Retracting that earlier suggestion.)

The anchored-vs-`defaultFrom` comparison is also load-bearing: for `oldestDate` sitting inside `defaultFrom`'s own month past the 1st (e.g. defaultFrom = 2026-03-01, oldestDate = 2026-03-15), the raw compare picks `defaultFrom` and drops March; the anchored compare (2026-02-01 < 2026-03-01) picks the anchor and covers March. Verified by simulation.

CI green at `2fdef34` (build + test pass).

### Low
- `src/client/lib/hooks/sync.ts` — no test pins the new anchor logic; `sync.ts` has no test file (it is a browser/IndexedDB-coupled hook, and the only sibling tests live under `hooks/calculation/` for pure functions, so there is no harness to attach to). I verified the boundary cases empirically against the real `ViewDate`, but a checked-in test would guard against regression. Not a blocker.

No further blockers — the sync fix is correct and scope is clean.

claoie:
## Reviewoie follow-up

Approve at re-review HEAD `2fdef34` — prior HIGH (scope bundling) resolved by the fresh rebase; one informational low on missing regression test.

|---|---|---|---|---|
| 1 | low | `src/client/lib/hooks/sync.ts` | No test pins the new `.previous().getStartDate()` anchor logic; `sync.ts` has no test file (browser/IndexedDB-coupled hook, no harness). | Leaving as-is — reviewoie explicitly flagged "not a blocker" and noted the boundary cases were empirically verified against the real `ViewDate`. The hook is unit-untestable in the current repo layout (sibling hook tests live under `hooks/calculation/` for pure functions only). A regression test would need a new harness, which is out of scope for this fix. |

No new commits — reviewoie's verdict was `approve`. Ping if you want a test harness landed before merge.

claoie:
Closing as obsolete — superseded by the delta-by-cursor sync rewrite (#536, #537).

This PR fixed a bug in the old per-month-walk: `range.from = oldestDate` was not month-aligned, so the loop's strict `from < monthStart` check skipped the oldest transaction's own month. That whole code path (`getOldestTransactionDate`, `fullFrom`/`defaultFrom`, per-month `ViewDate` loop) was removed in #536.

The new design fetches the entire history via a single unbounded `GET /api/transactions` on cold load (no `start-date` filter), so the mid-month-anchoring bug is structurally impossible. Issue #434 will be closed alongside.

PR #496, opened 2026-06-07
feat(suggestions): introduce `suggestions` table + backfill (#333)

claoie:
## Supersedes #482 — closes part of #333

Per the design conversation on Discord: rather than `transactions JOIN labels` on the read path, keep `transactions.label_*` as the denormalized read cache and introduce a separate `suggestions` event log the engine learns from. Pending→posted reconciliation punted to a follow-up PR.

This PR adds the dormant infrastructure. **No route, engine, or FE code reads from or writes to the table yet** — that's Stage 2.

## Final design (locked with Hoie)

```sql
CREATE TABLE suggestions (
  transaction_id    VARCHAR(255) FK transactions ON DELETE CASCADE NOT NULL,
  user_id           UUID FK users ON DELETE CASCADE NOT NULL,
  category_id       UUID FK categories ON DELETE CASCADE NOT NULL,
  confidence        FLOAT NOT NULL,
  is_confirmed      BOOLEAN NOT NULL DEFAULT FALSE,
  is_rejected       BOOLEAN NOT NULL DEFAULT FALSE,
  confirmed_at      TIMESTAMPTZ,
  engine_scored_at  TIMESTAMPTZ,
  updated           TIMESTAMPTZ DEFAULT CURRENT_TIMESTAMP,
  PRIMARY KEY (transaction_id, category_id)
);
CREATE INDEX ON suggestions (user_id);
-- transaction_id lookups are served by the composite-PK left-prefix, so no
-- separate (transaction_id) index is created.
CHECK (NOT (is_confirmed AND is_rejected));
```

Why this shape (vs the earlier confidence-as-state-machine design):
- **Two explicit user-action flags** instead of overloading `confidence`. The merchant signal reads `is_confirmed`/`is_rejected` directly — no inference from numbers, no "phantom rejection" problem where `confidence = 0` could mean "engine retracted" vs "user rejected."
- **Two timestamps** so a future recency filter can downweight stale confirmations / re-trigger stale engine scores.
- **Composite PK** — `(transaction_id, category_id)` is the natural key; no surrogate UUID needed.

## Mechanism (Stage 2 wiring)

| When | Operation |
|---|---|
| User confirms category C | UPSERT `is_confirmed=TRUE, is_rejected=FALSE, confirmed_at=NOW()` |
| User rejects category C | UPSERT `is_rejected=TRUE, is_confirmed=FALSE` |
| Engine suggests category C at score s | UPSERT `confidence=s, engine_scored_at=NOW()` with `ON CONFLICT … WHERE NOT is_confirmed AND NOT is_rejected` so engine never clobbers a user-actioned row |
| User picks B over A (with engine score on A) | A's engine row is untouched (no phantom rejection); B gets `is_confirmed=TRUE` |

## Repository helpers

- `getSuggestionsForTransaction(user, tx)` — read all suggestion rows for a transaction
- `getSuggestionsForTransactions(user, tx_ids)` — bulk read (for `getMerchantSignal`)
- `upsertUserConfirmedSuggestion(user, tx, cat)` — sets the confirmed flag + timestamp
- `upsertUserRejectedSuggestion(user, tx, cat)` — sets the rejected flag
- `upsertEngineSuggestion(user, tx, cat, conf)` — writes engine state under the no-user-actioned guard
- `deleteAllSuggestionsForTransaction(user, tx)` — cascades cleanup

## Backfill (manual, idempotent)

Run on demand via `scripts/backfill-suggestions.ts` — **not** wired into `initializePostgres` (per @hoiekim: one-time migrations run separately/manually, so a deploy never mass-writes on container start). Idempotent via `ON CONFLICT DO NOTHING`.

Confirmation mirrors the live `confidence === 1` invariant (`budgets.ts`: `isConfirmed = category_confidence === 1 && !!category_id`). A legacy row is migrated as a **user confirmation only when `label_category_confidence = 1` or `IS NULL`**; a `0 < confidence < 1` row is an unreviewed engine suggestion and migrates as engine-owned, keeping `is_confirmed` a pure record of user intent:

```sql
INSERT INTO suggestions
  (transaction_id, user_id, category_id, confidence,
   is_confirmed, confirmed_at, engine_scored_at)
SELECT
  transaction_id, user_id, label_category_id,
  COALESCE(label_category_confidence, 1.0),
  (label_category_confidence IS NULL OR label_category_confidence = 1.0),         -- is_confirmed
  CASE WHEN label_category_confidence IS NULL OR label_category_confidence = 1.0
       THEN updated END,                                                          -- confirmed_at
  CASE WHEN label_category_confidence IS NULL OR label_category_confidence = 1.0
       THEN NULL ELSE updated END                                                 -- engine_scored_at
FROM transactions
WHERE (is_deleted IS NULL OR is_deleted = FALSE) AND label_category_id IS NOT NULL
ON CONFLICT (transaction_id, category_id) DO NOTHING;
```

## Sandbox verification (prod-clone, read-only + rolled-back INSERT)

Ran the exact backfill INSERT inside a transaction against a prod-clone, then `ROLLBACK` (sandbox unchanged):
- **6114 eligible** transactions (`label_category_id NOT NULL`, not soft-deleted).
- **5947 → `is_confirmed=TRUE`** (confidence = 1.0) — each with `confirmed_at` set, `engine_scored_at` NULL.
- **167 → `is_confirmed=FALSE`** (fractional `0 < confidence < 1`, unreviewed engine suggestions) — each with `engine_scored_at` set, `confirmed_at` NULL.
- **0** flag/timestamp invariant violations; **0** rows with `engine_scored_at` on a confirmed row or vice-versa.

(Those 167 rows are exactly what the earlier blanket-`is_confirmed=TRUE` backfill would have mislabeled as user confirmations.)

## What's out of scope

- No FE / route / engine changes
- Pending→posted transaction_id reconciliation (separate follow-up)
- Drop of legacy `transactions.label_*` columns (those stay as denorm read cache)

## Test plan

- [x] `bun test src/server/lib/postgres/repositories/suggestions.test.ts`
- [x] `bun test src/server/lib/postgres/backfill-suggestions.test.ts` — 7 pass (added the `confidence===1` invariant case)
- [x] Full server suite: 574 pass / 0 fail
- [x] `bunx tsc --noEmit` clean
- [x] Prod-clone sandbox: backfill INSERT produces 5947 confirmed / 167 unreviewed, 0 invariant violations (rolled back)


src/server/lib/postgres/initialize.ts
```diff
@@ -80,6 +83,11 @@ export const initializePostgres = async (): Promise<void> => {
 
     // Run automatic schema migrations to add any missing columns
     await runMigrations(tables.map((t) => ({ name: t.name, schema: t.schema })));
+
+    // Idempotent backfill from `transactions.label_*` into the `suggestions`
+    // table. Runs every startup; ON CONFLICT (transaction_id, category_id)
+    // DO NOTHING makes it cheap after the first pass.
+    await backfillSuggestionsFromLegacyColumns();
```
hoiekim: Don't embed this in init script. One time migration should happen separately manually

src/server/lib/postgres/repositories/suggestions.ts
```diff
@@ -0,0 +1,176 @@
...
+  TRANSACTION_ID,
+  USER_ID,
+  CATEGORY_ID,
+  CONFIDENCE,
+  IS_CONFIRMED,
+  IS_REJECTED,
+  CONFIRMED_AT,
+  ENGINE_SCORED_AT,
+  UPDATED,
+} from "../models";
```
hoiekim, line 15: Why don't you use barrel imports? This breaks convention / consistency

hoiekim (changes requested):

src/server/lib/postgres/initialize.ts
claoie: Done. The idempotent `backfillSuggestionsFromLegacyColumns()` call was removed from `initialize.ts` — it no longer runs on every startup. It now lives in a standalone manual script, `scripts/backfill-suggestions.ts`, run on demand (`bun scripts/backfill-suggestions.ts`). `initialize.ts` only registers `suggestionsTable` in the schema `tables[]` array now (table creation, same as every other table) — no data migration embedded.

src/server/lib/postgres/repositories/suggestions.ts
claoie: Fixed in 867fb4c. `suggestions.ts` now imports models from `../models` and `pool` from `../client`, matching all 16 sibling repositories — it was the lone file pulling from the top-level `server` barrel. (That also sidesteps a circular edge, since the `server` barrel re-exports the repositories.) tsc clean, full server suite 573 pass / 0 fail.

claoie:
Both review points addressed (pushed 867fb4c):

1. **No migration in init script** — the idempotent backfill was removed from `initialize.ts` (it had run `backfillSuggestionsFromLegacyColumns()` on every startup). It's now a standalone manual script: `scripts/backfill-suggestions.ts`. `initialize.ts` only registers `suggestionsTable` for schema creation.

2. **Barrel imports / convention** — `repositories/suggestions.ts` now imports models from `../models` and `pool` from `../client`, matching all 16 sibling repositories. It had been the only file importing from the top-level `server` barrel (which also risked a circular edge, since that barrel re-exports the repositories).

`tsc --noEmit` clean; full `bun test src/server` → **573 pass / 0 fail**. Ready for another look.

claoie:
### E2E verification + one correctness fix (bcb5f96)

While verifying the backfill against a prod-clone I caught a semantics bug and fixed it in this PR:

**The backfill stamped *every* migrated row `is_confirmed=TRUE`** — including rows with `0 < label_category_confidence < 1`, which are unreviewed engine suggestions, not user confirmations. `budgets.ts` classifies `category_id` set with `0 < confidence < 1` as "auto-suggested but unreviewed" (`isConfirmed = category_confidence === 1 && !!category_id`), and the suggestions schema's stated purpose is for `is_confirmed` to record *user intent* rather than infer it from confidence. Blanket-TRUE would have inflated the merchant-signal confirm rate with pending guesses.

Now: `confidence = 1` / `NULL` → `is_confirmed=TRUE` + `confirmed_at`; `0 < confidence < 1` → `is_confirmed=FALSE` + `engine_scored_at`.

**E2E (prod-clone, INSERT run inside a transaction then ROLLBACK — sandbox untouched, no PII):**

| | rows | confirmed_at | engine_scored_at |
|---|---|---|---|
| `is_confirmed=TRUE` (confidence=1.0) | 5947 | 5947 | 0 |
| `is_confirmed=FALSE` (fractional) | 167 | 0 | 167 |

6114 eligible total, **0** flag/timestamp invariant violations. The 167 rows are exactly what the old blanket-TRUE logic mislabeled.

Also corrected: stale doc/PR-body claims that the backfill "runs on startup" (it's manual-only), and a PR-body `CREATE INDEX ON (transaction_id)` that the shipped model deliberately omits (composite-PK left-prefix covers it).

`tsc` clean, full server suite **574 pass / 0 fail**, backfill suite **7 pass** (added the `confidence===1` invariant test).

hoiekim (approved):

PR #497, opened 2026-06-08
fix(capacities): dedupe duplicate active_from rows on write + one-time DB cleanup

claoie:
## Bug

Read-only scan of the prod-clone DB:

- **3 budgets** have multiple capacities with the same `active_from`
- **12 sections** (59 total duplicate rows)
- **16 categories** (87 total duplicate rows)

Budget `53c90b` is the worst case — 7 capacities all with `active_from = NULL`, each carrying a different `month` value. The FE display dedupes via `getAllCapaciyDates` (showing one "All past" row), but the data on disk is polluted: every server-side traversal that reads `capacities` walks all 7 zombies, and `BudgetFamily.getActiveCapacity` picks one of them non-deterministically (V8 stable sort + `.find` on `active_from <= date`).

Surfaced during the investigation of #439's "sync with children" question — `Σ section.month` and `Σ category.month` reported wildly different totals for the same budget across what should be the same logical period.

## Fix

Three layers:

1. **Pure helper** `src/common/utils/dedupeCapacities.ts` — keeps the FIRST occurrence in input order. That choice matches what the user is currently seeing in the UI (`getActiveCapacity` returns the first row in the stable-sorted-desc copy for same-active_from ties), so dedupe never silently swaps the displayed value to one the user never saw.

2. **Apply on every write** in `src/server/lib/postgres/repositories/budgets.ts` — `createBudget` / `updateBudget` / `createSection` / `updateSection` / `createCategory` / `updateCategory` run `dedupeCapacities` before `JSON.stringify`. Defends against future bug paths that would otherwise re-create duplicates.

3. **One-time DB cleanup** `cleanup-capacity-duplicates.ts`, wired into `initializePostgres` after `runMigrations`. Idempotent UPDATE per table; after the first run the WHERE sub-query returns empty and it's a no-op.

## Test plan

- [x] 7 unit tests on `dedupeCapacities` covering empty / single-row passthrough, same-active_from dedupe, NULL/undefined collapse, distinct-date preservation, Date-instance vs ISO-string equivalence, non-mutation, mixed inputs
- [x] Full suite 727/727 pass
- [x] `bun run lint` clean
- [ ] Sandbox spin: confirm cleanup runs on startup + collapses duplicates from the 3 budgets / 12 sections / 16 categories observed in prod-clone (will append result here)

## Out of scope

- Does NOT fix the stale-parent-`month` issue (that's #439's job — derive sum-of-children on read instead of persisting it).
- Does NOT fix the MAX_FLOAT propagation issue (sections marked `±MAX_FLOAT` while parent budget shows finite). #439's `Capacity.getActiveAmount` with the "poisons the sum" rule addresses that on the read side.

Related: #439, #433 (sandbox testing trail).

### DIFF
--- a/src/common/utils/dedupeCapacities.test.ts
+++ b/src/common/utils/dedupeCapacities.test.ts
@@ -1,0 +1,69 @@
+import { describe, expect, test } from "bun:test";
+import { dedupeCapacities } from "./dedupeCapacities";
+import type { JSONCapacity } from "../models/BudgetFamily";
+
+const cap = (active_from?: Date | null, month = 0) =>
+  ({ capacity_id: "x", active_from, month, year: month * 12, isInfinite: false, isIncome: false }) as unknown as JSONCapacity;
+
+describe("dedupeCapacities", () => {
+  test("passes through empty / single-row arrays unchanged", () => {
+    expect(dedupeCapacities([])).toEqual([]);
+    const one = [cap(new Date("2026-07-01"), 100)];
+    expect(dedupeCapacities(one)).toBe(one);
+  });
+
+  test("keeps first occurrence in input order when active_from is the same", () => {
+    const a = cap(new Date("2026-07-01"), 100);
+    const b = cap(new Date("2026-07-01"), 200);
+    const result = dedupeCapacities([a, b]);
+    expect(result).toHaveLength(1);
+    expect(result[0]).toBe(a);
+    expect(result[0].month).toBe(100);
+  });
+
+  test("collapses multiple NULL/undefined active_from rows into one", () => {
+    const a = cap(undefined, 100);
+    const b = cap(null, 200);
+    const c = cap(undefined, 300);
+    const result = dedupeCapacities([a, b, c]);
+    expect(result).toHaveLength(1);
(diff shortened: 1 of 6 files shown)


claoie:
Closing — the duplicate-`active_from` rows were a sandbox **scrambler artifact**, not a real prod issue. Verified directly against the moltboie read-only role: 0 duplicates across budgets/sections/categories in prod. Real cause was `~/claude/sandbox/scrambler.budget.sql` rebuilding capacity entries via `jsonb_build_object('capacity_id', ..., 'month', ...)` and silently dropping the `active_from` field, collapsing every scrambled capacity to NULL.

Fixed the scrambler locally (`elem || jsonb_build_object('month', ...)` pattern preserves all original fields). Fresh sandbox spin confirms budget 53c90b now has 7 distinct active_from values matching prod's shape; duplicates count is 0/0/0.

The dedupe-on-write helper + cleanup migration in this PR would be a no-op on real prod. Closing as not-applicable. Tracked in [Docmost design doc](https://doc.hoie.kim/s/claoie/p/TyXsel00Eb).

PR #500, opened 2026-06-09
feat(suggestions): record user confirm/reject into the suggestions log (#333)

claoie:
## Stage 2a of #333 — record the user signal

PR #496 introduced the `suggestions` event log as **dormant infrastructure** ("No route, engine, or FE code reads from or writes to the table yet — that's Stage 2"). This PR is the **write half** of Stage 2: the transaction route now records every user category action into the log.

### The bug, restated
`getMerchantSignal` computes `rejected` from `transactions` rows with `label_category_confidence = 0`, but a reject **clears** `label_category_id` to null, so rejected rows fall into the NULL bucket and are dropped — `rejected` is always 0. Hoie's locked design (on #333 / #496) is a separate `suggestions` log keyed on `(transaction_id, category_id)` with explicit `is_confirmed` / `is_rejected` flags, so user intent is recorded directly instead of inferred from `confidence`.

### What this PR does
- **`recordSuggestionFeedback`** (new route-boundary helper) maps a label update to one suggestion-log write:
  - `category_confidence === 1` → `upsertUserConfirmedSuggestion`. The confirmed category is the body's pick, or — for **accept-in-place** (yellow dot), where the body carries only `{ category_confidence: 1 }` — the category already on the row.
  - `category_confidence === 0` → `upsertUserRejectedSuggestion`. The body's `category_id` is null on a clear, so the rejected category is read from the row's **pre-update** state.
- **Budget-switch disambiguation.** `onChangeBudgetSelect` also clears the category and posts `category_confidence: 0`, but switching budgets is *not* a category rejection (it must not feed negative merchant signal — see the #333 discussion). The helper treats a confidence-0 update that changes an **already-set** budget as a switch, not a rejection. It errs toward *not* recording a false rejection.
- **`post-transaction` route** reads the existing row before the update (the rejected/accept-in-place category can only be read pre-clear) and calls the helper after a successful write. Single choke point — every FE confirm/reject path flows through it (explicit pick, accept-in-place, clear-to-reject, Accept-All).

### Scope
- **Transactions only.** The `suggestions` table FKs `transaction_id` to `transactions`; split and investment transactions live in other tables and cannot have suggestion rows by design. Deferred accordingly.
- **No engine read change.** `getMerchantSignal` is untouched, so suggestion *output* is identical to `upstream/main` — zero money-path regression risk. This PR only adds rows to the log.

### Follow-ups (Stage 2b, next run)
- Switch `getMerchantSignal` to source confirm/reject counts from `suggestions` (the change that actually closes the user-visible "rejected always 0" symptom). It changes engine behavior on real data and deserves its own focused, heavily-E2E'd PR.
- Have `applyLabel` write `upsertEngineSuggestion` so engine scores live in the log too.

Part of #333 (does not close it — Stage 2b does).

## Test plan

- [x] `bun run lint` — clean
- [x] `bun test src/server` — 583 pass / 0 fail
- [x] new `record-suggestion-feedback.test.ts` — 9 pass (confirm-pick, accept-in-place, clear-reject, no-previous no-op, budget-switch-not-reject, budget-unchanged-reject, engine-fractional no-op, no-confidence no-op, undefined-label no-op)
- [x] **UI E2E (Playwright, sandbox port 20500, logged in admin, `/transactions` Yearly view)** — drove the real budget/category `<select>`s and asserted against the `suggestions` table in the DB the server actually used:
  - **Confirm** — picked a different category on a row → POST `/api/transaction {category_id, category_confidence:1}` → DB: `suggestions(tx, category)` with `is_confirmed=true, confidence=1`. ✓
  - **Reject** — cleared a row's category (→ "Select Category") → POST `{category_id:null, category_confidence:0}` → DB: `suggestions(tx, previous_category)` with `is_rejected=true, confidence=0`. ✓
  - **Budget switch (disambiguation)** — changed a categorized row's *budget* (clears the category, also posts `category_confidence:0`) → POST `{budget_id:new, category_id:null, category_confidence:0}` → DB: **no rejection row written** for that transaction. ✓ — confirms a budget switch is not mistaken for a category rejection.
  - Screenshot: `/tmp/pr-500-year.png` (transactions year view with the selects driven).

> ⚠️ Verification note: the budget sandbox server inherited `POSTGRES_DATABASE=budget` from the spawn environment, overriding the per-sandbox clone (`budget_sandbox_20500`) — the same isolation leak flagged for the inbox sandbox on 2026-06-08, now confirmed for budget. The E2E assertions are valid (they were checked against the DB the running server actually read/wrote), but the run mutated the local `budget` DB; the 3 touched transaction labels and the test suggestion rows were restored afterward. Fix belongs in `~/claude/scripts/up-sandbox-app.ts` (force/unset `POSTGRES_DATABASE` in the spawned server env).

### DIFF
--- a/src/server/lib/index.ts
+++ b/src/server/lib/index.ts
@@ -11,3 +11,4 @@ export * from "./validation";
 export * from "./alarm";
 export * from "./rate-limit";
 export * from "./infer-label-confidence";
+export * from "./record-suggestion-feedback";
--- a/src/server/lib/record-suggestion-feedback.test.ts
+++ b/src/server/lib/record-suggestion-feedback.test.ts
@@ -1,0 +1,161 @@
+// `recordSuggestionFeedback` decides which suggestion-log write a transaction
+// label update implies: a user confirmation (confidence 1), a user rejection
+// (confidence 0, not a budget switch), or nothing. The two upsert helpers it
+// calls run real SQL through `pool.query`, so the leaf mocks `pg` and the test
+// asserts on the captured INSERT — confirm vs reject is distinguished by the
+// `confirmed_at` column, which only the confirmed upsert writes.
+import { describe, test, expect, mock, beforeEach, afterAll } from "bun:test";
+import { restoreLeaves } from "test-helpers";
+
+// A fully-shaped row so the upserts' `new SuggestionModel(...)` type check
+// passes; the helper ignores the return value, so the contents are arbitrary.
+const okRow = {
+  transaction_id: "tx",
+  user_id: "u-1",
+  category_id: "cat",
+  confidence: 1,
+  is_confirmed: true,
+  is_rejected: false,
+  confirmed_at: null,
+  engine_scored_at: null,
+  updated: null,
+};
+
+const mockQuery = mock(async (_sql: string, _values?: unknown[]) => ({
+  rows: [okRow] as unknown[],
+  rowCount: 1 as number | null,
+}));
+
+class FakePool {
+  query = mockQuery;
+  end = async () => {};
+  connect = async () => ({ query: mockQuery, release: () => {} });
+}
+
+mock.module("pg", () => ({
+  Pool: FakePool,
+  types: { setTypeParser: () => {} },
+  default: { Pool: FakePool, types: { setTypeParser: () => {} } },
+}));
+
+const { recordSuggestionFeedback } = await import("./record-suggestion-feedback");
+
+afterAll(restoreLeaves);
+
+beforeEach(() => {
+  mockQuery.mockReset();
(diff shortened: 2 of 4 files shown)


claoie:
Closing — supersedes the design this PR was built on. PR #496's `suggestions` table is being replaced by `rejected_categories` in PR #501 per Hoie's 2026-06-09 call: confirmations are already in `transactions.label_*`, engine scores are also in `transactions.label_*` — only rejection history needed a separate table. Stage 2a (this PR's work) will be reopened on top of #501 with the new `addRejectedCategory` / `removeRejectedCategory` helpers.

PR #502, opened 2026-06-09
feat(rejected_categories): wire route + sync-plaid into the rejection signal (Stage 2a, #333)

claoie:
Stage 2a of the labels refactor — wires the dormant `rejected_categories` table (#501) into the user-action and Plaid-sync code paths so it starts accumulating real rejection signal.

## Surface

**Route layer (`POST /api/transaction`):** new helper `recordCategoryRejection` mirrors the label update into `rejected_categories` after the legacy `transactions.label_*` write succeeds. Four decision rules:

| Request body shape | Action |
|---|---|
| `label.category_id` absent | no-op (memo-only / budget-only update) |
| `label.budget_id` + `label.category_id: null` | NO rejection (category nullification is a side effect of budget switch — disambiguation called out by Hoie on PR #500's earlier shape) |
| `label.category_id: null` w/o budget change, prev category != null | UPSERT rejection on previous category |
| `label.category_id: <string>` | DELETE any prior rejection of that category for the same transaction (changed-my-mind) |

Mirror is best-effort: errors log a warning, do not bubble out of the route since the legacy update already succeeded.

**Plaid sync (pending → posted reconciliation):** in `sync-plaid.ts`, when `modelize` matches incoming via `byPendingId`, it records the (pending, posted) pair. After `upsertTransactions` creates the posted row, `migrateRejectedCategoriesOnPendingPosted(pending, posted)` runs in a single transaction:

```sql
BEGIN;
INSERT INTO rejected_categories (transaction_id, user_id, category_id, rejected_at)
SELECT $POSTED, user_id, category_id, rejected_at
FROM rejected_categories
WHERE transaction_id = $PENDING
ON CONFLICT (transaction_id, category_id) DO NOTHING;
DELETE FROM rejected_categories WHERE transaction_id = $PENDING;
COMMIT;
```

A ROLLBACK on any error leaves no half-migrated state.

## Out of scope (deliberate)

- **Wider orphan-pending-row issue.** The stored pending transaction itself, its splits, and any transaction_pairs still coexist with the posted row after Plaid transitions — that's pre-existing behaviour, not introduced or solved here. Comment in `modelize` calls this out.
- **`post-split-transaction` wiring.** Per-split rejections aren't well-defined (splits don't carry a merchant name); the merchant signal cares about parent-transaction-level rejection. Splits can be wired in a future PR if needed.
- **Engine read switch.** Stage 2b (next PR) rewrites `getMerchantSignal` to read rejection counts from `rejected_categories`. Until then, the engine's signal still uses the legacy `transactions.label_*` confidence-0 count.

## Tests

- `src/server/lib/record-category-rejection.test.ts` (new, 8 cases) — covers all four decision rules including the budget-switch disambiguation and the changed-my-mind cycle
- `src/server/lib/postgres/repositories/rejected-categories.test.ts` (+3 tests) — migrate happy-path, no-op when ids match, ROLLBACK on INSERT failure
- Full suite: **806/806 pass**
- `bun run typecheck`: clean
- `bun run lint`: clean

## E2E (to run on sandbox spin)

I'll drive the UI through these scenarios after the sandbox is up and post screenshots:
1. Confirm a category on a transaction → no `rejected_categories` row added.
2. Clear category on a transaction (without changing budget) → row added for the previous category.
3. Switch budget on a transaction (category clears as side-effect) → NO row added (disambiguation).
4. Pick the same category back after rejecting it → prior row deleted.

### DIFF
--- a/src/server/lib/compute-tools/sync-plaid.ts
+++ b/src/server/lib/compute-tools/sync-plaid.ts
@@ -63,28 +63,20 @@ export const buildTransactionLookupMaps = (
  * `rejected_categories` rows; future work may extend).
  *
  * Mechanism: Plaid's posted transaction carries
- * `pending_transaction_id` pointing back at the prior pending id. If
- * the stored set contains a row at that id, this incoming posted tx
- * supersedes it. Looking up `e.pending_transaction_id` in
- * `byTransactionId` catches the canonical shape exactly — no
- * compound-key false positives (recurring same-amount transactions),
- * no byPendingId backwards-direction confusion (re-emitted pending
- * data).
+ * `pending_transaction_id` pointing back at the prior pending id. The
+ * pair `(pending = e.pending_transaction_id, posted = e.transaction_id)`
+ * is the canonical supersession event — no need to cross-check against
+ * the stored set, since the back-pointer itself is the authoritative
+ * signal from Plaid. The migrate helper is idempotent + no-ops when
+ * the pending side has no rows, so an orphan back-pointer is harmless.
  */
 export const detectPendingPostedTransitions = (
   incoming: Pick<JSONTransaction, "transaction_id" | "pending_transaction_id">[],
-  lookupMaps: Pick<ReturnType<typeof buildTransactionLookupMaps>, "byTransactionId">,
 ): Array<{ pending: string; posted: string }> => {
   const transitions: Array<{ pending: string; posted: string }> = [];
   for (const e of incoming) {
-    if (!e.pending_transaction_id) continue;
-    const oldPending = lookupMaps.byTransactionId.get(e.pending_transaction_id);
-    if (oldPending && oldPending.transaction_id !== e.transaction_id) {
-      transitions.push({
-        pending: oldPending.transaction_id,
-        posted: e.transaction_id,
-      });
-    }
+    const { pending_transaction_id: pending, transaction_id: posted } = e;
+    if (pending && pending !== posted) transitions.push({ pending, posted });
   }
   return transitions;
 };
@@ -147,14 +139,6 @@ export const syncPlaidTransactions = async (item_id: string) => {
 
       const lookupMaps = buildTransactionLookupMaps(storedTransactions);
 
-      // Pending → posted transitions surfaced by modelize. Each entry is
-      // `{ pending: <old stored tx_id>, posted: <new incoming tx_id> }`.
-      // Used to migrate `rejected_categories` rows from the old pending
-      // id to the new posted id after the posted row is upserted. (The
-      // wider orphan-pending-row issue is out of scope for this PR — only
-      // the rejection-side migration is addressed here.)
-      const pendingPostedTransitions: Array<{ pending: string; posted: string }> = [];
-
       const modelize = (e: (typeof added)[0]) => {
         const result: JSONTransaction = { ...e, label: {} };
         const { authorized_date: auth_date, date } = e;
@@ -172,14 +156,10 @@ export const syncPlaidTransactions = async (item_id: string) => {
       // Pending → posted transitions across both added and modified
       // batches. Used to migrate `rejected_categories` rows from the old
       // pending id to the new posted id after the posted row is upserted.
-      // See `detectPendingPostedTransitions` for the canonical detection
-      // mechanism (back-pointer via `e.pending_transaction_id`, not
-      // findStoredTransaction id-inequality which has compound-key and
-      // byPendingId-direction false positives — issue surfaced in PR #502
-      // review pass).
-      pendingPostedTransitions.push(
-        ...detectPendingPostedTransitions([...added, ...modified], lookupMaps),
-      );
+      const pendingPostedTransitions = detectPendingPostedTransitions([
+        ...added,
+        ...modified,
+      ]);
 
       const updateJobs = [
         upsertTransactions(user, [...modeledAdded, ...modeledModified]),
--- a/src/server/lib/compute-tools/sync-plaid.ts
+++ b/src/server/lib/compute-tools/sync-plaid.ts
@@ -26,6 +26,7 @@ import {
   searchHoldingsByAccountId,
   MaskedUser,
   deleteSplitTransactionsByTransaction,
+  migrateRejectedCategoriesOnPendingPosted,
   logger,
 } from "server";
 import {
@@ -113,13 +114,33 @@ export const syncPlaidTransactions = async (item_id: string) => {
 
       const lookupMaps = buildTransactionLookupMaps(storedTransactions);
 
+      // Pending → posted transitions surfaced by modelize. Each entry is
+      // `{ pending: <old stored tx_id>, posted: <new incoming tx_id> }`.
+      // Used to migrate `rejected_categories` rows from the old pending
+      // id to the new posted id after the posted row is upserted. (The
+      // wider orphan-pending-row issue is out of scope for this PR — only
+      // the rejection-side migration is addressed here.)
+      const pendingPostedTransitions: Array<{ pending: string; posted: string }> = [];
+
       const modelize = (e: (typeof added)[0]) => {
         const result: JSONTransaction = { ...e, label: {} };
         const { authorized_date: auth_date, date } = e;
         if (auth_date) result.authorized_date = getDateTimeString(auth_date);
         if (date) result.date = getDateTimeString(date);
         const existing = findStoredTransaction(e, lookupMaps);
-        if (existing) result.label = existing.label;
+        if (existing) {
+          result.label = existing.label;
+          // Detect pending → posted: the incoming `e.transaction_id` (the
+          // posted id) differs from the stored `existing.transaction_id`
+          // (the pending id). Equality means a direct update under the
+          // same id, which needs no migration.
+          if (existing.transaction_id !== e.transaction_id) {
+            pendingPostedTransitions.push({
+              pending: existing.transaction_id,
+              posted: e.transaction_id,
+            });
+          }
+        }
         return result;
       };
 
@@ -137,7 +158,22 @@ export const syncPlaidTransactions = async (item_id: string) => {
 
       const partialItems = items.map(({ item_id, cursor }) => ({ item_id, cursor, updated }));
       return Promise.all(updateJobs)
-        .then(() => {
+        .then(async () => {
+          // Migrate rejected_categories rows from each pending → posted
+          // pair. Runs AFTER upsertTransactions so the posted FK target
+          // exists, and BEFORE counting so a failure is loud. Idempotent
+          // via ON CONFLICT DO NOTHING inside the helper.
+          for (const t of pendingPostedTransitions) {
+            try {
+              await migrateRejectedCategoriesOnPendingPosted(t.pending, t.posted);
+            } catch (err) {
+              logger.warn(
+                "Failed to migrate rejected_categories on pending→posted",
+                { pending: t.pending, posted: t.posted },
+                err,
+              );
+            }
+          }
           addedCount += added.length;
           modifiedCount += modified.length;
           removedCount += removed.length;
--- a/src/server/lib/compute-tools/sync-plaid.ts
+++ b/src/server/lib/compute-tools/sync-plaid.ts
@@ -174,19 +174,27 @@ export const syncPlaidTransactions = async (item_id: string) => {
         .then(async () => {
           // Migrate rejected_categories rows from each pending → posted
           // pair. Runs AFTER upsertTransactions so the posted FK target
-          // exists, and BEFORE counting so a failure is loud. Idempotent
-          // via ON CONFLICT DO NOTHING inside the helper.
-          for (const t of pendingPostedTransitions) {
-            try {
-              await migrateRejectedCategoriesOnPendingPosted(t.pending, t.posted);
-            } catch (err) {
+          // exists. Each migration touches a distinct pending_transaction_id
+          // with no shared rows across iterations, so parallelize with
+          // allSettled — serial would chain 4 round-trips per migration
+          // for first-sync / backfill batches. Failures are logged at warn
+          // and do not interrupt counting; idempotent via ON CONFLICT DO
+          // NOTHING inside the helper.
+          const migrations = await Promise.allSettled(
+            pendingPostedTransitions.map((t) =>
+              migrateRejectedCategoriesOnPendingPosted(t.pending, t.posted),
+            ),
+          );
+          migrations.forEach((m, i) => {
+            if (m.status === "rejected") {
+              const t = pendingPostedTransitions[i];
               logger.warn(
                 "Failed to migrate rejected_categories on pending→posted",
                 { pending: t.pending, posted: t.posted },
-                err,
+                m.reason,
               );
             }
-          }
+          });
           addedCount += added.length;
           modifiedCount += modified.length;
           removedCount += removed.length;
--- a/src/server/lib/compute-tools/sync-plaid.ts
+++ b/src/server/lib/compute-tools/sync-plaid.ts
@@ -56,6 +56,39 @@ export const buildTransactionLookupMaps = (
   return { byTransactionId, byPendingId, byCompoundKey };
 };
 
+/**
+ * Detect pending→posted transitions across a batch of incoming Plaid
+ * transactions. Returns `{pending, posted}` pairs for downstream
+ * migrations that need to follow the id rename (currently:
+ * `rejected_categories` rows; future work may extend).
+ *
+ * Mechanism: Plaid's posted transaction carries
+ * `pending_transaction_id` pointing back at the prior pending id. If
+ * the stored set contains a row at that id, this incoming posted tx
+ * supersedes it. Looking up `e.pending_transaction_id` in
+ * `byTransactionId` catches the canonical shape exactly — no
+ * compound-key false positives (recurring same-amount transactions),
+ * no byPendingId backwards-direction confusion (re-emitted pending
+ * data).
+ */
+export const detectPendingPostedTransitions = (
+  incoming: Pick<JSONTransaction, "transaction_id" | "pending_transaction_id">[],
+  lookupMaps: Pick<ReturnType<typeof buildTransactionLookupMaps>, "byTransactionId">,
+): Array<{ pending: string; posted: string }> => {
+  const transitions: Array<{ pending: string; posted: string }> = [];
+  for (const e of incoming) {
+    if (!e.pending_transaction_id) continue;
+    const oldPending = lookupMaps.byTransactionId.get(e.pending_transaction_id);
+    if (oldPending && oldPending.transaction_id !== e.transaction_id) {
+      transitions.push({
+        pending: oldPending.transaction_id,
+        posted: e.transaction_id,
+      });
+    }
+  }
+  return transitions;
+};
+
 /** Find the stored transaction matching an incoming Plaid transaction using O(1) map lookups. */
 export const findStoredTransaction = (
   incoming: Pick<JSONTransaction, "transaction_id" | "account_id" | "name" | "amount">,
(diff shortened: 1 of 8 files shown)


claoie:
## E2E verification (sandbox port 20502, against `budget_sandbox_20502`)

Drove `POST /api/transaction` through all four decision rules and asserted against the `rejected_categories` table after each call. Pattern + outcome only — no PII (tx IDs and category IDs prefix-redacted for the public record):

| Test | Body shape | Expected `rejected_categories` state | Result |
|---|---|---|---|
| 1. Confirm same category | `{ category_id: <orig> }` | no row added | ✓ 0 rows |
| 2. Clear category, no budget change | `{ category_id: null }` | row added for previous category | ✓ 1 row for orig category |
| 3. Switch budget (category clears as side effect) | `{ budget_id: <alt>, category_id: null }` | NO row added (disambiguation) | ✓ 0 rows — budget-switch disambiguation holds |
| 4. Pick previously-rejected category back | `{ category_id: <orig> }` after row-of-orig present | prior rejection row removed | ✓ 0 rows |

Seed transaction restored to its original `(budget_id, category_id)` state at the end of the run; sandbox DB has no test artifacts left behind.

Sanity-checked `migrateRejectedCategoriesOnPendingPosted` covered by the 3 new repo tests (happy path + no-op when ids match + ROLLBACK on INSERT failure). A live pending→posted scenario requires a Plaid mock that I don't have wired here; the helper's SQL + transaction semantics are pinned by the unit test, and the call site in `sync-plaid.ts` runs in a try/warn block so a real-prod failure won't break sync.

reviewoie:
Test comment

reviewoie:
## Code review — PR #502 (second pass, commit 8505e16)

**Verdict:** approve

**Findings:** 0 high, 0 med, 3 low (informational only — none block merge).

### High
No blockers. All four prior-pass findings verified fixed.

### Med
No blockers.

### Low (informational — would NOT block merge)

- `src/server/lib/postgres/repositories/rejected-categories.ts:84` — `migrateRejectedCategoriesOnPendingPosted` takes no `user_id`. Safe because `transactions.transaction_id` is `PRIMARY KEY` (globally unique across users); the `DELETE FROM rejected_categories WHERE transaction_id = $1` can only ever touch rows owned by the single user who owns that transaction. Other repo helpers (`addRejectedCategory`, `removeRejectedCategory`) carry the `user_id = $2` defense-in-depth guard documented in this very repo's own docstring ("if a future schema change ever lets two users share a transaction_id, this guard prevents cross-user clobber"). The migration helper inherits the same risk surface but lacks the same guard. Adding `AND user_id = $3` to the DELETE and threading the source-row `user_id` would close the gap. Not a present bug — flagging because it diverges from the in-file convention.

- `src/server/routes/accounts/post-transaction.ts:59` — `if (willTouchCategory && transaction.transaction_id)` short-circuits the `getPrevLabel` read on falsy `transaction_id`. `requireStringField` (the validator used at line 36) does NOT reject empty strings (no `.trim() === ""` check, unlike `requireQueryString`). The empty-string case would skip the prev-label read, fall through to the post-update `recordCategoryRejection(user, "", ...)`, and the rejection INSERT would fail the `transaction_id` FK constraint — which the `catch` swallows with a warn. Behavior is benign (no data corruption, no 500), but worth noting in case a future validation tightening removes the catch. The fix is one line: have `requireStringField` reject empties, matching `requireQueryString`.

- `src/server/lib/compute-tools/sync-plaid.test.ts:128-203` — Four of the six new `detectPendingPostedTransitions` tests (`byCompoundKey false-positive`, `re-emitted-pending direction guard`, `missing stored row defense`, `same-id no-op`) would pass against a `() => []` stub because none of the constructed inputs carry a `pending_transaction_id` that resolves to a stored row. They function as **design-pinning** tests — documenting that the new helper does NOT do compound-key matching or byPendingId chasing — rather than positive behavioral tests. Combined with the canonical case (which does fail under the stub) and the mixed-batch case the bidirectional coverage is sound, but the four pinning tests are weaker than they appear. If a future refactor reintroduces compound-key matching to the helper, only the first byCompoundKey test would actually catch it (and only because `result.toEqual([])` would no longer be satisfied if a transition got emitted). This is acceptable, just calling it out so the test-coverage claim is precise.

### Re-verification of fix landings

- **HIGH #1 + #2 (sync-plaid migration trigger):** `detectPendingPostedTransitions` correctly uses Plaid's `pending_transaction_id` back-pointer + `byTransactionId` lookup. The same-id defensive guard (`oldPending.transaction_id !== e.transaction_id`) is in place. The for-loop placement in `syncPlaidTransactions` is correct — transitions are gathered up front from `[...added, ...modified]` BEFORE `upsertTransactions` runs, and the migration helper is invoked AFTER `Promise.all(updateJobs)` resolves so the posted-row FK target exists. Per-pair `try/catch` around `migrateRejectedCategoriesOnPendingPosted` correctly logs and continues; counters increment regardless. Verified.

- **MED #1 (read-failure escape):** `getPrevLabel` call wrapped in its own try/catch at `post-transaction.ts:60-68`. On pool error the route logs a warn and falls through with `prevLabel = {null, null}`; downstream the helper's "prev was null → nothing to reject" branch is correct (`recordCategoryRejection` test at line 73 covers it). Verified.

- **MED #2 (budget-switch false positive):** `recordCategoryRejection` now takes a `PrevLabel` and the disambiguation at line 64-67 gates on `reqLabel.budget_id !== prevLabel.budget_id`. The new test "BUDGET-UNCHANGED rejection: body re-states current budget_id + category_id=null → still a genuine rejection" (record-category-rejection.test.ts:70-82) is the behavioral guard. Verified.

- **LOW #1 (test coverage):** 6 new `detectPendingPostedTransitions` tests + 1 new `recordCategoryRejection` "budget unchanged" case. Full suite 813/813 pass; typecheck + lint clean. Verified.

### Scope discipline

8 changed files, all on-topic for "wire `rejected_categories` into route + sync-plaid". Fix commit (`8505e16`) touches 5 files, every one mapped to one of the four findings. No drive-by edits.

### Tests
- `bun test src/server/lib/record-category-rejection.test.ts src/server/lib/compute-tools/sync-plaid.test.ts src/server/lib/postgres/repositories/rejected-categories.test.ts` → 39 pass / 0 fail
- `bun test` (full suite) → 813 pass / 0 fail
- `bun run typecheck` → clean
- `bun run lint` → clean

### Approval signal

Ship it. The two HIGH bugs from pass 1 have proper regression coverage. The three low-priority items above are documentation-quality concerns, not blockers.

src/server/lib/compute-tools/sync-plaid.ts
```diff
@@ -55,6 +56,39 @@ export const buildTransactionLookupMaps = (
...
+  const transitions: Array<{ pending: string; posted: string }> = [];
+  for (const e of incoming) {
+    if (!e.pending_transaction_id) continue;
+    const oldPending = lookupMaps.byTransactionId.get(e.pending_transaction_id);
+    if (oldPending && oldPending.transaction_id !== e.transaction_id) {
+      transitions.push({
+        pending: oldPending.transaction_id,
+        posted: e.transaction_id,
+      });
+    }
```
hoiekim: Isn't this redundant? Why don't you write it like this?
```
const transitions: Array<...> = [];
incoming.forEach((t) => {
  const { pending_transaction_id: pending, transaction_id: posted } = t;
  if (pending && pending !== posted) transitions.push({ pending, posted });
})
return transitions;
```
No need of `lookupMaps`

src/server/lib/compute-tools/sync-plaid.ts
```diff
@@ -127,6 +169,18 @@ export const syncPlaidTransactions = async (item_id: string) => {
...
+      // batches. Used to migrate `rejected_categories` rows from the old
+      // pending id to the new posted id after the posted row is upserted.
+      // See `detectPendingPostedTransitions` for the canonical detection
+      // mechanism (back-pointer via `e.pending_transaction_id`, not
+      // findStoredTransaction id-inequality which has compound-key and
+      // byPendingId-direction false positives — issue surfaced in PR #502
+      // review pass).
+      pendingPostedTransitions.push(
+        ...detectPendingPostedTransitions([...added, ...modified], lookupMaps),
+      );
```
hoiekim: Why did you spread it? That's an additional unnecessary `O(N)`. You could just do
```
const transitions = detectPendingPostedTransitions(...)
```

src/server/routes/accounts/post-transaction.ts
```diff
@@ -39,12 +41,56 @@ export const postTransactionRoute = new Route<TransactionPostResponse>(
...
+        try {
+          await recordCategoryRejection(user, transaction_id, reqLabel, prevLabel);
+        } catch (mirrorErr) {
+          logger.warn(
+            "Failed to mirror category change into rejected_categories",
+            { transactionId: transaction_id },
+            mirrorErr,
+          );
+        }
+      }
```
hoiekim: Make this block and `prevLabel` resolution as fire-and-forget operation. The API response is agnostic to the rejection operation's success/failure state. This way we don't regress the API latency.

src/server/lib/record-category-rejection.ts
```diff
@@ -0,0 +1,108 @@
...
+  // DIFFERENT from the previous budget AND `category_id: null` is a
+  // budget change, not a category rejection. A body that includes the
+  // unchanged budget_id falls through and the category clear is treated
+  // as a genuine rejection.
+  const isBudgetSwitch =
+    "budget_id" in reqLabel &&
+    newCategoryId === null &&
+    reqLabel.budget_id !== prevLabel.budget_id;
+
+  if (isBudgetSwitch) return;
```
hoiekim: There seems to be misunderstanding. When the user switch budget while there was a suggested budget & category already, the user is definitely rejecting the suggestion. We should write rejection in this case.

src/server/lib/record-category-rejection.ts
```diff
@@ -0,0 +1,108 @@
...
+      await addRejectedCategory(user, transaction_id, prevLabel.category_id);
+    }
+    return;
+  }
+
+  if (typeof newCategoryId === "string") {
+    // User picked or kept a category — clear any prior rejection of THIS
+    // category for THIS transaction. Cheap; no-op if no row matched.
+    await removeRejectedCategory(user, transaction_id, newCategoryId);
+  }
```
hoiekim, line 60: If the new category is different from suggested category, that's a rejection too. Implement this logic.

src/server/lib/record-category-rejection.ts
```diff
@@ -0,0 +1,108 @@
...
+export const getPrevLabel = async (
+  user: MaskedUser,
+  transaction_id: string,
+): Promise<PrevLabel> => {
+  const [tx] = await searchTransactionsById(user, [transaction_id]);
+  return {
+    category_id: tx?.label?.category_id ?? null,
+    budget_id: tx?.label?.budget_id ?? null,
+  };
+};
```
hoiekim, line 102: You should probably return confidence as well, to use it to determine if it's in suggested / confirmed state. If the previous label is in suggested state and the user is writing a different category label, we should record a rejection for the previous label.

claoie:
## Applied all 6 review comments — `507e693`

| Comment | Change |
|---|---|
| 1. `detectPendingPostedTransitions` redundant lookupMaps | dropped — back-pointer alone is authoritative; migrate helper is no-op on orphan; signature shrinks to `(incoming)` |
| 2. Spread on push | replaced with `const transitions = detect(...)` |
| 3. Fire-and-forget the rejection mirror | kicks off `getPrevLabel` in parallel with `updateTransactions`, chains `recordCategoryRejection` without awaiting; mirror runs off the latency path |
| 4. Budget switch on a SUGGESTED label IS a rejection | new disambiguation gated on `wasConfirmed(prev)` (only confirmed labels get the FE-side-effect pass) |
| 5. Picking different over a SUGGESTED label IS a rejection | new branch records rejection of the suggested category before doing the changed-my-mind cleanup on the picked one |
| 6. `getPrevLabel` returns `category_confidence` | needed for the suggested-vs-confirmed branches; route forwards it through unchanged |

## E2E verification (sandbox 20502)

7 scenarios, all pass — including the two NEW suggested-path cases your comments introduced:

| | Scenario | Expected | Result |
|---|---|---|---|
| A | Confirm same category | no rejection | ✓ |
| B | Clear, no budget change, prev confirmed | rejection of prev | ✓ |
| C | Budget switch, prev confirmed | no rejection (FE side effect) | ✓ |
| D | **Budget switch, prev SUGGESTED** | **rejection of prev (new)** | ✓ |
| E | **Pick different over SUGGESTED** | **rejection of prev (new)** | ✓ |
| F | Pick different over CONFIRMED | no rejection (re-categorize, not reject) | ✓ |
| G | Pick previously-rejected back | row removed (changed-my-mind) | ✓ |

Suite 815/815 (was 813; +2 net after dropping the lookupMaps-specific cases that no longer apply); typecheck + lint clean.

src/server/routes/accounts/post-transaction.ts
```diff
@@ -39,12 +41,42 @@ export const postTransactionRoute = new Route<TransactionPostResponse>(
...
+
+      // Kick off the previous-label read in parallel with the update.
+      // Both run on separate pool connections; the rejection mirror
+      // doesn't add to the API latency path.
+      const reqLabel = transaction.label;
+      const willTouchCategory = !!reqLabel && "category_id" in reqLabel;
+      const prevLabelPromise =
+        willTouchCategory && transaction.transaction_id
+          ? getPrevLabel(user, transaction.transaction_id)
+          : null;
```
hoiekim: I know I told you to do this, but I realized I made a mistake. `get` and `update` shouldn't run in parallel -- that's vulnerable. Await on `get`.

src/server/routes/accounts/post-transaction.ts
```diff
@@ -39,12 +41,42 @@ export const postTransactionRoute = new Route<TransactionPostResponse>(
...
+
       const response = await updateTransactions(user, [transaction]);
       const result = response[0];
       if (!result || result.status >= 400) {
         throw new Error("Database responded with an error.");
       }
       const transaction_id = result.update._id || "";
+
+      // Fire-and-forget the rejected_categories mirror. The API response
+      // is agnostic to the mirror's success/failure — per Hoie 2026-06-09,
```
hoiekim: Don't put my name on comments. This is an existing coding rule.

reviewoie:
## Code review — PR 502 (third pass, head e947a42)

**Verdict:** changes-requested — one HIGH (rule violation per your own 2026-06-09 instruction missed in the scrub commit). The rest are smaller polish/cleanup items.

I re-walked the whole diff against `upstream/main`, ran the 4-angle scan from Phase 2.H on this PR's own code, and adversarially poked the decision matrix in `recordCategoryRejection`. Tests + lint + typecheck green locally (`bun test src/server` → 582/582 pass). The 8 fixes from passes 1 + 2 all landed correctly:
- `detectPendingPostedTransitions` is back-pointer-authoritative, no `lookupMaps` param, returns the array directly without `push(...spread)`.
- `recordCategoryRejection` matrix matches the 5-row spec; verified by walking every body shape (`{category_id:null}`, `{category_id:null, budget_id:X}`, `{category_id:""}`, missing keys) against every prev confidence (`null`, `0`, `(0,1)`, `1`).
- Route does serial `await getPrevLabel` → `await updateTransactions` → fire-and-forget `recordCategoryRejection(...).catch(logger.warn)`. The read order is the race-safe one.
- `migrateRejectedCategoriesOnPendingPosted` transaction is correct: `pool.connect()` outside the try; finally always releases; ROLLBACK in catch is `.catch(()=>{})` so a no-op rollback can't mask the original error; single release on every path including a successful COMMIT, a ROLLBACK rethrow, or a COMMIT-throw rethrow.
- Decision matrix tests cover the 7 expected scenarios + the "body re-states unchanged budget" edge.

**Findings:** 6 total — 1 high, 1 med, 4 low.

### High

- `src/server/lib/record-category-rejection.test.ts:88` and `:149` — two test names still contain `(Hoie 2026-06-09)` attribution. Commit `e947a42` claims "scrub name attribution from comments" but missed these test name strings. Your rule from this PR's own thread (now in `feedback_comment_hygiene.md`): *"Don't put my name on comments. This is an existing coding rule."* — applies to all committed code including test name strings. Rename both to drop the parenthetical attribution; the rest of the test name already states the behavior being asserted. (No remaining `\bHoie\b` matches anywhere else in the diff.)

### Med

- `src/server/lib/record-category-rejection.ts:103-108` — `getPrevLabel`'s docstring is stale: *"The route layer kicks this off in parallel with the update so the rejection mirror can name the category being cleared..."*. The route now reads it serially, and the route's own inline comment at `post-transaction.ts:45-49` explicitly says *"running this read in parallel races against the write"*. Direct contradiction within the PR. Rewrite the helper docstring to "Called serially before the update — the update overwrites the columns we need to see, so a parallel read would race."

### Low

- `src/server/lib/compute-tools/sync-plaid.ts:177` — comment says *"BEFORE counting so a failure is loud"*, but the body's `try { await migrate... } catch (err) { logger.warn(...) }` swallows every failure and continues. A failure is `warn`-logged, never thrown — counters still advance. Either drop the "so a failure is loud" clause or rethrow.

- `src/server/lib/record-category-rejection.ts:113-117` — `getPrevLabel` calls `searchTransactionsById`, which is `SELECT *` on `transactions`. The function needs three columns (`label_category_id`, `label_budget_id`, `label_category_confidence`). On a hot POST /transaction path a tight column-list SELECT (or a dedicated `getTransactionLabel(user, id)` repo helper) trims the row payload. Not catastrophic — `transactions` has no HTML/blob fields — but is the kind of `SELECT *` you've called out in past reviews.

- `src/server/lib/compute-tools/sync-plaid.ts:179-189` — the `for (const t of pendingPostedTransitions)` loop awaits each `migrateRejectedCategoriesOnPendingPosted` serially. Each transition is on a distinct `pending_transaction_id`, so a `Promise.allSettled(...)` parallelization is safe (no contention). For typical sync sizes (1-10 transitions) this is noise; if a backfill / first-sync produces 100+ transitions, the serial cost is real (4 round-trips per migration). Either parallelize, or add a comment naming why serial is intentional.

- `src/server/lib/postgres/repositories/rejected-categories.test.ts:131-143` — the transaction-safety test covers "INSERT throws → ROLLBACK", but not "DELETE throws after INSERT succeeds" or "COMMIT throws". The whole point of wrapping in BEGIN/COMMIT is the multi-step guarantee; the test only exercises the first step's failure. Add at least one "DELETE throws → ROLLBACK reverts the INSERT" case.

src/server/lib/record-category-rejection.ts
```diff
@@ -0,0 +1,134 @@
...
+    if (
+      prevCategoryId &&
+      newCategoryId !== prevCategoryId &&
+      wasSuggested(prevLabel)
+    ) {
+      await addRejectedCategory(user, transaction_id, prevCategoryId);
+    }
+
+    // Clear any prior rejection of THIS category (changed-my-mind cycle).
+    await removeRejectedCategory(user, transaction_id, newCategoryId);
```
hoiekim: I think this is all we need. Probably no need to check `if (newCategoryId === null)` or `if (typeof newCategoryId === "string")` before this. Only need to skip `removeRejectedCategory` when `newCategoryId === null`. That'll simplify the code with the same result. What do you think?

hoiekim (changes requested):

reviewoie:
## Code review — PR 502 (fourth pass, head 4d2656f)

**Verdict:** approve — all 6 findings from the third-pass review (head e947a42) are resolved in commit `4d2656f8`. Incremental scope only: `sync-plaid.ts`, `rejected-categories.test.ts`, `record-category-rejection.ts(+test)`. No new findings.

I re-walked only the delta `e947a42 → 4d2656f` (the new commit landed after the last review). Verified each prior finding against the committed tree, re-ran the affected suites, and traced the new raw SQL against the schema and the removed-behavior invariant. Committed state is green: `bun test` on the three touched test files → 43/43 pass.

### Prior findings — all resolved

- **HIGH (test-name attribution):** both `(Hoie 2026-06-09)` strings dropped (`record-category-rejection.test.ts:88,149`). `grep -n '\bHoie\b'` across all PR-touched source + test files → zero matches.
- **MED (stale `getPrevLabel` docstring):** rewritten to state the serial-before-update ordering; no longer contradicts the route comment at `post-transaction.ts:45-49`.
- **LOW (sync-plaid "a failure is loud" comment):** the misleading clause is gone — comment now correctly says failures are warn-logged and don't interrupt counting.
- **LOW (`SELECT *` on the hot path):** `getPrevLabel` now issues a targeted three-column SELECT instead of routing through `searchTransactionsById`.
- **LOW (serial migrate loop):** parallelized via `Promise.allSettled` over distinct `pending_transaction_id`s; per-rejection warn preserved through `m.reason`.
- **LOW (transaction-safety test gaps):** two new cases added — "DELETE throws AFTER INSERT → ROLLBACK" and "COMMIT throws → ROLLBACK". Both names exist and assert the right SQL sequence; verified against the `throw err` rethrow path in `migrateRejectedCategoriesOnPendingPosted`.

### Verification notes on the new raw SQL (`getPrevLabel`)

- Column names `label_category_id` / `label_budget_id` / `label_category_confidence` match the `TransactionModel` declarations exactly — no runtime column-name crash (mocked-pg tests wouldn't catch that, so I checked the model).
- Soft-delete invariant preserved: the old path (`queryByIds`) applied `(is_deleted IS NULL OR is_deleted = FALSE)` for soft-delete tables; the new inline SQL carries the identical predicate. user_id scoping also preserved (cross-user reads return null fields, no leak).
- Route call site unchanged-signature and matches the canonical latency-path shape: serial `await getPrevLabel` → `await updateTransactions` → fire-and-forget `recordCategoryRejection(...).catch(warn)`, response independent of the mirror.

### Optional, non-blocking

- `getPrevLabel`'s inline `pool.query` lives in a non-repository lib file. The SELECT is correct and the column-projection win is real, but a plain by-id projected SELECT is the kind of standard shape you've previously preferred behind a repo helper (PR 300 raw-SQL pushback). A `getTransactionLabel(user, id)` in `repositories/transactions.ts` would keep raw SQL inside the repo layer. Strictly a placement preference — not a correctness issue, won't block.

No further blockers.

hoiekim (approved):

PR #505, opened 2026-06-10
fix(flow-chart): net per-section amounts so self-cancelling transfers don't double-count

claoie:
## What

The Sankey flow chart on the dashboard / chart pages was double-counting self-transfers. Both legs of a transfer pair (`+X` and `−X`) landed on opposite sides of the chart as ABSOLUTE magnitudes, even though they cancel — so the Transfers section showed up on both the income and expense sides at magnitude X, but the Budget page (which nets per category) showed it correctly at $0.

## Repro on prod data

For Hoie's prod budget (June 2026):

| Transfer pair | leg A | leg B |
|---|---|---|
| 1 | +A | −A |
| 2 | +B | −B |
| 3 | +C | −C |

All labeled to the Transfers budget/section/category. Signed sum = $0 (each pair self-cancels). Budget page showed $0. Flow chart showed $[redacted] on the expense side AND on the income side for the same Transfers section.

## Root cause

`getSankeyData` in `src/client/components/FlowChartRow/lib.ts` segregated each transaction by sign at ingest time into TWO maps (`expenseSections` and `incomeSections`), each holding absolute magnitudes. There was no point at which the +X and −X legs of a transfer pair could cancel — they ended up on opposite sides as same-magnitude entries.

`getBudgetData` (in `src/client/lib/hooks/calculation/budgets.ts`) does the right thing: each transaction's `amount += sorted_amount` on the section, so signed sums net naturally.

## Fix

Mirror the Budget page's aggregation:

1. Accumulate signed amounts per (section_id, budget_id) into single intermediate maps.
2. After processing all transactions, partition each section/budget by net sign:
   - `net < 0` → income side (magnitude = `|net|`)
   - `net > 0` → expense side (magnitude = `net`)
   - `net = 0` → drop entirely

Self-cancelling transfers (net = 0) disappear from the chart. Partial offsets — a refund inside a Shopping section — correctly reduce the bar to the remaining net spending instead of double-counting both legs.

## Tests (6 new in `src/client/components/FlowChartRow/lib.test.ts`)

- transfer pair → section drops from both sides; income/expense totals unchanged
- pure expense section → expense side only
- pure income section → income side only with absolute magnitude
- mixed section (refund inside Shopping) → routes to net-sign with reduced magnitude
- multi-section batch — independent partition
- `amount=0` transactions skipped (no zero rows)

Full suite 801/801 pass; typecheck + lint clean.

## E2E verification (to follow on sandbox spin)

Sandbox at port 20503 — will drive the Dashboard's flow chart against unscrambled prod data, screenshot, and compare against the Budget page's Transfers section reading $0.

*[Edited 2026-06-11: redacted real prod-budget dollar figures (transfer-pair magnitudes + section total) per §2a PII policy. Structure/logic unchanged.]*

### DIFF
--- a/src/client/components/FlowChartRow/lib.test.ts
+++ b/src/client/components/FlowChartRow/lib.test.ts
@@ -1,0 +1,189 @@
+import { describe, test, expect } from "bun:test";
+import { ViewDate } from "common";
+import {
+  Account,
+  Budget,
+  BudgetDictionary,
+  Category,
+  CategoryDictionary,
+  Section,
+  SectionDictionary,
+  Transaction,
+  TransactionDictionary,
+  InvestmentTransactionDictionary,
+} from "client";
+import { getSankeyData } from "./lib";
+
+// ---------------------------------------------------------------------------
+// Fixture helpers — minimal shapes that exercise getSankeyData's aggregation.
+// We're testing the partition-by-net logic, not the model classes themselves.
+// ---------------------------------------------------------------------------
+
+const TX_DATE = "2026-06-05T00:00:00";
+const view = new ViewDate("month", new Date(2026, 5, 15)); // June 2026
+
+const makeBudget = (id: string, name: string): Budget =>
+  new Budget({
+    budget_id: id,
+    name,
+    iso_currency_code: "USD",
+    capacities: [{ capacity_id: "c", month: 0 }],
+  });
+
+const makeSection = (id: string, name: string, budget_id: string): Section =>
+  new Section({
+    section_id: id,
+    budget_id,
+    name,
+    capacities: [{ capacity_id: "c", month: 0 }],
+  });
+
+const makeCategory = (id: string, name: string, section_id: string): Category =>
+  new Category({
+    category_id: id,
+    section_id,
+    name,
+    capacities: [{ capacity_id: "c", month: 0 }],
+  });
+
+const makeAccount = (id: string, budget_id: string): Account =>
+  new Account({
+    account_id: id,
+    name: "Account",
+    balances: { available: 0, current: 0, iso_currency_code: "USD" } as never,
+    label: { budget_id },
+    hide: false,
+  } as never);
+
+const makeTx = (id: string, account_id: string, amount: number, label: object): Transaction =>
+  new Transaction({
+    transaction_id: id,
+    account_id,
+    amount,
+    date: TX_DATE,
+    authorized_date: TX_DATE,
+    label: label as never,
+    iso_currency_code_at_purchase: "USD",
+  } as never);
+
+const sankeyTotalsBySectionId = (sankey: ReturnType<typeof getSankeyData>) => {
+  const out = new Map<string, { side: "income" | "expense"; amount: number }>();
+  // column1 = income sections; column5 = expense sections
+  sankey.graphData[0].forEach((r) => out.set(r.id, { side: "income", amount: r.amount }));
+  sankey.graphData[4].forEach((r) => out.set(r.id, { side: "expense", amount: r.amount }));
+  return out;
+};
+
+describe("getSankeyData — net-by-section partition", () => {
+  const buildScene = (txs: Transaction[]) => {
+    const budgets = new BudgetDictionary();
+    budgets.set("budget-expense", makeBudget("budget-expense", "Spending"));
+    budgets.set("budget-income", makeBudget("budget-income", "Income"));
+    budgets.set("budget-transfer", makeBudget("budget-transfer", "Transfers"));
+
+    const sections = new SectionDictionary();
+    sections.set("sec-shopping", makeSection("sec-shopping", "Shopping", "budget-expense"));
+    sections.set("sec-salary", makeSection("sec-salary", "Salary", "budget-income"));
+    sections.set("sec-transfer", makeSection("sec-transfer", "Transfers", "budget-transfer"));
+
+    const categories = new CategoryDictionary();
+    categories.set("cat-shopping", makeCategory("cat-shopping", "Misc", "sec-shopping"));
+    categories.set("cat-salary", makeCategory("cat-salary", "Paycheck", "sec-salary"));
+    categories.set("cat-transfer", makeCategory("cat-transfer", "Transfers", "sec-transfer"));
+
+    const accounts = [
+      makeAccount("acc-1", "budget-expense"),
+      makeAccount("acc-2", "budget-income"),
+    ];
+
+    const txDict = new TransactionDictionary();
+    txs.forEach((t) => txDict.set(t.transaction_id, t));
+
+    return getSankeyData(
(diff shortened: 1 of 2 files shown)


claoie:
## E2E verification (sandbox 20505, unscrambled prod-clone)

**Bundle verification:** `main-D2Q58GrT.js` contains the new `signedAmount`-based partition logic (13 occurrences). The pre-fix `incomeSections.set` direct-add was confirmed gone.

**Label-count diff:** the Transfers section/budget previously appeared 4 times in the Sankey (income SECTION + income BUDGET + expense SECTION + expense BUDGET, each with gross magnitudes). After this fix, it appears 2 times — once per aggregation level, on whichever side its NET resolves to.

May 2026 monthly view, FlowChartRow labels with x-coords + rect heights:

| x   | label              | h    | column                |
|-----|---------------------|------|------------------------|
| 110 | Unsorted            | 187  | income section         |
| 110 | Paychecks           | 58   | income section         |
| 110 | Extra               | 44   | income section         |
| 263 | **Transfers**       | 141  | income BUDGET          |
| 263 | Income              | 64   | income BUDGET          |
| 497 | Non-budgeted        | 187  | expense BUDGET         |
| 497 | Shared Budget       | 15   | expense BUDGET         |
| 497 | Baby Fund           | 58   | expense BUDGET         |
| 497 | Annie's Budget      | 34   | expense BUDGET         |
| 683 | Surplus             | 71   | (right rail marker)    |
| 683 | Ignored             | 199  | expense section        |
| 683 | **Transfers**       | 187  | expense SECTION        |
| ... | (rest of expense)   | ...  | expense section        |

Net signs verified against prod DB (chart's account_ids filter + May 2026):
- **Transfers BUDGET** = signed sum across all txs whose `label.budget_id = Transfers` OR `account.label.budget_id = Transfers`. Has some account-defaulted income legs from accounts whose default budget is "Transfers" → can net negative → appears on income side.
- **Transfers SECTION** = signed sum across txs whose category resolves to Transfers section = **+$[redacted]** → appears on expense side at proportional height.

The fact that the two appear on DIFFERENT sides is consistent: they're computed from different filter sets (budget by label_budget_id, section by category.section_id), and the account-default budget assignment introduces non-Transfer-section transactions into the budget aggregation. Pre-fix this was even messier because EACH aggregation was double-counted.

For self-cancelling transfer pairs within a section (e.g. June 2026: 3 pairs of ±$X all in scope), the section drops out entirely — verified by unit test `transfer pair (one section, +X and −X) → section drops from BOTH sides; income/expense unchanged` in `lib.test.ts`.

**Tests:** 6 new in `lib.test.ts` (transfer pair, pure expense, pure income, mixed/refund, multi-section, amount=0). Full suite 801/801, typecheck + lint clean.

**Screenshots** (kept locally): `/tmp/pr505/jun-monthly.png` shows the May 2026 monthly view; the Transfers BUDGET on income side is the small slice at x=263, and the Transfers SECTION on expense side is at x≈675.

If the residual "Transfers" appearance on TWO different sides feels wrong even with the net-based partition, the further fix would be to either (a) collapse budget aggregation through the section it actually maps to, or (b) exclude account-default-only assignments from the budget aggregation. Let me know if you want either as a follow-up.

*[Edited 2026-06-11: redacted a real prod-budget section total per §2a PII policy. Technical detail unchanged.]*

reviewoie:
## Code review — PR 505

**Verdict:** changes-requested (one med correctness regression with a verified repro)

Scope is clean: only `FlowChartRow/lib.ts` + new `lib.test.ts`, both directly on the netting fix. The transfer-cancellation logic itself is correct — a section/budget whose legs sum to 0 now drops from both sides, and partial offsets reduce to the net (verified: 6/6 tests pass, names accurate, repro confirms the +X/−X transfer pair vanishes). The concern is a second-order interaction the tests don't cover.

### Med
- `lib.ts:103-149` — **Section and budget are now netted independently, so a section can land on a different side of the chart than its parent budget — producing an orphaned section bar with no link and an income/expense total that contradicts it.** The fix partitions `sectionTotals` and `budgetTotals` separately by net sign; when a section's net sign differs from its budget's net sign, the section's `next` (budget_id) has no matching row in the adjacent budget column, so `Sankey.tsx`'s link filter (`row.next === targetColumn[i]?.id`) silently draws no link.

  Failure scenario (verified with a repro run against this branch vs `upstream/main`): one "Spending" budget, two sections — Shopping net `+100` (expense), Subscriptions net `−30` (refund-heavy, income). Budget nets `+70` → expense side only.
  - This branch: col1 (income sections) = `[sec-subs amt 30 next b-spend]`, but col2 (income budgets) = `[]` (b-spend is only in col4). `sec-subs` renders as a 30-bar on the income column with nothing to link to, while `tableData.income = 0`.
  - `upstream/main` on the same input: b-spend appears in BOTH income (30) and expense (100) budget columns, so `sec-subs` links cleanly and `income = 30`. No orphan.

  So the orphan + income-section-without-income-total is newly introduced by this PR. The PR body frames the fix as "mirror the Budget page's aggregation," but the Budget page nets per category/section only — it has no parallel section-vs-budget cross-link to keep consistent. Either net at one level, or drop a section row whose net sign disagrees with its resolved budget side (and exclude it from that side's section column), so the section column can't contradict the budget column. Add a test where a section nets opposite its budget.

### Low
- `lib.ts:88-89` — `intersectionBudgets` is now dead. After partitioning, `incomeBudgets` and `expenseBudgets` are disjoint by construction (each budget_id lands in exactly one by net sign), so `incomeBudgets.forEach(... expenseBudgets.has(k) ...)` never matches and the map is always empty. That makes the `isAIntersection`/`intersectionBudgetDiff` branches in all four column sorts (col1/col2/col4/col5) unreachable — ~30 lines of sort logic that can no longer fire. Not a correctness bug, but it's stale now that a budget can't appear on both sides; worth removing or noting so a future reader doesn't assume it's live.

### Informational
- E2E not yet run — PR body says "to follow on sandbox spin" (port 20503). The claoie-side sandbox/UI verification against the Dashboard flow chart vs the Budget page's Transfers `$0` reading is still owed before merge.

No further blockers.

claoie:
Closing per Hoie's feedback — the 'Transfers' budget/section is just a user-labeled category, not the 'real transfer pair' concept. Self-cancelling debits/credits within a labeled section are still legitimate cash movements the user intentionally categorized; the chart should show them, not silently cancel.

The actual transfer-pair data ( table, populated by the auto-detect transfer flow + the `POST /api/transfer-pair` route) is the right source for any 'don't double-count internal transfers' behavior — and that's NOT wired into the FE yet (no FE references to `/api/transfers` endpoints, no FE Data model for pairs). A future PR can wire those into the flow chart's filter set if/when that's the desired behavior.

For now, the chart's existing 'sum debits / sum credits per labeled section' aggregation is the correct user-facing semantic.

PR #506, opened 2026-06-11
feat(auto-suggest): read rejected count from rejected_categories

claoie:
## Summary

Stage 2b of the labels refactor (#333). Now that #501 retired the `transactions.label_category_confidence = 0.0` storage for rejections in favor of an explicit `rejected_categories` table, the engine-side `getMerchantSignal` SQL needs to source its `rejected` count from the new table.

Two-source signal:
- **accepted** — `transactions.label_category_confidence = 1.0` rows in the recent-N similar-merchant pool. Same `MERCHANT_SIGNAL_LIMIT` recency cap as before.
- **rejected** — `rejected_categories` rows for the winning category, joined to `transactions` for the merchant-similarity filter. **Unbounded** by the recency LIMIT — an explicit user "no, not this category" should stick until they re-pick that category (which triggers `removeRejectedCategory`).

The winning category is still picked by `accepted COUNT(*) DESC LIMIT 1` in a `WITH winning AS (...)` CTE. Rejections do NOT influence which category wins, only the gate (`evaluateSignal`) downstream. A category with 0 confirms in the recent pool can never win — same as the pre-2b shape.

## Test plan

- [x] Suite: 822 / 822 pass (16 existing auto-suggest tests + 6 new SQL-shape assertions, +800 elsewhere unchanged)
- [x] Typecheck clean
- [x] Lint clean
- [x] Sandbox E2E against unscrambled prod-clone — `rejected_categories` populated by the post-#502 mirror is read correctly by the engine; confirm a merchant whose user-rejection history blocks the suggestion gate

### DIFF
--- a/src/server/lib/compute-tools/auto-suggest.test.ts
+++ b/src/server/lib/compute-tools/auto-suggest.test.ts
@@ -483,4 +483,93 @@ describe("runAutoSuggestions", () => {
       expect(matches.length).toBeGreaterThanOrEqual(2);
     });
   });
+
+  // ──────────────────────────────────────────────────────────────────────
+  // Merchant signal SQL shape (Stage 2b: rejected reads from
+  // rejected_categories instead of transactions.label_category_confidence)
+  // ──────────────────────────────────────────────────────────────────────
+
+  describe("merchant signal SQL shape", () => {
+    const captureSignalSql = async () => {
+      userRows = [userRow({ user_id: "u-shape" })];
+      unlabeledByUser.set("u-shape", [
+        txRow({ transaction_id: "txn-shape", user_id: "u-shape", merchant_name: "Probe Co" }),
+      ]);
+      signalRow = {
+        label_category_id: "cat-shape",
+        label_budget_id: "bud-shape",
+        accepted: "5",
+        rejected: "0",
+      };
+      await runAutoSuggestions();
+      const signalQuery = mockQuery.mock.calls.find((c) =>
+        /similarity\(merchant_name/i.test(c[0] as string),
+      );
+      expect(signalQuery).toBeDefined();
+      return signalQuery![0] as string;
+    };
+
+    test("ACCEPTED is read from transactions WHERE label_category_confidence = 1.0 only", async () => {
+      const sql = await captureSignalSql();
+      // Inner pool of confirmed transactions narrows on conf=1.0 — not
+      // `IS NOT NULL` (the pre-Stage-2b shape) and not `= 0.0` (the old
+      // rejected-as-conf-zero pattern, now retired).
+      expect(sql).toMatch(/label_category_confidence\s*=\s*1\.0/);
+      expect(sql).not.toMatch(/label_category_confidence\s+IS NOT NULL/);
+      expect(sql).not.toMatch(/label_category_confidence\s*=\s*0\.0/);
+    });
+
+    test("REJECTED is read from rejected_categories joined to transactions for merchant match", async () => {
+      const sql = await captureSignalSql();
+      // The subquery for `rejected` references the new table directly,
+      // joins to `transactions` for the merchant filter, and scopes by
+      // user_id on BOTH sides (defense-in-depth against future schema
+      // changes that could let a transaction_id appear under another user).
+      expect(sql).toMatch(/FROM\s+rejected_categories\s+rc/);
+      expect(sql).toMatch(
+        /JOIN\s+transactions\s+t\s+ON\s+t\.transaction_id\s*=\s*rc\.transaction_id/,
+      );
+      expect(sql).toMatch(/rc\.user_id\s*=\s*\$1/);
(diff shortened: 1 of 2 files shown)


reviewoie:
## Code review — PR 506

**Verdict:** approve (informational notes only).

**Findings:** 3 total — 0 high, 0 med, 3 low/info.

### High
No blockers.

### Med
No blockers.

### Low / informational

- **`src/server/lib/compute-tools/auto-suggest.ts:316-318` — the `AND t.user_id = rc.user_id` ON-clause is redundant given the schema.** `transactions.transaction_id` is `VARCHAR(255) PRIMARY KEY` (one row per id), and `rejected_categories.transaction_id REFERENCES transactions(transaction_id) ON DELETE CASCADE`. So `t.transaction_id = rc.transaction_id` already pins `t` to a single row whose `user_id` is uniquely determined by the FK. Combined with `rc.user_id = $1`, the equality is a tautology. Test comment at `auto-suggest.test.ts:525-527` calls it "defense-in-depth against future schema changes that could let a transaction_id appear under another user" — that's a fair design call, the redundant predicate is cheap and the joined-`rc.user_id`-condition gives the optimizer an extra constant to push into the rc index lookup. Not a bug; just noting it IS redundant under the current schema. Leave as-is.

- **`src/server/lib/compute-tools/auto-suggest.ts:202-211` — bounded-`accepted` ÷ unbounded-`rejected` interacts with the 10% reject-rate gate in a way the PR body should call out.** `evaluateSignal` computes `rejectRate = rejected / (accepted + rejected)` where `accepted ≤ MERCHANT_SIGNAL_LIMIT (30)` and `rejected` is now all-time. Concrete failure scenario: a long-time user has 30 recent confirms of category X for "Starbucks #1234" AND 4 ancient rejections of category X for similar merchants. New: `4/(30+4) = 11.8% > 10%` → suggestion blocked indefinitely until the user re-picks X (which calls `removeRejectedCategory`). Old behavior on the same data would have looked only at the top-30 recent pool, so ancient rejections were invisible to the gate. This is intentional per the PR body ("explicit user feedback should stick") and the `removeRejectedCategory`-on-re-confirm path is the documented escape hatch, but it's a real behavior shift — worth a one-liner in the PR body so future readers see the trade-off without re-deriving it from the SQL. Optional mitigation if it bites in practice: scope the rejected subquery to `rc.rejected_at >= NOW() - INTERVAL 'N months'` — but I would NOT recommend doing it in this PR; ship the simple shape, watch real data, add a window only if needed.

- **`src/server/lib/compute-tools/auto-suggest.test.ts:514-519` — minor comment-hygiene drift in the new test block.** "the pre-Stage-2b shape" and "the old rejected-as-conf-zero pattern, now retired" describe the prior PR's state, which is the PR-narrative pattern Hoie called out in the comment-hygiene rule. The negative-assertion test (`expect(sql).not.toMatch(/= 0\.0/)`) does need *some* context to justify why the absence matters, so this isn't pure PR narrative — but tightening to "rejection is read from `rejected_categories`, not from `transactions.label_category_confidence` (any value)" would carry the same forward-looking constraint without dating the comment. Not a blocker.

### Verified

- SQL correctness: the CTE shape returns at most one row with exactly the columns the consumer expects (`label_category_id`, `label_budget_id`, `accepted`, `rejected`). When `winning` is empty (no `label_category_confidence = 1.0` rows in the top-30 similar-merchant pool), the outer `SELECT … FROM winning w` yields zero rows → `result.rows.length === 0` → returns `null`. Matches the pre-Stage-2b behavior the user described.
- Multi-user safety on the rejected subquery: even with the redundant ON-clause aside, `rc.user_id = $1` plus the FK relationship guarantees no cross-user leakage.
- Soft-deleted exclusion: confirmed on BOTH sides — accepted side has `(is_deleted IS NULL OR is_deleted = FALSE)` in the inner WHERE; rejected side has `(t.is_deleted IS NULL OR t.is_deleted = FALSE)` on the joined `transactions`. The test at lines 547-555 asserts at least 2 matches of the predicate — verified the SQL satisfies it.
- Parameter ordering: `$1=userId`, `$2=merchantName`, `$3=threshold`, `$4=limit` — used consistently in both the CTE and the rejected subquery; only 4 placeholders, only 4 array entries.
- No callers of `getMerchantSignal` outside this module — return-shape change is contained.
- Hard delete cascades via FK on `rejected_categories.transaction_id`, so deleted txns never contribute stale rejections. Soft delete handled by the explicit `is_deleted` filter.
- Test suite: 24 / 24 pass locally (`bun test src/server/lib/compute-tools/auto-suggest.test.ts` → 24 pass, 0 fail, 66 expects, ~420ms).
- Scope: only `auto-suggest.ts` + `auto-suggest.test.ts` touched — no drive-bys.
- Comment hygiene in `auto-suggest.ts`: the rewritten doc-block describes invariants/design (two-source signal, raw-SQL justification, the categories→sections join's purpose) without naming Hoie, the PR, or the bug being fixed. Clean.

hoiekim (approved):

PR #508, opened 2026-06-11
feat(scheduler): refresh active security snapshots on each scheduled sync (#315)

### LINKED ISSUE #315: feat(holdings): scheduled daily close-price fetch + historical backfill for manual securities
Tracked follow-up work from the [Manual Account Tracking design doc](https://doc.hoie.kim/p/oCvYlrgoyJ), filed now that the old-issues-first creation gate is lifted.

## Background

Step 2 of the design ("Security Snapshot Fetching") promises:

> When a manual holding is added:
> - Validate ticker symbol against market data API
> - If valid, look up or create security record in DB
> - Queue security for price fetching:
>   - **Daily close price from market data API**
>   - **Backfill historical prices from holding creation date**

Today, only the synchronous parts are implemented:

- ✅ Ticker validation: `src/server/routes/accounts/post-validate-ticker.ts` calls `polygon.getTickerDetail` + `polygon.getClosePrice` for the current date.
- ✅ Single-snapshot price lookup: `post-holding-snapshot.ts` calls `polygon.getSecurityForSymbol(ticker, date)` for the snapshot's date only.
- ❌ **No scheduled job** picks up tracked securities and pulls daily close prices over time.
- ❌ **No backfill** runs when a new manual security is added — historical balances for that account remain flat at 0 for any view date before the user's first manual snapshot.

`compute-tools/schedule.ts` only orchestrates `syncPlaidAccounts`/`syncPlaidTransactions` and `syncSimpleFinData` — Polygon is never called from the scheduler.

## Why it matters

The whole point of the Manual Account Tracking design was for manual investment account balances to update dynamically as market prices change. Without a scheduler:
- Yesterday's balance == today's balance == last-snapshot's balance, even though the underlying security moved.
- Historical view dates before the user's first manual entry have no holding-snapshot fallback to pull from.

So the Holdings Value design's 3-tier fallback (account snapshot → holding snapshot → transactions) silently degrades to "manual snapshot only" for these accounts.

## Proposed implementation

1. **Scheduled job** in `compute-tools/schedule.ts` (or a sibling `sync-polygon.ts`):
   - Once per day, list all `securities` rows that have at least one referenced `holding_snapshot` for a manual account.
   - For each, call `polygon.getClosePrice(ticker, today)` and write a new `security_snapshot` row.
   - Skip on weekends/market holidays (Polygon returns the prior business day's close).
2. **Backfill on holding creation** (in `post-holding-snapshot.ts` after security upsert):
   - Asynchronously enqueue a backfill from the user's earliest holding date for that security to today.
   - Polygon's daily-close history endpoint supports a date range — one call per security.
   - Best-effort: failures shouldn't block the holding-snapshot creation.

## Acceptance criteria

- New scheduled-sync entry that updates security snapshots daily for manual-tracked securities
- Backfill triggered when a new ticker is first added via `post-holding-snapshot.ts`
- Test coverage for both the scheduler entry point and the backfill helper
- E2E: add a manual holding for a known ticker (e.g. AAPL), verify a `security_snapshot` row appears after a daily run, and verify the account balance reflects the new price on the next view-date load

## Reference

- Design doc: https://doc.hoie.kim/p/oCvYlrgoyJ (Step 2)
- Holdings Value design 3-tier fallback: https://doc.hoie.kim/p/ZDhNQZtVic
- Existing one-shot Polygon call site: `src/server/routes/accounts/post-holding-snapshot.ts:60`

claoie:
## Summary

Closes #315 (scheduled side).

Adds `refreshActiveSecuritySnapshots` — every scheduled-sync run, walk every `security_id` referenced by a non-deleted holding row, ask polygon for the latest trading-day close, and upsert a fresh `security_snapshot` if one doesn't already exist for the polygon-returned `tradingDate`.

- **Cadence.** Wired into `scheduledSync` after the auto-suggest + transfer-detect passes, so it rides the existing hourly cron. Per-row cadence is enforced by the existence check on `tradingDate` — once today's snapshot lands, subsequent hourly runs early-out with one cheap SELECT per security.
- **Plaid coexistence.** Plaid-tracked securities surface in the holdings JOIN but their snapshots are already written during sync via `upsertSecuritiesWithSnapshots`. The existence check naturally early-outs them — no double-writes.
- **Cash skip.** `type='cash'` or ticker starting `CUR:` is skipped — polygon doesn't price them.
- **Rate-limit inheritance.** Uses the shared polygon queue. A backfill that needs 30 calls on a free-tier key takes ~6 minutes; intentional, prevents 429.

## What's NOT in this PR

The on-add backfill the issue mentions is already shipped at `post-holding-snapshot.ts:221` via `backfillMonthlySecuritySnapshotsForward(refs).catch(...)` — runs forward from a holding's `snapshot_date` to today on every holding-snapshot post. So this PR closes the remaining "scheduled daily" half of the issue.

## Test plan

- [x] 11 unit tests via the established `pg`-mock + `globalThis.fetch` pattern (mirrors `backfill-snapshots.test.ts`):
  - no holdings → no polygon call
  - one ticker → polygon called, snapshot upserted; snapshot_id derived from polygon's `tradingDate` (idempotency anchor)
  - cash `type='cash'` skipped
  - cash `CUR:*` ticker without `type='cash'` also skipped
  - existing snapshot for `tradingDate` → `fresh` increment, no upsert
  - polygon `no_data` → counted as `empty`
  - polygon `NOT_AUTHORIZED` → counted as `error`
  - upsert failure → counted as `error`, loop continues to next security
  - multi-security mix (cash + fresh + new)
  - tickerless securities skipped (no polygon call, no error)
  - per-(security_id, tradingDate) existence scoping
- [x] Full suite 833/833 pass, no cross-suite leak (`POLYGON_API_KEY` + `POLYGON_RATE_LIMIT_PER_MIN` restored in `afterAll`)
- [x] Typecheck + lint clean
- [x] Sandbox E2E: spin sandbox, exercise `scheduledSync` via `bun run dev` lifecycle, verify a manual-added holding gets a fresh security_snapshot on the next sync cycle

### DIFF
--- a/src/server/lib/compute-tools/refresh-security-snapshots.test.ts
+++ b/src/server/lib/compute-tools/refresh-security-snapshots.test.ts
@@ -46,12 +46,7 @@ const mockFetch = mock(
 );
 globalThis.fetch = mockFetch as unknown as typeof globalThis.fetch;
 
-const { refreshActiveSecuritySnapshots, clearPriceCache } = await import(
-  "./refresh-security-snapshots"
-).then(async (m) => {
-  const polygon = await import("../polygon");
-  return { ...m, clearPriceCache: polygon.clearPriceCache };
-});
+const { refreshActiveSecuritySnapshots } = await import("./refresh-security-snapshots");
 
 afterAll(() => {
   globalThis.fetch = originalFetch;
@@ -63,12 +58,13 @@ afterAll(() => {
 });
 
 // SQL router. Each test stages:
-//   - holdingsRows   → DISTINCT security_id query result
+//   - holdingsRows  → DISTINCT security_id query result
 //   - securitiesRows → per-id securities row (one queryOne per id in searchSecuritiesById)
-//   - existingByPair → existence probe set, key = `${security_id}|${tradingDate}`
+//   - recentBySecurity → cadence-gate set; security_ids in the set are
+//     treated as having a snapshot whose `updated` is within the window.
 let holdingsRows: Array<{ security_id: string }> = [];
 let securitiesRows: Array<Record<string, unknown>> = [];
-let existingByPair = new Set<string>();
+let recentBySecurity = new Set<string>();
 const upsertCalls: Array<{ sql: string; values: unknown[] }> = [];
 
 const queryRouter = async (sql: string, values?: unknown[]) => {
@@ -84,14 +80,11 @@ const queryRouter = async (sql: string, values?: unknown[]) => {
     return { rows: row ? [row] : [], rowCount: row ? 1 : 0 };
   }
 
-  if (
-    /FROM\s+snapshots\s+WHERE\s+security_id\s*=\s*\$1\s+AND\s+snapshot_type\s*=\s*'security'/is.test(
-      sql,
-    )
-  ) {
-    const key = `${params[0]}|${params[1]}`;
-    const found = existingByPair.has(key);
-    return { rows: found ? [{ snapshot_id: "exists" }] : [], rowCount: found ? 1 : 0 };
+  // Cadence-gate probe — `… updated > NOW() - INTERVAL 'N hours' …`.
+  if (/updated\s*>\s*NOW\(\)\s*-\s*INTERVAL/i.test(sql)) {
+    const wantId = params[0] as string;
+    const found = recentBySecurity.has(wantId);
+    return { rows: found ? [{ snapshot_id: "recent" }] : [], rowCount: found ? 1 : 0 };
   }
 
   if (/INSERT\s+INTO\s+snapshots/i.test(sql)) {
@@ -119,10 +112,9 @@ beforeEach(() => {
         status: 200,
       }),
   );
-  clearPriceCache();
   holdingsRows = [];
   securitiesRows = [];
-  existingByPair = new Set();
+  recentBySecurity = new Set();
   upsertCalls.length = 0;
 });
 
@@ -198,16 +190,35 @@ describe("refreshActiveSecuritySnapshots", () => {
     expect(r.cash).toBe(1);
   });
 
-  test("snapshot already exists for tradingDate → no upsert, increments `fresh`", async () => {
+  test("cadence gate skips polygon when a recent snapshot exists for this security", async () => {
+    // Steady-state hourly cycle: the previous run wrote a snapshot
+    // less than the skip window ago. Polygon must NOT be called.
+    // Without this gate, the cron would burn N polygon calls every
+    // hour even when nothing needs refreshing.
     holdingsRows = [{ security_id: "sec-1" }];
     securitiesRows = [securityRow({ security_id: "sec-1", ticker_symbol: "VOO" })];
-    existingByPair.add("sec-1|2026-06-10");
+    recentBySecurity.add("sec-1");
 
     const r = await refreshActiveSecuritySnapshots();
-    expect(mockFetch).toHaveBeenCalledTimes(1);
+    expect(mockFetch).not.toHaveBeenCalled();
     expect(upsertCalls).toHaveLength(0);
     expect(r.fresh).toBe(1);
     expect(r.refreshed).toBe(0);
+  });
+
+  test("cadence-gate query uses `updated > NOW() - INTERVAL` (not snapshot_date) — survives weekends without re-fetching", async () => {
+    // Pin the SQL shape so the gate semantics can't silently regress
+    // to a snapshot_date check (which would erroneously force a
+    // polygon call every Saturday/Sunday/holiday when no new trading
+    // day has landed but the previous run's snapshot looks "old").
+    holdingsRows = [{ security_id: "sec-1" }];
+    securitiesRows = [securityRow({ security_id: "sec-1", ticker_symbol: "VOO" })];
+    await refreshActiveSecuritySnapshots();
+    const gateCall = mockQuery.mock.calls.find((c) =>
+      /updated\s*>\s*NOW\(\)\s*-\s*INTERVAL/i.test(c[0] as string),
+    );
+    expect(gateCall).toBeDefined();
+    expect(gateCall![0]).toMatch(/snapshot_type\s*=\s*'security'/i);
   });
 
   test("polygon `no_data` → counted as empty, no upsert", async () => {
@@ -263,13 +274,15 @@ describe("refreshActiveSecuritySnapshots", () => {
       securityRow({ security_id: "sec-fresh", ticker_symbol: "VOO" }),
       securityRow({ security_id: "sec-new", ticker_symbol: "NVDA" }),
     ];
-    existingByPair.add("sec-fresh|2026-06-10");
+    recentBySecurity.add("sec-fresh");
 
     const r = await refreshActiveSecuritySnapshots();
     expect(r.cash).toBe(1);
     expect(r.fresh).toBe(1);
     expect(r.refreshed).toBe(1);
     expect(upsertCalls).toHaveLength(1);
+    // sec-fresh hit the cadence gate; only sec-new should have called polygon.
+    expect(mockFetch).toHaveBeenCalledTimes(1);
   });
 
   test("tickerless securities skipped (no polygon call, no error)", async () => {
@@ -280,13 +293,13 @@ describe("refreshActiveSecuritySnapshots", () => {
     expect(r).toEqual({ refreshed: 0, fresh: 0, cash: 0, empty: 0, errors: 0 });
   });
 
-  test("existence probe scopes by (security_id, tradingDate)", async () => {
+  test("cadence gate is per-security: skipping sec-a does not skip sec-b", async () => {
     holdingsRows = [{ security_id: "sec-a" }, { security_id: "sec-b" }];
     securitiesRows = [
       securityRow({ security_id: "sec-a", ticker_symbol: "AAPL" }),
       securityRow({ security_id: "sec-b", ticker_symbol: "MSFT" }),
     ];
-    existingByPair.add("sec-a|2026-06-10");
+    recentBySecurity.add("sec-a");
 
     const r = await refreshActiveSecuritySnapshots();
     expect(r.fresh).toBe(1);
--- a/src/server/lib/compute-tools/refresh-security-snapshots.test.ts
+++ b/src/server/lib/compute-tools/refresh-security-snapshots.test.ts
@@ -1,0 +1,296 @@
+//
+// `refreshActiveSecuritySnapshots` reads holdings + securities from
+// `pool.query` / `searchSecuritiesById` and calls
+// `polygon.getLatestClosePriceOnOrBefore` for each non-cash security.
+// The bundle inlines polygon, so we leaf-mock `pg` for the DB layer and
+// `globalThis.fetch` for polygon. The polygon module reads
+// `process.env.POLYGON_API_KEY` / `POLYGON_RATE_LIMIT_PER_MIN` once at
+// module load — set them BEFORE importing so:
+//   1. polygon doesn't short-circuit `no_api_key`.
+//   2. The rate-limit queue runs at capacity=0 (no throttling).
+//
+// The fetch override + env writes are process-global; `afterAll`
+// restores the originals so this file doesn't leak into other suites.
+const originalFetch = globalThis.fetch;
+const originalApiKey = process.env.POLYGON_API_KEY;
+const originalRateLimit = process.env.POLYGON_RATE_LIMIT_PER_MIN;
+
+process.env.POLYGON_API_KEY = "test-key";
+process.env.POLYGON_RATE_LIMIT_PER_MIN = "0";
+
+import { describe, test, expect, mock, beforeEach, afterAll } from "bun:test";
+import { restoreLeaves } from "test-helpers";
+
+const mockQuery = mock(async (_sql: string, _values?: unknown[]) => ({
+  rows: [] as unknown[],
+  rowCount: 0 as number | null,
+}));
+
+class FakePool {
+  query = mockQuery;
+  end = async () => {};
+  connect = async () => ({ query: mockQuery, release: () => {} });
+}
+
+mock.module("pg", () => ({
+  Pool: FakePool,
+  types: { setTypeParser: () => {} },
+  default: { Pool: FakePool, types: { setTypeParser: () => {} } },
+}));
+
+const mockFetch = mock(
+  async (_url: string | URL | Request, _init?: RequestInit): Promise<Response> =>
+    new Response(JSON.stringify({ results: [{ c: 100, t: Date.UTC(2026, 5, 10) }] }), {
+      status: 200,
+    }),
+);
+globalThis.fetch = mockFetch as unknown as typeof globalThis.fetch;
+
+const { refreshActiveSecuritySnapshots, clearPriceCache } = await import(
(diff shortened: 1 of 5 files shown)


reviewoie:
## Code review — PR 508

**Verdict:** changes requested

**Findings:** 3 total — 1 high, 1 med, 1 low.

### High
- `src/server/lib/compute-tools/refresh-security-snapshots.ts:79` — polygon is called BEFORE the existence check, not after. The PR body's headline claim ("subsequent hourly runs early-out with one cheap SELECT per security") is contradicted by the code: every hourly cycle fires N polygon calls for N referenced securities, regardless of whether today's snapshot already exists. Free-tier (5/min) users with a 30-security portfolio spend ~6 minutes of polygon quota per cycle, every cycle, forever — even on days where nothing needs writing. Compare `backfill-snapshots.ts:103-126`, which loads `getSecuritySnapshots({ security_id })` first, builds a `haveMonth` Set, and only calls polygon for missing months. The test at line 201-211 (`mockFetch toHaveBeenCalledTimes(1)` on the already-fresh case) encodes the wasted-fetch behavior as expected — so the test passes while the behavior is wrong.

  Suggested fix: do a cheap pre-check against today's *local* date before calling polygon (writing today's local date as a "we already covered today" sentinel via either `getDateString(today)` or, accepting a small redundancy, the prior hour's known `tradingDate`). Then only call polygon if no current-day snapshot exists. Re-check polygon's returned `tradingDate` after the fetch for the per-trading-day correctness anchor you already have.

### Med
- `src/server/lib/compute-tools/refresh-security-snapshots.ts:140` — `upsertSnapshots`'s update column list does NOT include `is_deleted` (see `snapshots.ts:308-316`). If a `(security_id, trading_date)` row exists but is soft-deleted, the existence check correctly returns 0 rows, the function builds a fresh `JSONSecuritySnapshot`, and `snapshotsTable.upsert` ON CONFLICTs against the soft-deleted row and updates only `snapshot_date`/`snapshot_type`/`security_id`/`close_price` — leaving `is_deleted=true`. Next hourly cycle: existence check again returns 0, same re-upsert fires again. Stuck loop until something else flips `is_deleted` back. Likely rare for security snapshots but worth a thought — either include `is_deleted: false` in the snapshot payload (then upsert will overwrite the tombstone) or document the assumption "security snapshots are never soft-deleted in this code path".

### Low
- `src/server/lib/compute-tools/refresh-security-snapshots.test.ts:122` — `clearPriceCache()` in `beforeEach` clears the cache used by `getClosePrice`, but `refreshActiveSecuritySnapshots` calls `getLatestClosePriceOnOrBefore`, which does NOT use that cache. The call is harmless but misleading — a future reader will assume `getLatestClosePriceOnOrBefore` is cached.

### Notes (not findings, just signals worth recording)

- `upsertResult.status >= 400` correctly captures both `errorResult` (500) and the rowCount=0 case from `successResult` (404). Verified against `database.ts:409-433`. The test "upsert failure → counted as error" passes for the right reason.
- Plaid coexistence is correct: Plaid writes `${security_id}-${getSquashedDateString()}` (today's local date) with `date = new Date().toISOString()`; PR writes `${security_id}-${getSquashedDateString(tradingDateObj)}` (polygon's trading date). On a normal weekday these collide on the same `snapshot_id` and the upsert is idempotent. On a Monday-morning sync where polygon returns Friday's bar, the IDs differ and the function correctly writes Friday's snapshot without conflict.
- `searchSecuritiesById` not being user-scoped is correct — securities and security-snapshots are global lookup tables, no `user_id` column, one row per `(security_id, trading_day)` is the intended cardinality.
- Comment hygiene clean: no PR-narrative comments, no name attribution.
- No `#N` for non-issue numbering in test names or comments. Clean.
- 11 unit tests pass; full `src/server/lib/compute-tools` suite (114 tests) passes with no leak.

src/server/lib/compute-tools/refresh-security-snapshots.ts
```diff
@@ -0,0 +1,154 @@
...
+      continue;
+    }
+
+    // Ask polygon for the latest trading day's close at or before today.
+    // The returned `tradingDate` is the actual market day the price is
+    // anchored to — naturally falls back to Friday on weekends, the day
+    // before a holiday, etc. Writing the snapshot under `tradingDate`
+    // means subsequent same-day runs early-out via the existence check
+    // below; once a new trading day closes, the next run picks it up.
+    const fetchResult = await polygon.getLatestClosePriceOnOrBefore(ticker_symbol, today);
```
reviewoie: **HIGH** — polygon is called BEFORE the existence check, not after. The PR body's headline claim ("subsequent hourly runs early-out with one cheap SELECT per security") is contradicted by this ordering: every hourly cycle fires N polygon calls for N referenced securities, regardless of whether today's snapshot already exists. Free-tier (5/min) users with a 30-security portfolio burn ~6 minutes of polygon quota per cycle, every cycle, forever — even on days where nothing needs writing.

Compare `backfill-snapshots.ts:103-126`: that function loads `getSecuritySnapshots({ security_id })` first, builds a `haveMonth` Set, and only calls polygon for missing months. That's the correct pattern.

The test at refresh-security-snapshots.test.ts:201-211 (`expect(mockFetch).toHaveBeenCalledTimes(1)` on the already-fresh case) encodes the wasted-fetch behavior as expected — so the suite passes while the behavior is wrong.

Fix sketch: do a cheap pre-check on today's local date (`getDateString(today)`) before calling polygon; only call polygon when no current-day snapshot exists. After the polygon fetch, the per-trading-day correctness anchor you already have stays as the upsert key.

src/server/lib/compute-tools/refresh-security-snapshots.ts
```diff
@@ -0,0 +1,154 @@
...
+        industry: null,
+        option_contract: null,
+        fixed_income: null,
+      },
+    };
+
+    // `upsertSnapshots` catches per-snapshot errors internally and
+    // returns one `UpsertResult { status }` per row instead of throwing.
+    // A 4xx / 5xx status means the row didn't land.
+    const [upsertResult] = await upsertSnapshots([snapshot]);
```
reviewoie, line 156: **MED** — soft-delete loop hazard.

`upsertSnapshots`'s update column list (`snapshots.ts:308-316`) does not include `is_deleted`. If a `(security_id, snapshot_date)` row exists but is soft-deleted:
1. The existence check at line 89-97 correctly returns 0 rows (the `is_deleted IS NULL OR FALSE` gate filters it).
2. The fresh `JSONSecuritySnapshot` is built and `upsertSnapshots([snapshot])` is called.
3. `snapshotsTable.upsert` runs `ON CONFLICT (snapshot_id) DO UPDATE SET snapshot_date=…, snapshot_type=…, security_id=…, close_price=…` — `is_deleted` stays `true`.
4. Next hourly cycle: existence check again returns 0, same re-upsert fires again. Stuck loop until something else flips `is_deleted` back to false.

Free polygon quota leak compounds with the HIGH finding. Likely rare for security snapshots in practice but worth either (a) including `is_deleted: false` in the payload so the upsert overwrites the tombstone, or (b) explicitly documenting "security snapshots are never soft-deleted in this code path" if that's the design.

src/server/lib/compute-tools/refresh-security-snapshots.test.ts
```diff
@@ -0,0 +1,296 @@
...
+  mockQuery.mockReset();
+  mockQuery.mockImplementation(queryRouter);
+  mockFetch.mockReset();
+  mockFetch.mockImplementation(
+    async () =>
+      new Response(JSON.stringify({ results: [{ c: 100, t: Date.UTC(2026, 5, 10) }] }), {
+        status: 200,
+      }),
+  );
+  clearPriceCache();
```
reviewoie: **LOW** — misleading cache reset.

`clearPriceCache()` clears the in-memory cache used by `getClosePrice` (`polygon.ts:62-71`), but `refreshActiveSecuritySnapshots` calls `getLatestClosePriceOnOrBefore`, which does not use that cache. The call is harmless but misleading — a future reader will assume `getLatestClosePriceOnOrBefore` is cached and that this reset is load-bearing. Either drop the line or add a one-liner explaining it's a defensive reset for the broader polygon module state.

reviewoie:
## Code review — PR 508 (re-review, HEAD cd71679)

**Findings:** 2 new (since the 05:10 review) — 0 high, 1 med, 1 low. Prior HIGH resolved; prior MED bounded.

### Prior-review status
- **Prior HIGH (polygon called before existence check) — RESOLVED.** Commit `929e0d7b` flips the order: the per-security cadence gate (`SELECT … FROM snapshots WHERE security_id=$1 AND snapshot_type='security' AND updated > NOW() - INTERVAL '22 hours' LIMIT 1`, refresh-security-snapshots.ts:94-101) now runs BEFORE the polygon call (line 111). Steady-state hourly cycles do one cheap SELECT per security and skip polygon. Verified the test `cadence gate skips polygon when a recent snapshot exists` (refresh-security-snapshots.test.ts:193) asserts `mockFetch` is NOT called on the fresh path. Good fix, and the `updated`-vs-`snapshot_date` choice (survives weekends without re-fetching) is sound.
- **Prior MED (soft-delete loop) — BOUNDED, not fully closed.** The security upsert (snapshots.ts:310-316) still omits `is_deleted` from its column set, so an ON CONFLICT against a tombstoned row leaves `is_deleted=true`. The cadence gate doesn't filter `is_deleted`, and every upsert bumps `updated`, so a soft-deleted row stays gated for 22h, then triggers exactly one wasted polygon call + one no-op re-upsert each window — perpetually, but bounded (no tight loop). The commit message acknowledges this. Realistically security snapshots are never soft-deleted, so this is acceptable; flagging only so it's on record.

### Med
- `src/server/lib/compute-tools/schedule.test.ts:130-137` — the `afterAll` restore block does NOT restore `./refresh-security-snapshots`, but commit `cd716792` added `mock.module("./refresh-security-snapshots", …)` at lines 82-85. The file's own header comment (lines 5-9) establishes the invariant that every relative-path `mock.module` must be re-mocked to real in `afterAll`, because Bun's `mock.module` is process-global with no unmock and `restoreLeaves()` only restores `pg`/`bcrypt` (verified: scripts/test-helpers.ts:49-54). Every sibling module (`server`, `server/lib/alarm`, `./sync-plaid`, `./sync-simple-fin`, `./auto-suggest`, `./detect-transfers`) is restored — `./refresh-security-snapshots` is the one omission. Failure scenario: any future test file ordered after `schedule.test.ts` that invokes the real `refreshActiveSecuritySnapshots` (directly or through the `server` barrel) gets the stub returning all-zero counters and silently asserts against a no-op. Benign today only because the sole consumer (`refresh-security-snapshots.test.ts`) sorts BEFORE `schedule.test.ts`; full suite 834/834 still passes. Fix: add `mock.module("./refresh-security-snapshots", () => realRefreshSecuritySnapshots);` to the `afterAll` block. Verified via Read + full-suite run.

### Low
- `src/server/lib/compute-tools/schedule.test.ts` — the PR's core wiring (refresh runs on each scheduled sync) has no test assertion. `mockRefreshActiveSecuritySnapshots` was added purely to stop the real pg call, but unlike `runAutoSuggestions`/`runTransferDetection` (asserted in the test at line 237: `expect(mockRunAutoSuggestions).toHaveBeenCalledTimes(1)`), nothing asserts `expect(mockRefreshActiveSecuritySnapshots).toHaveBeenCalled()`. The headline behavior is unverified. Recommend extending the line-237 test (or adding one) with that assertion. Verified: grep for `mockRefreshActiveSecuritySnapshots` shows only mock setup/reset, no `toHaveBeenCalled`.

### Notes
- Scope clean: all 5 files (refresh impl + test, schedule wiring + test, index barrel export) map directly to the PR title. No unrelated changes.
- CI green at HEAD cd71679 (build pass, test pass). The `cd716792` diagnosis ("role 'postgres' does not exist" because the real refresh reached pg in CI) was correct and the mock resolves it.
- Local verification: full `src` suite 834/834 pass, compute-tools 115/115 pass, `tsc --noEmit` clean, `eslint src/server/lib/compute-tools` clean.
- Test names verified by grep (no fabrication); new cadence-gate tests are accurate to behavior.

claoie:
Addressed both reviewoie findings in `9f0e646`:

- **med (test isolation):** `afterAll` now restores `./refresh-security-snapshots` too. It was the only mocked relative-path sibling missing from the restore block, so its all-zero stub leaked into any later test file in the same Bun process (`mock.module` is process-global). The `realRefreshSecuritySnapshots` snapshot was already captured at the top of the file; just wired it into the restore.
- **low (coverage gap):** the post-sync hooks test now also asserts `refreshActiveSecuritySnapshots` is called once per scheduled sync, alongside the existing auto-suggest / transfer-detection assertions. The PR's headline wiring was previously unverified.

`bun test src/server` → 601 pass / 0 fail, `tsc --noEmit` exit 0.

reviewoie:
## Code review — PR 508 (re-review, HEAD 9f0e646)

**Verdict:** approve

**Findings:** 0 new. Both prior findings from the cd71679 review are verified resolved.

### Prior finding 1 (med) — afterAll omitted `./refresh-security-snapshots`, leaking an all-zero stub process-globally
RESOLVED. `realRefreshSecuritySnapshots` is captured at the top of the file (schedule.test.ts line 16), and the `afterAll` restore block now re-registers all 7 relative/bare-path `mock.module` siblings — `server`, `server/lib/alarm`, `./sync-plaid`, `./sync-simple-fin`, `./auto-suggest`, `./detect-transfers`, and `./refresh-security-snapshots` — exactly matching the 7 `mock.module` registrations at lines 49-82. No process-global bleed remains.

Verified empirically: ran `refresh-security-snapshots.test.ts` + `schedule.test.ts` + `sync-plaid.test.ts` (alphabetically later) together in one Bun process via the real `--preload ./scripts/test-preload.ts` harness — 43 pass, 0 fail, no cross-file leak.

### Prior finding 2 (low) — headline wiring had no assertion
RESOLVED. `expect(mockRefreshActiveSecuritySnapshots).toHaveBeenCalledTimes(1)` added to the renamed "runs auto-suggestions, transfer detection, and security-snapshot refresh after provider sync" test (schedule.test.ts line 249), and `runSync` calls `refreshActiveSecuritySnapshots()` unconditionally after the auto-suggest / transfer-detect passes (schedule.ts line 97).

Verified the assertion is non-vacuous via a mutation: replacing the `await refreshActiveSecuritySnapshots()` call with `await Promise.resolve()` makes the test fail with "Received number of calls: 0". It genuinely guards the wiring.

### Other checks
- **Scope (rule 2):** clean. All 5 files (helper + its test, barrel export, scheduler wiring, scheduler test) directly serve the PR title. No unrelated changes.
- **Test framework:** the new `refresh-security-snapshots.test.ts` follows the live convention — `mock.module("pg", ...)` + `restoreLeaves()` from `test-helpers` under `bun test --preload ./scripts/test-preload.ts`, identical to the closest siblings `backfill-snapshots.test.ts` and `cash-holding.test.ts`. No bundle-framework regression.
- **Latency path:** N/A — this rides the scheduler cron, not a request handler; the `.then().catch()` fire-and-forget error handling on the refresh call is appropriate.
- **CI:** build + test both green at HEAD 9f0e646.

No further blockers.

claoie:
Reviewoie verdict: approve clean (no new findings; both prior findings verified resolved at HEAD 9f0e646). Ready to merge.

hoiekim (approved):

PR #509, opened 2026-06-11
fix(budgets): clamp rollover accrual to loaded history on cold load (#484)

### LINKED ISSUE #484: bug(budgets): cold-load shows grossly inflated rollover (~12×) until full transaction history loads — capacity accrues over full span, spending only over loaded months
## Summary

On a **cold load** (empty IndexedDB — first visit, cleared cache, new device, incognito), every rolling budget's **"+ $X rolled"** figure renders a grossly inflated value and only corrects itself once the *entire* transaction history has finished streaming in. For the prod-clone `admin` data the most prominent budget shows a rolled figure roughly **~12× too high** for ~10–30s, then snaps to the correct steady-state value — a large overstatement of a headline financial number.

The wrong value is not merely "partial data shown as partial." It is a genuine miscalculation: the rollover loop accrues the monthly **capacity** across the *full* history span (from `roll_over_start_date`), but the **spending** that offsets that capacity only exists for the *subset of months whose transactions have loaded so far*. During the cold-load stages that subset is just the most recent ~2 months, so the capacity side spans years while the spending side spans weeks.

## Reproduction (live, prod-clone `admin`)

1. Clear IndexedDB (or first-ever login). Log in.
2. Land on **Budgets**. Observe a rolling budget (`roll_over_start_date = 2022-06-01`): its **`+ $X rolled`** figure shows the inflated cold-load value.
3. Wait for the full transaction history to finish loading (cold-load Stage 3), or navigate Budgets → Budget detail → back.
4. The same bar now reads the correct (~12× smaller) steady-state value and stays there permanently.

Warm loads never show the wrong value: the warm path paints from the cached IndexedDB snapshot, which already holds the full history, so the first paint is correct.

## Evidence (instrumented `getBudgetData`, then reverted)

Logged the budget (`76ef06a3-…`) inside the rollover loop across the cold-load calculate() passes (dollar figures redacted):

```
[0] txns=104   splits=57 iters=48 juneRolled=<inflated>
[1] txns=104   splits=57 iters=48 juneRolled=<inflated>
[2] txns=<full> splits=57 iters=48 juneRolled=<correct>
[3] txns=<full> splits=57 iters=48 juneRolled=<correct>
```

Identical loop bounds (`iters=48`, July 2022 → June 2026) every pass. The only variable is the loaded transaction count: a small recent subset (Stage 1/2) vs the full history (Stage 3). Capacity accrual is constant at −(monthly capacity) × 48; with only the recent subset loaded almost no spending offsets it (→ inflated), with the full history the real spending offsets it (→ correct, ~12× smaller). The Stage-3 value is the correct steady-state value (computed by the production code on essentially-complete data).

## Root cause

`src/client/lib/hooks/calculation/budgets.ts:146-163` — the rollover accumulation loop:

```ts
const endDate = new ViewDate("month");
budgetData.forEach((history, budgetLikeId) => {
  ...
  const startDate = new ViewDate("month", roll_over_start_date).next();
  while (startDate.getEndDate() <= endDate.getEndDate()) {
    const previousCapacity = getActiveCapacity(previousDate.getEndDate());
    history.add(startDate.getEndDate(), {
      rolled_over_amount: previousSummary.rolled_over_amount - previousCapacity.month, // capacity for EVERY month, full history
    });
    startDate.next();
  }
});
```

The loop **unconditionally** subtracts a month of capacity for every month from `roll_over_start_date` to *now*, regardless of whether that month's transactions are present in `transactions`. The spending side (`processTransaction` → `budgetData.add(..., { rolled_over_amount: amountAfterSplit })`, lines 83/102/118/134) only contributes for transactions actually loaded.

`calculate(data)` re-runs on every `data` change (`src/client/components/App/Utility.tsx:37-40`), and the cold-sync path (`src/client/lib/hooks/sync.ts`) streams transactions in stages (Stage 1: none → Stage 2: ~2 recent months → Stage 3: full history). So the loop runs repeatedly against a transaction set that is complete on the *capacity* axis but truncated on the *spending* axis until Stage 3 lands.

The rollover figure is fundamentally only correct when the **complete** transaction history back to `roll_over_start_date` is in memory — an invariant the staged/windowed sync architecture does not guarantee during cold load.

## Impact

- **Budgets page** (`LabeledBar` `src/client/components/LabeledBar/index.tsx:37-40,155-165`): every rolling budget shows an inflated "+ $X rolled" for the cold-load window.
- **Dashboard projection charts** (`BalanceChartRow` `src/client/components/BalanceChartRow/index.tsx:72-73` reads `rolled_over_amount` for `roll_over` budgets): Retirement / Cash-Flow / balance projections are driven by the inflated rollover during cold load, so those curves are wrong too.
- Recurs on every cache-cleared / first-time / incognito load; the larger the history and the older `roll_over_start_date`, the larger the overstatement.

Not user state (re-curating data can't fix it) and not the stale-snapshot family — it's a deterministic calc/loading-order defect.

## Proposed fix (sketch)

Pick one; (1) is smallest and matches existing status-gating patterns:

1. **Suppress the rolled figure until the cold load is complete.** Gate the `rolled_over_amount` display (and its use in `BalanceChartRow`) on a "history fully loaded" status so a rolling budget shows a skeleton/`—` instead of a wrong number until Stage 3 finishes. Warm loads start complete, so they're unaffected.
2. **Clamp capacity accrual to the loaded-history window.** Start the accrual at `max(roll_over_start_date, earliest_loaded_transaction_month)` so the capacity and spending axes always span the same window. (Changes the semantic — rollover would understate until full history loads, but never *overstate*.)
3. Compute rollover from a complete-history source. (Against the FE-only-analytics preference; least likely wanted.)

Relatedly, `budgets.ts` currently has **zero unit-test coverage** (#429) — a rollover-on-partial-data case would have caught this.

---
*Filed by `sop-daily-user-testing`. Root-caused via live instrumentation on the prod-clone `admin` dataset; instrumentation reverted, working tree clean.*

*[Edited 2026-06-05: redacted real dollar amounts, the personal budget name, and the account/transaction counts from the original body per the daily security review §2a PII policy — public repo. The bug's technical content (~12× cold-load overstatement, root cause, fix sketch) is unchanged.]*

claoie:
### The bug
On a **cold load** (empty IndexedDB — first visit, cleared cache, new device, incognito) every rolling budget's "+ $X rolled" figure rendered grossly inflated (~12× too high on the prod-clone `admin` data) for the ~10–30s until the full transaction history finished streaming, then snapped to the correct value.

It was a genuine miscalculation, not "partial data shown as partial": the rollover loop in `getBudgetData` accrued a month of **capacity** for *every* month from `roll_over_start_date` to now, but the **spending** that offsets it only existed for the subset of months whose transactions had loaded. During cold-load Stage 1/2 the capacity axis spanned years while the spending axis spanned weeks. The Dashboard projection charts (`BalanceChartRow` reads `rolled_over_amount` for `roll_over` budgets) were driven by the same inflated number.

### Why a plain clamp in the calc isn't enough
The obvious "clamp accrual to the earliest loaded transaction month" would **break steady state** for any budget whose `roll_over_start_date` predates its oldest transaction — those genuinely empty early months legitimately accrue full capacity into the carry, and clamping would silently drop them even on a complete load. The calc can't distinguish "no transactions because the month isn't loaded yet" from "no transactions because none ever existed" without a load-completeness signal.

### The fix
- New `Status.isTransactionHistoryPartial` (default **false** → every warm/complete Data path is treated as complete, so steady-state values are byte-identical).
- `sync.ts` sets it `true` on the two cold-load intermediate stages (Stage 1, Stage 2). Stage 3 (full history) and the warm/cached path leave it false. `reduceStatuses` ORs it across sources.
- `getBudgetData` takes the flag; when set, it clamps the accrual to start no earlier than the earliest loaded transaction month, so capacity and spending span the same window.

Result: during cold load the figure **understates** slightly and converges upward to the correct value (never the alarming 12× overstatement); once history is complete it's unchanged. No component changes — both `LabeledBar` and `BalanceChartRow` read the corrected `budgetData`.

### Tests
New `src/client/lib/hooks/calculation/budgets.rollover.test.ts` (4 cases, pure-function, no mocks):
- complete load accrues capacity over the full span (steady-state unchanged)
- partial load clamps to the loaded window — magnitude bounded, never overstates
- **equivalence**: `clamp(partial, old start)` === a complete run whose `roll_over_start_date` *is* the earliest loaded month (time-independent assertion that pins the clamp semantics)
- Stage 1 (no loaded transactions) shows `0`, not an inflated carry

Verified the clamp tests **fail** on the pre-fix source (`partial` === `complete`, overstated) and **pass** after.

### Verification
- `bun test …/budgets.rollover.test.ts` → 4 pass
- `bun test src` (full suite) → **826 pass / 0 fail**
- `bun run typecheck` → clean
- A sandbox is being spun for cold-load UI verification (clear IndexedDB → Budgets page → watch the rolled figure converge instead of starting at ~12×).

### DIFF
--- a/src/client/lib/models/miscellaneous.ts
+++ b/src/client/lib/models/miscellaneous.ts
@@ -96,4 +96,10 @@ export class Status {
   isInit = false;
   isLoading = false;
   isError = false;
+  // True while a cold sync is in flight (no IDB cache to paint from, history
+  // arrives in stages). Consumers whose calc spans the full history — the
+  // rollover accrual in getBudgetData — clamp to the loaded window while
+  // this is set, so the accrual axis and the spending axis cover the same
+  // months. Warm syncs keep this false (IDB cache had everything).
+  isColdSync = false;
 }
--- a/src/client/components/App/lib/hooks.ts
+++ b/src/client/components/App/lib/hooks.ts
@@ -75,6 +75,7 @@ export const useData = () => {
           sections,
           categories,
           transfers,
+          data.status.isColdSync,
         );
 
         const capacityData = getCapacityData(budgets, sections, categories);
--- a/src/client/lib/hooks/calculation/budgets.rollover.test.ts
+++ b/src/client/lib/hooks/calculation/budgets.rollover.test.ts
@@ -4,12 +4,15 @@
 // accrued a month of capacity for EVERY month from roll_over_start_date to now,
 // but spending only existed for the subset of months whose transactions had
 // streamed in — so the figure overstated (~12× in prod) until the full history
-// loaded. getBudgetData now takes `isTransactionHistoryPartial`; when set it
-// clamps the accrual to the earliest loaded transaction month so the capacity
-// and spending axes span the same window.
+// loaded. getBudgetData now takes `isColdSync`; when set, it skips the accrual
+// loop entirely so the figure stays at 0 until Stage 4 commits the full
+// history. Reason: under the new delta-by-cursor sync the loaded subset is
+// keyed by `updated`, not `date`, so back-edited rows from years ago appear in
+// Stage 2 with no relation to "the last N months of history" — the axis
+// mismatch is unfixable until the full history lands.
 
 import { describe, test, expect } from "bun:test";
-import { ViewDate, LocalDate } from "common";
+import { LocalDate, ViewDate } from "common";
 import { getBudgetData } from "./budgets";
 import {
   TransactionDictionary,
@@ -20,60 +23,66 @@ import {
   CategoryDictionary,
   TransferDictionary,
 } from "../../models/Data";
-import { Account } from "../../models/Account";
 import { Transaction } from "../../models/Transaction";
 import { Budget } from "../../models/Budget";
+import { Account } from "../../models/Account";
 
-const ACCOUNT_ID = "acc-1";
 const MONTHLY_CAPACITY = 100;
 
-const makeAccount = () => {
-  const dict = new AccountDictionary();
-  dict.set(ACCOUNT_ID, new Account({ account_id: ACCOUNT_ID, hide: false }));
-  return dict;
-};
-
-const makeBudget = (rollOverStart: string) => {
-  const dict = new BudgetDictionary();
-  const budget = new Budget({
+const makeBudget = (rollOverStart: string): BudgetDictionary => {
+  const b = new Budget({
     budget_id: "bud-1",
-    name: "Test",
+    user_id: "u1",
+    name: "Rolling Budget",
+    iso_currency_code: "USD",
+    capacities: [{ active_from: null, children: {}, year: MONTHLY_CAPACITY * 12, month: MONTHLY_CAPACITY, week: 0, day: 0 }],
     roll_over: true,
     roll_over_start_date: rollOverStart,
-    capacities: [{ month: MONTHLY_CAPACITY, is_synced: false }],
   });
-  dict.set(budget.budget_id, budget);
+  const dict = new BudgetDictionary();
+  dict.set(b.budget_id, b);
   return dict;
 };
 
-// One unconfirmed transaction (unsorted path) labeled directly to the budget,
-// dated `monthsAgo` months before the current month.
-const makeTransactions = (monthsAgo: number) => {
-  const dict = new TransactionDictionary();
-  let vd = new ViewDate("month");
-  for (let i = 0; i < monthsAgo; i++) vd = vd.previous();
-  const dateStr = vd.getEndDate().toISOString().slice(0, 10);
-  dict.set(
-    "txn-1",
-    new Transaction({
-      transaction_id: "txn-1",
-      account_id: ACCOUNT_ID,
-      amount: 30,
-      date: dateStr,
-      label: { budget_id: "bud-1" },
-    }),
-  );
-  return dict;
-};
-
-const emptySplits = () => new SplitTransactionDictionary();
 const emptySections = () => new SectionDictionary();
 const emptyCategories = () => new CategoryDictionary();
+const emptySplits = () => new SplitTransactionDictionary();
+
+const makeAccount = (): AccountDictionary => {
+  const a = new Account({
+    account_id: "acc-1",
+    name: "Checking",
+    type: "depository" as never,
+    subtype: "checking" as never,
+    label: { budget_id: "bud-1", category_id: null },
+  } as never);
+  const dict = new AccountDictionary();
+  dict.set(a.account_id, a);
+  return dict;
+};
+
+const makeTransactions = (monthsAgo: number): TransactionDictionary => {
+  const dict = new TransactionDictionary();
+  const d = new ViewDate("month").previous();
+  for (let i = 1; i < monthsAgo; i++) d.previous();
+  const t = new Transaction({
+    transaction_id: `txn-${monthsAgo}`,
+    account_id: "acc-1",
+    name: "Coffee",
+    merchant_name: "Cafe",
+    amount: 50,
+    date: d.getStartDate().toISOString().slice(0, 10),
+    pending: false,
+    label: { budget_id: "bud-1", category_id: null },
+  } as never);
+  dict.set(t.transaction_id, t);
+  return dict;
+};
 
 const rolledThisMonth = (
   rollOverStart: string,
   transactions: TransactionDictionary,
-  partial: boolean,
+  isColdSync: boolean,
 ) => {
   const { budgetData } = getBudgetData(
     transactions,
@@ -83,18 +92,16 @@ const rolledThisMonth = (
     emptySections(),
     emptyCategories(),
     new TransferDictionary(),
-    partial,
+    isColdSync,
   );
   return budgetData.get("bud-1", new ViewDate("month").getEndDate()).rolled_over_amount;
 };
 
-describe("getBudgetData rollover cold-load clamp (#484)", () => {
-  // Transaction one month back simulates a Stage-2 cold load: only the most
-  // recent month is in memory while roll_over_start_date sits years earlier.
+describe("getBudgetData rollover cold-sync skip (#484)", () => {
   const txns = makeTransactions(1);
   const OLD_START = "2022-06-01";
 
-  test("complete load accrues capacity over the full span (steady-state, unchanged)", () => {
+  test("warm load accrues capacity over the full span (steady-state, unchanged)", () => {
     const complete = rolledThisMonth(OLD_START, txns, false);
     const start = new ViewDate("month", new LocalDate(OLD_START)).next();
     const now = new ViewDate("month");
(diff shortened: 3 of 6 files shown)


reviewoie:
## Code review — PR 509

**Verdict:** approve

Scope is clean: all 6 files relate directly to the rollover cold-load clamp (`budgets.ts` clamp, `sync.ts` flag on cold stages, `miscellaneous.ts` `Status` field, `status.ts` OR-reduce, `hooks.ts` plumbing, the new test). No unrelated changes.

Verified locally against the PR head (`99f003d`):
- `budgets.rollover.test.ts` → 4 pass; full `calculation/` suite → 70 pass / 0 fail; `bun run typecheck` clean.
- Removed the clamp in a throwaway worktree → the partial-overstatement and equivalence tests fail with `-4770` (the multi-year inflated accrual) vs the clamped `-70`. The test genuinely catches the bug, not a tautology.
- Steady-state safety confirmed: `isTransactionHistoryPartial` defaults false on the `Status` class, Stage 3 (`finalData`) and the warm path never set it, so `getBudgetData`'s accrual start is byte-identical on warm/complete loads. CI build + test both green.

### Low
- `src/client/lib/hooks/sync.ts:512` — pre-existing, not introduced here, but worth noting: the warm-path opening `setData((oldData) => new Data(oldData))` copies `oldData.status` by reference (`assign`), so a `true` partial flag from an in-flight cold load could be inherited if a warm `_sync` interleaves. In practice a warm load implies a prior persisted sync (oldData is steady-state, partial=false), so it doesn't fire today. Flagging only because the new flag now rides on that copied status object — no action needed.
- `src/client/lib/hooks/calculation/budgets.ts:159` — `earliestLoadedMonth` is computed once globally across all budgets, while each budget's coverage could in principle differ. This is correct for the cold-load stages (every budget is fed the same recent window), but the global is an implicit coupling to the staged-sync shape. Fine as-is; a one-line comment noting the assumption would future-proof it.

No further blockers.

claoie:
**Verification (claoie)**

Checked out the branch (`99f003d`) and verified the cold-load rollover clamp behaviorally, not just by compile:

- `bun test src/client/lib/hooks/calculation` → **70 pass / 0 fail**.
- Confirmed the new tests actually *guard* the clamp (not no-ops): forcing the `isTransactionHistoryPartial` branch off made exactly the 2 cold-load tests fail —
  `clamp(partial, old start) === complete run started at the earliest loaded month` — and restoring it returned to 70 pass.
- `tsc --noEmit` exit 0.

**UI E2E not run, with reason:** the clamp only diverges from baseline under a cold load that is still streaming history (`isTransactionHistoryPartial`), and the corrected rollover figure is not visually distinguishable from a correct figure without first reproducing the overstatement condition. There is also no admin login credential available in this environment. The numeric unit tests pin the exact behavior (clamped accrual starts at the earliest loaded month vs. accruing unloaded older months), which is the load-bearing assertion here.

src/client/lib/models/miscellaneous.ts
```diff
@@ -96,4 +96,10 @@ export class Status {
   isInit = false;
   isLoading = false;
   isError = false;
+  // True only on cold-load intermediate stages whose transaction history is
+  // still being streamed in (recent months loaded, older months not yet).
+  // Rollover accrual must not span months whose spending hasn't arrived, or it
+  // overstates the carried figure (#484). Defaults false: every warm/complete
+  // Data path is treated as complete, so steady-state values are unchanged.
+  isTransactionHistoryPartial = false;
```
hoiekim: This is ugly. Why can't you use `isLoading`? If you still need a new flag, how about `isColdSync`?

hoiekim (changes requested):

reviewoie:
## Code review — PR 509

### High
| # | File:line | Summary | Failure scenario |
|---|-----------|---------|------------------|
| 1 | `src/client/lib/hooks/calculation/budgets.ts:198-203` | `effectiveStart.next()` mutates `earliestLoadedMonth` across iterations of `budgetData.forEach`, so only the FIRST rollover budget gets its accrual; every subsequent rollover budget gets zero carry. | Cold load with two rollover budgets, each with a recent-window transaction. Iteration 1 (`bud-1`): `effectiveStart = earliestLoadedMonth`; `.next()` mutates `earliestLoadedMonth` to "now"; while-loop runs once and advances it to "now+1". Iteration 2 (`bud-2`): `earliestLoadedMonth` now points at "now+1" (past `endDate`); `effectiveStart = earliestLoadedMonth`; `startDate = effectiveStart.next()` = "now+2"; `while (startDate <= endDate)` is false → loop never executes → no rollover carry written to `bud-2` at "now". Verified with a 2-budget repro under the same test harness: `bud-1` rolled = -70, `bud-2` rolled = +40 (just the unsorted-pre-pass add — no -100 capacity subtraction). Patching to `effectiveStart.clone().next()` restores `bud-2` to -60 as expected. The existing test only exercises one budget, so it doesn't catch this. Same shape applies cross-section / cross-category in mixed sets. Fix: `const startDate = effectiveStart.clone().next();` and add a multi-rollover-budget test. |

### Med
| # | File:line | Summary | Failure scenario |
|---|-----------|---------|------------------|
| 2 | `src/client/lib/hooks/calculation/budgets.ts:39-43`, `sync.ts:644-647`, `sync.ts:740-742`, `models/miscellaneous.ts:97-103` | Inline `(#484)` PR-narrative tags inside committed comments. `feedback_comment_hygiene.md` rule: future readers shouldn't need to chase a PR link to understand a constraint. The `Stage 4`/`cold sync`/`clamp to loaded window` *constraint* statements are fine — the `(#484)` reference and `Defaults false so steady-state values are byte-identical` (a PR-justification, not a future-reader constraint) are not. | Six months from now a reader sees `(#484)` in five spots, has to open the PR to understand what bug existed, then has to decide whether the comment still applies. Strip the PR refs; keep the forward-looking constraint sentences. |

### Low
| # | File:line | Summary | Failure scenario |
|---|-----------|---------|------------------|
| 3 | `src/client/lib/hooks/calculation/budgets.rollover.test.ts:7` | Stale header comment refers to a parameter name that no longer exists. | Header says *"getBudgetData now takes `isTransactionHistoryPartial`"* but the param was renamed to `isColdSync` (the rename Hoie asked for in inline review 4539599999). Update the comment to match the shipped name. |
| 4 | `src/client/lib/hooks/calculation/budgets.ts:189` | When `isColdSync=true` and there are zero loaded transactions, `earliestLoadedMonth = endDate` (same reference). Today this is safe — `budgetData.forEach` is empty when no transactions exist — but the aliasing is a latent footgun the moment some other code path adds rollover entries to `budgetData` without going through `transactions.forEach`. Pair with finding 1 fix: `earliestLoadedMonth = earliest ? new ViewDate("month", earliest) : endDate.clone();`. |

claoie:
## Status update (HEAD 1a347a07) — DRAFT

Rebased onto current upstream/main (was on the pre-#536/#537 sync staging). Renamed `isTransactionHistoryPartial` → `isColdSync` per the inline review. Wired it into the new Stage 1 / Stage 2 / Stage 4 staging.

**Reviewoie round 1 findings (4 total):**
- HIGH (cross-iteration ViewDate mutation): addressed in current commit — `effectiveStart.clone().next()` so `earliestLoadedMonth` doesn't advance across budgetData.forEach iterations.
- MED (PR-narrative `(#484)` tags in comments): stripped.
- LOW (stale header comment in test): updated.
- LOW (latent endDate aliasing): noted; depends on a future code path adding to budgetData independently of transactions — not introducing today, but flagged.

**E2E uncovered a real regression that the unit tests miss.** Throttled the Stage 4 unbounded fetch by 3s on a sandbox cold load to observe the streaming window. Observed:

```
t=   625ms  tx=   0  rollover=+$0
t=  1045ms  tx=   0  rollover=+$XX,XXX  ← inflated, not clamped
...                                       (stays inflated for 3s)
t=  3994ms  tx=6458  rollover=+$513     ← Stage 4 commits, correct
```

Stage 1 paint at t=625ms correctly shows $0 (empty transactions → clamp to current month → no carry). But between Stage 1 and Stage 4, the figure jumps to the un-clamped value. The unit tests pass (4/4) so the clamp math is correct given the inputs; the wiring is what's broken — either `isColdSync` isn't reaching the calc, or the calc is running against different inputs than expected during the Stage 2 → Stage 4 window.

Leaving as DRAFT while I dig into the wiring. Next step: instrument `getBudgetData` with a debug log to trace `isColdSync` + `transactions.size` at each render during the throttled streaming window.

CI green (typecheck + lint + 4/4 rollover tests + 640/640 server tests).

*[Edited 2026-06-22: redacted a >$1k budget rollover figure per security policy — value is a computed bug artifact, not a real balance.]*

claoie:
## Status update — READY for review (HEAD 957a58d0)

### What changed since the last update

Root cause of the inflation found via debug-instrumented E2E:

```
[browser] getBudgetData isColdSync= true tx.size= 0    budgetData.size= 0   ← Stage 1 paint
[browser] getBudgetData isColdSync= true tx.size= 1012 budgetData.size= 56  ← Stage 2 paint
[browser]   earliestLoadedMonth= 2021-07-01            ← FIVE YEARS AGO
[browser] getBudgetData isColdSync= false tx.size= 6458 budgetData.size= 61 ← Stage 4 paint
```

The new delta-by-cursor sync filters by `updated`, not `date`. Stage 2's recent-2-month fetch returns `{transactions whose updated is recent}` — which includes back-edited rows whose `date` is years old (1012 transactions in Stage 2 with earliest date = 2021-07-01). `min(transactions.date) = 2021-07-01` doesn't represent "loaded window" at all — it's just "the oldest row that happens to have been edited recently." The clamp inferred "5 years of history loaded" and accrued capacity over 5 years → $XX,XXX inflated.

The original "clamp to earliest loaded month" design fits a sync model that loads contiguous date windows (the pre-#536 per-month walk). Under the new `updated`-keyed delta, no contiguous date window is ever loaded mid-stream. The clamp is unfixable without a different signal — and the simplest, predictable signal is "while cold, just show $0 rollover." Carry computes correctly the moment Stage 4 commits.

### The fix (this commit)

`getBudgetData(isColdSync=true)` now:
1. Zeros out any `rolled_over_amount` that `processTransaction` wrote during its forward-carry pass (back-edited transactions would otherwise add their amount into next-month's rollover under the new sync).
2. Skips the multi-year accrual loop entirely.
3. Returns immediately. Warm syncs (`isColdSync=false`) keep the loop unchanged, so steady-state values are byte-identical.

The four unit tests in `budgets.rollover.test.ts` now cover the new semantic:
- warm load accrues full span (steady-state, unchanged)
- cold sync skips accrual entirely — rollover = 0
- cold + zero transactions also = 0 (Stage 1)
- cold's result is independent of which transactions happen to be loaded (key claim — the old clamp violated this)

### E2E verification (sandbox port 20509, Stage 4 throttled +3s to observe the streaming window)

```
t=  626ms  tx=   0  Dashboard=$0  All Budgets=$0   ← Stage 1, isColdSync=true
t= 1123ms  tx=   0  Dashboard=$0  All Budgets=$0   ← Stage 2 paint (back-edited rows loaded; would have shown $XX,XXX before)
…throughout the 3 s streaming window: $0 / $0
t= 4034ms  tx=6458  Dashboard=$513  All Budgets=$513   ← Stage 4 commits, correct value
```

CI: typecheck + lint clean. 4/4 rollover tests pass.

*[Edited 2026-06-22: redacted a >$1k budget rollover figure per security policy — value is a computed bug artifact, not a real balance.]*

hoiekim (approved):

Issue #318, opened 2026-05-06
AccountsTable Balance row silently shows today's balance for past dates with no history


moltboie:
## Bug

In `src/client/components/AccountsTable/Balance.tsx`, the displayed amount in the three non-Credit branches (CryptoExchange, Investment, default) is rendered as:

```tsx
{dynamicAmount ? numberToCommaString(dynamicAmount) : currentString}
```

For past viewDates where the account has no balance history (or a recorded zero), `dynamicAmount` evaluates to `0` and the ternary falls back to `currentString` — which is today's `balances.current`. This silently substitutes today's balance for the missing historical value.

The leading comment in `Balance.tsx` already documents the desired semantics:

```ts
// For future dates (e.g. year-end of the current year), fall back to the
// current balance rather than 0 so the row shows a meaningful value.
// For past dates with no history, show 0 (data was never recorded).
const fallback = viewDateDate > today ? getAccountBalance(account) : 0;
const dynamicAmount = balanceHistory.get(viewDateDate) ?? fallback;
```

`dynamicAmount` already encodes "0 for past, today's for future". The additional `dynamicAmount ? ... : currentString` ternary contradicts that intent specifically when dynamicAmount is 0.

The same-page AccountsDonut correctly contributes 0 for those accounts:

```ts
// AccountsPage.tsx
const fallback = viewDateDate > today ? getAccountBalance(a) : 0;
const value = balanceHistory.get(viewDateDate) || fallback;
balanceTotal += value;
```

So the donut total and the sum of table row balances disagree on past months — same data shown two different ways on the same page.

## Concrete repro (local prod-snapshot DB, today = 2026-05-06)

Account: **IBM - RBA** (`fda9c`, earliest snapshot 2025-09-30)
- today `balances.current` = $14,423.98
- `balanceHistory.get(Dec 31 2024)` = undefined → `dynamicAmount` = 0

| viewDate | display (pre-fix) | display (post-fix) | donut contribution |
|----------|-------------------|--------------------|--------------------|
| Dec 2024 | $14,423.98 (today's) | $0.00 | $0 |

## Aggregate impact (AccountsPage, viewDate = December 2024)

| metric | value |
|--------|-------|
| Donut center total | $536,199 |
| Pre-fix sum of non-credit table rows | $610,294.88 |
| **Discrepancy** | **+$74,095.88** |
| Post-fix sum of non-credit table rows | $536,199.21 (rounding noise only) |

[...]

Issue #312, opened 2026-05-04
AccountsTable Balance widget for CryptoExchange shows today's current and ignores viewDate history


moltboie:
## Summary
The CryptoExchange subtype branch in `src/client/components/AccountsTable/Balance.tsx` (lines 49–58 on upstream/main `45aaa5a`) renders today's `balances.current` in both the amount cell and the Changes widget, regardless of the selected `viewDate`. When the user views a past month, the row shows today's balance — and the Changes widget compares today's balance against the previous-period historical balance, mixing time periods.

This is the same shape of bug that PR #285 fixed for the Investment branch (line 60-69) and PR #311 (open) is fixing for the default depository/loan branch (line 71-81). The CryptoExchange branch is the third instance of the same pattern.

## Repro (local prod-snapshot DB)
Only crypto exchange account in the snapshot:
- `account_id = QY6XDvjOKxFvADwrPyYZCwRqOQP9K8uJBBzr7` (Binance, type=`investment`, subtype=`crypto exchange`)
- `balances.current` (today) = `$1,835.77`
- snapshot at `2026-01-31` = `$2,134.53`
- snapshot at `2025-12-31` = `$2,366.09`

Set `viewDate = January 2026` (a past month). Render the `<Balance>` component:

| | Display amount | Changes widget |
|---|---|---|
| Pre-fix (current) | `$1,835.77` (today's `current!`) | `-$530.32` (today − Dec, mixing time periods) |
| Expected (consistent with Investment / PR #311) | `$2,134.53` (Jan-end historical) | `-$231.56` (Jan − Dec) |

For the same account on the same `viewDate`, AccountsPage's donut slice already correctly shows `$2,134.53` (`AccountsPage/index.tsx:46` uses `balanceHistory.get(viewDateDate)`), so the donut and the table row disagree on the same page.

## Verification
A render test using `react-dom/server.renderToStaticMarkup` reproduces the bug:

```
HTML: <div class="Balance"><div>$1,835.77</div><div class="Changes colored negative">-$530.32</div></div>
```

After applying the proposed fix:
```
HTML: <div class="Balance"><div>$2,134.53</div><div class="Changes colored negative">-$231.56</div></div>
```

## Root cause
Lines 49–58 of `Balance.tsx`:

```tsx
} else if (subtype === AccountSubtype.CryptoExchange) {
  return (
    <div className="Balance">
      <div>
        {symbol}
        {currentString}                                       {/* always today's `current` */}
      </div>
[...]

Issue #310, opened 2026-05-02
bug: AccountsTable Balance shows wrong change for depository accounts on past view dates (today's current vs historical previous)


moltboie:
## Summary
On the **Accounts** page, the per-account `Changes` widget for **depository** accounts mixes time periods when the user picks a past `viewDate`: the displayed balance is the historical value for the selected period, but the change indicator compares **today's `current` balance** against the previous period's historical balance.

This was a partial fix in PR #285 — that PR fixed the same issue for **Investment** accounts (using `dynamicAmount || current!` instead of `current!`), but the **default (depository / loan / etc.)** branch in `src/client/components/AccountsTable/Balance.tsx` was not updated and still uses `current!` directly.

## Reproduction (against local prod-snapshot DB, today = 2026-05-02)

Account: `TOTAL CHECKING` (`wzNkq3qLKvheKRMP7ZNViw3y6qoYaJhykpbDL`, depository)

| month | last snapshot in month | balance |
|-------|------------------------|---------|
| 2025-08 | 2025-08-31 | \$2,925.81 |
| 2025-09 | 2025-09-22 | \$3,045.44 |
| 2026-01 | 2026-01-27 | \$6,325.63 |
| 2026-02 | 2026-02-24 | \$6,778.10 |
| 2026-03 | 2026-03-30 | \$7,445.44 |

Today's `balances_current` = \$7,445.44.

**Switch viewDate to February 2026:**
- Display: `\$6,778.10` (correct — historical Feb)
- Changes shown: `current (\$7,445.44) − prev (\$6,325.63)` = **+\$1,119.81** (WRONG)
- Correct change Feb→Jan: \$6,778.10 − \$6,325.63 = **+\$452.47**
- Off by \$667.34

**Switch viewDate to September 2025:**
- Display: \$3,045.44 (correct historical Sep)
- Changes shown: \$7,445.44 − \$2,925.81 = **+\$4,519.63** (WRONG)
- Correct: \$3,045.44 − \$2,925.81 = **+\$119.63**
- Off by \$4,400

## Root Cause

`src/client/components/AccountsTable/Balance.tsx`, default (else) branch:

```tsx
{!!previousAmount && <Changes currentAmount={current!} previousAmount={previousAmount} />}
```

[...]

PR #516, opened 2026-06-14
fix(accounts): credit rows and donut honor viewDate (#515)

### LINKED ISSUE #515: bug(accounts): credit-card rows ignore viewDate (always show live balance), disagreeing with the donut's past-month credit total
## Summary

On the **Accounts** page, credit-card account rows ignore the viewDate selector. Every other account type (depository, investment, crypto-exchange) re-renders its **historical** balance when you step the month/year selector back, but credit rows are frozen on the **live** balance for all months. This also makes the page internally inconsistent: for any non-current month the donut headline reports **$0 outstanding** in N Credits, while the credit rows below still show the full live outstanding balances.

This is the one account type that never received the viewDate-awareness fixes already applied to depository accounts (#310), crypto/investment accounts (#312), and the zero-history fallback (#318).

## Reproduction

1. Log in, open **Accounts**.
2. At the current month, note the donut headline "**outstanding in N Credits**" total and the per-row "spent" amounts on the credit cards — they agree.
3. Step the month selector back to any past month (e.g. Sep 2025).
4. Observe:
   - Every depository / investment row updates to its historical balance, and the donut center total changes accordingly.
   - **All credit-card rows stay pinned to the live balance** — unchanged across every month.
   - The donut headline now shows the credit total as a neutral "-" (i.e. **$0 outstanding**).

So at any past month the credit rows and the donut headline disagree about the same accounts at the same viewDate.

## Expected vs Actual

- **Expected:** credit rows reflect the snapshot balance for the selected month (credit accounts *do* have historical `account_balance` snapshots), and the donut headline and the rows agree. For the current month / future, fall back to the live balance — exactly the rule the other account types already use.
- **Actual:** credit rows always render the live `balances.current` / `balances.available`; the donut headline reports $0 outstanding for past months. The two surfaces contradict each other on every non-current month.

## Root cause

`src/client/components/AccountsTable/Balance.tsx`
- Line 17 / 30 build `currentString` / `availableString` straight from the live `balances.current` / `balances.available`.
- Lines 29-48: the `type === AccountType.Credit` branch returns early using only those live strings. It never reads the viewDate-aware `dynamicAmount` (computed at line 26 as `balanceHistory.get(viewDateDate) ?? fallback`) that every other branch renders (investment/crypto at lines 66-76, depository at 77-85).

`src/client/components/AccountsDonut/BalanceInfo.tsx:45-49` + `src/client/pages/AccountsPage/index.tsx:36-40`
- `totalCredit` is summed from live `balances.current`, and the headline only renders it when `!viewDateSpan` (current month). For past months the outstanding credit collapses to $0 in the headline while the rows stay live — the inconsistent half of the same root gap.

The non-credit path was migrated to viewDate-awareness across #310 / #312 / #318; the credit branch predates those and was never updated.

## Evidence this is system behavior, not persisted user state

- A credit card with a populated snapshot history (~80 account_balance snapshots spanning Aug 2025 → Mar 2026) has a latest-per-month balance that varies materially month to month (verified against the snapshots — the monthly values are all distinct; the most recent month, `$<live>`, is the live balance, and earlier months such as Sep 2025 (`$<sep>`) and Jan 2026 (`$<jan>`) are each several-fold different). All dollar values redacted; the load-bearing fact is `$<sep>` ≠ `$<jan>` ≠ `$<live>`.
- The Accounts row renders the single live value `$<live>` at **every** selected month. Verified directly in the UI: at the Sep 2025 view the row showed `$<live>`, not that month's snapshot balance `$<sep>`.
- No UI action re-curates this — it is a fixed render path keyed off `balances.current` rather than the snapshot history. (Not a #302-class persisted-user-choice case.)

## Proposed fix sketch

1. In `Balance.tsx`, give the credit branch the same `dynamicAmount` treatment as the other types: render `dynamicAmount` as "spent" (deriving "available" from the credit limit minus the historical balance when a stored historical `available` isn't present), keeping the live values only when `viewDateDate >= today` — mirroring the existing `fallback = viewDateDate > today ? getAccountBalance(account) : 0` rule.
2. Align the donut so the two surfaces agree for past months: either sum the historical credit balances (`balanceHistory.get(viewDateDate)` per credit account) for the outstanding headline, or blank the credit rows to match the headline. Pick one so the table and the donut report the same credit total at every viewDate.

## Severity

Minor / UX — balances are still correct for the current month, and account totals (assets) are unaffected; the defect is that the credit column does not honor the time selector and contradicts the donut for historical months. Distinct from the open cold-load past-month issues (#510 / #428).

claoie:
## Summary

Closes #515. Credit-card account rows ignored the viewDate selector: every other account type (depository, investment, crypto-exchange) re-renders its **historical** snapshot balance as you step the month selector back, but the credit branch in `Balance.tsx` returned early using only the live `balances.current` / `balances.available`. The page was also internally inconsistent — `AccountsPage` summed `totalCredit` from live `balances.current`, and `BalanceInfo` only rendered the outstanding-credit headline at the current month, so for any past month the donut reported **$0 outstanding** while the rows still showed full live balances.

The non-credit paths were migrated to viewDate-awareness across #310 / #312 / #318; the credit branch predates those and was never updated.

## Changes

- **`AccountsTable/Balance.tsx`** — the credit "spent" value now uses the viewDate-aware `dynamicAmount` (historical snapshot for past months, live balance for current/future — identical to the other account types). "available" is derived from the credit limit minus the outstanding for past months so the two figures stay coherent (`spent + available = limit`); current/future months and SimpleFin accounts (which report no limit) keep the live value.
- **`pages/AccountsPage/index.tsx`** — `totalCredit` is summed from the viewDate-aware balance history (`balanceHistory.get(viewDateDate) ?? fallback`) instead of live `balances.current`, mirroring the existing non-credit donut loop directly below it.
- **`AccountsDonut/BalanceInfo.tsx`** — render the outstanding-credit headline for all months (dropped the `!viewDateSpan` gate) now that `totalCredit` is viewDate-aware, so the donut and the rows agree at every viewDate.

Scope is the one bug: credit rows honoring viewDate + the donut/rows agreement. No change to non-credit rows, the calculation hook, or current-month behavior (`BalanceHistory.get` buckets by year-month, so at the current month it returns the value set at `today` = the live balance).

## Test plan

- [x] `bun run lint` — clean
- [x] `bun test` — 822 pass / 0 fail
- [x] **UI E2E** — Playwright, fix sandbox `:20516` vs `upstream/main` baseline `:20000` (same deterministic scramble → apples-to-apples), logged in as `admin`, **Accounts** page. Read-only (no DB writes). Four credit accounts present (limits known), stepped the selector Jun 2026 → Feb 2026:
  - **Current month (Jun 2026) — no regression.** Both fix and baseline identical: `61dObo` spent **$[redacted]** (= live), donut headline **-$[redacted] outstanding in 4 Credits** (= sum of the 4 rows' spent).
  - **Past month (Feb 2026):**
    - **Baseline (bug):** all credit rows frozen on the live balance — `61dObo` still **$[redacted]**, `QPLNzz` **$0.00**, `vj76b7` **$[redacted]** — and the donut headline collapses to **"–" ($0 outstanding)**. Rows and donut disagree.
    - **Fix:** rows render the historical balance — `61dObo` spent **$[redacted]** (available **$[redacted]** = limit $[redacted] − $[redacted], coherent), `QPLNzz` **$[redacted]**, `vj76b7` **$[redacted]** — and the donut headline shows **-$[redacted] outstanding** = exactly the sum of the 4 historical spents. Rows and donut **agree**.
  - Screenshots (read back): `/tmp/pr-516-fix-past.png` (donut "-$[redacted]"), `/tmp/pr-516-baseline-past.png` (donut "–"), plus `*-current.png` for both.

*[Edited 2026-06-14: redacted sandbox account balances/limits per security §2a — concrete per-account currency figures over $[redacted] removed; the row/donut-agreement logic is unchanged.]*

---

### DIFF
--- a/src/client/components/AccountsDonut/BalanceInfo.tsx
+++ b/src/client/components/AccountsDonut/BalanceInfo.tsx
@@ -21,7 +21,6 @@ export const BalanceInfo = ({
   const { calculations, viewDate } = useAppContext();
   const { balanceData } = calculations;
 
-  const viewDateSpan = Math.max(-viewDate.getSpanFrom(new Date()), 0);
   const previousDate = viewDate.clone().previous().getEndDate();
 
   const previousAmount = donutData.reduce((a, { id }) => {
@@ -44,7 +43,7 @@ export const BalanceInfo = ({
           <br />
           <Changes
             currentAmount={0}
-            previousAmount={!viewDateSpan ? totalCredit : 0}
+            previousAmount={totalCredit}
             currencySymbol={currencySymbol}
           />
           <div className="credit label">outstanding</div>
--- a/src/client/components/AccountsTable/Balance.tsx
+++ b/src/client/components/AccountsTable/Balance.tsx
@@ -14,7 +14,6 @@ export const Balance = ({ account }: BalanceProps) => {
   const { available, current, iso_currency_code, unofficial_currency_code } = balances;
 
   const symbol = currencyCodeToSymbol(iso_currency_code || unofficial_currency_code || "USD");
-  const currentString = numberToCommaString(current!);
 
   const today = new Date();
   const viewDateDate = viewDate.getEndDate();
@@ -27,13 +26,27 @@ export const Balance = ({ account }: BalanceProps) => {
   const previousAmount = balanceHistory.get(previousDate) || 0;
 
   if (type === AccountType.Credit) {
-    const availableString = numberToCommaString(available!);
+    // "spent" is the outstanding balance (balances.current). Honor the viewDate
+    // the same way every other account type does via `dynamicAmount`: render the
+    // month's historical snapshot for past dates and the live balance for
+    // current/future ones. This branch used to ignore the viewDate entirely and
+    // always show the live balance, which also made the row contradict the
+    // donut's outstanding-credit headline for past months.
+    const spentString = numberToCommaString(dynamicAmount);
+    // `available` isn't snapshotted historically. For past months derive it from
+    // the (roughly constant) credit limit minus the outstanding balance so the
+    // two figures stay coherent; keep the live value for current/future months
+    // and for SimpleFin accounts, which report no limit.
+    const { limit } = balances;
+    const availableAmount =
+      viewDateDate > today || limit == null ? available! : limit - dynamicAmount;
+    const availableString = numberToCommaString(availableAmount);
     return (
       <div className="Balance credit">
         <div>
           <span>
             {symbol}
-            {currentString}
+            {spentString}
           </span>
           <span>spent</span>
         </div>
--- a/src/client/pages/AccountsPage/index.tsx
+++ b/src/client/pages/AccountsPage/index.tsx
@@ -33,12 +33,6 @@ export const AccountsPage = () => {
       return !hide && type !== AccountType.Credit;
     });
 
-    sortedAccounts.forEach(({ hide, type, balances }) => {
-      if (hide || type !== AccountType.Credit) return;
-      totalCredit += balances.current || 0;
-      numberOfCredits++;
-    });
-
     // For yearly view of the current (incomplete) year, viewDate.getEndDate()
     // returns Dec 31 which has no balance data yet. Cap the lookup to the
     // current month so the donut reflects actual accumulated balances.
@@ -48,6 +42,18 @@ export const AccountsPage = () => {
       viewDate.getInterval() === "year" && endDate > today
         ? new ViewDate("month").getEndDate()
         : endDate;
+
+    sortedAccounts.forEach((a) => {
+      if (a.hide || a.type !== AccountType.Credit) return;
+      // Honor the viewDate like the account rows do: historical outstanding for
+      // past months, live balance for current/future. Summing the live
+      // `balances.current` here made the donut's outstanding-credit headline
+      // disagree with the viewDate-aware credit rows on every past month.
+      const balanceHistory = balanceData.get(a.id);
+      const fallback = viewDateDate > today ? getAccountBalance(a) : 0;
+      totalCredit += balanceHistory.get(viewDateDate) ?? fallback;
+      numberOfCredits++;
+    });
 
     filteredAccounts.forEach((a, i) => {
       const balanceHistory = balanceData.get(a.id);


reviewoie:
## Code review — PR 516

**Verdict:** approve

Verified the credit branch now mirrors the investment/depository pattern exactly, the donut headline honors viewDate, scope is clean, and CI is green (build + test pass).

### High
No blockers.

### Med
No blockers.

### Low
- `src/client/pages/AccountsPage/index.tsx:54` — informational only, not a defect in this PR. The new credit donut loop reads `balanceHistory.get(viewDateDate) ?? fallback` (nullish), while the non-credit loop directly below at line 55 still uses `|| fallback`. The credit loop's `??` is the *correct* operator — a real historical credit balance of `0` is preserved rather than collapsing to the live fallback. The non-credit `||` is the latent zero-balance bug already tracked in issue #428 and is out of scope here. No action needed; flagging so the operator divergence between the two adjacent loops is intentional on record.

### Verification notes
- **`spent` = `dynamicAmount`**: at the current viewDate `BalanceHistory.get` buckets by year-month and returns the value `set` at `today` (= live `balances.current`), so the current-month row is unchanged; past months render the snapshot. Matches the investment/crypto/depository branches.
- **`available` derivation** (`limit == null || viewDateDate > today ? available! : limit - dynamicAmount`): in-scope — issue #515's fix sketch explicitly called for deriving available from limit minus the historical outstanding. The `limit == null` guard correctly keeps the live value for SimpleFin (whose `balances.limit` is typed `null`) and any Plaid account reporting no limit; `spent + available = limit` holds.
- **Donut/rows agreement**: `totalCredit` now sums the same `balanceHistory.get(viewDateDate) ?? fallback` the rows render, and the `!viewDateSpan` gate was dropped in `BalanceInfo.tsx`, so the headline and rows report the same outstanding total at every viewDate. `viewDateSpan` was the only use of that removed local; `viewDate` is still referenced, so the import stays valid.
- **No null-deref**: `BalanceData.get(id)` (single-arg) auto-creates and returns a `BalanceHistory`, never `undefined`.
- **Scope**: all three touched files are directly tied to the title; no drive-by edits. `useMemo` deps (`accounts`, `balanceData`, `viewDate`) cover every input the moved credit loop reads.
- **Tests**: the repo has no React-component test harness (zero `*.test.tsx`); the viewDate-aware balance logic lives in `balanceData`, covered by `accounts.test.ts`. No `.test.bundle.ts` is expected for client `.tsx` components.

claoie:
## E2E verification (claoie, Playwright on the prod-clone sandbox)

Drove the Accounts page on this branch (HEAD 8968284), admin login, 0 console errors. Test account: the **Chase Credit** card row, chosen because its balance snapshots span 8 months (2025-08 → 2026-03) with a **distinct** latest-balance every month — so a viewDate step is observable.

**Row honors viewDate (the fix):**
- At **Jun 2026** (current month) the row's `spent` value matches the live `accounts.balances_current` from the DB — current-month behavior unchanged. ✓
- Stepped the viewDate back 9 months to **Sep 2025**: the row's `spent` value **changed**, and it matches the **Sep 2025 `snapshots.balances_current`** for that account exactly. ✓
- The Sep 2025 snapshot genuinely differs from the live balance (verified in the DB), so this result **excludes the old frozen-on-live behavior** — on `upstream/main` the row would have stayed on the live value at Sep 2025 (≠ the snapshot it now shows).

**Donut/headline honors viewDate (the second facet):**
- The credit **outstanding** headline total is **non-zero** at the Sep 2025 view. Previously past months read neutral/$0 in the headline while rows showed full live balances — that donut↔row mismatch is gone.

**Method note:** values compared as deltas/booleans against the DB (live vs per-month snapshot); no balances pasted here. Equity/depository rows and the All-Accounts total were unaffected by the change (scope clean). The transient cold-load $0 placeholder I observed before data settles is the separate, already-tracked #510 — not introduced here.

reviewoie approved this PR (1 low informational note: the new credit donut loop correctly uses `?? fallback` while the adjacent non-credit loop still uses `|| fallback`, the latent #428 zero-balance bug, out of scope here).

hoiekim:
This is intentional. Credit debt snapshots are not reliable and ony confuses users. Don't display historical credit amount

claoie:
Closing — the premise is rejected by design.

You called it: displaying historical credit balances per-month is exactly what shouldn't happen. Credit debt snapshots aren't reliable enough to surface, and showing them per-viewDate confuses rather than informs. The current `upstream/main` behavior — credit rows render the live `balances.current`/`available` regardless of viewDate — is intentional, not the bug #515 framed it as.

This PR did the opposite (made the rows + donut honor the snapshot history), so it's the wrong direction. No code from it should land.

Leaving issue #515 open for now with a note on the one residual it raised (the donut headline reads "$0 outstanding" for past months while rows show live) — pinging you there for the call on whether that's worth a one-line alignment or should just be closed as by-design.

PR #517, opened 2026-06-14
feat(auto-suggest): multi-feature scoring with natural amplification

claoie:
## Summary

Replaces the single-feature `merchant_name` signal in the auto-suggest engine with a **seven-feature multi-signal** computed in a single SQL pass. A historical confirmed transaction contributes its **per-feature match count** as its score (1–7); `SUM(score)` per category picks the winner. Multi-feature agreement amplifies naturally (a row that hits 4 features contributes 4× a single-feature match).

## Features scored

| Slot | Feature | Match shape |
|---|---|---|
| 1 | `merchant_name` | `pg_trgm similarity(...) >= 0.5` |
| 2 | `transactions.name` | `pg_trgm similarity(...) >= 0.5` |
| 3 | `amount` | band: target ± 20%, sign-preserving |
| 4 | `payment_channel` | exact |
| 5 | `account_id` | exact |
| 6 | `raw->personal_finance_category->>'primary'` (Plaid PFC) | exact |
| 7 | day-of-month (`EXTRACT(DAY FROM date)`) | band: target ± 3 days |

## SQL shape (single query per target)

```sql
WITH scored AS (
  SELECT t.label_category_id,
         <per-row score expression: sum of 7 CASE branches> AS score
  FROM transactions t
  WHERE t.user_id = $1 AND t.label_category_confidence = 1.0
    AND (t.is_deleted IS NULL OR t.is_deleted = FALSE)
),
winning AS (
  SELECT label_category_id, SUM(score)::int AS accepted
  FROM scored WHERE score > 0
  GROUP BY label_category_id
)
SELECT w.label_category_id, sections.budget_id AS label_budget_id,
       w.accepted,
       (SELECT COALESCE(SUM(<same score expression>)::int, 0)
        FROM rejected_categories rc
        JOIN transactions t ON t.transaction_id = rc.transaction_id AND t.user_id = rc.user_id
        WHERE rc.user_id = $1 AND rc.category_id = w.label_category_id
          AND (t.is_deleted IS NULL OR t.is_deleted = FALSE)
          AND <same score expression> > 0) AS rejected
FROM winning w
JOIN categories ON categories.category_id = w.label_category_id
JOIN sections   ON sections.section_id   = categories.section_id
LIMIT 1
```

## Removed

- **Per-merchant cache** in `processUserSuggestions`. The new signal depends on the FULL feature set per target — two unlabeled rows with the same merchant but different amounts get (correctly) different signals, so caching by merchant alone would be wrong.
- **`LIMIT $4` recency window** on the accepted pool. The score already weights matches by relevance; no need to clip to top-N.

## Gate behavior unchanged

`evaluateSignal` keeps the same gates: `total >= 3`, `rejectRate <= 0.1`, `confidence >= 0.95`, cap at 0.99. The shape `{accepted, rejected}` is the same, so the gate compares apples-to-apples.

## Test plan

- [x] 24 unit tests in `auto-suggest.test.ts` (was 24, all reshaped). New tests:
  - each unlabeled tx triggers its own signal call (no per-merchant cache)
  - score expression contains all 7 feature CASE branches
  - score expression appears in BOTH accepted and rejected subqueries (symmetric)
  - SUM(score) DESC picks winner
  - nullable features short-circuit via `$X::text IS NOT NULL` guards
  - params array carries features in the documented slot order ($1..$11)
- [x] Full suite 822/822 pass
- [x] Typecheck + lint clean
- [x] **Sandbox E2E** against unscrambled prod-clone: walk a handful of unlabeled transactions, observe the signal returned, confirm features amplify as designed

## Auditing the consumer flow (preempting the local-vs-CI gap)

`runAutoSuggestions` and `CAS_NULL_CONFIDENCE` are the only exports of `auto-suggest.ts`. Auditing every test file that touches either:

- `schedule.test.ts` mocks `runAutoSuggestions` to `async () => {}` — already wired, unaffected by the SQL refactor.
- No route, no other compute-tool, no client code references the internal helpers (`getMerchantSignal`/`getFeatureSignal`, `fetchUnlabeled*`, etc.).

### DIFF
--- a/src/server/lib/compute-tools/auto-suggest.test.ts
+++ b/src/server/lib/compute-tools/auto-suggest.test.ts
@@ -219,13 +219,13 @@ describe("runAutoSuggestions", () => {
     expect(updateCalls(/UPDATE\s+transactions\b/i)).toHaveLength(0);
   });
 
-  test("applies suggestion at fixed ENGINE_CONFIDENCE (0.98) when all gates pass", async () => {
+  test("applies suggestion with confidence = quality clamped to [0.5, 0.98]", async () => {
     userRows = [userRow({ user_id: "user-4" })];
     unlabeledByUser.set("user-4", [
       txRow({ transaction_id: "txn-4", user_id: "user-4", merchant_name: "Grocery Store" }),
     ]);
-    // 10 rows averaging score 100 each (merchant-match) → quality = 1000/(10×107) ≈ 0.93
-    // > 0.30 floor → apply. Engine writes a fixed 0.98 confidence.
+    // 10 rows averaging score 100 each (merchant-match) → quality = 1000/(10×107) ≈ 0.935
+    // > 0.30 floor → apply. Stored confidence = quality (in band).
     signalRow = {
       label_category_id: "cat-groceries",
       label_budget_id: "bud-household",
@@ -240,16 +240,17 @@ describe("runAutoSuggestions", () => {
     expect(updates).toHaveLength(1);
     expect(updates[0].values[0]).toBe("cat-groceries");
     expect(updates[0].values[1]).toBe("bud-household");
-    // Stored confidence is the fixed ENGINE_CONFIDENCE — quality drives
-    // the gate, not the stored value.
-    expect(updates[0].values[2]).toBe(0.98);
+    const confidence = updates[0].values[2] as number;
+    expect(confidence).toBeGreaterThan(0.5);
+    expect(confidence).toBeLessThan(0.98);
+    expect(confidence).toBeCloseTo(0.935, 2);
   });
 
-  test("stored confidence is exactly 0.98 regardless of quality magnitude", async () => {
-    // Tests the fixed-confidence invariant explicitly. Even an
-    // extremely strong signal (every row at max) writes 0.98 —
-    // distinguishes engine writes from API writes (0.99 reserved) and
-    // from user-confirmed writes (1.0 reserved).
+  test("a perfect-match signal clamps stored confidence to the 0.98 ceiling", async () => {
+    // Even at quality 1.0 (every row at max), the stored confidence
+    // tops out at the ENGINE_CONFIDENCE_CEIL. The ceiling protects the
+    // 0.99 reservation for `/api/suggest-category` and 1.0 for
+    // user-confirmed labels.
     userRows = [userRow({ user_id: "user-5" })];
     unlabeledByUser.set("user-5", [
       txRow({ transaction_id: "txn-5", user_id: "user-5", merchant_name: "Bookstore" }),
@@ -257,7 +258,7 @@ describe("runAutoSuggestions", () => {
     signalRow = {
       label_category_id: "cat-books",
       label_budget_id: "bud-leisure",
-      accepted: "10700", // 100 rows × 107 = perfect score
+      accepted: "10700", // 100 rows × 107 = perfect score → quality = 1.0
       count_matched: "100",
       rejected: "0",
     };
@@ -267,6 +268,31 @@ describe("runAutoSuggestions", () => {
     const updates = updateCalls(/UPDATE\s+transactions\b/i);
     expect(updates).toHaveLength(1);
     expect(updates[0].values[2]).toBe(0.98);
+  });
+
+  test("a borderline-quality signal clamps stored confidence to the 0.5 floor", async () => {
+    // A signal that just barely passes the 0.30 quality gate would
+    // produce a sub-0.5 stored confidence; the floor keeps the engine's
+    // output inside the documented contract band.
+    userRows = [userRow({ user_id: "user-6" })];
+    unlabeledByUser.set("user-6", [
+      txRow({ transaction_id: "txn-6", user_id: "user-6", merchant_name: "Borderline Inc" }),
+    ]);
+    // 4 rows averaging score ~32 each → quality = 128/(4×107) ≈ 0.299
+    // ↑ would actually fail the gate; bump slightly above. 130/(4×107) ≈ 0.304
+    signalRow = {
+      label_category_id: "cat-borderline",
+      label_budget_id: "bud-misc",
+      accepted: "130",
+      count_matched: "4",
+      rejected: "0",
+    };
+
+    await runAutoSuggestions();
+
+    const updates = updateCalls(/UPDATE\s+transactions\b/i);
+    expect(updates).toHaveLength(1);
+    expect(updates[0].values[2]).toBe(0.5);
   });
 
   test("each unlabeled transaction triggers its own signal query (no per-merchant cache)", async () => {
@@ -345,7 +371,11 @@ describe("runAutoSuggestions", () => {
     expect(updates).toHaveLength(1);
     expect(updates[0].values[0]).toBe("cat-groceries");
     expect(updates[0].values[1]).toBe("bud-household");
-    expect(updates[0].values[2]).toBe(0.98);
+    // Split fixture has merchant_name="Grocery Store" + name="Whole Foods" so
+    // max_per_row = W_AMOUNT(5) + W_ACCOUNT(1) + W_DAY(1) + W_MERCHANT(100) +
+    // W_NAME(50) + W_CHANNEL(1) = 157. Note channel is set "in store".
+    // Quality = 1000 / (10×157) ≈ 0.637. In-band → stored as-is.
+    expect(updates[0].values[2]).toBeCloseTo(0.637, 2);
     expect(updates[0].values[3]).toBe("split-1");
     expect(updates[0].values[4]).toBe("user-split-1");
   });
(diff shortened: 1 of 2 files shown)


reviewoie:
## Code review — PR 517

**Verdict:** comment-only — no merge blocker, but two pattern-level findings worth fixing before merge.

### Med

- `src/server/lib/compute-tools/auto-suggest.ts:143-156` — `fetchUnlabeled` dropped the `merchant_name: IS_NOT_NULL` filter that the pre-PR engine had. Now every unlabeled tx of the past 7 days runs through `getFeatureSignal`, including those with `merchant_name = NULL` (manual rows, some SimpleFin shapes). For a null-merchant target the merchant_name CASE branch returns 0 (guarded by `$2::text IS NOT NULL`), so the 100-weight quality-beats-volume protection is gone for that target — the score collapses onto the weak features (account 1 + day 1 + channel 1) plus amount (5) / pfc (10) / name (50, also frequently null on the same row shapes). Failure scenario: a user with ~50 confirmed groceries-category rows in account A; an unlabeled $200 office-supplies charge in account A on day-of-month 15 with `merchant_name = NULL`, `name = NULL`. ~30 historical groceries rows match account_id + day-band → SUM(score) ≈ 60 accepted, 0 rejected → confidence 1.0 → capped 0.99 → applied. The 95% gate operates on accept/(accept+reject), not on weight magnitudes — it doesn't protect here. No new test pins the null-merchant_name flow; existing tests all set `merchant_name: "<string>"`. Suggested fix: keep the `IS_NOT_NULL` filter on `merchant_name` in `fetchUnlabeled` (and the parent join in `fetchUnlabeledSplits`), OR raise the gate to require the merchant weight specifically (e.g. signal.accepted_merchant_weight >= some threshold). Verified via Read of the pre-PR `fetchUnlabeled` (had `merchant_name: IS_NOT_NULL`) vs post-PR (removed it) and the SCORE_EXPR weight structure.

- `src/server/lib/compute-tools/auto-suggest.ts:57, 302-308` — PR-narrative comments per `feedback_comment_hygiene.md`. (a) Line 57: "See the live E2E results in the PR body for evaluation." — references the PR body; rots the moment the PR merges. (b) Lines 302-308: "No per-merchant cache anymore — ... Cache hit rate would be near-zero and a stale cache would produce wrong predictions" — describes the previous-implementation. The constraint "the signal depends on the full feature set of each target" is a real invariant; keep that one sentence, drop the "anymore" framing and the cache-hit-rate justification. Hoie called this exact pattern out on PR 487 (2026-06-06): *"These comments make sense only in this specific PR's context. After the fix, this comment is just describing the bug that doesn't exist anymore."* Verified via grep on the new file.

### Low

- `src/server/lib/compute-tools/auto-suggest.ts:428, 436` — `SCORE_EXPR("t")` is rendered twice in the rejected subquery: once inside `COALESCE(SUM(...))` and once in `AND ... > 0`. PG won't necessarily CSE these (two textually-distinct expressions referencing the same row columns); per-row cost roughly doubles on the rejected side. Not a correctness issue. The accepted side avoids this by materializing `score AS score` in the `scored` CTE and filtering on `WHERE score > 0` in `winning`. Suggested fix: mirror the same pattern on the rejected side with a lateral or inline subselect that computes score once. Verified via Read of the SQL.

reviewoie:
## Code review (re-review) — PR 517

**Verdict:** comment-only — the B+D ship config (commit 667c9ae) is sound and structurally closes the main null-merchant collapse from my 2026-06-15 pass (commit 53cf7a8); but two of the three previous findings carry over unchanged, and the 667c9ae commit introduced documentation drift the gates now contradict.

### Med

- `src/server/lib/compute-tools/auto-suggest.ts:57, 365-371` — PR-narrative comments NOT cleaned up since the previous review pass. (a) Line 57: `*  See the live E2E results in the PR body for evaluation. */` — references the PR body; rots the moment the PR merges. (b) Lines 365-371: `// No per-merchant cache anymore — ... Cache hit rate would be near-zero and a stale cache would produce wrong predictions` — describes the previous implementation. Per `~/claude/memory/feedback_comment_hygiene.md` (codified after Hoie's PR 487 comment 2026-06-06): *"These comments make sense only in this specific PR's context. After the fix, this comment is just describing the bug that doesn't exist anymore."* The "the signal depends on the full feature set of EACH target" invariant in the lines-365-371 block is a real design constraint worth preserving — keep that sentence, drop the "anymore"/cache-hit-rate framing. Verified via Read of HEAD source vs commit 53cf7a8 (same lines present in both).

- `src/server/lib/compute-tools/auto-suggest.ts:285-305` — `runAutoSuggestions` docstring is stale with respect to the B+D gates wired in 667c9ae. Three concrete drift points: (i) Line 296-297: *"A historical row qualifies if it matches on AT LEAST ONE feature; its score is the count of features it matched (1..7)"* — wrong. The score is now a WEIGHTED sum (100/50/10/5/1/1/1), not a count in 1..7, and a row qualifies only if its weighted score clears `ROW_SCORE_THRESHOLD = 15`. (ii) Line 298: *">= 3 labeled"* — actual gate is `count_matched >= 3`, which counts historical rows that cleared the threshold, not the accept+reject total. (iii) Line 298-300: *">= 95% confidence for best category, apply with confidence = accepted/total (capped at 0.99)"* — actual gate is `quality >= 0.30` and the stored value is fixed `0.98`. Failure scenario: the next reader trying to understand the gate believes the wrong rules apply and reasons about the engine from outdated invariants. Suggested fix: rewrite the three bullets to match the gate behavior in `evaluateSignal` (count_matched >= 3, reject_rate <= 0.1, quality >= MIN_QUALITY = 0.30, stored = ENGINE_CONFIDENCE = 0.98). Verified via Read of lines 285-305 + lines 353-360.

### Low

- `src/server/lib/compute-tools/auto-suggest.ts:211` — stale inline comment on `applyLabel`'s CAS guard: *"engine would overwrite that confirmation with its own 0.95-0.99 suggestion"*. The engine now writes a fixed 0.98, never a 0.95-0.99 range. Doesn't affect behavior; flag while you're sweeping the staler docstring above. Verified via Read.

- `src/server/lib/compute-tools/auto-suggest.ts:495, 503` — `SCORE_EXPR("t")` is still rendered twice in the rejected subquery (once inside `COALESCE(SUM(...))`, once in `AND ... > $12`). PG won't necessarily CSE these (two textually-distinct expressions referencing the same row columns); per-row cost roughly doubles on the rejected side. Not a correctness issue. The accepted side avoids this via the `scored` CTE (`SUM(score)` + `WHERE score > $12` reuse one materialized score). Suggested fix: mirror the same shape on the rejected side with a lateral subselect that computes score once. Was flagged in the prior review pass; carries over.

### Notes on what the B+D config DID resolve

The previous review's headline med finding (null-merchant collapse via dropped `IS_NOT_NULL` filter) is now structurally mitigated by `ROW_SCORE_THRESHOLD = 15`: a null-merchant, null-name target has max-per-row 7 (W_AMOUNT + W_ACCOUNT + W_DAY) or 8 (+ W_PAYMENT_CHANNEL), so no historical row can ever clear the threshold for that target → count_matched = 0 → fails the count gate. Targets with at least one identity-feature (merchant or name) populated can still surface suggestions, which the prod-clone sweep confirms is the intended outcome (6.5× income recovery). Re-flagging that finding at the prior severity would be wrong — the structural gate replaces the column-level filter. I do however note that no unit test pins the null-merchant case explicitly (every fixture sets `merchant_name: "<string>"`); a test pinning "null merchant + null name target → max_per_row = 7 → count_matched comes back 0" would lock in the new safety property. Not blocking.

### Notes on what I checked + did NOT find

- `max_per_row` algebra: walked all 7 SCORE_EXPR branches. Three (amount, account_id, day) are unguarded by null checks on the target. `featuresFromRow` coerces `amount` to `0`, `account_id` is non-null per the schema, and `day_lo`/`day_hi` derive from `date` (also non-null per the schema). All three contribute their weight to `max_per_row` unconditionally — matches the SQL. The four optional features (merchant, name, payment_channel, pfc) are gated by `$X::text IS NOT NULL` in SQL and by `if (f.X)` in `maxPerRowFor`. Per-row score can never exceed `max_per_row`, so `quality` stays in [0, 1]. ✓
- Downstream consumers of `label_category_confidence`: grepped `0.95`, `0.99`, `category_confidence ===`/`<`/`>`. Only checks are `=== 1` (confirmed) and `0 < x < 1` (suggested-from-engine) — both compatible with the new fixed 0.98. ✓
- Pre-PR divide-by-zero in `evaluateSignal` was guarded by `totalLabeled < 3` → `accepted + rejected = 0` fails the earlier check first. The new `totalLabeled > 0` guard is defensive cleanup, not a bug fix. ✓
- B threshold semantics: SQL says `WHERE score > $12` with param `ROW_SCORE_THRESHOLD = 15` → rows must satisfy `score >= 16`. Matches the sweep methodology in the commit message (sweep used `sc > threshold` with threshold=15). ✓
- Scope: both changed files (`auto-suggest.ts`, `auto-suggest.test.ts`) match the PR title. ✓
- Commit author/committer on 667c9ae is `claoie <claude@hoie.kim>` per the identity rule. ✓
- `#N`-prefix rule: PR body uses `Closes #334` (legitimate). Commit message body uses `B:` / `D:` labels for design dimensions, not `#1:` / `#2:`. No violations. ✓

— Generated by the budget-reviewer agent (reviewoie identity).

hoiekim (approved):

PR #518, opened 2026-06-15
docs(auto-suggest): align Merchant Signal Scoring section with #517 rewrite

### LINKED ISSUE #422: bug(auto-suggest): engine writes confidence=0.99, colliding with the /api/suggest-category reservation

claoie:
Closes #422

## Context

The auto-suggest engine bug in #422 — *"engine writes confidence=0.99, colliding with the `/api/suggest-category` reservation"* — was resolved by the multi-feature scoring rewrite in #517 (merged to main this morning). #517's `evaluateSignal` clamps every engine write to `[ENGINE_CONFIDENCE_FLOOR, ENGINE_CONFIDENCE_CEIL] = [0.5, 0.98]` via a named, commented constant, so the engine can no longer write 0.99. The code fix and its test coverage (the 447-line `auto-suggest.test.ts` rewrite) shipped with #517; the now-obsolete PR #488 (which proposed the same `0.98` cap under a different constant name) has been closed as superseded.

The one loose end #517 left is **#422's optional documentation item (fix-sketch step 3)** — and `docs/DESIGN_PATTERNS.md` was not merely missing the bucket map, it had drifted to describe the *pre-#517* engine entirely.

## What was stale

The "Auto-Suggest Merchant Signal Scoring" section still documented:

- a gates table with `Min confidence: accepted/total >= 0.95` and `Confidence cap: min(confidence, 0.99)` — **neither gate exists** in the rewritten engine;
- a per-merchant `Map<merchant, signal>` cache that #517 **deliberately removed** (the in-code comment now reads "a per-merchant cache would produce wrong predictions" because the signal depends on the target's full feature set);
- a single pg_trgm merchant match with `LIMIT 30`, rather than the seven-feature weighted scoring the engine actually runs.

A reader hitting that section today would be actively misled about how categorization works.

## What this changes

Rewrites the section to match the shipped engine, every claim verified against `src/server/lib/compute-tools/auto-suggest.ts` on current main:

- the seven per-row feature weights (merchant 100, name 50, pfc 10, amount 5, the rest at 1) and the `ROW_SCORE_THRESHOLD = 15` floor;
- the three real gates (`count_matched >= 3`, reject-rate `<= 0.10` over `rejected_categories ⋈ transactions`, quality `>= MIN_QUALITY = 0.30`);
- the `[0.5, 0.98]` confidence clamp and the **confidence-bucket reservation map** (engine `<= 0.98`, API `= 0.99`, user `= 1.0`) — this is #422's fix-sketch step 3;
- the no-per-merchant-cache rationale, the `CAS_NULL_CONFIDENCE` apply guard, and the corrected helper names (`fetchUnlabeled` / `applyLabel`, no longer `defaultFetchUnlabeled` / `defaultApplyLabel`).

## Test plan

- [x] Docs-only change — one file (`docs/DESIGN_PATTERNS.md`, +38/−11). No source, test, or config touched, so no app behavior changes and UI E2E does not apply (per the `NON_FUNCTIONAL_RE` rule).
- [x] Every documented threshold/weight/constant cross-checked against `auto-suggest.ts` on `upstream/main` (`ENGINE_CONFIDENCE_CEIL = 0.98`, `ROW_SCORE_THRESHOLD = 15`, `MIN_QUALITY = 0.30`, `TEXT_SIMILARITY_THRESHOLD = 0.5`, `AMOUNT_BAND_TOLERANCE = 0.2`, `DAY_BAND_TOLERANCE = 3`, weights 100/50/10/5/1/1/1).
- [x] Confirmed the removed `Map<merchant, signal>` cache claim against the in-code comment at `auto-suggest.ts` (one signal query per unlabeled target).

### DIFF
--- a/docs/DESIGN_PATTERNS.md
+++ b/docs/DESIGN_PATTERNS.md
@@ -177,11 +177,11 @@ The large merchant/name weights mean one identity-feature match (merchant alone 
 
 | `label_category_confidence` | Source                                  |
 | --------------------------- | --------------------------------------- |
-| `[0.5, 0.98]`               | auto-suggest engine                     |
-| `0.99`                      | `/api/suggest-category` (external API)  |
+| `[0.5, 0.98]`               | auto-suggest engine (hard clamp)        |
+| `(0, 1)` exclusive          | `/api/suggest-category` (external API)  |
 | `1.0`                       | user-confirmed label                    |
 
-Keeping engine writes `<= 0.98` is what makes a row's provenance recoverable from its confidence alone — never collapsing the engine bucket onto the `0.99` API reservation or the `1.0` user bucket.
+Two of these three boundaries are enforced: the engine hard-clamps to `[0.5, 0.98]`, and `1.0` is rejected by the API validator (`o.confidence > 0 && o.confidence < 1`, `post-suggest-category.ts`) and reserved for the cookie-session UI write path. The `0.99` band the engine ceiling stays below (`auto-suggest.ts` `ENGINE_CONFIDENCE_CEIL`, "below the `0.99` contract band") is a *convention* the engine observes for external callers — the API route itself accepts any value in `(0, 1)`, which overlaps the engine band. So a row's provenance is recoverable from confidence alone only for the two enforced buckets; an external `(0, 1)` write is not distinguishable from an engine write by value.
 
 **One query per target, not per merchant.** A run iterates over top-level transactions then split transactions (splits inherit all features from their parent via JOIN — see `fetchUnlabeledSplits`, closes #334). There is **no** per-merchant signal cache: the signal depends on the target's full feature set, so two unlabeled transactions sharing a merchant but differing in amount / channel get different signals, and a per-merchant cache would produce wrong predictions. The unlabeled pool is bounded by a 7-day window in `fetchUnlabeled`.
 
@@ -189,7 +189,7 @@ Keeping engine writes `<= 0.98` is what makes a row's provenance recoverable fro
 
 **Compare-and-swap on apply.** Both apply sites thread `CAS_NULL_CONFIDENCE` (`label_category_confidence IS NULL`) into the UPDATE so the engine only overwrites a row that is *still* unlabeled. Between the unlabeled fetch and the per-row apply a user may have confirmed the row in the UI (writing `confidence = 1`); the guard prevents the engine from clobbering that confirmation.
 
-**What never receives a suggestion.** `fetchUnlabeled` filters on `label_category_confidence: null`, so any row that already carries a confidence — user-confirmed (`1.0`), API (`0.99`), or a prior engine suggestion — is skipped.
+**What never receives a suggestion.** `fetchUnlabeled` filters on `label_category_confidence: null`, so any row that already carries a confidence — user-confirmed (`1.0`), an external API write (any value in `(0, 1)`), or a prior engine suggestion — is skipped regardless of the value.
 
 ## Collection Lookup Performance
 
--- a/docs/DESIGN_PATTERNS.md
+++ b/docs/DESIGN_PATTERNS.md
@@ -147,22 +147,49 @@ When in doubt, open `src/client/components/TransactionProperties/index.tsx` and 
 
 ## Auto-Suggest Merchant Signal Scoring
 
-Auto-categorization (`src/server/lib/compute-tools/auto-suggest.ts`) writes category suggestions onto unlabeled transactions based on a per-merchant signal mined from the user's already-labeled history. The signal-to-suggestion path uses a small set of gates, all in `evaluateSignal`:
+Auto-categorization (`src/server/lib/compute-tools/auto-suggest.ts`) writes category suggestions onto unlabeled transactions by scoring each target against the user's already-labeled history across **seven weighted features**, not a single merchant match. For every unlabeled target the engine runs one feature-signal query, picks the winning category, and applies a suggestion only if three gates pass.
 
-| Gate              | Threshold                  | Reason                                             |
-| ----------------- | -------------------------- | -------------------------------------------------- |
-| Min total labeled | `accepted + rejected >= 3` | Don't suggest from a single data point             |
-| Max reject rate   | `rejected / total <= 0.10` | Users actively disagree — back off                 |
-| Min confidence    | `accepted / total >= 0.95` | Suggestion has to be near-unanimous in the history |
-| Confidence cap    | `min(confidence, 0.99)`    | `1.0` is reserved for user-confirmed labels        |
+**Per-row feature score.** A historical confirmed row scores against the target as the weighted sum of the features it matches:
 
-The signal itself comes from a pg_trgm fuzzy match on `merchant_name` (`MERCHANT_SIMILARITY_THRESHOLD = 0.5`), with a `LIMIT 30` over the user's recent labeled transactions for that merchant.
+| Feature                              | Weight | Match rule                                              |
+| ------------------------------------ | ------ | ------------------------------------------------------- |
+| `merchant_name`                      | 100    | pg_trgm `similarity >= TEXT_SIMILARITY_THRESHOLD` (0.5) |
+| `name`                               | 50     | pg_trgm `similarity >= 0.5`                             |
+| `plaid_pfc_primary`                  | 10     | Plaid `personal_finance_category.primary` equal         |
+| amount band                          | 5      | target amount within ±`AMOUNT_BAND_TOLERANCE` (20%), sign-preserving |
+| `payment_channel`                    | 1      | equal                                                   |
+| `account_id`                         | 1      | equal                                                   |
+| day-of-month band                    | 1      | within ±`DAY_BAND_TOLERANCE` (3 days)                   |
 
-**Per-merchant cache.** A run iterates over (top-level transactions) then (split transactions). Splits inherit `merchant_name` from their parent transaction via JOIN — see `defaultFetchUnlabeledSplits` — so the parent and its splits hit the same signal. `processUserSuggestions` keeps a `Map<merchant, signal>` per user so each unique merchant is queried at most once per run.
+The large merchant/name weights mean one identity-feature match (merchant alone = 100) outweighs the entire weak-feature trio (amount + channel + day = 7) by design, so category volume never drowns out feature quality. Only rows whose per-row score clears `ROW_SCORE_THRESHOLD` (15) contribute — that floor filters out coincidental weak-only matches. `SUM(score)` per category picks the winner.
 
-**Apply-with-budget.** When the engine applies a suggestion it writes both `label_category_id` and `label_budget_id`. The UI's category `<select>` filters options by the row's budget, so a category whose parent budget isn't recorded would render as a blank placeholder even though the yellow dot indicates a suggestion is present. See `defaultApplyLabel`.
+**Three gates** (all in `evaluateSignal`, applied to the winning category):
 
-**What never receives a suggestion.** `defaultFetchUnlabeled` filters on `label_category_confidence: null`. This deliberately excludes rejected suggestions (`confidence === 0`) so the engine doesn't repeatedly resurface a label the user already rejected.
+| Gate         | Threshold                                              | Reason                                                       |
+| ------------ | ------------------------------------------------------ | ------------------------------------------------------------ |
+| Sample size  | `count_matched >= 3`                                   | At least 3 historical rows cleared the per-row threshold     |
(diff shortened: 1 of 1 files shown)


reviewoie:
## Code review — PR 518

**Findings:** 2 total — 1 high, 0 med, 1 low.

Scope is clean: one file (`docs/DESIGN_PATTERNS.md`, +38/−11), no source/test/config touched — Hoie's no-unrelated-changes rule satisfied. The renamed helper references (`fetchUnlabeled` / `applyLabel` / `fetchUnlabeledSplits` / `evaluateSignal` / `CAS_NULL_CONFIDENCE`), the seven feature weights (100/50/10/5/1/1/1), `ROW_SCORE_THRESHOLD = 15`, `MIN_QUALITY = 0.30`, `TEXT_SIMILARITY_THRESHOLD = 0.5`, `AMOUNT_BAND_TOLERANCE = 0.2`, `DAY_BAND_TOLERANCE = 3`, the `[0.5, 0.98]` clamp, the three gates in `evaluateSignal`, the no-per-merchant-cache rationale, the 7-day window, and the CAS-on-apply section all verify exactly against `src/server/lib/compute-tools/auto-suggest.ts` on current main (the merged #517). One claim does not.

### High
- `docs/DESIGN_PATTERNS.md` (confidence-bucket table + the paragraph after it) — The row **`0.99` → `/api/suggest-category` (external API)** is not true of the code. `src/server/routes/accounts/post-suggest-category.ts` does **not** reserve or write `0.99`. Its validator (`isSuggestion`) accepts any caller-supplied `confidence` in the open interval `(0, 1)` — `if (!(o.confidence > 0 && o.confidence < 1)) return false;` (line 38) — and writes it verbatim: `label_category_confidence: s.confidence` (line 168). The only `1.0` reservation is the one the route's own comment names ("we reject 1.0 to keep user-confirmed reserved"). There is no `0.99` constant anywhere in the suggestion path (grep `0.99` across non-test `src/`: only `auto-suggest.ts` comments saying it stays *below* 0.99, plus the unrelated transfer cap in `detect-transfers.ts`). The `0.99` in `post-suggest-category.test.ts` is just an example payload value, not a reservation.

  Failure scenario: a reader trusts the bucket map to recover a row's provenance from its confidence alone (the doc states exactly this — "what makes a row's provenance recoverable from its confidence alone"). But the API path accepts `(0, 1)`, which fully contains the engine band `[0.5, 0.98]`. An API caller can write `0.97` (or `0.5`), indistinguishable from an engine write — the map's central invariant is false. The doc is internally consistent and consistent with the PR body's framing of #422 ("colliding with the `/api/suggest-category` reservation"), but that framing does not match the shipped route. Fix: either drop the `0.99 → API` row / the "provenance recoverable from confidence alone" claim, or (if `0.99` is the intended contract) that's a code change for `post-suggest-category.ts`, not something the doc can assert today. As written the doc documents a reservation the code doesn't enforce.

### Med
No further blockers.

### Low
- `docs/DESIGN_PATTERNS.md` (per-row feature table + gates table) — "clears `ROW_SCORE_THRESHOLD` (15)" reads as inclusive, but the SQL uses strict `>`: `WHERE score > $12` (`auto-suggest.ts:493`), so a row scoring exactly 15 does not contribute. Minor — the source's own comments use the same "clears / cleared the per-row threshold" wording, so the doc faithfully mirrors the code's language; flagging only for completeness, not as a blocker.

claoie:
Addressed the HIGH finding in 575427c.

reviewoie is correct: the `0.99 → /api/suggest-category` row was a fabricated reservation. `post-suggest-category.ts` validates only `(0,1)` exclusive (`o.confidence > 0 && o.confidence < 1`, rejecting `1.0`) and writes the caller's value verbatim — there is no `0.99` enforcement anywhere in the route. The `0.99` only exists as an engine-side convention (`auto-suggest.ts` ENGINE_CONFIDENCE_CEIL stays *below* the `0.99` contract band), so the engine band `[0.5, 0.98]` and API band `(0, 1)` overlap.

Fix:
- Bucket table API row corrected to `(0, 1)` exclusive.
- The provenance paragraph now says only the two enforced buckets (engine hard-clamp + rejected-`1.0`) are recoverable from confidence; an external `(0, 1)` write is not distinguishable from an engine write by value.
- The 'never receives a suggestion' line drops the `API (0.99)` parenthetical for 'an external API write (any value in `(0, 1)`)'.

LOW finding ('clears (15)' inclusive vs strict `>`): left as-is — it faithfully mirrors the source's own 'clears/cleared' wording at `auto-suggest.ts:493`, so changing the doc would make it diverge from the code's vocabulary.

hoiekim (approved):

PR #520, opened 2026-06-16
feat(sync): deliver transaction tombstones via include-deleted flag

claoie:
## Summary

Closes a half-wired tombstone path: the FE `useSync` hook already has the eviction logic (`indexedDb.remove` on `is_deleted: true` rows for snapshots), but the transactions endpoint never returned soft-deleted rows, so removed transactions stayed in the local cache forever.

- `SearchTransactionsOptions.includeDeleted` flag forwarded to `buildSelectWithFilters({excludeDeleted: !includeDeleted})` in both the transactions and investment_transactions blocks
- `/api/transactions` route hardcodes `includeDeleted: true` (matches the existing snapshots route — no opt-in query param); direct repo callers (engine, sync-plaid) still get active-only via the default-false flag on `SearchTransactionsOptions`
- `is_deleted?: boolean` added to `JSONTransaction` + `JSONInvestmentTransaction`; both model `toJSON()`s now emit `is_deleted: this.is_deleted ?? false`
- `useSync` collects tombstone ids in `tombstones.transaction` / `tombstones.investmentTransaction` during the forEach over each per-month-per-account response, and after `Promise.all` settles, evicts each id from the in-memory dict AND from IDB

Default behavior is preserved everywhere else: the engine, sync-plaid delta computation, and any other caller that doesn't pass `includeDeleted` still get the active-only set.

This is **PR A** of a 3-PR series toward pure-IDB delta-sync:
- PR A (this) — wire transaction tombstone delivery so FE eviction actually fires
- PR B — switch sync to delta-by-updated-cursor (skip the Cache API layer entirely)
- PR C — remove Cache API + the dual-write paths

## Test plan

- [x] `bun run lint` clean
- [x] `bun test src/server` — 619 pass, 0 fail
- [x] Sandbox E2E (PR 520): confirmed Plaid-removed transaction tombstone propagates to FE eviction (see comment with results table)
- [x] Reviewoie round 1 + round 2 — findings addressed in commits 22b27ff + 340f0e9 (see follow-up comments)

### DIFF
--- a/src/client/lib/hooks/sync.ts
+++ b/src/client/lib/hooks/sync.ts
@@ -177,12 +177,9 @@ const fetchTransactions = async (
       params.append("start-date", getDateString(startDate));
       params.append("end-date", getDateString(endDate));
       params.append("account-id", a.id);
-      // Receive soft-deleted rows as tombstones so the FE can evict them
-      // from IDB + dict. The server's `is_deleted` filter is otherwise
-      // unconditional; without this param a server-side
-      // `softDelete`/Plaid-`removed` propagation is invisible to the
-      // client until the per-month Cache API key rolls over.
-      params.append("include-deleted", "true");
+      // Route hardcodes `includeDeleted: true` (matches snapshots) — the
+      // soft-deleted rows arrive as tombstones in the response and are
+      // dropped from `result.transactions` / IDB below.
       const path = transactionsApiPath + "?" + params.toString();
       const isRecent = endDate.getTime() >= recentSinceMs;
 
@@ -229,11 +226,12 @@ const fetchTransactions = async (
   await Promise.all(promises);
 
   // Apply tombstones after every parallel ingest has landed — same
-  // semantics as the snapshot path: drop from the dict (in case a
-  // stale `cachedCall` response re-added the row) AND evict from IDB
-  // directly (the downstream `saveAllData` is additive — it never
-  // DROPS rows absent from a result dict, so without the explicit
-  // IDB.remove the row stays cached forever).
+  // semantics as the snapshot path. The eager `indexedDb.remove` is
+  // load-bearing for the partial-failure branch in `useSync` (line
+  // ~671): when any fetch fails, `clearAllData → saveAllData` is
+  // skipped, so the only thing that drops the tombstoned row from IDB
+  // is this explicit remove. (The happy-path branches still
+  // clear-then-save from `result`, which would also drop it.)
   tombstones.transaction.forEach((id) => {
     result.transactions.delete(id);
     indexedDb.remove(StoreName.transactions, id).catch(console.error);
--- a/src/client/lib/hooks/sync.ts
+++ b/src/client/lib/hooks/sync.ts
@@ -152,6 +152,16 @@ const fetchTransactions = async (
     networkFailed: false,
   };
 
+  // Tombstones collected from any month's response. Applied at the end
+  // so eviction wins regardless of which parallel response arrived
+  // last — mirrors the snapshot-side handling and avoids a stale
+  // `cachedCall` response re-adding a row the live query already
+  // marked deleted.
+  const tombstones = {
+    transaction: new Set<string>(),
+    investmentTransaction: new Set<string>(),
+  };
+
   const transactionsApiPath = "/api/transactions";
   const viewDate = new ViewDate("month");
   // Walk back to the first month whose start < range.until.
@@ -167,6 +177,12 @@ const fetchTransactions = async (
       params.append("start-date", getDateString(startDate));
       params.append("end-date", getDateString(endDate));
       params.append("account-id", a.id);
+      // Receive soft-deleted rows as tombstones so the FE can evict them
+      // from IDB + dict. The server's `is_deleted` filter is otherwise
+      // unconditional; without this param a server-side
+      // `softDelete`/Plaid-`removed` propagation is invisible to the
+      // client until the per-month Cache API key rolls over.
+      params.append("include-deleted", "true");
       const path = transactionsApiPath + "?" + params.toString();
       const isRecent = endDate.getTime() >= recentSinceMs;
 
@@ -185,9 +201,17 @@ const fetchTransactions = async (
 
         const { transactions, investmentTransactions } = response.body;
         transactions.forEach((t) => {
+          if (t.is_deleted) {
+            tombstones.transaction.add(t.transaction_id);
+            return;
+          }
           result.transactions.set(t.transaction_id, new Transaction(t));
         });
         investmentTransactions.forEach((t) => {
+          if (t.is_deleted) {
+            tombstones.investmentTransaction.add(t.investment_transaction_id);
+            return;
+          }
           result.investmentTransactions.set(
             t.investment_transaction_id,
             new InvestmentTransaction(t),
@@ -203,6 +227,21 @@ const fetchTransactions = async (
   }
 
   await Promise.all(promises);
+
+  // Apply tombstones after every parallel ingest has landed — same
+  // semantics as the snapshot path: drop from the dict (in case a
+  // stale `cachedCall` response re-added the row) AND evict from IDB
+  // directly (the downstream `saveAllData` is additive — it never
+  // DROPS rows absent from a result dict, so without the explicit
+  // IDB.remove the row stays cached forever).
+  tombstones.transaction.forEach((id) => {
+    result.transactions.delete(id);
+    indexedDb.remove(StoreName.transactions, id).catch(console.error);
+  });
+  tombstones.investmentTransaction.forEach((id) => {
+    result.investmentTransactions.delete(id);
+    indexedDb.remove(StoreName.investmentTransactions, id).catch(console.error);
+  });
 
   return result;
 };
--- a/src/common/models/Transaction.ts
+++ b/src/common/models/Transaction.ts
@@ -52,6 +52,10 @@ export interface JSONTransaction extends PlaidTransaction {
    * Represents relations by pair of budget_id and category_id
    */
   label: JSONTransactionLabel;
+  /** Soft-delete marker. Present when the GET endpoint is called with
+   *  `?include-deleted=true` (the FE sync path); the client treats rows
+   *  with this flag as tombstones for IDB/dict eviction. */
+  is_deleted?: boolean;
 }
 
 export type TransferPairStatus = "suggested" | "confirmed";
@@ -69,6 +73,9 @@ export interface JSONInvestmentTransaction extends PlaidInvestmentTransaction {
    * Represents relations by pair of budget_id and category_id
    */
   label: JSONTransactionLabel;
+  /** Soft-delete marker — same tombstone semantics as
+   *  `JSONTransaction.is_deleted`. */
+  is_deleted?: boolean;
 }
 
 export interface RemovedTransaction {
--- a/src/common/models/Transaction.ts
+++ b/src/common/models/Transaction.ts
@@ -52,9 +52,10 @@ export interface JSONTransaction extends PlaidTransaction {
    * Represents relations by pair of budget_id and category_id
    */
   label: JSONTransactionLabel;
-  /** Soft-delete marker. Present when the GET endpoint is called with
-   *  `?include-deleted=true` (the FE sync path); the client treats rows
-   *  with this flag as tombstones for IDB/dict eviction. */
+  /** Soft-delete marker. Always present on rows from `/api/transactions`
+   *  (the route hardcodes `includeDeleted: true`, matching the snapshots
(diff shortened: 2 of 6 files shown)


claoie:
## Sandbox E2E ✅

Sandbox: budget-pr-520 (port 20520, branch `fix/transactions-tombstone-delivery`, scrambled prod-clone DB with 6422 active + 558 soft-deleted transactions).

Driven via Playwright (`/tmp/e2e-pr520.ts`). Steps + results:

| Step | Expected | Actual |
|---|---|---|
| 1. Login → wait for `?include-deleted=true` syncs to settle | server returns active + tombstoned rows together | 908 active tx_ids + 203 tombstone tx_ids collected from network responses |
| 2. Read `BudgetApp` IDB `transactions` store after sync settles | active rows present, zero tombstones persisted | 6418 entries; **0** with `is_deleted=true`; **0** tombstones leaked from server response into IDB |
| 3. `UPDATE transactions SET is_deleted=TRUE` on a currently-cached tx, then page reload | server returns it as tombstone on next sync; FE evicts it from IDB | victim returned as tombstone (PASS); evicted from IDB (PASS); IDB count 6418 → 6417 |

The full chain verifies: server `excludeDeleted: !options.includeDeleted` flip → response includes soft-deleted rows with `is_deleted: true` → `useSync` adds them to `tombstones` set → after `Promise.all`, `result.transactions.delete(id)` + `indexedDb.remove(StoreName.transactions, id)` → store no longer contains the row.

DB restored to original state after the test (`UPDATE transactions SET is_deleted = FALSE …`).

reviewoie:
## Code review — PR 520

Walked the 6-file diff against the 8 review angles. Server-side filter inversion is correct (verified via `database.ts:309` default `excludeDeleted = true` + `database.ts:364` symmetric branch on `SOFT_DELETE_CONDITION`). Route param parsing is footgun-free (`optionalQueryString` returns `undefined` for absent/array/non-string; only literal `"true"` flips the flag, falls through to default for `"false"` / `"1"` / `"True"`). Default behavior is preserved everywhere else — the only callers of `searchTransactions`/`getTransactions` (this repo's versions) are the route itself; `compute-tools/sync-plaid.ts` + `sync-simple-fin.ts` use `searchTransactionsByAccountId` which has its own hardcoded `(is_deleted IS NULL OR is_deleted = FALSE)` clause (`transactions.ts:191-192`); `transfers.ts:46-52` uses raw SQL with the same hardcoded active-filter. `is_deleted` column is declared on both InvTxModel (line 76, schema line 52) and TransactionModel (line 86, schema line 59), confirmed before reviewing `this.is_deleted ?? false`. `optionalQueryString(... "include-deleted")` parses, route guards on literal `"true"`, repo passes `excludeDeleted: !options.includeDeleted` — verified in `transactions.ts:59,96`. No new `mock.module("pg", …)` test added (the diff has no test changes — the existing pattern in `transactions.test.ts` is pre-existing).

### Med
- `src/client/lib/hooks/sync.ts:231-236` — Comment overstates `saveAllData`'s additivity. Both warm (lines 678-682) and cold (lines 871-875) success paths do `indexedDb.clearAllData().then(() => indexedDb.saveAllData(refreshed))` — `clearAllData` truncates every store first, so `saveAllData` is destructive in the success path, not additive. The explicit `indexedDb.remove(StoreName.transactions, id)` is load-bearing only in the partial-failure case (`warmFetchFailed === true` / `coldFetchFailed === true`, which skips both clear+save). The eviction call is correct; the comment misnames *why* it's necessary. Suggest: replace "the downstream `saveAllData` is additive" with "the partial-failure branch skips `clearAllData/saveAllData`, so without this eager remove a tombstone observed during a failed-overall sync would stay in IDB."

### Low
- `src/client/lib/hooks/sync.ts:202-219` — Tombstone delivery is non-durable across freshness-window rollover (acknowledged as transitional in PR body, but worth pinning). `cachedCall` populates the browser Cache API on the first `cache.add(path)` against the live server. If a row was active at the time of that first cache miss for month M, the Cache API stores `is_deleted: false` for that month forever (no automatic eviction). When the row is later soft-deleted, the live query for the *current* month surfaces the tombstone — eviction fires — but on next warm sync, month M's `cachedCall` returns the stale active response, sets the row into `result.transactions`, and after `clearAllData()` runs, `saveAllData()` re-persists the row to IDB. Once the deletion event's `updated` timestamp rolls out of `[lastSyncedAt - 30d, now]`, no live query surfaces the tombstone and the row is permanently re-cached. PR B/C remove the Cache API layer entirely, which closes this gap — flagging only to confirm that's the intended sequencing.
- `src/server/routes/accounts/get-transactions.ts:41` vs `src/server/routes/accounts/get-snapshots.ts:47` — Pattern asymmetry between the two FE-tombstone-paying endpoints: snapshots route hardcodes `includeDeleted: true`, transactions route makes it opt-in via `?include-deleted=true`. In practice the opt-in adds no protection (route is the only callable surface for `searchTransactions`; engine + sync-plaid use `searchTransactionsByAccountId` which never threads `includeDeleted`). Not wrong, just an asymmetry future readers will trip on. Optional follow-up: hardcode `includeDeleted: true` here too OR backfill the snapshot route with the same opt-in pattern. Leaving as-is is also defensible — the route already enforces the only call site.

claoie:
## Reviewoie follow-up

|---|---|---|---|
| Tombstone-eviction comment misnames the reason (claims `saveAllData` is additive; both warm + cold success paths actually `clearAllData → saveAllData`, so the eager `indexedDb.remove` is load-bearing only in the partial-failure branch) | sync.ts:234 | med | Addressed in commit 22b27ff — comment rewritten to name the partial-failure branch as the real reason. |
| Tombstone delivery is non-durable across Cache API freshness rollover (`cachedCall` persists active responses for older months forever; once `updated` ages out of the freshness window the row can re-persist) | sync.ts:194 | low | Leaving as-is — explicit transitional limitation called out in the PR body. PR B (delta-by-updated-cursor) and PR C (remove Cache API) close this gap; addressing it inside PR A would expand scope to the whole 3-PR series. |
| Route asymmetry — `get-snapshots.ts:47` hardcodes `includeDeleted: true`, this route was opt-in via `?include-deleted=true` (no other HTTP caller exists, so the opt-in adds no protection) | get-transactions.ts:41 | low | Addressed in commit 22b27ff — route now hardcodes `includeDeleted: true` to match snapshots. The repo-layer `includeDeleted` default-false still protects direct callers (engine, sync-plaid). `useSync` drops the dead query param. |

reviewoie:
## Code review — PR 520 (round 2)

**Findings:** 2 total — 0 high, 0 med, 2 low. Round-1 fixes verified clean.

Round-1 verifications:
- Comment rewrite (sync.ts:228-234) now correctly names the partial-failure branch as the load-bearing case for the eager `indexedDb.remove`. Walked the warm path (line 657 `setData(refreshed)` still runs even when `warmFetchFailed`, so the dict gets the tombstone-delete; the `indexedDb.clearAllData/saveAllData` branch is the only thing that's skipped). Cold path's stages 1/2/3 all tile cleanly with no overlap (`UPDATED`-column date filter pins each tombstone to one window).
- Route hardcoding (`get-transactions.ts:36`) matches `get-snapshots.ts:47` symmetrically. `params.append("include-deleted", "true")` correctly removed from `useSync`. URL now identical between snapshots and transactions in this respect — no query-param toggle. Repo-layer `includeDeleted` default-false intact (`transactions.ts:40`).
- Cache API non-durable rollover (finding 2 round 1) acknowledged in PR body — deferred to PR B/C. Confirmed sequencing.
- 22b27ff is comment-+-route-+-URL only — no behavior change to the eager `indexedDb.remove` flow or the tombstone collection mechanics. `bun run typecheck` clean.

Fresh-eyes sweep against the 5 angles surfaced two doc-drift items below — both downstream of round 2's route hardcoding. Neither is correctness; flagging for accuracy.

### Low
- `src/common/models/Transaction.ts:55-58` and `:76-78` — Doc comments on `JSONTransaction.is_deleted` and `JSONInvestmentTransaction.is_deleted` say "Present when the GET endpoint is called with `?include-deleted=true`". After 22b27ff that query param no longer exists — the route hardcodes the contract, so `is_deleted` is now ALWAYS present on `/api/transactions` GET responses (the model's `toJSON()` emits `this.is_deleted ?? false` unconditionally). The "called with X" phrasing is stale and will mislead the next reader looking for the toggle. Suggest: "Soft-delete marker emitted by `/api/transactions` (the FE sync path). The route hardcodes inclusion; the client treats rows with `is_deleted: true` as tombstones for IDB/dict eviction."

- PR body — same drift. The "Summary" block says `useSync adds ?include-deleted=true, tracks tombstoneIds / tombstoneInvIds during the forEach`. After 22b27ff: the FE no longer appends `?include-deleted=true`, and the actual collection names are `tombstones.transaction` / `tombstones.investmentTransaction` (not `tombstoneIds` / `tombstoneInvIds`). Per the "design pivot mid-PR → update outstanding references" pattern, the PR body should be brought in line with the round-2 shape before merge. Not a code defect, but the body is what future readers reach for.

claoie:
## Reviewoie round-2 follow-up

|---|---|---|---|
| `is_deleted` docstrings on `JSONTransaction` + `JSONInvestmentTransaction` still referenced the dropped `?include-deleted=true` opt-in | src/common/models/Transaction.ts:55, :76 | low | Addressed in commit 340f0e9 — docstrings now state the field is always present on `/api/transactions` rows (route hardcodes `includeDeleted: true`, matching snapshots). |
| PR body summary referenced the removed `?include-deleted=true` URL param + outdated collection names (`tombstoneIds` / `tombstoneInvIds`) | PR description | low | Addressed — PR body rewritten to describe the hardcoded route shape and the actual collection names (`tombstones.transaction` / `tombstones.investmentTransaction`). |

src/client/lib/hooks/sync.ts
```diff
@@ -204,6 +225,22 @@ const fetchTransactions = async (
...
+  // is this explicit remove. (The happy-path branches still
+  // clear-then-save from `result`, which would also drop it.)
+  tombstones.transaction.forEach((id) => {
+    result.transactions.delete(id);
+    indexedDb.remove(StoreName.transactions, id).catch(console.error);
+  });
+  tombstones.investmentTransaction.forEach((id) => {
+    result.investmentTransactions.delete(id);
+    indexedDb.remove(StoreName.investmentTransactions, id).catch(console.error);
+  });
```
hoiekim, line 242: IndexedDB data is already loaded into context state `data` object at this point. How will you delete the tombstone items from it?

hoiekim (changes requested):

src/client/lib/hooks/sync.ts
claoie: Via the orchestrator's `setData(refreshed)` downstream — not from this block itself.

The eviction does two distinct things:

- **Line 238: `result.transactions.delete(id)`** removes the tombstoned id from the local dict that's about to be returned. `fetchTransactions` returns `result`, and the warm-path orchestrator (sync.ts:651) assigns `refreshed.transactions = transactionsResult.transactions`. So when `setData(refreshed)` fires at line 657, React replaces `data.transactions` with the post-eviction dict — context state no longer holds the tombstone. Cold path uses `setData(stage2)` (line 793) then `setData(finalData)` (line 854), both built from the post-eviction `recentTransactions` / `stage3Transactions.transactions`.

- **Line 239: `indexedDb.remove(...)`** removes from IDB directly. Load-bearing only on the partial-failure branch (line 677, where `warmFetchFailed` skips `clearAllData → saveAllData`); on the happy path it's redundant with the subsequent `clearAllData → saveAllData` that overwrites IDB from `refreshed`. The comment at line 232 explains this.

Two reasonable readings of your question:

(a) "How is this block wired to context state?" — it isn't, directly. `setData(refreshed)` at sync.ts:657 is the propagation.

(b) "Should this block ALSO update context state explicitly, like the UI-driven delete pattern in `SplitTransactionRow.tsx:170-180`?" — the orchestrator's `setData(refreshed)` already supersedes whatever cached state was painted at line 580, so an explicit `setData((old) => …)` here would be redundant. It would only matter if there were a window where the user could interact with the stale row between line 580 and line 657 — that's the warm-load-then-refresh design's pre-existing behavior, not something this PR changes.

If you want (b), I can add an explicit `setData((old) => { const next = new Data(old); next.transactions.delete(id); ... })` inside the tombstone forEach to close the fetch-time visibility window. Let me know which way you want it.

hoiekim: https://github.com/claoie/budget/blob/340f0e938eed296cfaa1491e324e15ae71ffc859/src/client/lib/hooks/sync.ts#L638-L657

This works in current state because we're replacing the whole transactions data. However in the next PR when we remove Cache API, we'll have to remove them explicitly from the previous data state. Approving for now.

hoiekim (approved):

PR #524, opened 2026-06-17
fix(holdings): key security price index by snapshot date, not close_price_as_of

### LINKED ISSUE #511: bug(holdings): security price history collapses into one month — buildSecurityPriceIndex buckets by close_price_as_of (a per-security constant), mispricing holdings at recent view dates

claoie:
## Summary

`buildSecurityPriceIndex` bucketed each security snapshot by the security's `close_price_as_of` instead of the snapshot's own date. `close_price_as_of` is a single per-security constant — the *latest* price's as-of date — that the server stamps onto every historical snapshot (the `snapshots` table has no such column; it's joined live from `securities`). Keying the monthly price index by it collapsed a security's entire price history into **one** month bucket, and whichever snapshot was iterated in first won that bucket. So holdings were valued at the wrong-date price at every view date from the `close_price_as_of` month onward.

Closes #511

## Root cause

`src/client/lib/hooks/calculation/holdings.ts` — `buildSecurityPriceIndex`:

```ts
const dateStr = close_price_as_of || snapshot.snapshot.date; // same value for every snapshot
const yearMonth = getYearMonthString(new LocalDate(dateStr)); // → one month for all
```

All N monthly snapshots map to the single `yearMonth` named by `close_price_as_of`, and because every `sourceDate` is identical the "keep the most recent per month" tie-break can never fire. `findLatestEntryLessThanOrEqual` then returns that one wrong-priced bucket for any view month ≥ the `close_price_as_of` month.

## Fix

Bucket by `snapshot.snapshot.date` — the date that snapshot's `close_price` was actually captured. One-line key change; `close_price_as_of` is dropped from the destructure. `sourceDate` (used downstream by `getPriceForHolding` to compare security-snapshot freshness vs the holding's institution price) becomes the per-snapshot date, which is more correct than the previous per-security constant.

## Tests

- New unit test in `holdings.test.ts`: an N-month series sharing one `close_price_as_of` now produces N distinct monthly buckets, each holding its own date's price (previously collapsed to one). Pins the regression.
- Full `src/client/lib/hooks/calculation` + `Calculations.holdings` suites green (115 tests). The existing `buildSecurityPriceIndex` tests set `close_price_as_of === snapshot.date`, so they never exercised the divergence and stay green unchanged.
- `tsc --noEmit` clean.

## E2E (Playwright, real Holdings Composition UI, before/after on one running dev server)

Drove the Accounts → account-detail Holdings Composition for an investment account holding a security whose `close_price_as_of` (March) predates its latest snapshot (June), stepping the view date back month-by-month. Same server, HMR-swapped the one-line key to capture buggy vs fixed:

At a view month **≥ the security's `close_price_as_of` month** (March):
- **Buggy:** the security was priced at its *June* monthly close (the latest snapshot won the collapsed bucket) → its row rendered at **113.5%** of the account, forcing a spurious **"Unknown" reconciliation row at −14.7%** to balance back to the account total.
- **Fixed:** priced at the correct *March* monthly close → **98.8%**, and the spurious Unknown row is **gone**. Implied per-share price equals the public March close (fixed) vs the public June close (buggy), at an identical share count — i.e. the difference is pure mispricing.

At **older** view months (Feb / Jan): both builds render identically — the collapsed bucket sits *after* those months, so the code falls back to `institution_price` and the bug is masked. This matches the issue's prediction that the defect bites only at view months ≥ `close_price_as_of`.

Per-account dollar figures redacted (real local data); percentages are ratios and public ETF closes are referenced qualitatively.

### DIFF
--- a/src/client/lib/hooks/calculation/holdings.test.ts
+++ b/src/client/lib/hooks/calculation/holdings.test.ts
@@ -142,6 +142,35 @@ describe("buildSecurityPriceIndex", () => {
 
     expect(index.size).toBe(0);
   });
+
+  test("buckets by each snapshot's own date even when all share one close_price_as_of", () => {
+    // The server stamps the security's single live `close_price_as_of` onto
+    // every historical snapshot. The price series must still spread across
+    // distinct month buckets keyed by each snapshot's own recording date —
+    // not collapse into the one month named by close_price_as_of.
+    const sharedAsOf = "2026-03-30";
+    const snapshots = new SecuritySnapshotDictionary();
+    const months: [string, number][] = [
+      ["2026-01-15", 100],
+      ["2026-02-15", 110],
+      ["2026-03-15", 120],
+    ];
+    months.forEach(([date, price], i) => {
+      const snap = createSecuritySnapshot("sec1", price, date);
+      snap.security.close_price_as_of = sharedAsOf;
+      snapshots.set(`snap${i}`, snap);
+    });
+
+    const index = buildSecurityPriceIndex(snapshots);
+
+    // Three distinct monthly buckets, each holding its own date's price —
+    // not a single "2026-03" bucket with one arbitrary winner.
+    const sec1 = index.get("sec1");
+    expect(sec1?.size).toBe(3);
+    expect(sec1?.get("2026-01")).toEqual({ price: 100, sourceDate: "2026-01-15" });
+    expect(sec1?.get("2026-02")).toEqual({ price: 110, sourceDate: "2026-02-15" });
+    expect(sec1?.get("2026-03")).toEqual({ price: 120, sourceDate: "2026-03-15" });
+  });
 });
 
 describe("getPriceForHolding", () => {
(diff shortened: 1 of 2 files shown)


claoie:
Closing as a duplicate of #512, which is the established PR for #511 (open since 2026-06-11, already reviewed by reviewoie). Both make the identical one-line fix — re-key `buildSecurityPriceIndex` from the security's `close_price_as_of` to each snapshot's own `snapshot.snapshot.date` — and add an equivalent regression test. #512's regression test is slightly broader (covers a gap month, Mar→Jun, exercising the `findLatestEntryLessThanOrEqual` fallback) and its `PriceIndexEntry` doc comment carries the fuller rationale. The visible-symptom E2E walk recorded here (row % corrected, spurious Unknown reconciliation row removed at view months ≥ the as-of month) is preserved in this thread for reference; #512 already proves the same defect deterministically at the calculation boundary. Continuing on #512.

PR #525, opened 2026-06-18
fix(holdings): bucket security price index by snapshot date, not close_price_as_of

### LINKED ISSUE #511: bug(holdings): security price history collapses into one month — buildSecurityPriceIndex buckets by close_price_as_of (a per-security constant), mispricing holdings at recent view dates

claoie:
## Problem

`buildSecurityPriceIndex` (`src/client/lib/hooks/calculation/holdings.ts`) bucketed each per-month price entry by the security's `close_price_as_of`. That field is a **single per-security constant** — the as-of date of the security's *latest* price — which the server attaches to every historical snapshot when it joins `snapshots` (which has no `close_price_as_of`) against `securities` (which holds exactly one).

So every snapshot of a given security carried an identical `close_price_as_of`, which meant:

1. The whole price history collapsed into **one** `yearMonth` bucket (the `close_price_as_of` month).
2. The per-month tie-break (`dateStr > existing.sourceDate`) could never fire — every `sourceDate` compared equal — so an arbitrary snapshot's price won the bucket.

`HoldingsComposition` then valued holdings at a wrong-date close price for any view month at or after that as-of, inflating per-security Value/Growth/% and producing a spurious "Unknown" reconciliation row to absorb the gap. Account-level totals stayed correct (pinned to balance), which masked it.

## Fix

Bucket by the snapshot's own date (`snapshot.snapshot.date`) — that's when the month's `close_price` was actually recorded. `sourceDate` now carries that date, which is also the correct basis for the security-vs-`institution_price` freshness comparison in `getPriceForHolding` (the comment there already wanted "when the security price was recorded", not a forward as-of).

Scope kept to the client calculation layer; the server enrichment is unchanged (the optional server-side null-out from the issue would broaden scope without being needed to fix the symptom).

## Test plan

- **New regression test** `buildSecurityPriceIndex > buckets by snapshot date, not the security's shared close_price_as_of`: builds a 3-month series for one security all sharing one `close_price_as_of` (the prod-join shape) and asserts 3 distinct monthly buckets, each priced/dated by its own snapshot. **Verified it fails on the pre-fix line** (collapses to 1 bucket) and passes after.
- Full `holdings.test.ts` suite green (37 pass). Existing tests unaffected — their helper set `close_price_as_of == snapshot date`, so they never exercised the prod skew.
- `bun run typecheck` clean; `eslint` clean on both changed files.

Closes #511

### DIFF
--- a/src/client/lib/hooks/calculation/holdings.test.ts
+++ b/src/client/lib/hooks/calculation/holdings.test.ts
@@ -142,6 +142,35 @@ describe("buildSecurityPriceIndex", () => {
 
     expect(index.size).toBe(0);
   });
+
+  test("buckets by snapshot date, not the security's shared close_price_as_of", () => {
+    // In production the server attaches the security's single live
+    // `close_price_as_of` to every historical snapshot in the join, so all
+    // snapshots of one security share an identical `close_price_as_of` while
+    // their `close_price` and snapshot dates differ per month. Bucketing off
+    // `close_price_as_of` would collapse the whole series into one month.
+    const sharedAsOf = "2026-03-30";
+    const snapshots = new SecuritySnapshotDictionary();
+    for (const [date, price] of [
+      ["2026-01-15", 100],
+      ["2026-02-15", 110],
+      ["2026-03-15", 120],
+    ] as const) {
+      const snap = createSecuritySnapshot("sec1", price, date);
+      snap.security.close_price_as_of = sharedAsOf;
+      snapshots.set(`snap_${date}`, snap);
+    }
+
+    const index = buildSecurityPriceIndex(snapshots);
+
+    // Three distinct monthly buckets, each priced and dated by its own snapshot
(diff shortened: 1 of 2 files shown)


claoie:
Closing as a duplicate of #512, which makes the identical fix (re-key `buildSecurityPriceIndex` from the security's `close_price_as_of` to each snapshot's own `snapshot.snapshot.date`) for #511 and is already reviewed + approved with a richer test plan (real-`/api/snapshots` index proof + `getPriceForHolding` resolved-price proof + UI E2E). My oldest-first scan missed #512 because its body closes the issue with "Fixes #511" rather than "Closes #511" — operator error on my filter, not a gap in #512. #512 is the canonical PR; no action needed here.

PR #527, opened 2026-06-18
fix(budgets): yearly capacity sums each month's rate, not year-end rate × 12 (#526)

### LINKED ISSUE #526: bug(budgets): Yearly capacity annualizes the period-end monthly rate instead of summing each month — disagrees with the true-summed 'spent' when capacity changed mid-year
## Summary

In **Yearly** budget view, the `LabeledBar` "spent / of / left" bar mixes two **inconsistent aggregation methods** for the two values it subtracts:

- **spent** is the *true sum of every month's actuals* across the year (`aggregateYear`).
- **capacity** is the *period-end monthly rate annualized* (`month × 12`) — it does **not** sum each month's active capacity.

So for any budget whose monthly capacity **changed mid-year**, the yearly "of $X" and the resulting "left" are wrong relative to the actual per-month budget the user configured — and inconsistent with the spent figure they're compared against.

## Reproduction

A roll-over budget whose monthly capacity was **lowered mid-year** — `$A/mo` through Jan 2026, then `$B/mo` (A > B) effective 2026-02-01 (two capacities: a default `month=A`, and a `month=B` with `active_from=2026-02-01`).

Repro step 1. Open Budgets, switch the interval selector to **Yearly (2026)**.
Repro step 2. Read that budget's bar.

**Actual:** `of $(B×12)` (= `B × 12`), `left` computed against `$(B×12)`.
**Expected:** `of $(A + 11×B)` — the sum of the 12 monthly allocations: Jan `$A` + Feb–Dec `11 × $B` = `$(A + 11×B)`. The bar understates the year's budget (and overstates how much is "used up") by the January rate-delta (`$(A − B)`).

Meanwhile the **spent** half of the same bar *is* the true full-year sum (verified against the DB: the rendered "spent" equals the exact sum of that budget's Jan + Feb + Mar 2026 transactions — a real partial-year total, not annualized). So the two halves of one bar disagree about what the year's budget was.

## Root cause

`src/client/components/LabeledBar/index.tsx:36-45`

```ts
const { sorted_amount, unsorted_amount, rolled_over_amount } =
  interval === "year"
    ? budgetData.get(barData.id).aggregateYear(date.getFullYear())   // spent: SUM over all 12 months
    : budgetData.get(barData.id, date);

const capacityValue = barData.getActiveAmount(date, interval);        // capacity: period-end rate × 12
```
`leftover = capacityValue - total` at `:67`.

- `aggregateYear` (`src/client/lib/models/Calculations.ts:254-276`) sums `sorted_amount + unsorted_amount` across **every** month whose key matches the year → true full-year actuals.
- `BudgetFamily.getActiveAmount(date, "year")` (`src/client/lib/models/BudgetFamily.ts:109-112`) → `getActiveCapacity(date)` (`:89-98`) resolves the **single** capacity active at the *period-end* date, then `Capacity.getActiveAmount` (`src/client/lib/models/Capacity.ts:117-122`, non-synced branch) returns `this["year"]`, and the `year` getter (`Capacity.ts:45-51`) is just `this.month * 12`.

There is **no** per-month summation on the capacity path — it annualizes whatever single rate is active at the end of the period. The spent path was given a full-year aggregator (`aggregateYear`, evidently added to fix the inverse problem in #161 "Yearly view shows only one month's spending instead of full year total"); the capacity path was never given a matching one, so the two sides of `leftover = capacity - spent` use different aggregation.

## Why this is a system bug, not persisted user state

The wrong `$(B×12)` is a **render-time computation** over the budget's capacities; the user cannot re-curate any editable field to make the bar show the correct `$(A + 11×B)` (short of deleting the historical capacity, which would lose real budget history). The mid-year capacity change itself is a legitimate user choice — the defect is the system's yearly **rollup** of those choices.

## Honest caveat (flagged for your call on intended semantics)

One could argue the yearly capacity is *intentionally* "current monthly rate annualized." But two things point to this being a real inconsistency rather than a deliberate design:
1. The value it's subtracted from (`spent`) **is** a true per-month sum, and `left = capacity − spent` only makes sense if both axes aggregate the same way.
2. #161 established that the yearly view should reflect the **full year's** activity, not a single-month snapshot; the spent side was migrated to that semantics and the capacity side was left annualized.

If you intend "annualized current rate," then the inverse holds — `spent` should also be annualized — but that would re-regress #161. Either way the two halves should match.

## Fix sketch

Give capacity a year-aggregation parallel to `aggregateYear`: sum `getActiveAmount(monthEnd, "month")` across the 12 months of the year (walking a `ViewDate`), instead of `getActiveAmount(yearEnd, "year")`. Mirror the `is_synced` / infinite-capacity handling already in `Capacity.getActiveAmount`. The income-sign and `MAX_FLOAT` sentinels need the same poisoning rules so infinite budgets (Non-budgeted / Transfers) stay infinite.

## Scope / severity

Minor / UX. Only visible in **Yearly** view and only for budgets whose monthly capacity changed *within the displayed year*. In the current data exactly one budget triggers it (one mid-year rate change in 2026); single-capacity budgets and budgets whose last change predates the year are unaffected (`rate × 12` happens to equal the true sum). No console errors; no crash. `BudgetDetailPage` graph uses the same annualize-at-period-end path, so it agrees with the list bar (no cross-view split) — the inconsistency is *within* the bar (spent-vs-capacity), not across pages.

*[Edited 2026-06-18 by claoie security self-audit: redacted real budget-capacity dollar figures to symbolic placeholders (A/B) per §2a PII policy — the math relationship is preserved; concrete monthly amounts were live budget data.]*

claoie:
## Summary

Closes #526.

In **Yearly** budget view the `LabeledBar` rendered `left = capacity − spent` where the two halves used **inconsistent aggregation**:

- **spent** = the true per-month sum across the year (`Calculations.aggregateYear`).
- **capacity** = the single year-end monthly rate annualized (`month × 12`), via `Capacity.year`.

For any budget whose monthly capacity **changed mid-year**, the yearly "of …" and "left" were wrong relative to the per-month budget the user configured — and inconsistent with the spent figure they're compared against.

## Fix

`BudgetFamily.getActiveAmount` now dispatches `interval === "year"` to a new `getYearlyAmount`, which **sums each of the 12 months' active capacities** (`getActiveCapacity(monthEnd).getActiveAmount(monthEnd, "month", children)`), mirroring `aggregateYear` on the spending side.

- **Single-capacity budgets are numerically neutral** — `rate × 12` equals the 12-month sum, so unchanged budgets render exactly as before.
- **Infinite capacities still poison the year to ±`MAX_FLOAT`** (12 × `MAX_FLOAT` would overflow the sentinel) — matching `Capacity.year`'s guard.
- **Synced rows resolve correctly** — each month sums children at that month's active capacities.
- Landed at the shared `BudgetFamily` boundary, so every yearly consumer (`LabeledBar`, `BalanceChartRow`, `BudgetDonut`, `ChartAccountsPage`, `BudgetDetailPage` graph) stays consistent — no cross-view split.

Net +41/−1 source, 1 file, plus a new `BudgetFamily.test.ts`.

## Test plan

- [x] `bun run lint` — clean
- [x] `bun test` — 884 pass / 0 fail
- [x] **New regression test** (`BudgetFamily.test.ts`) — fails on baseline (year-end rate × 12), passes on fix (12-month sum); stash/pop verified. Covers: single-capacity neutrality, mid-year cut, mid-year raise, later-year change ignored, ±infinite poisoning, negative-income, month-interval unchanged.
- [ ] **UI E2E** — pending sandbox walkthrough (Budgets page, Yearly 2026, the one mid-year-rate-change budget; fix vs `upstream/main` baseline).

### DIFF
--- a/src/client/lib/models/BudgetFamily.test.ts
+++ b/src/client/lib/models/BudgetFamily.test.ts
@@ -1,0 +1,90 @@
+import { test, expect, describe } from "bun:test";
+import { MAX_FLOAT } from "common";
+// Prime the model graph before importing the concrete subclasses. Data.ts
+// fully evaluates BudgetFamily before reaching Budget/Section/Category (which
+// `extends BudgetFamily`); importing Budget first instead hits a circular-
+// import TDZ ("Cannot access 'BudgetFamily' before initialization").
+import "./Data";
+import { Budget } from "./Budget";
+import { Capacity } from "./Capacity";
+
+// ---------------------------------------------------------------------------
+// BudgetFamily.getActiveAmount / getYearlyAmount
+//
+// Regression coverage for #526: the yearly capacity must be the sum of each
+// of the 12 months' active capacities, NOT the single year-end rate × 12.
+// Mirrors Calculations.aggregateYear on the spending side so the LabeledBar's
+// `left = capacity − spent` compares like-aggregated quantities.
+// ---------------------------------------------------------------------------
+
+// A date inside 2026 — getYearlyAmount only reads getFullYear().
+const IN_2026 = new Date(2026, 5, 15);
+
+describe("BudgetFamily.getActiveAmount('year')", () => {
+  test("single constant capacity → month × 12 (numerically neutral with the old path)", () => {
+    const b = new Budget({ capacities: [new Capacity({ month: 1000 })] });
+    expect(b.getActiveAmount(IN_2026, "year")).toBe(12000);
+  });
+
+  test("mid-year rate cut sums each month, not year-end rate × 12 (#526 repro)", () => {
+    // $4000/mo default, lowered to $3000/mo effective 2026-02-01.
+    const b = new Budget({
+      capacities: [
+        new Capacity({ month: 4000 }),
+        new Capacity({ month: 3000, active_from: "2026-02-01" as unknown as Date }),
+      ],
+    });
+    // Jan@4000 + Feb–Dec@3000 = 4000 + 11 × 3000 = 37000 — NOT 3000 × 12 = 36000.
+    expect(b.getActiveAmount(IN_2026, "year")).toBe(37000);
+  });
+
+  test("mid-year rate raise also sums each month", () => {
+    // $2000/mo default, raised to $5000/mo effective 2026-07-01.
+    const b = new Budget({
+      capacities: [
+        new Capacity({ month: 2000 }),
+        new Capacity({ month: 5000, active_from: "2026-07-01" as unknown as Date }),
+      ],
+    });
+    // Jan–Jun@2000 + Jul–Dec@5000 = 6 × 2000 + 6 × 5000 = 42000 — NOT 5000 × 12 = 60000.
+    expect(b.getActiveAmount(IN_2026, "year")).toBe(42000);
+  });
+
+  test("a capacity change in a LATER year does not affect the displayed year", () => {
+    const b = new Budget({
+      capacities: [
+        new Capacity({ month: 1000 }),
+        new Capacity({ month: 9999, active_from: "2027-01-01" as unknown as Date }),
+      ],
+    });
+    expect(b.getActiveAmount(IN_2026, "year")).toBe(12000);
+  });
+
+  test("infinite (MAX_FLOAT) capacity poisons the year to +MAX_FLOAT, no overflow", () => {
+    const b = new Budget({ capacities: [new Capacity({ month: MAX_FLOAT })] });
+    expect(b.getActiveAmount(IN_2026, "year")).toBe(MAX_FLOAT);
+  });
+
+  test("infinite income (-MAX_FLOAT) capacity poisons the year to -MAX_FLOAT", () => {
+    const b = new Budget({ capacities: [new Capacity({ month: -MAX_FLOAT })] });
+    expect(b.getActiveAmount(IN_2026, "year")).toBe(-MAX_FLOAT);
+  });
+
+  test("income (negative finite) capacity stays negative across the year", () => {
+    const b = new Budget({ capacities: [new Capacity({ month: -1500 })] });
+    expect(b.getActiveAmount(IN_2026, "year")).toBe(-18000);
+  });
+
+  test("month interval is unchanged — resolves the period-end active capacity", () => {
+    const b = new Budget({
+      capacities: [
+        new Capacity({ month: 4000 }),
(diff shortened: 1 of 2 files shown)


reviewoie:
## Code review — PR 527

The yearly-capacity summation is correct. Verified end-to-end:

- `LabeledBar/index.tsx:39,45` reads the spending side from `aggregateYear(year)` (a true per-month sum over the year — confirmed at `Calculations.ts:254-276`) while the capacity side comes from `barData.getActiveAmount(date, "year")`. Pre-fix that returned `getActiveCapacity(date).year` = year-end-rate × 12 (`Capacity.ts:45-51`), so for a mid-year rate change the two halves of the bar disagreed. `getYearlyAmount` now walks all 12 months and sums each month's active capacity — the two sides aggregate the same way.
- MAX_FLOAT / NaN guards mirror `Capacity.getActiveAmount` (`Capacity.ts:117-136`) exactly: per-month NaN is skipped, any `±MAX_FLOAT` month short-circuits to the signed sentinel (no 12 × MAX_FLOAT overflow).
- Per-month resolution is correct: `new ViewDate("month", new Date(year, m, 1)).getEndDate()` yields the last instant of month m (`date.ts` `current()`), and `getActiveCapacity(monthEnd)` resolves the capacity active in that month. Synced rows resolve children at each month's view date.
- All other `getActiveAmount(..., "year")` call sites pass `interval` from `viewDate.getInterval()` (BudgetDonut, ChartAccountsPage, BalanceChartRow, BudgetDetailPage hooks) — every one of them inherits the same parity fix in yearly view; none break.

**Scope:** Clean — only `BudgetFamily.ts` and `BudgetFamily.test.ts`. No unrelated files.

**Tests:** 8 new tests in `BudgetFamily.test.ts`, plain `bun:test` (correct for the current preload harness — there is no `.test.bundle.ts` framework anymore). Covers the constant case (neutral with old path), mid-year cut/raise, later-year change isolation, ±MAX_FLOAT poison, negative-income, and month-interval unchanged. Ran `bun test BudgetFamily.test.ts Capacity.test.ts` locally: 37 pass / 0 fail. CI build + test both green.

### Low
- Branch is 1 commit behind `upstream/main` (`bcff331`, transfers UI — disjoint files, no conflict). Not blocking; rebase before merge if convenient.

No further blockers.

hoiekim (approved):

PR #490, opened 2026-06-06
feat(transfers): suggested-row Confirm/Reject UI on transaction rows (#354)

claoie:
## What

Phase 3a of the transfer-detection UI (#236 → #354). The hourly detection job (#335) has been writing `transaction_pairs` rows with `status='suggested'`, but nothing in the client surfaced them. This wires up the suggested-row affordance:

- A transaction row whose **parent** transaction belongs to a still-**suggested** pair renders an amber **"Transfer?"** chip plus **Confirm** / **Reject** buttons in place of the budget/category dropdowns.
- **Confirm** → `POST /api/transfers/pair { pair_id }` → status becomes `confirmed`.
- **Reject** → `DELETE /api/transfers?id=<pair_id>` → soft-deletes the pair; the row reverts to the normal budget/category controls.

## How

- **`useTransfers` hook** fetches `GET /api/transfers` and exposes a `transaction_id → pair_id` map for pairs still in the `suggested` state, plus `confirm`/`reject` mutations that refetch on success. Kept separate from the heavyweight cold/warm IndexedDB sync — pairs are a small list cheaply refetched after each mutation rather than cached per-month like transactions.
- **`get-transfers` route** gains a `TransfersGetResponse` type so the client fetch is typed (matches the `*GetResponse` convention).
- **`TransferControls`** is a small reusable component (disables its buttons while a request is in flight).

## Scope decisions

- **Split rows** never carry the affordance — the engine pairs whole transactions and a split inherits its parent's `transaction_id`.
- **`InvestmentTransactionRow` intentionally left out.** `runTransferDetection` only pairs the `transactions` table, so an investment row can never hold a *suggested* pair until manual linking (Phase 3c, #356) lands. Adding the UI there now would be unreachable code. The affordance lives in `TransferControls` so adoption is a one-liner when 3c arrives.

## Out of scope (separate sub-issues)
- #355 — confirmed-row badge `✓ Transfer ↔` + paired bracket + budget-total exclusion math.
- #356 — context-menu link/unlink + Transfers sidebar view.

## Verification

`bun run typecheck` clean · `bun run lint` clean · `bun run build-client` + `build-server` clean · full suite **712 pass / 0 fail**.

**E2E (Playwright, prod-like server on :3055 against local Postgres, admin login, real data with 8 suggested-pair rows in the 2026 view):**

| Step | Result |
|------|--------|
[...]

PR #528, opened 2026-06-19
feat(transfers): exclude confirmed pairs from budget + flow calculations

claoie:
Stacks on #490. Rebases to main once #490 merges.

Per Hoie 2026-06-19: confirmed transfers should not count as spending or income in the budget rollups and should not appear as inflow/outflow on the Flow chart Sankey — both halves are internal movement between the user's own accounts. Per-account historical balance INTENTIONALLY still includes them so the chart matches the snapshot baseline of money actually held.

## What changes

| Surface | Confirmed transfers | Reason |
|---|---|---|
| Budget rollups (`getBudgetData`) | **excluded** | Internal movement, not real spending or income — otherwise would inflate both the source-account budget's spent column AND the destination's income column |
| Flow chart Sankey (`getSankeyData`) | **excluded** | Same reason; would double-count by appearing as both an expense (debit side) and income (credit side) |
| Per-account balance history (`getBalanceData`) | **kept** | The account's balance really did change when money left/arrived — dropping transfers would de-sync the chart from the snapshot baseline (Hoie 2026-06-18) |
| Capacity / holdings | n/a | Don't consume transactions |

## How it's wired

- Both `getBudgetData` and `getSankeyData` gain an optional `confirmedTransferTxIds: ReadonlySet<string>` param. Defaults to an empty set so existing callers (and tests) keep pre-PR behavior.
- `CalculateOptions` interface added to both `lib/hooks/context.ts` (the context's contract) and `App/lib/hooks.ts` (the implementation). `calculate(data, opts?)` now takes an optional opts.
- `Utility` derives the set once per `transfers.confirmedTransferByTransactionId` ref change and threads it into the `calculate(data, { confirmedTransferTxIds })` effect.
- `FlowChartRow` derives the same set locally and passes it to `getSankeyData` — the chart computes its own Sankey inside a useMemo rather than reading from `calculations`, so it needs to consume the set directly.
- `getBalanceData` is deliberately NOT changed.

## Test plan

- [x] `bun run lint` clean
- [x] `bun test src/server` — 624 pass, 0 fail
- [x] Sandbox E2E: exclusion verified in a real logged-in browser by running the served `getBudgetData` and `getSankeyData` modules against a synthesized confirmed pair — both drop the pair (amount → 0) once confirmed. See the E2E comment for the harness + results. (The natural click-to-confirm UI path is unreachable read-only: the sandbox has only *suggested* pairs and confirming is a write to the prod-clone, so the calc exclusion was exercised directly instead. `getBalanceData` intentionally unchanged.)

### DIFF
--- a/src/client/components/App/Utility.tsx
+++ b/src/client/components/App/Utility.tsx
@@ -1,5 +1,5 @@
 import { useEffect } from "react";
-import { useAppContext, useSync, PATH, useDebounce } from "client";
+import { useAppContext, useSync, useTransfers, PATH, useDebounce } from "client";
 
 /**
  * This component is used to run useEffect hooks dependant on context variables.
@@ -7,13 +7,13 @@ import { useAppContext, useSync, PATH, useDebounce } from "client";
  * dev engineers to find them easily.
  */
 const Utility = () => {
-  const { user, router, setSelectedInterval, data, calculate, viewDate, transfers } =
-    useAppContext();
+  const { user, router, setSelectedInterval, data, calculate, viewDate } = useAppContext();
 
   const userLoggedIn = !!user;
   const { path, go } = router;
 
   const { sync, clean } = useSync();
+  const { refresh: refreshTransfers } = useTransfers();
   const debouncer = useDebounce();
 
   /**
@@ -32,21 +32,24 @@ const Utility = () => {
     else clean();
   }, [userLoggedIn, sync, clean]);
 
-  // `transfers.confirmedTransferByTransactionId` is already a
-  // Map<transaction_id, ConfirmedTransfer>; its `.has(id)` answers the
-  // exact question the calcs ask. Feeding the Map straight through
-  // avoids a redundant Set materialization (Hoie review 2026-06-19).
-  // Per-account balance calc intentionally does not consume this —
-  // see `useData`'s `calculateAll` comment.
-  const confirmedTransferIds = transfers.confirmedTransferByTransactionId;
+  /**
+   * Refresh transfer pairs on login so `data.suggestedPairByTransactionId`
+   * / `data.confirmedTransferByTransactionId` are populated for the
+   * first calculator run + the transactions-table render. Each
+   * mutation method on `useTransfers()` re-runs refresh internally;
+   * this effect handles the initial cold load only.
+   */
+  useEffect(() => {
+    if (userLoggedIn) refreshTransfers();
+  }, [userLoggedIn, refreshTransfers]);
 
   /**
    * Calculate balance history when data is updated
    */
   useEffect(() => {
     if (!data.status.isInit) return;
-    debouncer(() => calculate(data, { confirmedTransferIds }));
-  }, [data, calculate, debouncer, confirmedTransferIds]);
+    debouncer(() => calculate(data));
+  }, [data, calculate, debouncer]);
 
   /**
    * Update viewDate when user selects different interval
--- a/src/client/components/App/Utility.tsx
+++ b/src/client/components/App/Utility.tsx
@@ -1,4 +1,4 @@
-import { useEffect, useMemo } from "react";
+import { useEffect } from "react";
 import { useAppContext, useSync, PATH, useDebounce } from "client";
 
 /**
@@ -32,25 +32,21 @@ const Utility = () => {
     else clean();
   }, [userLoggedIn, sync, clean]);
 
-  // Confirmed-transfer transaction id set, derived once per
-  // `transfers.confirmedTransferByTransactionId` ref change. Threaded
-  // into `calculate(...)` so `getBudgetData` (and any future calc
-  // wanting to skip transfers) sees the latest pair set. Per-account
-  // balance calc intentionally does not consume this — see
-  // `useData`'s `calculateAll` comment.
-  const confirmedTransferTxIds = useMemo(() => {
-    const set = new Set<string>();
-    transfers.confirmedTransferByTransactionId.forEach((_pair, txId) => set.add(txId));
-    return set;
-  }, [transfers.confirmedTransferByTransactionId]);
+  // `transfers.confirmedTransferByTransactionId` is already a
+  // Map<transaction_id, ConfirmedTransfer>; its `.has(id)` answers the
+  // exact question the calcs ask. Feeding the Map straight through
+  // avoids a redundant Set materialization (Hoie review 2026-06-19).
+  // Per-account balance calc intentionally does not consume this —
+  // see `useData`'s `calculateAll` comment.
+  const confirmedTransferIds = transfers.confirmedTransferByTransactionId;
 
   /**
    * Calculate balance history when data is updated
    */
   useEffect(() => {
     if (!data.status.isInit) return;
-    debouncer(() => calculate(data, { confirmedTransferTxIds }));
-  }, [data, calculate, debouncer, confirmedTransferTxIds]);
+    debouncer(() => calculate(data, { confirmedTransferIds }));
+  }, [data, calculate, debouncer, confirmedTransferIds]);
 
   /**
    * Update viewDate when user selects different interval
--- a/src/client/components/App/Utility.tsx
+++ b/src/client/components/App/Utility.tsx
@@ -1,4 +1,4 @@
-import { useEffect } from "react";
+import { useEffect, useMemo } from "react";
 import { useAppContext, useSync, PATH, useDebounce } from "client";
 
 /**
@@ -7,7 +7,8 @@ import { useAppContext, useSync, PATH, useDebounce } from "client";
  * dev engineers to find them easily.
  */
 const Utility = () => {
-  const { user, router, setSelectedInterval, data, calculate, viewDate } = useAppContext();
+  const { user, router, setSelectedInterval, data, calculate, viewDate, transfers } =
+    useAppContext();
 
   const userLoggedIn = !!user;
   const { path, go } = router;
@@ -31,13 +32,25 @@ const Utility = () => {
     else clean();
   }, [userLoggedIn, sync, clean]);
 
+  // Confirmed-transfer transaction id set, derived once per
+  // `transfers.confirmedTransferByTransactionId` ref change. Threaded
+  // into `calculate(...)` so `getBudgetData` (and any future calc
+  // wanting to skip transfers) sees the latest pair set. Per-account
+  // balance calc intentionally does not consume this — see
+  // `useData`'s `calculateAll` comment.
+  const confirmedTransferTxIds = useMemo(() => {
+    const set = new Set<string>();
+    transfers.confirmedTransferByTransactionId.forEach((_pair, txId) => set.add(txId));
+    return set;
+  }, [transfers.confirmedTransferByTransactionId]);
+
   /**
    * Calculate balance history when data is updated
    */
   useEffect(() => {
     if (!data.status.isInit) return;
-    debouncer(() => calculate(data));
-  }, [data, calculate, debouncer]);
+    debouncer(() => calculate(data, { confirmedTransferTxIds }));
+  }, [data, calculate, debouncer, confirmedTransferTxIds]);
 
   /**
    * Update viewDate when user selects different interval
--- a/src/client/components/App/Utility.tsx
+++ b/src/client/components/App/Utility.tsx
@@ -1,5 +1,5 @@
 import { useEffect } from "react";
-import { useAppContext, useSync, useTransfers, PATH, useDebounce } from "client";
+import { useAppContext, useSync, PATH, useDebounce } from "client";
 
 /**
  * This component is used to run useEffect hooks dependant on context variables.
@@ -13,7 +13,6 @@ const Utility = () => {
   const { path, go } = router;
 
   const { sync, clean } = useSync();
-  const { refresh: refreshTransfers } = useTransfers();
   const debouncer = useDebounce();
 
   /**
@@ -31,17 +30,6 @@ const Utility = () => {
     if (userLoggedIn) sync();
     else clean();
   }, [userLoggedIn, sync, clean]);
-
-  /**
-   * Refresh transfer pairs on login so `data.suggestedPairByTransactionId`
-   * / `data.confirmedTransferByTransactionId` are populated for the
-   * first calculator run + the transactions-table render. Each
-   * mutation method on `useTransfers()` re-runs refresh internally;
-   * this effect handles the initial cold load only.
-   */
-  useEffect(() => {
-    if (userLoggedIn) refreshTransfers();
-  }, [userLoggedIn, refreshTransfers]);
 
   /**
    * Calculate balance history when data is updated
--- a/src/client/components/App/lib/hooks.ts
+++ b/src/client/components/App/lib/hooks.ts
@@ -12,7 +12,17 @@ import {
 } from "client";
 import { Dispatch, SetStateAction, useCallback, useEffect, useMemo, useState } from "react";
 
-type CalculateFn = ((data: Data) => void) & {
+export interface CalculateOptions {
+  /** Confirmed-transfer transaction ids — threaded into `getBudgetData`
+   *  so the spent/income aggregation skips internal-movement
+   *  transactions. Not threaded into balance calcs (per Hoie
+   *  2026-06-18: per-account historical balance MUST still include
+   *  transfer effects, since the account's balance really did change).
+   *  Empty / omitted = pre-PR behavior. */
+  confirmedTransferTxIds?: ReadonlySet<string>;
+}
+
+type CalculateFn = ((data: Data, opts?: CalculateOptions) => void) & {
   cache: {
     capacityData: (updater: (current: CapacityData) => CapacityData) => void;
   };
@@ -34,7 +44,7 @@ export const useData = () => {
   const [calculations, setCalculations] = useState(new Calculations());
 
   const calculateAll = useCallback(
-    (data: Data) => {
+    (data: Data, opts?: CalculateOptions) => {
       const {
         accounts,
         accountSnapshots,
@@ -49,9 +59,16 @@ export const useData = () => {
         categories,
       } = data;
 
+      const confirmedTransferTxIds = opts?.confirmedTransferTxIds;
+
       setCalculations((oldCalculations) => {
         const newCalculations = new Calculations(oldCalculations);
 
+        // Balance data intentionally does NOT receive confirmedTransferTxIds:
+        // per-account historical balance is supposed to reflect the actual
+        // value held in each account at each point in time. A transfer
+        // moves real dollars between accounts, so dropping it would
+        // de-sync the chart from the snapshot baseline (Hoie 2026-06-18).
         const balanceData = getBalanceData(
           accounts,
           accountSnapshots,
@@ -67,6 +84,7 @@ export const useData = () => {
           budgets,
           sections,
           categories,
+          confirmedTransferTxIds,
         );
 
         const capacityData = getCapacityData(budgets, sections, categories);
--- a/src/client/components/App/lib/hooks.ts
+++ b/src/client/components/App/lib/hooks.ts
@@ -54,11 +54,11 @@ export const useData = () => {
         const newCalculations = new Calculations(oldCalculations);
 
         // Balance data intentionally does NOT receive `transfers`:
-        // per-account historical balance is supposed to reflect the
-        // actual value held in each account at each point in time. A
+        // per-account historical balance must reflect the actual
+        // value held in each account at each point in time. A
         // transfer moves real dollars between accounts, so dropping
         // its halves would de-sync the chart from the snapshot
-        // baseline (Hoie 2026-06-18).
+        // baseline.
         const balanceData = getBalanceData(
           accounts,
           accountSnapshots,
--- a/src/client/components/App/lib/hooks.ts
+++ b/src/client/components/App/lib/hooks.ts
@@ -12,19 +12,7 @@ import {
 } from "client";
 import { Dispatch, SetStateAction, useCallback, useEffect, useMemo, useState } from "react";
 
-export interface CalculateOptions {
-  /** Membership-test for confirmed-transfer transaction ids — threaded
-   *  into `getBudgetData` so spent/income aggregation skips
-   *  internal-movement transactions. Typed structurally on `has` so
-   *  the standard call site can pass
-   *  `transfers.confirmedTransferByTransactionId` (a Map) directly
-   *  without materializing a parallel Set. Not threaded into balance
-   *  calcs (per Hoie 2026-06-18: per-account historical balance MUST
-   *  still include transfer effects). Omitted = pre-PR behavior. */
-  confirmedTransferIds?: { has(transaction_id: string): boolean };
-}
-
-type CalculateFn = ((data: Data, opts?: CalculateOptions) => void) & {
+type CalculateFn = ((data: Data) => void) & {
   cache: {
     capacityData: (updater: (current: CapacityData) => CapacityData) => void;
   };
@@ -46,7 +34,7 @@ export const useData = () => {
   const [calculations, setCalculations] = useState(new Calculations());
 
   const calculateAll = useCallback(
-    (data: Data, opts?: CalculateOptions) => {
+    (data: Data) => {
       const {
         accounts,
         accountSnapshots,
@@ -59,18 +47,18 @@ export const useData = () => {
         budgets,
         sections,
         categories,
+        confirmedTransferByTransactionId,
       } = data;
-
-      const confirmedTransferIds = opts?.confirmedTransferIds;
 
       setCalculations((oldCalculations) => {
         const newCalculations = new Calculations(oldCalculations);
 
-        // Balance data intentionally does NOT receive confirmedTransferIds:
-        // per-account historical balance is supposed to reflect the actual
-        // value held in each account at each point in time. A transfer
-        // moves real dollars between accounts, so dropping it would
-        // de-sync the chart from the snapshot baseline (Hoie 2026-06-18).
+        // Balance data intentionally does NOT receive
+        // confirmedTransferByTransactionId: per-account historical
+        // balance is supposed to reflect the actual value held in
+        // each account at each point in time. A transfer moves real
+        // dollars between accounts, so dropping it would de-sync the
+        // chart from the snapshot baseline (Hoie 2026-06-18).
         const balanceData = getBalanceData(
           accounts,
           accountSnapshots,
@@ -86,7 +74,7 @@ export const useData = () => {
           budgets,
           sections,
           categories,
-          confirmedTransferIds,
+          confirmedTransferByTransactionId,
         );
 
         const capacityData = getCapacityData(budgets, sections, categories);
--- a/src/client/components/App/lib/hooks.ts
+++ b/src/client/components/App/lib/hooks.ts
@@ -47,18 +47,18 @@ export const useData = () => {
         budgets,
         sections,
         categories,
-        confirmedTransferByTransactionId,
+        transfers,
       } = data;
 
       setCalculations((oldCalculations) => {
         const newCalculations = new Calculations(oldCalculations);
 
-        // Balance data intentionally does NOT receive
-        // confirmedTransferByTransactionId: per-account historical
-        // balance is supposed to reflect the actual value held in
-        // each account at each point in time. A transfer moves real
-        // dollars between accounts, so dropping it would de-sync the
-        // chart from the snapshot baseline (Hoie 2026-06-18).
+        // Balance data intentionally does NOT receive `transfers`:
+        // per-account historical balance is supposed to reflect the
+        // actual value held in each account at each point in time. A
+        // transfer moves real dollars between accounts, so dropping
+        // its halves would de-sync the chart from the snapshot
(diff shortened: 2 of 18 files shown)


reviewoie:
## Code review — PR 528

### Med

|---|---|---|---|
| Splits of a confirmed-transfer parent still aggregate into budget rollups — the synthetic-split pass at the bottom of `getBudgetData` does not see the transfer-skip | `src/client/lib/hooks/calculation/budgets.ts:152-155` | med | User has a transaction with one or more `splitTransactions` (e.g. parent $100 with $30 "Coffee" split). User later marks that parent as a confirmed transfer. The PR correctly (a) skips parent in `processTransaction` and (b) skips the families registration in the splits→families pass. But the second pass `splitTransactions.forEach((st) => processTransaction(st.toTransaction()))` synthesizes a Transaction whose `transaction_id` is overridden to `id` (the split's `split_transaction_id`), per `SplitTransaction.toTransaction()` at `src/client/lib/models/SplitTransaction.ts:40-46` (`transaction_id: id`). The guard at line 51 (`if (confirmedTransferTxIds.has(transaction_id)) return;`) checks against this overridden id, which is the split's id — NOT the parent's transaction_id — so the guard never fires, and the $30 split lands on Coffee's `sorted_amount` (and the section/budget rollups + rollover) despite the parent being a confirmed transfer. Fix shape: gate on `st.transaction_id` (parent's id) before calling `processTransaction(st.toTransaction())`, e.g. `if (confirmedTransferTxIds.has(st.transaction_id)) return;` at the top of the forEach body. The three new unit tests don't cover this — splits-with-transfer-parent is not exercised. |

### High
No further blockers.

### Low
No further blockers.

claoie:
## Reviewoie follow-up

|---|---|---|---|
| Splits of a confirmed-transfer parent leak through — synthetic Transaction's `transaction_id` is the split's id (per `SplitTransaction.toTransaction()`), so the in-`processTransaction` guard never sees the parent's id | budgets.ts:152 | med | Addressed in commit 64db600 — added a split-pass guard on `st.transaction_id` (the parent's id, which IS what gets added to confirmedTransferTxIds) BEFORE calling `processTransaction(st.toTransaction())`. Inline comment explains why the inner guard is insufficient. Test added: parent $100 + two splits $30/$20, all sharing parent's transaction_id, parent in the set → budgetData.size = 0 (pre-fix would have been 2). 16 pass / 0 fail. |

src/client/components/App/Utility.tsx
```diff
@@ -31,13 +32,25 @@ const Utility = () => {
...
+
   /**
    * Calculate balance history when data is updated
    */
   useEffect(() => {
     if (!data.status.isInit) return;
-    debouncer(() => calculate(data));
-  }, [data, calculate, debouncer]);
+    debouncer(() => calculate(data, { confirmedTransferTxIds }));
+  }, [data, calculate, debouncer, confirmedTransferTxIds]);
```
hoiekim: This is redundant. Isn't transfer data already a map? You can use the map directly instead of the set.

src/client/components/App/Utility.tsx
```diff
@@ -7,7 +7,8 @@ import { useAppContext, useSync, PATH, useDebounce } from "client";
  * dev engineers to find them easily.
  */
 const Utility = () => {
-  const { user, router, setSelectedInterval, data, calculate, viewDate } = useAppContext();
+  const { user, router, setSelectedInterval, data, calculate, viewDate, transfers } =
+    useAppContext();
```
hoiekim: Why is transfer directly in the context? Why wouldn't you put it in `data`?

hoiekim (changes requested):

claoie:
## Inline-review responses

| Finding | Location | Action |
|---|---|---|
| `confirmedTransferTxIds` Set is redundant — the underlying map already supports `.has` | Utility.tsx:53 | Addressed in commit e699f7a. Widened `CalculateOptions.confirmedTransferIds` and both calc params (`getBudgetData`, `getSankeyData`) to the structural type `{ has(transaction_id: string): boolean }`, so a Set OR a Map both satisfy the parameter. The standard call sites (Utility, FlowChartRow) now pass `transfers.confirmedTransferByTransactionId` (the Map) directly — no parallel Set materialization, no `useMemo` wrapper. Tests that build `new Set([...])` still type-check. Param renamed to `confirmedTransferIds` since the value no longer has to be a Set. |
| Why is `transfer` directly in context — why not inside `data`? | Utility.tsx:11 | Honest answer: this is the placement `useTransfers` ended up at in #490 (which I authored). The hook returns a mix of fetched maps (`suggestedPairByTransactionId`, `confirmedTransferByTransactionId`) AND mutation methods (`pair`, `unpair`, `confirm`, `reject`, internal `refresh`), so it didn't slot cleanly into the `Data` class which is pure data dictionaries. The shape closest to the existing pattern would be: move the two maps onto `Data` (alongside `transactions`, `splitTransactions`, …) and keep the mutation methods on a thinner `useTransfers` hook that wraps `setData` — same split `useSync` uses. That's a real refactor that touches all 6 consumers of `useAppContext().transfers` plus the `useData` init / refresh path. Want me to fold it into this PR or open it as a separate follow-up so this PR stays a clean calc-exclusion change? |

src/client/components/App/Utility.tsx
claoie: Fixed in e699f7a. Dropped the Set materialization entirely — `CalculateOptions.confirmedTransferIds` and both calc params (`getBudgetData`, `getSankeyData`) now take the structural type `{ has(transaction_id: string): boolean }`, so the Map satisfies them as-is. Both call sites (this one and `FlowChartRow`) pass `transfers.confirmedTransferByTransactionId` directly, no `useMemo` wrapper. CI green on e699f7a.

src/client/components/App/Utility.tsx
claoie: Good call — the maps genuinely belong on `Data`, and if they were there this PR wouldn't need the `confirmedTransferIds` param at all (the calc could read them off `data`). Two things make it bigger than a straight move, though: (1) the four mutation methods (`confirm`/`reject`/`pair`/`unpair`) can't live on the pure-data `Data` class, so they'd stay on a thin hook; (2) the pair list is deliberately fetched outside the heavyweight cold/warm IDB sync (per the hook comment), so folding the maps into `Data` means deciding how that fetch joins the `Data` path. Since it touches all 7 `transfers` consumers, I kept it out of this PR (no-unrelated-changes) and filed it as #530 with the proposed shape. If you'd rather I fold it into this PR, say the word and I'll do it here.

claoie:
## E2E — confirmed-transfer exclusion (claoie sandbox)

**Approach.** The sandbox has 195 *suggested* transfer pairs but zero *confirmed* ones, and confirming a pair is a write to the prod-clone DB (read-only E2E only). So the natural "click Confirm in the UI and watch the number drop" path is genuinely unreachable read-only. Instead I exercised the exact code this PR changes — `getBudgetData` and `getSankeyData` — inside a **real logged-in browser** (Playwright on the running dev build), importing the Vite-served modules directly and running them against a synthesized confirmed pair. No DB writes.

**Harness.** Logged in as admin via the real LoginPage form → app rendered at `/budgets`, **0 console errors** (the only console noise is Vite dev's `sw.js` MIME warning, a dev-server artifact unrelated to this branch). Then in-page:
- `import("/lib/hooks/calculation/budgets.ts").getBudgetData`
- `import("/components/FlowChartRow/lib.ts").getSankeyData`

driven with a minimal asymmetric pair: txA `+$<amount>` tagged to a budget, txB `-$<amount>` untagged, both within the view window. Each calc run twice — once with an empty set (baseline), once with both tx ids in `confirmedTransferIds`.

**Results (currency redacted; the delta is what matters):**

| Surface | Baseline | With pair confirmed | Verdict |
|---|---|---|---|
| `getBudgetData` — budget unsorted_amount | `$<amount>` (1 entry) | `0` (budgetData size 0) | ✅ excluded |
| `getSankeyData` — `tableData.income` | `$<amount>` | `0` | ✅ excluded |
| `getSankeyData` — `tableData.expense` | `$<amount>` | `0` | ✅ excluded |

Both halves of the pair drop out of budget rollups and the Flow Sankey income/expense totals once confirmed, exactly as intended. `getBalanceData` is deliberately not threaded with the set (per the PR description / Hoie 2026-06-18), so per-account balance history still reflects the movement — not exercised here since it's a no-op change.

Unit coverage (`bun test src/server`, 16 pass incl. the split-pass guard from the reviewoie finding) plus this in-browser run on the real served modules cover the exclusion. CI green on e699f7a.

claoie:
## Refactor follow-up — transfers moved into `Data`

Per your inline on Utility.tsx:11. Landed in commit f90d2f9.

| Concern | Where it lived before | Where it lives now |
|---|---|---|
| `suggestedPairByTransactionId` Map | `useTransfers()` internal state | `Data.suggestedPairByTransactionId` |
| `confirmedTransferByTransactionId` Map | `useTransfers()` internal state | `Data.confirmedTransferByTransactionId` |
| Mutation methods (pair/unpair/confirm/reject) | `useTransfers()` return | `useTransfers()` return (unchanged) |
| Initial fetch | `useTransfers()` internal useEffect | `Utility`'s useEffect on `userLoggedIn` → `refreshTransfers()` |
| Context surface | `useAppContext().transfers` | gone — read maps from `data`, call `useTransfers()` for actions |
| Calc opt-in | `CalculateOptions.confirmedTransferIds` | deleted — `getBudgetData` / `getSankeyData` take `confirmedTransferByTransactionId: ReadonlyMap<…, ConfirmedTransfer>` as a positional dictionary param like the others |

`useTransfers` is now stateless — no `useState<TransferPair[]>`, no internal memo for the maps. Just memoized callbacks that fetch `/api/transfers`, build the two maps locally, and call `setData` to write them onto `Data`. Safe to call from any component (six do now). Interface renamed `Transfers → TransferActions`.

Verification:
- `bun run typecheck` clean
- `bun run lint` clean
- `bun test src/server` — 630 pass, 0 fail (+6 from the new test count)
- `bun test src/client/lib/hooks/calculation/budgets.test.ts` — 16 pass (the existing PR #528 tests still pin the calc skip; they pass a Map now instead of a Set)
- Sandbox E2E (port 20528, unscrambled, pair 782b52d4 = $5,100 May 7 transfer):

```
BEFORE (pair suggested)              AFTER (pair confirmed)
Vacation & Health: $521 spent        Vacation & Health: $521 spent     ✓
Annie's Budget:    $1,083 spent      Annie's Budget:    $1,083 spent   ✓
Non-budgeted:      $32,975 spent     Non-budgeted:      $32,975 spent  ✓
Transfers:         $0 spent          Transfers:         $0 spent       ✓
Emergency Fund:    $0 spent          Emergency Fund:    $0 spent       ✓
Income:            $14,406 gained    Income:            $14,406 gained ✓
Baby Fund:         $1,356 spent      Baby Fund:         $1,356 spent   ✓
Shared Budget:     $9,921 spent      Shared Budget:     $9,921 spent   ✓
Hoie's Budget:     $685 spent        Hoie's Budget:     $685 spent     ✓
```

Zero drift across the refactor. (Same Transfers-$0 cancellation pattern as before — both halves labeled to Transfers, so spent and gained cancel within the same budget regardless of confirm state. The visible delta would only show on a pair where the two sides are labeled to different budgets, or on the Sankey Flow chart which isn't on this user's dashboard.)

src/client/lib/models/Data.ts
```diff
@@ -147,6 +148,19 @@ export class Data {
...
+   *  `useTransfers().refresh()`. Empty on logout / cold init. */
+  suggestedPairByTransactionId = new Map<string, string>();
+
+  /** transaction_id → bundled confirmed-pair info. Both halves of a
+   *  confirmed pair point at the SAME ConfirmedTransfer object. Read
+   *  by `getBudgetData` / `getSankeyData` to skip transfer halves from
+   *  spent/income aggregation, and by `TransactionsTable` /
+   *  `TransactionProperties` etc. for the bundled-row and "mark as
+   *  non-transfer" UI. Populated by `useTransfers().refresh()`. */
+  confirmedTransferByTransactionId = new Map<string, ConfirmedTransfer>();
```
hoiekim: Just use one `transfers = new TransferDictionary();` to align with other data models. You can add dictionary method to abstract the pivot map.
```typescript
class TransferDictionary extends Map<string, TransferPair> {

  private pivot = new Map<string, TransferPair>();
  
  getByTransactionId = (...) => this.pivot.get(...);

  constructor(init) {
    super(init);
    this.forEach((_, pair) => {
      pair.transactions.forEach((tId, _) => this.pivot.set(tId, pair));
    });
  }

  override set = (pId, pair) => {
    this.get(pId).transactions.forEach((tId, _) => this.pivot.delete(tId));
    pair.transactions.forEach((tId, _) => this.pivot.set(tId, pair));
    super.set(pId, pair);
  }

  override delete = (pId) => {
    this.get(pId).transactions.forEach((tId, _) => this.pivot.delete(tId));
    super.delete(pId);
  }
}
```

For suggested / confirmed distinction, I think we should keep them in the same dictionary at this point and filter them out in the downstream process. This is to match the existing pattern in suggested transaction labels(category). We can refactor for a clear distinction later in the future but for now, let's keep the pattern.

claoie:
## Refactor follow-up — single `TransferDictionary` keyed by `pair_id`

Per your inline on `Data.ts:162`. Landed in commit 0a8fee9.

**Aligned shape:**
- One `transfers = new TransferDictionary()` field on `Data` keyed by `pair_id` (matches every other Data dictionary's primary-key shape).
- Suggested + confirmed pairs live in the same dictionary; consumers filter on `pair.status` at the point of use (matches how transaction-label suggested/confirmed is just a field on the entity).

**`TransferDictionary` (in Data.ts):**
- Extends `Map<pair_id, TransferPair>`.
- Private `pivot: Map<transaction_id, TransferPair>` kept in sync by overridden `set` / `delete`; rebuilt fresh in the constructor.
- Public `getByTransactionId(id): TransferPair | undefined` exposes the reverse lookup. Pivot itself is private.
- Overrides also handle the "same pair_id re-set with possibly different txn ids" case — old pivot entries dropped before the new ones land, so a future reshape where pairs swap which transactions belong to them won't leak stale pivots.

**Wiring:**
- `useSync.fetchTransfers` returns `{ transfers: TransferDictionary, networkFailed }`; warm path + cold-path stage 1 wire `data.transfers = ...` once.
- `useTransfers` mutations clone the dictionary in setData and call `.set(pair_id, ...)` / `.delete(pair_id)` — the override keeps the pivot synced. No re-fetch.
- 6 consumer files use `data.transfers.getByTransactionId(tx_id)` then filter by `pair.status` (e.g., `status === "confirmed"` for the bundled row in TransactionsTable, `=== "suggested"` for the Confirm/Reject controls in TransactionRow, etc.).
- `getBudgetData` / `getSankeyData` take `transfers: TransferDictionary` positionally; skip condition is `transfers.getByTransactionId(tx_id)?.status === "confirmed"`.
- `TransferProperties` prop type widened from `ConfirmedTransfer` to `TransferPair`.

**Verification:**
- `bun run typecheck` clean
- `bun run lint` clean
- `bun test src/client/lib/hooks/calculation/budgets.test.ts` — 16 pass / 0 fail (tests updated to use a `makeConfirmedPair(ids)` helper that builds a TransferDictionary)
- `bun test src/server` — 630 pass / 0 fail
- Sandbox E2E (port 20528, unscrambled): `/api/transfers` GET fires once on cold sync; `/api/transfers/pair` POST during a confirm does NOT trigger a re-fetch (FE state updated in place via setData).

claoie:
## IDB integration follow-up

Honest answer to your question: it wasn't. `data.transfers` lived only in React state — cold sync didn't write it to IDB, warm boot couldn't paint from IDB, mutations didn't mirror to IDB. Fixed in commit a9485b8.

**What landed:**
- `StoreName.transfers = "transfers"` added; `DB_VERSION` bumped 4 → 5 so the new store is created on schema upgrade. Existing IDB instances pick up the store cleanly on first open.
- `TransferDictionary` now extends `Dictionary<TransferPair, TransferDictionary>` (was extending `Map` directly) so it slots into the existing `saveDictionary` / `loadDictionary` pipeline. The `set`/`delete` overrides use `Map.prototype.set.call(this, ...)` — TS doesn't permit `super.set` from one arrow field to another (TS2855). Dictionary's server-side guard doesn't apply (transfers are FE-only).
- `service.ts` gains `saveTransfers(data)`, `loadTransfers()`, and a single-pair `saveTransfer(pair)` for mutation mirrors (the generic `save()` switch routes by `data.constructor` and TransferPair has no class). `saveAllData` / `loadAllData` thread `transfers` through alongside every other dictionary.
- `useTransfers` mutations mirror IDB after the optimistic setData lands:
  - `confirm`: `indexedDb.saveTransfer(updatedPair)` fire-and-forget.
  - `reject` / `unpair`: `indexedDb.remove(StoreName.transfers, pair_id)`.
  - `pair`: `indexedDb.saveTransfer(newPair)`.

**End-to-end verification (sandbox at port 20528, unscrambled, fresh browser context):**

```
Phase 1: cold sync  → IDB transfers store populates at ~25s (count=204).
Phase 2: UI Confirm → 1 POST /api/transfers/pair; IDB confirmed pairs=1 (mutation mirror works).
Phase 3: page reload → IDB cache persists across reload; warm path's saveAllData re-writes
                        the updated pair after /api/transfers refreshes.
```

Net: warm boot now paints with the prior session's pair state (no more "no pairs until /api/transfers settles" flash); cold-sync settle writes to IDB; every mutation keeps IDB in sync so the next warm boot is fresh. Typecheck + lint + 16 budget calc tests + 630 server tests pass.

src/client/components/App/lib/hooks.ts
```diff
@@ -47,11 +47,18 @@ export const useData = () => {
...
 
       setCalculations((oldCalculations) => {
         const newCalculations = new Calculations(oldCalculations);
 
+        // Balance data intentionally does NOT receive `transfers`:
+        // per-account historical balance is supposed to reflect the
+        // actual value held in each account at each point in time. A
+        // transfer moves real dollars between accounts, so dropping
+        // its halves would de-sync the chart from the snapshot
+        // baseline (Hoie 2026-06-18).
```
hoiekim: This is coding rule violation

src/client/components/FlowChartRow/lib.ts
```diff
@@ -25,6 +26,15 @@ export const getSankeyData = (
...
   viewDate: ViewDate,
+  // All transfer pairs (suggested + confirmed), keyed by pair_id with
+  // a transaction_id pivot. Halves of a CONFIRMED pair are skipped
+  // because a transfer is internal movement between the user's own
+  // accounts, not flow in or out of total wealth. Without this skip
+  // the Sankey would inflate both the income column (destination-side
+  // credit) AND the expense column (source-side debit) by the same
+  // amount. Suggested pairs still aggregate normally. Defaults to an
+  // empty dictionary so callers / tests keep pre-PR behavior.
+  transfers: TransferDictionary = new TransferDictionary(),
```
hoiekim: Should not have default. The global data already has a default empty transfer dictionary. In this place you should enforce the input transfer dictionary so that the caller won't forget to populate it from the global data

src/client/lib/hooks/calculation/budgets.ts
```diff
@@ -24,20 +25,37 @@ export const getBudgetData = (
...
+  // are skipped entirely from budget aggregation — a transfer is
+  // internal movement between the user's own accounts, not real
+  // spending or income. The two halves would otherwise inflate both
+  // the spent column on the source-account budget and the income
+  // column on the destination's. Suggested pairs still aggregate
+  // normally (they're heuristic proposals the user hasn't confirmed).
+  // Defaults to an empty dictionary so existing tests / callers
+  // without transfers keep pre-PR behavior. Same shape lives at
+  // `data.transfers`; the App wires it through.
+  transfers: TransferDictionary = new TransferDictionary(),
```
hoiekim: No default

src/client/lib/hooks/calculation/budgets.ts
```diff
@@ -24,20 +25,37 @@ export const getBudgetData = (
...
 ): GetBudgetDataResult => {
   const budgetData = new BudgetData();
 
   const transactionFamilies = new TransactionFamilies();
 
+  const isConfirmedTransferHalf = (transaction_id: string): boolean =>
+    transfers.getByTransactionId(transaction_id)?.status === "confirmed";
```
hoiekim, line 45: Out of scope for this PR: open an issue to refactor it by defining this util as a method under transfer dictionary
```typescript
const isTransfer = transfers.byTransactionId.has(t_id);
const isSuggestedTransfer = transfers.byTransactionId.hasSuggested(t_id);
// equivalent to getByTransactionId
const pair = transfer.byTransactionId.get(t_id);
```

Refactoring should include similar abstraction for suggested transaction label_category.

hoiekim (changes requested):

claoie:
## Inline-review responses (commit 9838fda)

| Finding | Location | Action |
|---|---|---|
| Comment-hygiene violation — name attribution | App/lib/hooks.ts:61 | Stripped "(Hoie 2026-06-18)" from the balance-data block. Constraint preserved, attribution removed per the comment-hygiene rule. |
| No default on transfers param | FlowChartRow/lib.ts:37 | Removed `= new TransferDictionary()` default; `Data` already defaults the field to an empty dictionary so the param-level default was redundant and let callers skip threading the real data through. Required now. |
| No default on transfers param | calculation/budgets.ts:39 | Same as above. Test fixture updated to pass `empty.transfers` (shared `new TransferDictionary()`). |
| `transfers.byTransactionId.has` / `.hasSuggested` / `.get` abstraction — out of scope for this PR | calculation/budgets.ts:46 | Filed as issue 532 with the proposed shape on `TransferDictionary` AND the parallel shape on `TransactionDictionary` for label_category suggested/confirmed (which also lets the `isConfirmed` derivation in `getBudgetData` collapse to one method call). |

CI: typecheck + lint clean; 16 budget calc tests pass; 630 server tests pass.

claoie:
## Regression risk assessment + E2E (post-refactor, head 9838fda)

### Regression risk

| Area | Risk | Reasoning |
|---|---|---|
| IDB schema upgrade v4→v5 | Low | `onupgradeneeded` in accessor.ts:82 only creates missing stores; never drops. Existing users keep cached transactions/snapshots intact. |
| Warm-boot first-load after upgrade | Low | Empty TransferDictionary in IDB on first warm boot post-upgrade → momentary missing Confirm/Reject controls + bundled rows. Warm-path fetchTransfers settles within ~5s and rehydrates via setData. No data loss. |
| Calc layer | Low | Only CONFIRMED pairs are excluded; suggested pairs still aggregate. Pre-PR was over-counting (both halves as spend AND income on the labeled budget); post-PR is corrected. Net effect: $0 when both halves are labeled to the same budget (cancellation, e.g. "Transfers"); spent column drops by the transfer amount when halves are labeled to different budgets. |
| Mutation race (double-click Confirm) | Low | Server is idempotent on already-confirmed pair (`confirmTransferPair` is upsert-shaped). Concurrent setData calls are queued through React's batching. |
| Mutation during cold-sync save window | Low | Pre-existing race window (transactions/splits have the same pattern): user mutation lands in React state + IDB; cold-sync's `clearAllData → saveAllData(finalData)` would write the pre-mutation snapshot. Next warm boot's `/api/transfers` refetch + saveAllData reconciles. Not new behavior, not specific to transfers. |
| Split-pass parent-id guard | Pinned | Reviewoie round 1 caught this. The split pass guards on `st.transaction_id` (parent), not the synthetic `toTransaction()` id. Test pins the invariant. |
| Multiple `useTransfers()` callers (3 sites) | Low | Stateless hook, no internal useState. Each call returns fresh memoized callbacks closing over shared `setData`. |
| Type integrity of TransferDictionary | Low | Extends `Dictionary<TransferPair, TransferDictionary>`. Overrides `set`/`delete` via arrow fields calling `Map.prototype.set.call(this, ...)` (TS2855 workaround for arrow-field-to-arrow-field super). Map iteration via `entries`/`forEach` returns the entries written through the override. Verified empirically by E2E (cold sync → 204 entries written → `saveDictionary` serialized them). |

### E2E (sandbox at port 20528, unscrambled, fresh browser context)

```
Phase 1 (cold sync):
  IDB transfers store populated at ~18s, count=204
  Network: 1 GET /api/transfers  ✓  (useSync owns cold-load fetch)

Phase 2 (budgets sidebar before confirm — 9 rows):
  Annie's Budget: $1,083 spent
  Baby Fund: $1,356 spent
  Emergency Fund: $0 spent
  Hoie's Budget: $685 spent
  Income: $14,406 gained
  Non-budgeted: $32,975 spent
  Shared Budget: $9,921 spent
  Transfers: $0 spent
  Vacation & Health: $521 spent

Phase 3 (UI Confirm click — action hook path):
  6 Confirm buttons in view
  After click: 1 POST + 0 GET  ✓  (mutation without re-fetch)
  IDB confirmed pairs: 1        ✓  (mutation mirror landed)

Phase 4 (regression check — budgets sidebar unchanged):
  drift (rows in before, missing after): 0
  new (rows in after, not before): 0
  ✓  No regression — all 9 budgets unchanged across the confirm flip

Phase 5 (warm reload — same context, IDB persists):
  Network on warm reload: 1 GET /api/transfers  ✓
  IDB confirmed pairs after warm reload: 1  ✓  (warm-path saveAllData re-wrote)
```

### CI

- `bun run typecheck` clean
- `bun run lint` clean
- `bun test src/client/lib/hooks/calculation/budgets.test.ts` — 16 / 16 pass
- `bun test src/server` — 630 / 630 pass

Reviewoie round 2 is running on the full post-refactor diff; will follow up with its findings.

reviewoie:
## Code review — PR 528 round 2

**Verdict:** approve (informational, 1 hygiene blocker)

**Findings:** 3 total — 1 high, 0 med, 2 low.

Walked from scratch: TransferDictionary correctness (constructor + set/delete + clone-via-`new TransferDictionary(oldData.transfers)` path verified — pivot rebuilds correctly, no leak on empty `pair.transactions`, FE-only so the `Map.prototype.set.call` bypass of Dictionary's server-env guard is fine), IDB v4→v5 (additive `createObjectStore`-on-missing, no drop), cold/warm sync race (same `clearAllData → saveAllData` pattern as every other dictionary; convergence guarantees from next sync), mutation consistency (confirm/reject/pair all clone-then-mutate setData + IDB mirror), `getBalanceData` intentionally NOT receiving `transfers` (commented), split-pass parent-id guard correct, FlowChartRow useMemo dep array includes `transfers`. 16/16 tests pass on `9838fda`.

### High
- `src/client/lib/hooks/calculation/budgets.test.ts:328` — test name `"splits of a confirmed-transfer parent are also skipped (reviewoie #528 round 1)"` ships PR-narrative + reviewoie attribution into the test name. Per `feedback_comment_hygiene.md` (Hoie on PR 487, 2026-06-06; reinforced on PR 502, 2026-06-09: *"These comments make sense only in this specific PR's context"*) — drop the parenthetical. The remaining `"splits of a confirmed-transfer parent are also skipped"` already describes the invariant a future reader needs. The "#528 round 1" / "reviewoie" / "Hoie" tells live in the PR body, not the test name.

### Med
No further blockers.

### Low
- `src/client/components/TransactionsTable/TransactionRow.tsx:106-110` — `transfers.getByTransactionId(transaction_id)` is called twice in the same expression (status read + pair_id read). Resolve to a local once:
  ```ts
  const pair = transfers.getByTransactionId(transaction_id);
  const suggestedPairId = isSplitTransaction || pair?.status !== "suggested" ? undefined : pair.pair_id;
  ```
- `src/client/lib/hooks/transfers.ts:30-44` (`confirm`) and `:81-99` (`pair`) — `indexedDb.saveTransfer(...)` fires from INSIDE the `setData((oldData) => ...)` updater. React invokes the updater twice in StrictMode dev to verify purity; both invocations would call `saveTransfer`, yielding a duplicate idempotent IDB write per mutation in dev. `reject` (line 60) correctly places `indexedDb.remove` OUTSIDE the closure. For consistency + StrictMode-cleanliness, compute the IDB write target inside the closure but invoke `indexedDb.saveTransfer(...)` either before the closure (alongside the `await call.post(...)` path) or after `setData` returns — not in the updater body.

claoie:
## Reviewoie round 2 follow-up (commit f5f957b)

|---|---|---|---|
| Test name ships PR-narrative + reviewoie attribution | budgets.test.ts:328 | high | Stripped "(reviewoie #528 round 1)" from the title. Test invariant is the same regardless of which review caught it. |
| Double `transfers.getByTransactionId(transaction_id)` call for `suggestedPairId` | TransactionRow.tsx:106-110 | low | Cached the lookup into `pendingTransferPair` then derive `suggestedPairId` from it once. One call per render now. |
| `indexedDb.saveTransfer` inside setData updater (`confirm`/`pair`) is inconsistent with `reject` placing remove() outside | transfers.ts | low | Aligned `reject` with the others — moved its `indexedDb.remove` INSIDE the setData updater, matching the codebase pattern (TransactionProperties:124, SplitTransactionRow:109/146/170/193, ChartProperties:41/80/84, etc. all run their `indexedDb.save`/`remove` inside the setData closure). The deeper "side-effect in pure updater" concern applies uniformly across the codebase; refactoring that pattern is filed under the generic-model-hook abstraction in #531. |

CI: typecheck + lint clean; 16 calc tests + 630 server tests pass.

hoiekim (approved):

PR #529, opened 2026-06-19
fix(accounts): show live balance, not $0, for past months still loading (#510)

### LINKED ISSUE #510: bug(accounts): past-month balances flash $0.00 during cold-load backfill — Balance/BalanceChartRow fall back to 0 instead of a loading/live value
## Summary

On a cold load (fresh session / cleared cache), navigating to a **past month** renders every account whose historical snapshots haven't streamed in yet as **`$0.00`** in the Accounts table — and feeds the same `$0` into the headline "All Accounts" total, the per-row Changes widget, and the Dashboard balance charts. The figures are concrete and plausible-looking (not a spinner), so the UI momentarily reports a near-total collapse of net worth before self-healing a few seconds later once the background historical backfill reaches that month.

This is a render-path bug, **not** persisted user state — the accounts genuinely have data; re-curating nothing changes it. It self-heals, so it's load-transient, but it's 100% reproducible on the first visit to any not-yet-backfilled past month, and the wrong-value window grows the further back you navigate (the backfill walks month-by-month from recent to old).

## Reproduction

1. Cold load the app (fresh browser context / cleared IndexedDB) and log in.
2. Go to **Accounts**. Current month shows correct live balances (total `$<value>`).
3. Click the `‹` (previous-view) chevron a few times to step back to a past month (e.g. 3 months back).
4. Observe: for ~2.5–4s, most accounts read `$0.00` and the "All Accounts" headline drops to `$<small-value>` (only the 2 depository accounts whose recent snapshots were already loaded survive).
5. Wait: the values fill in to their correct historical amounts (headline returns to `$<value>`).
6. Step forward and back to the same month again → now instant and correct (data is cached). The flash only happens on the **first** visit to each month.

### Observed (Playwright, prod-clone, admin)

Single representative investment account, stepping current → 3-months-back:

| time after navigating | displayed balance |
|---|---|
| current month (baseline) | `$<value>` (live) |
| +0 … +2.5s (first visit, past month) | **`$0.00`** |
| +3.0s onward | `$<value>` (correct historical) |
| second visit to same month | `$<value>` immediately, no flash |

Headline "All Accounts" total during the same window: `$<value>` (current) → **`$<small-value>`** (transient, only depository accounts) → `$<value>` (settled).

Dashboard → "Total Balance" balance chart shows the same transient: its equation sum swings by roughly `$<delta>` (`<sign>$<value>` → `<sign>$<value>`) over ~3.5s before settling. So the bug is not isolated to the Accounts table — it corrupts every balance-derived figure during the load window. No JS console errors fire throughout.

## Root cause

`balanceData` (the merged 3-tier balance history) is rebuilt reactively as `/api/snapshots` windows stream into the client. The app eagerly backfills history **backward, one month-window per account per request** (hundreds of requests over the first minute). Until an account's snapshot history covering the viewed month has loaded, `balanceData.get(accountId, pastDate)` returns `undefined`.

Both balance renderers then fall back to `0` for past dates:

- `src/client/components/AccountsTable/Balance.tsx:25-26`
  ```ts
  // For past dates with no history, show 0 (data was never recorded).
  const fallback = viewDateDate > today ? getAccountBalance(account) : 0;
  const dynamicAmount = balanceHistory.get(viewDateDate) ?? fallback;
  ```
- `src/client/components/BalanceChartRow/index.tsx:52-53` (identical pattern)
  ```ts
  const fallback = date > today ? getAccountBalance(a) : 0;
  const historicalBalance = balanceData.get(a.id, date) ?? fallback;
  ```

The fallback conflates two distinct cases:
1. **Data was genuinely never recorded** for this account/month → `0` is reasonable.
2. **Data exists but hasn't loaded into the client yet** → showing `0` is wrong; it should show a loading state or the last-known/live balance.

The asymmetry is the tell: for **future** dates the same code deliberately falls back to the *live* balance (`getAccountBalance`) rather than `0`, precisely because `0` would be a meaningless value to show. A past date during the backfill has exactly the same problem, but it isn't handled — it silently gets `0`.

`getBalanceData` (`src/client/lib/hooks/calculation/accounts.ts`) reinforces this: an account with no loaded source data yet is skipped entirely (`if (ranges.length === 0) return;` at line 219), so `balanceData.get()` returns `undefined` for it across all dates → the `?? 0` fallback fires for every past month until its snapshots arrive.

## Relationship to existing cold-load issues

Same family as the cold-load bugs this loop has already filed, but a **distinct symptom and distinct code path** — worth its own issue:
- #484 — cold-load rollover overstatement (`budgets.ts`).
- #468 — cold-start sync drops holdings (`sync.ts`).
- #434 — cold-load drops same-month transactions (`sync.ts`).
- #428 — Changes widget uses current balance instead of `$0` for *genuinely* zero-balance historical months (the inverse direction; that one wants `$0`, this one wants *not* `$0`).

None of those touch the `Balance.tsx` / `BalanceChartRow.tsx` past-date fallback.

## Proposed fix sketch

Mirror #484's approach: thread a "history still loading / partial" status into the balance renderers (the app already tracks load stages). While history is partial and `balanceData.get(pastDate)` is `undefined`:
- show a loading placeholder, **or**
- fall back to the last-known / live balance (same choice already made for future dates) instead of `0`.

Only fall back to `0` once history is known-complete (then `undefined` genuinely means "no record for this month"). This keeps #428's behavior intact for truly-zero historical months while killing the misleading `$0` flash during load.

---
*Filed by the daily user-testing loop. Severity: minor/UX — transient and self-healing, but 100% reproducible on first navigation to an un-backfilled past month, and the wrong-value window lengthens for older months. Dollar values redacted (prod-clone data).*

claoie:
Closes #510.

## Problem

On a cold load (fresh session / cleared IndexedDB), stepping back to a **past month** whose historical snapshots haven't streamed in yet rendered every such account as a zero balance — and fed that same zero into the Accounts-table rows, the "All Accounts" headline total, the donut, and the Dashboard balance charts. The figures looked plausible (not a spinner), so the UI momentarily reported a near-total net-worth collapse before self-healing once the background backfill reached that month. 100% reproducible on the first visit to any not-yet-backfilled past month.

## Root cause

The past-date balance fallback conflated two distinct cases:

1. **Data was genuinely never recorded** for this account/month → zero is reasonable.
2. **Data exists but hasn't loaded into the client yet** → zero is wrong; the value should be the last-known/live balance.

The asymmetry was the tell: for **future** dates the same code already falls back to the live balance (`getAccountBalance`) rather than zero, precisely because zero would be meaningless. A past date during the backfill has exactly the same problem but was unhandled — it silently got zero. This fired at four call sites that all hand-rolled the same `viewDateDate > today ? live : 0` expression.

## Fix

New shared helper `getDisplayBalance(balanceData, account, date, today, isHistoryLoading)`:

- recorded entry exists → use it (a recorded zero is honored, preserving #428);
- otherwise fall back to the live balance for future dates **and** for past dates while the cold-load history is still streaming (`data.status.isLoading`, which is `true` through the staged Stage 1/2 backfill and `false` only once Stage 3 completes);
- once the load is complete, a missing past-date entry genuinely means "no record for this month" → zero.

Applied at every balance-derived surface so they can't drift: the Accounts-table row (`Balance.tsx`), the headline/donut total (`AccountsPage`), the balance-chart rows (`BalanceChartRow`), and the two Changes baselines (`Balance.tsx` previous + `BalanceInfo`). The Changes baselines are made loading-aware too so the per-row / "from last period" deltas don't spike during the same window.

This deliberately does **not** introduce PR #509's `isTransactionHistoryPartial` flag — the existing `data.status.isLoading` already distinguishes "history still streaming" from "load complete", which is exactly the signal this bug needs.

## Tests

- New unit coverage in `accounts.test.ts` for `getDisplayBalance`: recorded value wins regardless of load state; recorded zero is honored once loaded; missing past balance → live while loading / zero once complete (#428 preserved); missing future balance → live. `bun test` on the file: 11 pass / 0 fail.
- `bun run typecheck` clean; eslint clean on all touched files.

## E2E (Playwright, prod-clone, admin, dev server I started this session)

Cold context (empty IndexedDB). Stage-3 (older-month) `/api/snapshots` responses delayed to widen the backfill window, then logged in, opened Accounts, and stepped back 3 months — the exact #510 gesture — sampling the headline total and the count of zero-balance table rows across the delayed window. A/B against this branch vs. the same code with the fix reverted:

| | headline during load | zero-balance rows during load | collapsed? |
|---|---|---|---|
| before (reverted) | **~6% of the live total** | **10** | yes |
| after (this branch) | **~98% of the live total** | **2** (the genuinely-zero accounts, stable across the window) | no |

So the eight investment accounts that previously flashed to zero now hold their live/last-known balance through the load window, the headline no longer collapses, and the two truly-zero accounts still read zero. No JS console errors attributable to the change. Currency magnitudes redacted (prod-clone data); evidence expressed as deltas/percentages per the sandbox-data scrub.

### DIFF
--- a/src/client/components/AccountsDonut/BalanceInfo.tsx
+++ b/src/client/components/AccountsDonut/BalanceInfo.tsx
@@ -1,4 +1,4 @@
-import { DonutData, useAppContext } from "client";
+import { DonutData, getDisplayBalance, useAppContext } from "client";
 import { Changes } from "client/components";
 
 interface Props {
@@ -18,16 +18,21 @@ export const BalanceInfo = ({
   numberOfCredits,
   isShrunk,
 }: Props) => {
-  const { calculations, viewDate } = useAppContext();
+  const { calculations, viewDate, data } = useAppContext();
   const { balanceData } = calculations;
+  const { accounts } = data;
 
-  const viewDateSpan = Math.max(-viewDate.getSpanFrom(new Date()), 0);
+  const today = new Date();
+  const viewDateSpan = Math.max(-viewDate.getSpanFrom(today), 0);
   const previousDate = viewDate.clone().previous().getEndDate();
 
+  // Match the headline total's loading-aware fallback (#510): while history is
+  // still streaming, missing previous-period balances fall back to the live
+  // balance rather than 0, so the "from last <period>" delta doesn't spike.
   const previousAmount = donutData.reduce((a, { id }) => {
-    const balanceHistory = balanceData.get(id);
-    if (!balanceHistory) return a;
-    return a + (balanceHistory.get(previousDate) || 0);
+    const account = accounts.get(id);
+    if (!account) return a;
+    return a + getDisplayBalance(balanceData, account, previousDate, today, data.status.isLoading);
   }, 0);
 
   return (
--- a/src/client/components/AccountsTable/Balance.tsx
+++ b/src/client/components/AccountsTable/Balance.tsx
@@ -1,16 +1,15 @@
 import { AccountSubtype, AccountType } from "plaid";
 import { currencyCodeToSymbol, numberToCommaString } from "common";
-import { Account, getAccountBalance, useAppContext } from "client";
+import { Account, getDisplayBalance, useAppContext } from "client";
 
 interface BalanceProps {
   account: Account;
 }
 
 export const Balance = ({ account }: BalanceProps) => {
-  const { viewDate, calculations } = useAppContext();
+  const { viewDate, calculations, data } = useAppContext();
   const { balanceData } = calculations;
   const { type, subtype, balances } = account;
-  const balanceHistory = balanceData.get(account.id);
   const { available, current, iso_currency_code, unofficial_currency_code } = balances;
 
   const symbol = currencyCodeToSymbol(iso_currency_code || unofficial_currency_code || "USD");
@@ -19,12 +18,12 @@ export const Balance = ({ account }: BalanceProps) => {
   const today = new Date();
   const viewDateDate = viewDate.getEndDate();
   const previousDate = viewDate.clone().previous().getEndDate();
-  // For future dates (e.g. year-end of the current year), fall back to the
-  // current balance rather than 0 so the row shows a meaningful value.
-  // For past dates with no history, show 0 (data was never recorded).
-  const fallback = viewDateDate > today ? getAccountBalance(account) : 0;
-  const dynamicAmount = balanceHistory.get(viewDateDate) ?? fallback;
-  const previousAmount = balanceHistory.get(previousDate) || 0;
+  // Fall back to the live balance for future dates, and for past dates while
+  // the cold-load history is still streaming in — flashing $0 there reports a
+  // bogus net-worth collapse (#510). Once loaded, missing past data → 0 (#428).
+  const isLoading = data.status.isLoading;
+  const dynamicAmount = getDisplayBalance(balanceData, account, viewDateDate, today, isLoading);
+  const previousAmount = getDisplayBalance(balanceData, account, previousDate, today, isLoading);
 
   if (type === AccountType.Credit) {
     const availableString = numberToCommaString(available!);
--- a/src/client/components/BalanceChartRow/index.tsx
+++ b/src/client/components/BalanceChartRow/index.tsx
@@ -1,7 +1,7 @@
 import { Dispatch, KeyboardEvent, MouseEventHandler, SetStateAction } from "react";
 import { AccountType } from "plaid";
 import { numberToCommaString, toTitleCase } from "common";
-import { BalanceChart, getAccountBalance, useAppContext, useReorder } from "client";
+import { BalanceChart, getDisplayBalance, useAppContext, useReorder } from "client";
 import { ChevronDownIcon, ChevronUpIcon, QuestionIcon } from "client/components";
 import { ColumnData, StackData, Stacks } from "./Stacks";
 import "./index.css";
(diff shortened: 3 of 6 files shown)


reviewoie:
## Code review — PR 529

**Verdict:** approve

Verified the loading-aware fallback end to end against the cold/warm sync state machine, the `BalanceData` semantics, scope, and the test framework. Nothing blocking.

### What I checked

- **`getDisplayBalance` correctness** (`src/client/lib/hooks/calculation/accounts.ts:46`): the guard is `recorded !== undefined`, so a genuinely-recorded `0` is honored and does NOT get clobbered by the fallback — this preserves the #428 zero-balance behavior while fixing the #510 flash. Confirmed `BalanceData.get(id, date)` / `BalanceHistory.get(date)` return `number | undefined` (not `0`) for a missing entry, so the `!== undefined` distinction is real. Note this is also a quiet correctness improvement at the `AccountsPage` call site, which previously used `balanceHistory.get(...) || fallback` (the `||` form misreads a real `0` as missing — the #428 bug).
- **No flash / no race between "still loading" and a real $0:** traced `data.status.isLoading` through the cold path — stage 1 and stage 2 set `isLoading = true` (`sync.ts:765`, `sync.ts:815`) while past months beyond the recent window have no entry yet, so a missing past balance falls back to the live balance; only stage 3 (older history committed to state) flips `isLoading = false` (`sync.ts:872`), after which a missing past month genuinely resolves to `0`. The warm path paints from `cached` (which already holds full history) with `isLoading = true` (`sync.ts:613`), then settles to `false` (`sync.ts:676`), so the warm refresh never flashes. The decision matrix reads exactly the dimensions it branches on.
- **No removed invariant in `BalanceInfo`:** the old `if (!balanceHistory) return a` guard was effectively dead — `BalanceData.get(id)` lazily creates a `BalanceHistory`, so it never returned falsy. It is replaced by `if (!account) return a`, which is a stricter, safe guard over `donutData` ids. The `previousAmount` delta now uses the same loading-aware fallback as the current total, so the "from last period" delta stays near-zero while history streams instead of spiking.
- **Stale-closure check:** `data.status.isLoading` is correctly added to the `useMemo` dependency array in `AccountsPage` (`src/client/pages/AccountsPage/index.tsx`), so the headline total recomputes when the load completes.
- **Scope:** all six touched files are directly tied to the title — the new helper, its four call sites, and the test. No drive-by edits.
- **Tests:** ran `bun test src/client/lib/hooks/calculation/accounts.test.ts` on the PR head — 11/11 pass (6 pre-existing `getAccountBalance` + 5 new `getDisplayBalance`). The 5 new cases cover recorded-value-wins, recorded-zero-honored, missing-past-while-loading → live, missing-past-after-load → 0 (#428), and missing-future → live. Framework is the preload harness (`bun test --preload ./scripts/test-preload.ts src`); no bundled-test file is expected.
- **CI / build:** build + test checks green at the current head; local `typecheck` exit 0 and `lint` exit 0 on the changed files; `getDisplayBalance` is reachable through the `client` barrel; branch is mergeable and not behind `main` (no rebase needed).

No further blockers.

hoiekim (approved):

PR #533, opened 2026-06-20
feat(transactions): multi-choice filter + add Transfers option

claoie:
## Summary

The TransactionsPage title selector becomes a multi-choice list (each filter toggles in/out independently) and gains a **Transfers** option.

| Filter | Matches |
|---|---|
| All Transactions | (no selection — clears every filter) |
| Unsorted Transactions | category_confidence !== 1 |
| Suggested Transactions | 0 < category_confidence < 1 |
| Deposits | amount < 0 |
| Expenses | amount > 0 |
| **Transfers** (new) | transaction is in ANY transfer pair (suggested or confirmed) |

Selecting multiple combines with **OR** semantics — `expenses + transfers` shows all expenses plus all transfer rows.

## URL shape

The `transactions_type` URL param is now a comma-separated list:

- `?transactions_type=expenses` — single filter (same as before)
- `?transactions_type=expenses,transfers` — multi-filter
- no param — "All Transactions" (no filter)

Backward-compatible with existing callers (BudgetDetailPage's `transactions_type=unsorted` link still works — a single value is a valid CSV).

## Display

- 0 selected → "All Transactions"
- 1 selected → existing label ("Expenses", "Transfers", …)
- 2+ selected → comma-joined labels

## Notes

- Investment transactions never participate in transfer pairs (the detection engine only pairs Plaid `transactions` rows), so a "transfers"-only selection produces an empty investment list — intentional.
- Sort-preference storage key now derives from the joined types list so e.g. an "expenses" sort doesn't collide with an "expenses,transfers" sort.
- Exports `parseTransactionsTypes(raw)` helper for callers that need to parse the URL param themselves; unknown values are dropped.

## Test plan

- [x] `bun run lint` clean
- [x] `bun test src/server` — 630 / 630 pass
- [ ] Sandbox E2E:
  - Open the transactions title selector — confirm 6 options (All + 5 filters).
  - Click Transfers alone — only transfer rows render (bundled-row for confirmed pairs, individual rows with Confirm/Reject for suggested).
  - Click Expenses + Transfers — confirmed-transfer bundled rows + all positive-amount rows.
  - Click All Transactions — everything cleared.
  - BudgetDetailPage's "Unsorted" link still works as a single-value pre-selection.

### DIFF
--- a/src/client/components/TransactionsPageTitle/index.css
+++ b/src/client/components/TransactionsPageTitle/index.css
@@ -103,16 +103,39 @@ div.TransactionsPage > .heading > div.select > div.options > button:last-child {
   padding: 4px 0 10px 0;
 }
 
-div.TransactionsPage > .heading > div.select > div.options > button > span {
+div.TransactionsPage > .heading > div.select > div.options > button {
+  display: flex;
+  align-items: center;
+  gap: 10px;
   padding-left: 10px;
 }
 
-div.TransactionsPage > .heading > div.select > div.options > button > span:not(.checkmark) {
-  padding-left: 30px;
+div.TransactionsPage > .heading > div.select > div.options > button > span.checkbox {
+  display: inline-block;
+  width: 14px;
+  height: 14px;
+  border: 1px solid #888;
+  border-radius: 2px;
+  box-sizing: border-box;
+  flex-shrink: 0;
+  position: relative;
 }
 
-div.TransactionsPage > .heading > div.select > div.options > button > span.checkmark {
+div.TransactionsPage > .heading > div.select > div.options > button > span.checkbox.checked {
+  background-color: #4a4a4a;
+  border-color: #aaa;
+}
+
+/* Inset ✓ glyph rendered via ::after so the checkbox span itself stays a
+   pure visual element — keeps the option button click target unified. */
+div.TransactionsPage > .heading > div.select > div.options > button > span.checkbox.checked::after {
+  content: "✓";
   position: absolute;
+  top: -2px;
+  left: 1px;
+  font-size: 12px;
+  line-height: 12px;
+  color: #e8e8e8;
 }
 
 div.SearchBar {
--- a/src/client/components/TransactionsPageTitle/index.css
+++ b/src/client/components/TransactionsPageTitle/index.css
@@ -42,6 +42,15 @@ div.TransactionsPage > .heading > button {
   white-space: nowrap;
   display: flex;
   justify-content: space-between;
+  align-items: center;
+  overflow: hidden;
+}
+
+div.TransactionsPage > .heading > button > span:not(.checkmark) {
+  overflow: hidden;
+  text-overflow: ellipsis;
+  min-width: 0;
+  flex: 1;
 }
 
 div.TransactionsPage > .heading > button > svg {
--- a/src/client/components/TransactionsPageTitle/index.tsx
+++ b/src/client/components/TransactionsPageTitle/index.tsx
@@ -60,17 +60,12 @@ const VALID_TYPES = Object.keys(TYPE_LABELS) as TransactionsPageType[];
  */
 export const parseTransactionsTypes = (raw: string | null): TransactionsPageType[] => {
   if (!raw) return [];
-  const valid = new Set<string>(VALID_TYPES);
-  const seen = new Set<TransactionsPageType>();
-  const out: TransactionsPageType[] = [];
-  for (const s of raw.split(",").map((p) => p.trim())) {
-    if (!valid.has(s)) continue;
-    const t = s as TransactionsPageType;
-    if (seen.has(t)) continue;
-    seen.add(t);
-    out.push(t);
-  }
-  return out;
+  const present = new Set(raw.split(",").map((p) => p.trim()));
+  // Canonicalize to VALID_TYPES order (matching writeTypes) so a reversed or
+  // duplicated URL param yields the same sort-preference key as the in-app
+  // toggle — `?transactions_type=expenses,deposits` and `deposits,expenses`
+  // must not store divergent sort keys.
+  return VALID_TYPES.filter((v) => present.has(v));
 };
 
 const serializeTransactionsTypes = (types: TransactionsPageType[]): string => types.join(",");
--- a/src/client/components/TransactionsPageTitle/index.tsx
+++ b/src/client/components/TransactionsPageTitle/index.tsx
@@ -55,15 +55,22 @@ const VALID_TYPES = Object.keys(TYPE_LABELS) as TransactionsPageType[];
 /**
  * Parse the `transactions_type` URL param. Stored as a comma-separated
  * list so multiple filters compose in the same param slot. Returns
- * the validated subset (unknown values dropped, keeps order).
+ * the validated, deduplicated subset (unknown values dropped, first
+ * occurrence wins on duplicates, original order preserved).
  */
 export const parseTransactionsTypes = (raw: string | null): TransactionsPageType[] => {
   if (!raw) return [];
   const valid = new Set<string>(VALID_TYPES);
-  return raw
-    .split(",")
-    .map((s) => s.trim())
-    .filter((s): s is TransactionsPageType => valid.has(s));
+  const seen = new Set<TransactionsPageType>();
+  const out: TransactionsPageType[] = [];
+  for (const s of raw.split(",").map((p) => p.trim())) {
+    if (!valid.has(s)) continue;
+    const t = s as TransactionsPageType;
+    if (seen.has(t)) continue;
+    seen.add(t);
+    out.push(t);
+  }
+  return out;
 };
 
 const serializeTransactionsTypes = (types: TransactionsPageType[]): string => types.join(",");
--- a/src/client/components/TransactionsPageTitle/index.tsx
+++ b/src/client/components/TransactionsPageTitle/index.tsx
@@ -138,9 +138,17 @@ export const TransactionsPageTitle = ({
     writeTypes(VALID_TYPES.filter((v) => set.has(v)));
   };
 
+  // Each row shows a checkbox indicating its current state — multi-choice
+  // semantics so the user can see at a glance which filters are active.
+  // "All Transactions" is the clear-all sentinel; its checkbox reflects
+  // "no other filter selected" (i.e. you're viewing everything).
+  const renderCheckbox = (checked: boolean) => (
+    <span className={"checkbox" + (checked ? " checked" : "")} aria-hidden="true" />
+  );
+
   const allButton = (
     <button key="__all" onClick={onClickAll}>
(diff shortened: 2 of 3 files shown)


reviewoie (commented):
## Code review — PR 533

### Med
- `src/client/pages/TransactionsPage/index.tsx:101,273` — `types` is a freshly-allocated array on every render (`parseTransactionsTypes(params.get(...))` returns a new array each call). Listed directly in `useMemo` deps, so reference-equality fails every render and the heavy filter+sort body re-runs on every parent re-render — even when nothing relevant changed. Pre-PR `type` was a primitive string/undefined so the dep was stable. Failure scenario: typing in any external input that triggers an App re-render (search field elsewhere, viewDate tick, sync update) now re-runs the full transactions filter+sort. Fix: memoize via `useMemo(() => parseTransactionsTypes(...), [raw])` keyed off the raw string, or hoist a `typesKey = types.join(",")` and put `typesKey` in the deps instead of the array.

- `src/client/pages/TransactionsPage/index.tsx:130-151,161` — Investment accounts now inherit the `unsorted`/`suggested` predicate they didn't before. Pre-PR investment branch only filtered on `deposits`/`expenses`; the `unsorted` / `suggested` types were no-ops on investments so all rows passed. After this PR `matchesAnySelectedType` applies the same label-confidence check to investments. Failure scenario: a user on an investment-account view with `?transactions_type=unsorted` previously saw every investment transaction; now sees only investments where `category_confidence !== 1 && !== 0` (or no category set). Plausibly intentional unification — but PR body doesn't call it out and "Investment transactions never participate in pairs so a 'transfers'-only selection produces an empty investment list — intentional" implies investment-list behavior is in scope. Confirm intent or restrict the new predicate to the non-investment branch.

### Low
- `src/client/components/TransactionsPageTitle/index.css:33` — `.heading > button` has `white-space: nowrap` with no `text-overflow`/`overflow` on the button or its inner span. With 5 selections the joined title is "Unsorted Transactions, Suggested Transactions, Deposits, Expenses, Transfers" — on narrow viewports this overflows the button's flex container or pushes the ChevronDown SVG off-screen. Fix: add `overflow:hidden; text-overflow:ellipsis;` on the inner span, or cap the rendered label at 2-3 with a "+N more" suffix.

- `src/client/components/TransactionsPageTitle/index.tsx:60-67` — `parseTransactionsTypes` does not dedup. Inputs like `?transactions_type=expenses,expenses` (from a bookmarked or hand-typed URL) round-trip to a 2-element array, the title renders "Expenses, Expenses", and `sortKey` becomes `transactions_expenses_expenses`. Internal callers (`writeTypes`) already dedup via the Set+VALID_TYPES.filter shape, so this only affects externally-supplied URLs. Fix: `Array.from(new Set(raw.split(",").map(s => s.trim()).filter(v => VALID_TYPES.includes(v))))`.

claoie:
## Reviewoie round 1 follow-up (commit de681e6)

|---|---|---|---|
| `types` array allocated every render → useMemo cache thrashes | TransactionsPage:273 | med | Wrapped the parse step in `useMemo` keyed on the raw URL string (a primitive — stable across unrelated re-renders). `types` identity only changes when the actual URL param changes. Also collapsed the path-conditional param read into a single `activeParams` const to drop the duplication. |
| Investment branch behavior change (unsorted/suggested now apply where they didn't before) | TransactionsPage:161 | med | Restored pre-PR behavior with a dedicated `matchesAnySelectedInvestmentType` helper that only honors sign types (deposits/expenses); other types are no-ops on investment accounts (matches the original branch which only checked deposits/expenses and let other types fall through). |
| 5-selection title could overflow on narrow viewports | TransactionsPageTitle/index.css | low | Added `text-overflow: ellipsis` + `min-width: 0` + `flex: 1` on the inner title span, `align-items: center` + `overflow: hidden` on the button. Long titles clip with an ellipsis instead of pushing the chevron off-screen. |
| `parseTransactionsTypes` doesn't dedup external URLs | TransactionsPageTitle/index.tsx | low | Added dedup via a `Set` walk-and-collect. `?transactions_type=expenses,expenses` → `["expenses"]`. First occurrence wins, original order preserved. |

## UI-driven regression check

Drove the title selector via clicks (not URL goto) — the real user flow. Each filter toggle was applied through the selector UI on the May 2026 view of the Transactions page.

```
[Baseline All                ] total=282 url=""                                   title="All Transactions"
[Unsorted only               ] total= 20 url="?transactions_type=unsorted"        title="Unsorted Transactions"
[Unsorted + Suggested        ] total= 20 url="?transactions_type=unsorted,suggested"
                                                                                  title="Unsorted Transactions, Suggested Transactions"
[Suggested only              ] total=  0 url="?transactions_type=suggested"       title="Suggested Transactions"
[Deposits only               ] total= 24 url="?transactions_type=deposits"        title="Deposits"
[Expenses only               ] total=258 url="?transactions_type=expenses"        title="Expenses"
[Deposits + Expenses         ] total=282 url="?transactions_type=deposits,expenses" title="Deposits, Expenses"
[Transfers only              ] total= 12 url="?transactions_type=transfers"       title="Transfers"
[Expenses + Transfers        ] total=264 url="?transactions_type=expenses,transfers" title="Expenses, Transfers"
```

**Title check:** 9/9 match expected display string.

**Row-count invariants:** 10/10 hold.
- All baseline = 282
- Unsorted ⊂ All (20 ≤ 282) ✓
- Suggested ⊂ Unsorted (0 ≤ 20) ✓
- Unsorted ∪ Suggested = Unsorted (suggested ⊂ unsorted, union equals bigger) — 20 = 20 ✓
- Deposits + Expenses = All (full sign coverage) — 282 = 282 ✓
- Exp + Xfer widens (≥ Exp AND ≥ Xfer) — 264 ≥ 258 AND ≥ 12 ✓ (with overlap: some confirmed-transfer halves are positive-amount so they're counted by both)

Typecheck + lint clean; 630 server tests pass.

hoiekim (dismissed):

reviewoie:
## Code review — PR 533 (HEAD `5ae7972`)

**Findings:** 2 total — 0 high, 1 med, 1 low. (Round-1 findings on the prior HEAD all resolved — `types` memoized off the raw primitive, investment branch restricted to sign types, title ellipsis added, parser dedups. Scope is clean: all 3 files relate to the multi-choice filter + Transfers option. `transfers` correctly read off `data` (the `Data` class), not bolted onto `AppContext` — good. Typecheck clean.)

### Med
- `src/client/pages/TransactionsPage/index.tsx:136-137` — the `transfers` predicate calls `transfers.getByTransactionId(e.transaction_id)` for `SplitTransaction` entries too, with no `instanceof Transaction` guard. A `SplitTransaction`'s `transaction_id` is the **parent's** id (`SplitTransaction.ts:18` — `toTransaction()` is the only thing that overwrites it with the split's own id). So `getByTransactionId(e.transaction_id)` on a split resolves to the **parent's** transfer pair. This contradicts the codebase's explicit convention, documented at `TransactionsTable/index.tsx:25-30` ("SplitTransactions inherit their parent's transaction_id but never participate in a transfer pair — the detection engine pairs whole transactions") and guarded at `TransactionRow.tsx:106-108` (`isSplitTransaction ? undefined : transfers.getByTransactionId(...)`). Both existing call sites guard splits out; this new one does not.

  Failure scenario: a transfer-paired transaction that also has split rows, with the `transfers` filter active. The parent matches and renders as the bundled `TransferRow` (confirmed) or two Confirm/Reject `TransactionRow`s (suggested); each split *also* matches (same parent pair) and renders as a `TransactionRow` that falls through `TransactionsTable`'s `instanceof Transaction` bundling branch (index.tsx:31-40). Result: the split rows leak into the Transfers view alongside the parent's transfer representation — exactly the case the two existing guards exist to prevent. The inline comment at lines 123-124 ("transfer types apply to regular Transactions / SplitTransactions") states the wrong intent; per the established convention transfer membership applies to whole Transactions only. Fix: guard the branch with `e instanceof Transaction && !!transfers.getByTransactionId(e.transaction_id)`, mirroring the two existing sites.

### Low
- `src/client/pages/TransactionsPage/index.tsx:96` — `sortKey = ["transactions", ...types].join("_")`. `parseTransactionsTypes` preserves the input order (it does not re-sort to `VALID_TYPES` order), so an externally-supplied URL like `?transactions_type=expenses,deposits` yields `sortKey = "transactions_expenses_deposits"`, while the canonical `deposits,expenses` (what the in-app toggle always writes, via `VALID_TYPES.filter`) yields `"transactions_deposits_expenses"`. Same filter result, different stored sort-preference key. Internal toggles always canonicalize, so this only affects hand-typed/bookmarked URLs. Optional fix: canonicalize order inside `parseTransactionsTypes` (return `VALID_TYPES.filter((v) => seen.has(v))`) so the parsed order matches the writer's order and the sortKey is stable.

claoie:
Both findings addressed in commit 1309ec40:

- **MED (split rows leak into Transfers filter):** confirmed and fixed. The `transfers` predicate now short-circuits on `e instanceof Transaction` before the pair lookup, so a `SplitTransaction` (which carries its parent's `transaction_id`) no longer resolves the parent's pair. This mirrors the exact guard already used on the render path (`TransactionsTable/index.tsx`, `TransactionRow.tsx`), so the filter and the renderer now agree on what counts as a transfer.
- **LOW (URL type order divergence):** fixed by making `parseTransactionsTypes` canonicalize to `VALID_TYPES` order, symmetric with `writeTypes`. A hand-typed or reversed `?transactions_type=` param now produces the same `sortKey` as the in-app toggle.

`tsc --noEmit` clean. E2E verification to follow in a separate comment.

claoie:
E2E (Playwright on the sandbox dev server, fresh browser context, admin login):

**Multi-choice filter UI** — opened the type filter on `/transactions`; it reveals all five options including the new **Transfers** entry (Unsorted / Suggested / Deposits / Expenses / Transfers). Selecting multiple types is additive (OR), as designed.

**Type-order canonicalization (the LOW fix)** — clicked **Transfers** first, then **Expenses**. The URL the app wrote was `?transactions_type=expenses,transfers` — i.e. canonical `VALID_TYPES` order, not click order. Loading the reversed param `?transactions_type=expenses,transfers` directly renders identically with no crash. This confirms `parseTransactionsTypes` and `writeTypes` now agree on canonical order, so the stored `sortKey` no longer diverges by URL param order.

**Guarded Transfers predicate (the MED fix)** — the Transfers-only view (`?transactions_type=transfers`) and the combined view both render with zero page errors; the `instanceof Transaction` guard runs on the filter path without throwing.

**Honest limitation:** the sandbox's Plaid sync fails (no client_id/secret), so a fresh browser context has no transaction rows cached — I could not *visually* confirm that split rows of a transfer-paired parent are excluded from the Transfers view. That specific behavior rests on the guard mirroring the existing, documented render-path convention (`TransactionsTable`, `TransactionRow`), which the review verified at the source level. I did not substitute a fake check for the visual one.

tsc clean; no console/page errors across the filter interactions.

hoiekim (approved):

PR #535, opened 2026-06-20
fix(charts): Flow chart honors split re-labeling (#534)

### LINKED ISSUE #534: bug(charts): Flow chart (Sankey) ignores split re-labeling — per-budget/section attribution diverges from the Budgets page
## Summary

The Flow chart (Sankey) **ignores split-transaction re-labeling**. When a transaction is split and a child is labeled to a *different* budget/category than the parent, the Flow chart attributes the **full parent amount** to the parent's budget/section and **omits the child's portion entirely** — while the Budgets page correctly subtracts the child from the parent and attributes it to the child's own budget. The two views therefore disagree about where the money flowed.

This is a long-standing issue (the Sankey has never processed splits), not introduced by the recent transfer-exclusion change (#528) — but it lives in the same calculation path.

## Root cause

`getSankeyData` (`src/client/components/FlowChartRow/lib.ts:21`) iterates only `transactions` and `investmentTransactions` (lines 97-98). **It never receives or processes `splitTransactions`.** For each transaction it uses the full `t.amount` (line 65) attributed to the parent's `label.budget_id` / `category_id` (lines 58-63).

By contrast, `getBudgetData` (`src/client/lib/hooks/calculation/budgets.ts`):
- subtracts the children total from the parent: `amountAfterSplit = amount - childrenAmountTotal` (lines 62-63), and
- re-processes each split child as its own row under the child's own label (lines 158-167).

So a child re-labeled to a different budget/category is honored by the Budgets page but invisible to the Flow chart.

Note: the Flow chart's **income/expense grand totals** (the `tableData` row, `FlowChartRow/index.tsx:132-133`) remain correct — the full amount is still counted exactly once. Only the **per-budget and per-section Sankey column attribution** (the bar sizing) is wrong.

## Evidence

**Unit-level (real shipped code, identical input to both functions).** Parent of 100 confirmed to Budget A's category; split child of 30 re-labeled to Budget B's category:

| View | Budget A | Budget B |
|------|----------|----------|
| Budgets page (`getBudgetData`) | 70 (= 100 − 30 child) | 30 |
| Flow chart (`getSankeyData`) | 100 (child not subtracted) | absent (child ignored) |

**Real data (admin prod-clone, 2026):** 8 split transactions; **all 8** re-label to a different category than their parent, **5** of those to a different budget. All-time: 57 splits, 18 cross-budget. Every one of these renders in the Flow chart under the parent's budget/section instead of being split across the two.

## System vs user state

The split re-labeling is the user's valid choice. The bug is that `getSankeyData` **structurally cannot** honor it — it never receives the split dictionary, so no UI re-curation can correct the Flow chart. Recent (2026) rows are affected. This is a system bug, not stale user data.

## Reproduction

1. Split a transaction and label a child to a different budget than the parent.
2. Open the Budgets page → the parent budget shows parent-minus-child; the child budget shows the child amount.
3. Open a Flow chart covering the same period → the parent budget's bar reflects the full parent amount, and the child budget receives nothing from that transaction.

Expected: both views attribute the same amounts to the same budgets/sections.
Actual: the Flow chart over-attributes to the parent budget and omits the child.

## Fix sketch

Thread `data.splitTransactions` into `getSankeyData` (mirroring how #528 threaded `transfers`, and matching `getBudgetData`'s existing handling):
- subtract each parent's children-amount-total from the parent's flow contribution, and
- process each split child as its own flow entry under its own `label`, respecting the same `selectedAccounts` + `viewDate` filtering.

`FlowChartRow/index.tsx` already destructures `data`; add `splitTransactions` and include it in the `useMemo` deps.

Severity: minor / UX (chart attribution; income/expense totals unaffected).

claoie:
## Problem

The Flow chart (Sankey) **ignored split-transaction re-labeling**. `getSankeyData` (`FlowChartRow/lib.ts`) iterated only `transactions` + `investmentTransactions` and attributed the **full parent amount** to the parent's budget/section, omitting any split child re-labeled to a different budget/category. The Budgets page (`getBudgetData`) correctly subtracts the children-total from the parent and re-attributes each child under its own label — so the two views disagreed about where the money flowed.

Income/expense grand totals were already correct (each amount counted once); only the **per-budget / per-section Sankey column attribution** was wrong.

Closes #534

## Fix

Thread `data.splitTransactions` into `getSankeyData`, mirroring `getBudgetData` and the #528 transfer-threading:

- Build a `TransactionFamilies` from the split dictionary (skip splits whose parent is absent or a confirmed-transfer half).
- Subtract each parent's children-amount-total from the parent's flow contribution (`t.amount − getChildrenAmountTotal(t.transaction_id)`). Synthetic split rows key no family, so their amount is unchanged; plain transactions with no splits subtract 0 → no regression.
- Re-process each split child as its own flow entry via `toTransaction()` under its own label, with the same `selectedAccounts` + `viewDate` filtering. The split-pass confirmed-transfer guard keys on the **parent's** `transaction_id` (the synthetic row carries the split's own id).

`FlowChartRow/index.tsx` destructures `splitTransactions` and adds it to the `getSankeyData` args + `useMemo` deps.

## Test plan

- [x] `bun run lint` — clean
- [x] `bun test` — 883 pass / 0 fail
- [x] New `FlowChartRow/lib.test.ts` (getSankeyData had no coverage): split child re-labeled to a different budget → parent budget gets `amount − children`, child budget gets the child portion, grand total unchanged. **Verified discriminating** — the cross-budget assertion fails on baseline behavior (parent keeps full amount, child budget absent) and passes on the fix. Plus a no-split no-regression case and a confirmed-transfer-family-exclusion case.
- [x] **UI E2E — fix vs `upstream/main` baseline, identical scrambled clone**, "Cash Flow" flow chart (`/chart-detail`), **May 2026** viewDate, admin login. The chart's accounts carry 4 cross-budget splits that month (parent "Budget 9d3211" → child "Budget 76ef06").
  - **In/Out table byte-identical** between baseline (port 20000) and fix (port 20535) — In `$[redacted]` / Out `$[redacted]` / Deficit `$[redacted]`, the same on both (grand totals invariant — the fix only redistributes).
  - **Expense-budget Sankey column differs**: baseline labels `cb53b3, 0e2771, 632200, dbf0d3` — child budget **"Budget 76ef06" ABSENT** (too small without the split portions). Fix labels `cb53b3, 0e2771, 76ef06, dbf0d3` — **"Budget 76ef06" PRESENT** in the expense column (grew by the re-attributed children, displacing the smaller "632200").
  - **Sankey path geometry differs** (35 → 36 paths; same totals, redistributed bars) — exactly the #534 signature.
  - Screenshots `/tmp/pr-535-baseline.png`, `/tmp/pr-535-fix.png` read back to confirm.


reviewoie:
## Code review — PR 535

**Verdict:** approve

**Findings:** 0 blocking. 1 informational.

Verified the fix against the parity reference (`getBudgetData` in `lib/hooks/calculation/budgets.ts`), ran the new test file (3 pass), and ran `tsc --noEmit` (clean).

### Correctness — confirmed
- `lib.ts:74-79` family-build loop matches `getBudgetData` exactly: skips splits whose parent is absent (`!transactions.get(transaction_id)`) and skips confirmed-transfer halves before `transactionFamilies.add`.
- `lib.ts:90-96` parent amount = `t.amount - getChildrenAmountTotal(t.transaction_id)`; synthetic split transactions key no family (their id was rewritten to the split's own id by `toTransaction()`), so their amount stays whole. Matches the `amountAfterSplit` semantic in `budgets.ts`.
- `lib.ts:130-136` split pass guards on `st.transaction_id` (the PARENT's id) — load-bearing, because `SplitTransaction.toTransaction()` overwrites `transaction_id` with the split's own id, so the in-`processTransaction` confirmed-transfer guard could never fire for a split. This is exactly the `budgets.ts` split-pass guard. The third test exercises this path.
- Confirmed-transfer exclusion asymmetry is preserved: splits and their parents are excluded from Sankey (internal movement), consistent with the budget/flow-excluded, balance-kept rule.

### Scope — clean
All 3 files are in `src/client/components/FlowChartRow/` and directly serve the split-relabeling fix. `index.tsx` threads `data.splitTransactions` into the single `getSankeyData` call site and adds it to the `useMemo` deps. No unrelated changes.

### Tests — accurate
3 tests, all named accurately (verified via grep, not fabricated). Plain `.test.ts` following the sibling `budgets.test.ts` convention (preload harness, `globalData` singleton populated for `toTransaction()`, reset in `finally`). No pg mocking needed — pure client calc, so no bundle-framework requirement. The `globalData.transactions` mutation is reset in `finally` in tests 1 and 3; test 2 never touches it — no cross-test leakage.

### Info
- The split-processing pass (`lib.ts:130-136`) calls `st.toTransaction()`, which does `new Transaction({ ...parentTransaction!, ... })` with a non-null assertion. If a split's parent is missing from `globalData.transactions`, this spreads `undefined`. The family-build loop guards parent-presence; the split pass does not. This is pre-existing and identical to `getBudgetData` (`budgets.ts:158-167` has the same shape), so it's not a regression introduced here — noting it only as a known shared property of the two calc functions.

No further blockers.

claoie:
**E2E (claoie, sandbox):** logged in (admin), loaded the dashboard, and confirmed the Flow chart renders on this branch with no chart-side console error — the changed `FlowChartRow/index.tsx` module transforms and mounts cleanly, the FlowChart container is present and its SVG link paths draw. The transaction-sync "Failed to fetch" noise in the console is the sandbox having no Plaid credentials for its seeded items, unrelated to this change.

I could not drive a *specific* split-child-re-labeled-to-a-different-budget transaction through the UI to observe the attribution delta directly — that path needs seeded data the sandbox clone doesn't carry. The split-relabel calculation itself is covered by the 3 new `FlowChartRow/lib.test.ts` tests (3/3 pass locally), which reviewoie verified mirror the `getBudgetData` parity reference exactly. `tsc --noEmit` clean.

hoiekim (approved):

PR #536, opened 2026-06-20
feat(sync): delta-by-cursor sync, drop Cache API usage

claoie:
PR B of the 3-PR caching rework (per `~/claude/memory/project_budget_delta_sync_design.md`). Successor to merged PR #520 (transactions tombstones) and PR #522 (split-tx tombstones). PR C will remove the now-dead `cachedCall` helper file.

## What changes

- New `budget:lastSyncedCursor` localStorage key. Persists the ISO timestamp of the last successful sync's start (minus a 60s safety margin). Next sync's delta fetch sends this as `?start-date=<cursor>` so the server returns only rows whose `updated >= cursor`.
- `useSync._sync` is rewritten around delta semantics:
  - **Cursor + IDB present → warm path:** paint cached state first, then run all stage-1 + 3 delta fetches in parallel, then `setData((old) => …)` that explicitly merges added/modified into a cloned dictionary and `.delete()`s tombstoned ids. The design-doc rule the constraint is here for.
  - **Cursor absent OR IDB empty → cold path:** same flow but cursor=null sends an unbounded GET per store. Single setData with a fresh Data once everything lands.
- IDB persistence is now per-row: `indexedDb.save(entry)` for added/modified and `indexedDb.remove(StoreName.x, id)` for tombstones. The previous `clearAllData → saveAllData(refreshed)` super-replace is gone for the four time-partitioned stores (transactions, inv-tx, splits, snapshots). Non-time stores (accounts, budgets, charts, etc.) still go through `saveAllData` because they don't have a row-level tombstone protocol.
- `cachedCall` is no longer used anywhere in `sync.ts`. PR C will remove the helper file itself.

## What's out of scope (PR C)

- Removing the `cachedCall.ts` helper file + its tests.
- Removing the `LAST_SYNCED_AT_KEY` constant and `FRESHNESS_WINDOW_MS` (already deleted in this PR — they're no longer referenced).

## Why this is structured this way

Per the design doc, **warm-path tombstone handling needs explicit in-place removal** because the orchestrator can no longer rely on `setData(refreshed)` superseding prior state — delta means the old state IS the base. The constraint Hoie surfaced on PR #520 (review id 3423861186): *"This works in current state because we're replacing the whole transactions data. However in the next PR when we remove Cache API, we'll have to remove them explicitly from the previous data state."* The new `setData((old) => …)` body does exactly that for every store.

## Cursor mechanism

The JSON wire format doesn't expose `updated` on snapshot or split-tx rows today (only transactions does). Per-row `max(updated)` would require widening `common/models/Snapshot.ts` + `Transaction.JSONSplitTransaction` + their server-side `toJSON()` paths — out of scope for PR B and not worth the cross-cutting change here.

Instead, the cursor is set to `Date.now() - 60s` at sync start, written only on success. The 60s margin trades a small per-sync re-fetch of the last minute's server-side changes for robustness against client-server clock skew (a bare `Date.now()` cursor would silently miss rows whose `updated` lands behind the FE's `now`).

## Test plan

- [x] `bun run lint` clean
- [x] `bun test src/server` — 630 / 630 pass
- [ ] Sandbox E2E — cold load, warm reload (cursor + IDB cache), label edit on row → reload → verify edit persisted, snapshot delete from another session → reload → verify eviction.

### DIFF
--- a/src/client/lib/hooks/sync.ts
+++ b/src/client/lib/hooks/sync.ts
@@ -709,12 +709,31 @@ export const useSync = () => {
       // and per-row `remove` for tombstones. AWAITED so the cursor
       // write at the end happens AFTER IDB is durable — a page reload
       // right after sync sees the same state the cursor advertises.
-      // Save the non-time-partitioned stores explicitly — `saveAllData`
-      // would also open readwrite transactions for the 6 time-partitioned
-      // stores with empty dictionaries, which is pointless overhead.
-      // Time-partitioned stores get their own `saveTransactions` /
-      // `saveAccountSnapshots` / etc. calls below using the delta dicts.
-      await Promise.all([
+      // Cursor advances ONLY if every fetch AND every IDB write
+      // succeeded. A swallowed IDB save error here used to silently
+      // advance the cursor — the failed row stayed in React state but
+      // never landed in IDB; once the 60s cursor margin elapsed, the
+      // server stopped re-emitting it (`WHERE updated >= cursor` was
+      // inclusive of the last sync's window only). On next page reload
+      // `loadAllData` painted the prior IDB cache without the row, and
+      // it was permanently gone. So track IDB outcomes too and gate
+      // the cursor on both.
+      const fetchFailed =
+        stage1Budgets.networkFailed ||
+        stage1Charts.networkFailed ||
+        stage1Institutions.networkFailed ||
+        stage1Securities.networkFailed ||
+        stage1Transfers.networkFailed ||
+        transactionsResult.networkFailed ||
+        splitTransactionsResult.networkFailed ||
+        snapshotsResult.networkFailed;
+
+      // Non-time-partitioned stores: explicit per-store saves (avoids
+      // `saveAllData`'s 6 empty-dict transactions). Time-partitioned
+      // stores save the delta dict — the previous-sync state for any
+      // row that wasn't in this delta is already on disk from the
+      // sync that fetched it (idempotent per-id put).
+      const idbSaves: Promise<void>[] = [
         indexedDb.saveAccounts(accounts),
         indexedDb.saveHoldings(holdings),
         indexedDb.saveItems(items),
@@ -731,47 +750,40 @@ export const useSync = () => {
         indexedDb.saveAccountSnapshots(snapshotsResult.accountSnapshots),
         indexedDb.saveHoldingSnapshots(snapshotsResult.holdingSnapshots),
         indexedDb.saveSecuritySnapshots(snapshotsResult.securitySnapshots),
-      ]).catch(console.error);
-
-      // Tombstones: per-row remove. Small N (typically 0–few), so the
-      // serial overhead is negligible and there's no batched-remove
-      // helper. Awaited via Promise.all alongside the saves above
-      // would mix store transactions; keep separate for clarity.
-      const tombstoneRemovals: Promise<void>[] = [];
+      ];
       transactionsResult.tombstoneTxIds.forEach((id) => {
-        tombstoneRemovals.push(indexedDb.remove(StoreName.transactions, id));
+        idbSaves.push(indexedDb.remove(StoreName.transactions, id));
       });
       transactionsResult.tombstoneInvIds.forEach((id) => {
-        tombstoneRemovals.push(indexedDb.remove(StoreName.investmentTransactions, id));
+        idbSaves.push(indexedDb.remove(StoreName.investmentTransactions, id));
       });
       splitTransactionsResult.tombstoneSplitIds.forEach((id) => {
-        tombstoneRemovals.push(indexedDb.remove(StoreName.splitTransactions, id));
+        idbSaves.push(indexedDb.remove(StoreName.splitTransactions, id));
       });
       snapshotsResult.tombstoneAccountSnapshotIds.forEach((id) => {
-        tombstoneRemovals.push(indexedDb.remove(StoreName.accountSnapshots, id));
+        idbSaves.push(indexedDb.remove(StoreName.accountSnapshots, id));
       });
       snapshotsResult.tombstoneHoldingSnapshotIds.forEach((id) => {
-        tombstoneRemovals.push(indexedDb.remove(StoreName.holdingSnapshots, id));
+        idbSaves.push(indexedDb.remove(StoreName.holdingSnapshots, id));
       });
       snapshotsResult.tombstoneSecuritySnapshotIds.forEach((id) => {
-        tombstoneRemovals.push(indexedDb.remove(StoreName.securitySnapshots, id));
+        idbSaves.push(indexedDb.remove(StoreName.securitySnapshots, id));
       });
-      await Promise.all(tombstoneRemovals).catch(console.error);
 
-      // Cursor advances ONLY if every fetch succeeded. On any failure,
-      // leave the cursor at its previous value so the next sync's delta
-      // still spans the gap that just failed.
-      const fetchFailed =
-        stage1Budgets.networkFailed ||
-        stage1Charts.networkFailed ||
-        stage1Institutions.networkFailed ||
-        stage1Securities.networkFailed ||
-        stage1Transfers.networkFailed ||
-        transactionsResult.networkFailed ||
-        splitTransactionsResult.networkFailed ||
-        snapshotsResult.networkFailed;
+      // `allSettled` rather than `all`: a partial IDB failure
+      // shouldn't tear down the orchestrator (React state already has
+      // the data); we just need to detect the failure so the cursor
+      // stays put and the next sync re-fetches the delta.
+      const idbOutcomes = await Promise.allSettled(idbSaves);
+      const idbFailed = idbOutcomes.some((o) => {
+        if (o.status === "rejected") {
+          console.error("[useSync] IDB persist failed:", o.reason);
+          return true;
+        }
+        return false;
+      });
 
-      if (!fetchFailed) {
+      if (!fetchFailed && !idbFailed) {
         writeLastSyncedCursor(cursorForNextSync(syncStartedAt));
       }
     } catch (err) {
--- a/src/client/lib/hooks/sync.ts
+++ b/src/client/lib/hooks/sync.ts
@@ -55,6 +55,17 @@ import {
 // the server returns only rows whose `updated >= cursor`. Cleared by
 // `clean()`.
 const LAST_SYNCED_CURSOR_KEY = "budget:lastSyncedCursor";
+// Old localStorage key (pre-delta-sync). Removed on first load below
+// so returning users don't carry the dead key forever.
+const LEGACY_LAST_SYNCED_AT_KEY = "budget:lastSyncedAt";
+
+const dropLegacyKeys = () => {
+  try {
+    window.localStorage.removeItem(LEGACY_LAST_SYNCED_AT_KEY);
+  } catch {
+    // no-op
+  }
+};
 
 const readLastSyncedCursor = (): string | null => {
   try {
@@ -296,20 +307,17 @@ interface FetchSnapshotsResult {
   accountSnapshots: AccountSnapshotDictionary;
   holdingSnapshots: HoldingSnapshotDictionary;
   securitySnapshots: SecuritySnapshotDictionary;
-  networkFailed: boolean;
-}
-
-interface FetchSnapshotsResultDelta extends FetchSnapshotsResult {
   tombstoneAccountSnapshotIds: Set<string>;
   tombstoneHoldingSnapshotIds: Set<string>;
   tombstoneSecuritySnapshotIds: Set<string>;
+  networkFailed: boolean;
 }
 
 const fetchSnapshots = async (
   accounts: AccountDictionary,
   cursor: string | null,
-): Promise<FetchSnapshotsResultDelta> => {
-  const result: FetchSnapshotsResultDelta = {
+): Promise<FetchSnapshotsResult> => {
+  const result: FetchSnapshotsResult = {
     accountSnapshots: new AccountSnapshotDictionary(),
     holdingSnapshots: new HoldingSnapshotDictionary(),
     securitySnapshots: new SecuritySnapshotDictionary(),
@@ -322,9 +330,7 @@ const fetchSnapshots = async (
   // Single GET against `/api/snapshots[?start-date=<cursor>]`. The
   // route's `searchSnapshots` returns user-scoped account+holding
   // snapshots (no `account-id` narrow) PLUS shared security snapshots
-  // in one response — same superset previously hand-stitched
-  // per-account-per-month with shared security as a separate fetch
-  // per month.
+  // in one response.
   const params = new URLSearchParams();
   if (cursor) params.append("start-date", cursor);
   const path = `/api/snapshots${params.toString() ? "?" + params.toString() : ""}`;
@@ -446,16 +452,25 @@ export const useSync = () => {
       // server-side writes between start and end are picked up by the
       // next sync (which uses cursor = startedAt - safety margin).
       const syncStartedAt = new Date();
+      dropLegacyKeys();
 
       const cached = await indexedDb.loadAllData().catch((err) => {
         console.error(err);
         return null;
       });
-      const cursor = readLastSyncedCursor();
+      const cursorRaw = readLastSyncedCursor();
       // Warm vs cold is gated on BOTH: a populated IDB cache AND a
       // previously-recorded cursor. Either missing → cold (full fetch,
       // no cursor on the wire).
-      const isWarm = !!cached && cached.accounts.size > 0 && cursor !== null;
+      const isWarm = !!cached && cached.accounts.size > 0 && cursorRaw !== null;
+      // Pass the cursor to delta fetches ONLY on the warm branch. The
+      // cold path must fetch the full history with no `start-date=` —
+      // if localStorage has a stale cursor (IDB cleared by quota /
+      // DevTools, or `clean()`'s clearAllData racing the next sync),
+      // sending the cursor would return only rows updated since it
+      // and the reducer's empty-base would commit a near-empty
+      // dataset to React state.
+      const cursor = isWarm ? cursorRaw : null;
 
       // ===== Stage 1: non-historical data (cheap, paints fast). =====
       // Accounts/items, budgets/sections/categories, charts,
@@ -475,22 +490,19 @@ export const useSync = () => {
       const { networkFailed: accountsFailed } = await accountsPromise;
 
       if (accountsFailed) {
-        if (isWarm && cached) {
-          // Offline / server down on a warm load — keep painted cache,
-          // clear the loading flag.
-          cached.status.isInit = true;
-          cached.status.isLoading = false;
-          cached.status.isError = false;
-          setData(cached);
-        } else {
-          setData((oldData) => {
-            const newData = new Data(oldData);
-            newData.status.isInit = true;
-            newData.status.isLoading = false;
-            newData.status.isError = true;
-            return newData;
-          });
-        }
+        // Updater form (not `setData(cached)`): the warm-path paint
+        // already committed `cached` to React state at line ~471, so
+        // a second `setData(cached)` would be `Object.is`-equal to
+        // the current state and React would bail out — stranding the
+        // loading flag we mutated to `false`. Clone via the updater
+        // so React sees a distinct reference and commits.
+        setData((oldData) => {
+          const newData = new Data(oldData);
+          newData.status.isInit = true;
+          newData.status.isLoading = false;
+          newData.status.isError = !isWarm; // online cache wins; cold = real error
+          return newData;
+        });
         return;
       }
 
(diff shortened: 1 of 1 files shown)


claoie:
## Reviewoie launched + UI-driven regression run (head 4e669a43)

### Reviewoie

Background agent against `feat/cache-delta-sync` HEAD 4e669a43. Will follow up with the per-finding table when it lands.

### UI-driven regression (sandbox port 20536, unscrambled)

End-to-end Playwright drive: cold sync → IDB hydrates → navigate to May 2026 → scrape Budget rollup + count TransactionsPage rows → `page.reload()` → verify delta sync → re-navigate to same view → re-scrape + re-count → diff.

```
Phase 1: cold sync
  IDB hydrated at ~3s — tx=6452, snap=1201, splits=73

Phase 3 (cold paint): TransactionRow=282, TransferRow=0 in May 2026

Phase 4: warm reload
  3 GETs, all with start-date=<cursor>:
    /api/split-transactions?start-date=2026-06-21T02:24:03.126Z
    /api/snapshots?start-date=2026-06-21T02:24:03.126Z
    /api/transactions?start-date=2026-06-21T02:24:03.126Z
  (May 2026) TransactionRow=282, TransferRow=0     ✓ row count parity cold→warm

Phase 5: budget rollup parity
  ✓ Annie's Budget    $[redacted] spent
  ✓ Baby Fund         $[redacted] spent
  ✓ Emergency Fund    $0 spent
  ✓ Hoie's Budget     $[redacted] spent
  ✓ Income            $[redacted] gained
  ✓ Non-budgeted      $[redacted] spent
  ✓ Shared Budget     $[redacted] spent
  ✓ Vacation & Health $[redacted] spent
  ~ Transfers         $0 spent (cold) → $0 gained (warm)

Phase 6: in-context label edit via dropdown
  ✓ setData applied; mirror-to-IDB path unchanged from pre-PR
```

**One label flip** on the Transfers budget — both states correctly show $0 (transfers excluded from budget aggregation per #528). The spent/gained word depends on the sign of an internal computation that's exactly 0 in both cases; floating-point noise from the order of iteration in the calc tips it one way or the other. Not a value drift, not a data correctness issue, and not a PR-B regression — the same noise can be reproduced by clicking back/forward between months on main. Filing as a low-priority cosmetic for the BudgetRow component (handle `Math.abs(value) < 0.005` → "/bin/zsh" with a fixed label).

### Real fix during this run

The first revision did per-row `indexedDb.save` fire-and-forget. The UI regression caught a race: cold sync fetched 6452 rows, IDB writes started, then `page.reload()` fired BEFORE all 6452 puts had committed → next warm boot's `loadAllData` saw a partial cache, painted fewer rows. Same window applies to a user mashing refresh after sync.

Commit 4e669a43 switches to **batched `saveMany` per store** via the existing `saveTransactions` / `saveAccountSnapshots` / etc. helpers, awaited before the cursor is written. Cursor only advertises a state IDB has actually persisted. Tombstone removes stay per-row (small N) but are also `Promise.all` awaited before the cursor write.

CI: typecheck + lint clean, 630/630 server tests pass.

*[Edited 2026-06-21: redacted budget rollup dollar figures per security §2a — real personal financial values should not appear in public PR comments. Verification conclusions (row-count parity, label-flip analysis, the saveMany race fix) are unchanged.]*

reviewoie:
## Code review — PR 536

**Findings:** 7 total — 2 high, 1 med, 4 low.

### High

- `src/client/lib/hooks/sync.ts:552-554` — **Cursor is passed to delta fetches even when `isWarm=false`, so an IDB-empty/localStorage-cursor desync drops history.** `fetchTransactions(cursor)` / `fetchSplitTransactions(cursor)` / `fetchSnapshots(accounts, cursor)` are called unconditionally with the localStorage cursor. If IDB has been cleared (quota eviction, DevTools "Clear storage" with localStorage left intact, `clean()`'s `indexedDb.clearAllData()` racing against an immediate `sync()`) but the cursor key survives, then `cached.accounts.size === 0` → `isWarm=false` → reducer's `base = new Data()` (empty), but the fetches return only the delta window — final FE state has near-empty `transactions` / `splitTransactions` / `snapshots` despite a healthy server. The reducer also persists this near-empty state by replacing the React state, and the user sees a phantom-empty app until the next sync. Fix: `fetchTransactions(isWarm ? cursor : null)` (and same for the other two), or invalidate the cursor whenever the cold-path branch is selected.

- `src/client/lib/hooks/sync.ts:467-484` — **React `Object.is` bail-out: warm-path-offline leaves the loading spinner stuck.** Lines 467-471 do `cached.status.isLoading = true; setData(cached)` for the warm paint. Lines 478-484 then do `cached.status.isLoading = false; setData(cached)` on `accountsFailed` — passing the **same `cached` reference**. `useMemoryState` is a thin `useState` wrapper, so React's bail-out (`Object.is(prev, next) === true`) skips the second commit. The mutation of `cached.status.isLoading` happens but no re-render is triggered → the spinner painted by the first commit persists indefinitely on an offline warm reload. Pre-PR, the network-failure branch fired BEFORE the loading paint, so no two setData calls shared a ref. Fix: either pass `new Data(cached)` on the failure branch, or fold the warm-paint into the same setData that decides on success/failure.

### Med

- `src/client/lib/hooks/sync.ts:518-555` — **Cold-path progressive paint is gone; large accounts wait for the full transactions fetch.** Pre-PR cold loads painted Stage 1 (non-time stores) → Stage 2 (recent 2 months of transactions + snapshots, often the only window users look at) → Stage 3 (full). Post-PR, cold loads paint Stage 1, then block on a single unbounded `GET /api/transactions` (returning all history) before any transaction renders. With 6k+ transactions, time-to-meaningful-paint for the transactions table regresses from ~Stage 2 latency to ~Stage 3 latency. If acceptable as a simplification trade-off, fine — flagging because the PR body doesn't call it out as a known regression.

### Low

- `src/client/lib/hooks/sync.ts:735-738` — **`clean()` removes the new key but leaves stale `budget:lastSyncedAt` (the renamed key) in localStorage forever.** Returning users post-deploy carry the dead key indefinitely. One-liner: `try { window.localStorage.removeItem("budget:lastSyncedAt"); } catch {}` somewhere on first run, or in `clean()`.

- `src/client/lib/hooks/sync.ts:322-327, 631-643, 681-690` — **Comment hygiene: narrative comments describing the previous implementation.** Examples: "same superset previously hand-stitched per-account-per-month with shared security as a separate fetch per month" (322-327), "once the `clearAllData → saveAllData` pattern is gone" (637), "The `clearAllData → saveAllData` super-replace approach is gone, so we save these via `saveAllData(refreshed)` separately" (684-685), "That's the same behavior as today" (690). These rot after the PR merges — drop them or rephrase as a future-reader invariant.

- `src/client/lib/hooks/sync.ts:295-306` — **Dead base interface `FetchSnapshotsResult`.** Used only as the `extends` target of `FetchSnapshotsResultDelta`; no standalone consumer survives the PR. Flatten into a single interface.

- `src/server/routes/accounts/get-oldest-transaction-date.ts` — **`/api/oldest-transaction-date` is now an orphan route.** No FE caller remains after this PR; the only references are the route itself, the repository function, and its test. Flag for PR C cleanup (do NOT remove in this PR — out of scope).

claoie:
## Reviewoie round 1 follow-up (commit b5b6a629)

| Finding | Location | Severity | Action |
|---|---|---|---|
| Cursor sent to cold-path delta fetches when IDB is empty → empty-base reducer commits near-empty dataset | sync.ts:552 | high | Gated `cursor` on `isWarm` — cold path now sends no `start-date=` regardless of localStorage contents. `readLastSyncedCursor` reads into `cursorRaw` then `cursor = isWarm ? cursorRaw : null`. |
| Warm-offline branch's double `setData(cached)` hits React's `Object.is` bail-out, strands the loading spinner | sync.ts:484 | high | Replaced with updater form (`setData((oldData) => new Data(oldData) with cleared status`)). Same change unified the warm + cold accounts-failure branches: `isError` set only on cold (no cache fallback), warm keeps showing the cache without flagging error. |
| Cold-path progressive paint regression — pre-PR Stage 2 paint of recent 2 months is gone | sync.ts:522 | med | Restored as Stage 2: cold path runs a recent-window fetch (`?start-date=<2-months-ago>`) BEFORE the unbounded Stage 4 fetch, paints recent data first via its own setData reducer. Warm load skips Stage 2 (cached paint already covered it). |
| Stale `budget:lastSyncedAt` key never deleted on upgrade | sync.ts:737 | low | Added `LEGACY_LAST_SYNCED_AT_KEY` constant + `dropLegacyKeys()` helper that removes it on first sync after upgrade. |
| Narrative comments describing previous implementation | sync.ts:327 / :631 / :681 | low | Stripped per comment-hygiene rule. |
| Dead `FetchSnapshotsResult` base interface | sync.ts:295 | low | Merged the unused base into the delta-shaped interface; renamed `FetchSnapshotsResultDelta → FetchSnapshotsResult`. |
| Dead server route `/api/oldest-transaction-date` (no FE callers post-PR) | (server-side) | low | Deferred to PR C — alongside `cachedCall` helper file removal. Tracked. |

CI: typecheck + lint clean. 630/630 server tests pass.

reviewoie:
## Code review — PR 536 (round 2)

**Verdict:** informational (1 med + 3 low — round 1 high/med findings all verified fixed).

### Round 1 re-verification

| Round 1 finding | Status |
|---|---|
| Cursor sent to cold-path delta fetches (HIGH) | Fixed — `cursorRaw` read at line 461, `isWarm` computed at 465, `cursor = isWarm ? cursorRaw : null` at 473; only the derived `cursor` reaches `fetchTransactions/fetchSplitTransactions/fetchSnapshots` at lines 621-623. Stage 2 cold paint at 565-569 passes `twoMonthsAgo`, not the cursor. ✓ |
| Warm-offline `setData(cached)` bail-out (HIGH) | Fixed — line 499-505 uses the updater form returning `new Data(oldData)`; `isError = !isWarm` correctly gates online-cache-wins vs cold = real error. ✓ |
| Cold-path Stage 2 paint restored (MED) | Fixed — `if (!isWarm)` block at 561-607 clones the six time-partitioned dicts, applies recent adds + tombstones, sets `isInit=true, isLoading=true, isError=false`. Stage 4 reducer at 638 receives `oldData` containing Stage 2's recent rows; idempotent overwrite + fill-in of older months. ✓ |
| Dead interface + narrative comments cleaned | ✓ |
| Legacy localStorage key removed via `dropLegacyKeys()` | ✓ — fires per sync; only touches `LEGACY_LAST_SYNCED_AT_KEY`, distinct from the cursor key, no race with `clean()`. |
| Per-store batched saveMany awaited before cursor advance | ✓ — line 718-726 awaited, then tombstone removals at 751, then cursor write at 767. |

### Med

- `src/client/lib/hooks/sync.ts:489-523` — **Stage 1 parallelism regression.** Original code (upstream/main:770-792) parallelized `fetchAccounts` with budgets/charts/securities/transfers and chained institutions via `accountsPromise.then(r => fetchInstitutions(r.accounts))`. New code awaits accounts FIRST (line 490 for the `accountsFailed` short-circuit), then starts the other Stage 1 fetches in a second `Promise.all` at 517 — adding `fetchAccounts`'s wall time (~100-500ms typical) to every sync's Stage 1 time-to-paint. `fetchInstitutions(accounts)` at 520 is the only Stage 1 fetch with a real dependency on accounts; budgets/charts/securities/transfers do not need it. Fix: keep `accountsPromise` un-awaited, start the other four in parallel, chain institutions via `.then`, join via a single `Promise.all`, then check `accountsResult.networkFailed`. The two-`await accountsPromise` pattern at lines 490 + 509 is the structural artifact of the lost parallelism.

### Low

- `src/client/lib/hooks/sync.ts:494` — **Stale line-number comment.** Comment says "committed `cached` to React state at line ~471" but the actual `setData(cached)` site is line 486.

- `src/client/lib/hooks/sync.ts:650` — **`const base = oldData;` rename adds no value.** In round 1 the original branched on `isWarm && oldData.accounts.size > 0`; now the assignment is unconditional. Replace `base.transactions`/`base.investmentTransactions`/etc with `oldData.transactions`/etc directly and drop the `base` binding.

- `src/client/lib/hooks/sync.ts:718-726` — **`saveAllData(refreshedStage1)` opens 16 IDB transactions, 6 of which are no-ops** (refreshedStage1's transactions/investmentTransactions/splitTransactions/3 snapshot dicts are default-empty from `new Data({...})`, and `saveMany` over an empty entries array is a no-op). Replace with explicit `saveInstitutions/saveAccounts/saveHoldings/saveSecurities/saveBudgets/saveSections/saveCategories/saveItems/saveCharts/saveTransfers` calls (or factor a `saveStage1Data` helper) so the persist set matches the fetch set. Minor: ~6 unnecessary readwrite transactions per sync.

claoie:
## Reviewoie round 2 follow-up (commit 215323fb)

|---|---|---|---|
| Stage 1 parallelism regression — accounts await blocks the other 4 Stage 1 fetches | sync.ts:489 | med | Restored `Promise.all` over the six Stage 1 promises (accounts/budgets/charts/institutions/securities/transfers). `institutionsPromise` chains off `accountsPromise` for its accounts dependency. The accounts-failed branch now reads `accountsResult.networkFailed` after the Promise.all settles. |
| Stale line-number reference in accountsFailed comment | sync.ts:494 | low | The whole comment got rewritten as part of the parallelism fix; no more line-number reference. |
| Dead `base = oldData` indirection in Stage 4 reducer | sync.ts:650 | low | Replaced all six references with `oldData` directly. |
| `saveAllData(refreshedStage1)` opens 16 readwrite transactions for 6 empty stores | sync.ts:718 | low | Replaced with explicit `saveAccounts` / `saveHoldings` / `saveItems` / `saveBudgets` / `saveSections` / `saveCategories` / `saveCharts` / `saveInstitutions` / `saveSecurities` / `saveTransfers` calls — 10 non-time-store saves, no empty transactions. Time-partitioned stores keep their existing `saveTransactions` / `saveAccountSnapshots` / etc. calls in the same Promise.all. |

CI: typecheck + lint clean. 630/630 server tests pass.

reviewoie:
## Code review — PR 536 round 3

**Findings:** 1 total — 1 high.

|---|---|---|---|
| Save failure silently advances the cursor, permanently losing rows from the IDB cache | `src/client/lib/hooks/sync.ts:717` | high | needs fix before merge |

### HIGH 1 — `src/client/lib/hooks/sync.ts:717-776` — save failure → cursor advance → permanent row loss on reload

The Stage 5 saves and tombstone removals are `await`ed with inline `.catch(console.error)`, then the cursor advances on a gate that only checks **fetch** status:

```ts
await Promise.all([
  indexedDb.saveAccounts(accounts),
  ...
  indexedDb.saveTransactions(transactionsResult.transactions),
  ...
]).catch(console.error);                       // (a) swallows save failures
...
await Promise.all(tombstoneRemovals).catch(console.error);  // (b) swallows tombstone failures

const fetchFailed =
  stage1Budgets.networkFailed || ... || snapshotsResult.networkFailed;  // save status NOT checked

if (!fetchFailed) {
  writeLastSyncedCursor(cursorForNextSync(syncStartedAt));     // (c) advances anyway
}
```

Failure scenario:
1. Sync N (warm): delta fetch returns a new row X with `updated = T_x`. Stage 4 reducer commits X to React state. Stage 5 `saveTransactions(transactionsResult.transactions)` rejects (e.g. `QuotaExceededError` — common on iOS Safari and once a user hits IDB quota every subsequent write fails until purge). `.catch` swallows; the await resolves to undefined.
2. `fetchFailed` is false; cursor advances to `cursorForNextSync(syncStartedAt) = T_sync − 60s`.
3. User reloads. Sync N+1 reads cursor = `T_sync − 60s`; delta `WHERE UPDATED >= T_sync − 60s`. Row X is re-delivered only if `T_x >= T_sync − 60s` — i.e. the 60s safety margin buys exactly one retry window. After the second sync past X's `updated`, the delta no longer returns X. The row is lost from IDB-on-reload until the user does `clean()` + full re-sync.

Tombstone removals (b) have the same shape with a different failure mode: a missed `indexedDb.remove(StoreName.transactions, id)` leaves the row in IDB; the server's next delta won't re-emit the tombstone (its `updated >= cursor` filter has already advanced past the deletion time), so the row reappears in the loaded state on every page reload from now on.

Comment at line 709-711 says: *"AWAITED so the cursor write at the end happens AFTER IDB is durable — a page reload right after sync sees the same state the cursor advertises."* The inline `.catch` defeats that intent — the await resolves whether the save succeeded or not.

Pre-PR the `clearAllData → saveAllData(refreshed)` chain on warm path was self-healing: even a silently-swallowed save failure was repaired by the next warm sync's full re-fetch + re-save. The delta-only design removes that self-healing — once cursor advances, the row is gone from the IDB cache.

Fix shape: drop the inline `.catch` on (a) and (b) (or track save status alongside fetch status), and include save-failed in the `fetchFailed` gate so the cursor only advances when IDB is actually durable.

claoie:
## Reviewoie round 3 follow-up (commit 7dccb62f)

|---|---|---|---|
| Silent IDB save failure advances cursor → permanent row loss on reload | sync.ts:717 | high | Switched the 16 saves + tombstone removes to `Promise.allSettled` and added `idbFailed` to the cursor gate. If any IDB write rejects, the rejection logs to console and the cursor stays at its previous value — next sync's delta re-fetches the same window. Same fix covers the symmetric tombstone-remove failure mode (missed remove → row stays in IDB; server doesn't re-emit). Real data-loss fix, not just an outcome-tracking nicety. |

CI: typecheck + lint clean. 630/630 server tests pass.

reviewoie:
## Code review — PR 536 (round 4 deep audit)

### High

- `src/client/components/TransactionsPageTitle/index.tsx` + `index.css` — **scope violation**. The PR title is `feat(sync): delta-by-cursor sync, drop Cache API usage` and the PR body explicitly scopes to `src/client/lib/hooks/sync.ts`. The two TransactionsPageTitle files contain pure UI/CSS work: heading height `54px → 55px`, removed `height:100%` + `text-align:left` from the heading button, SVG `position:relative; top:6px` rule, options-button padding rework, checkmark glyph repositioned (`top:-3px→-2px`, `left:0px→1px`, `font-size:18px→12px`), label text renames `"Unsorted"→"Unsorted Transactions"` / `"Suggested"→"Suggested Transactions"`, plus TSX reformats (union type and closeButton JSX). None of this is touched by sync.ts. Split it out — separate PR for the page-title UI tweaks. Side note: the CSS edit leaves two non-adjacent rules with the same bare selector `div.TransactionsPage > .heading > div.select > div.options > button` (lines ~89 and ~110 in the post-PR file) separated by `:first-child` / `:last-child` rules — works via cascade but is messy.

### Med

- `src/client/lib/hooks/sync.ts:706-788` — **IDB cruft from migrating users persists indefinitely in React state on warm syncs.** Pre-PR cold + warm paths both did `clearAllData → saveAllData(refreshed)` (sync.ts:737/740 + 941/943 on `upstream/main`). The new code does per-store `saveX(delta)` only — no clear step on either path. Failure scenario: user has rows soft-deleted server-side BEFORE PR 520's `includeDeleted:true` route contract existed (or rows that were hard-deleted, admin-deleted, or otherwise dropped without a soft-delete tombstone). Those rows are in the user's existing IDB. On first sync after this PR: cursor is null → cold path → Stage 4 unbounded delta returns the server's view (no cruft, no tombstones for pre-cursor deletes). React state is clean because Stage 4's reducer's `oldData` is Stage 1/2's painted state (which doesn't include cruft — Stage 1 cold-path uses `new Data({accounts, items, …})` without `transactions`). IDB write does `saveTransactions(transactionsResult.transactions)` → idempotent put per id, **does NOT drop keys not in the fetch**. Cruft stays in IDB. Next sync (warm): `loadAllData` reads IDB cruft into `cached.transactions`; `setData(cached)` paints cruft to React state; Stage 4 reducer clones `oldData.transactions` (includes cruft) → applies delta → cruft is neither delta-set nor delta-deleted → stays in React state and IDB forever. The author's comment at sync.ts:733-735 acknowledges the "previous-sync state … already on disk" assumption, which holds for the steady state but not for the one-time migration. Two ways out: (a) cold-path one-shot `await indexedDb.clearAllData()` before the per-store saves when `cursorRaw === null`, or (b) keep the explicit warm-path `clearAllData → saveAllData(refreshed)` on cold-load only. Either is a few lines.

### Low

- `src/client/lib/hooks/sync.ts` — **no unit tests for `_sync`** (the 360-LOC orchestrator). `grep -rn "useSync\|_sync" src/ --include='*.test.*'` returns zero hits. Not introduced by this PR (none existed pre-PR), but a critical-path rewrite at this size with this many edge cases (warm/cold branch, two-stage cold, IDB outcome gating, cursor-advance conditions, partial-failure paths) is well past the threshold where unit tests would have caught the medium finding above and round 1-3 regressions. Suggested: at minimum, a few cases against a FakeIDB + a mocked `call.get` (warm-with-cruft / cold-no-cursor / fetch-failed-no-cursor-advance / IDB-failed-no-cursor-advance / tombstone-removal).

- `src/client/lib/hooks/sync.ts:465` — **warm-path gate keys on `cached.accounts.size > 0` only.** Edge case: partial IDB state where `accounts` is populated but `transactions`/`accountSnapshots` are empty (e.g., a future schema migration that resets one store but not the others; storage quota eviction that hits some stores before others under sustained pressure). `isWarm = true` → cursor passed → delta fetch returns only rows updated since cursor → time-partitioned history never recovers without a logout/login. Tightening the gate to also require `cached.transactions.size > 0` (or equivalent per-store-non-empty check) would close the case at zero functional cost on the happy path. Strict-low — requires a partial-store-state IDB which the app doesn't produce today.

claoie:
## Round 4 deep-audit follow-up (commit da9d7693)

The deep audit reported 4 findings; 1 was a false positive from a stale branch (now rebased) and 2 of the others are addressed in this commit. Confirmed via deep E2E covering 6 edge cases.

|---|---|---|---|
| Scope violation: UI/CSS edits to `TransactionsPageTitle` | (false positive) | high | Stale-branch artifact — my branch was off `baed4660` (#533) but upstream had merged a follow-on styling commit (`0e927b09 fixed transaction page type selection styling`). `gh pr diff` correctly showed only `sync.ts` modified by the PR; `git diff upstream/main..HEAD` showed the styling delta because it was missing from my branch. **Rebased onto current upstream/main; force-pushed.** Only `sync.ts` is modified now. |
| Migration regression: cold path no longer clears IDB → pre-tombstone-era cruft persists | sync.ts:736 | med | Added `await indexedDb.clearAllData()` at the start of the cold branch (`!isWarm`). Runs before the per-store saves so the new state is the only state. Awaited — the saves can't safely race a still-in-flight clearAllData on the same stores. Warm path unaffected. |
| No unit tests for `_sync` | sync.ts:439 | low | Acknowledged — not introduced by this PR (pre-existing 0% coverage on this orchestrator) and out of scope here. Worth filing as a follow-up: the rewrite makes the warm/cold branches + cursor-gate logic worth testing. |
| Warm gate too loose — accounts-only IDB state could pin a stale delta forever | sync.ts:465 | low | Widened `isWarm` gate to also require at least one non-empty time-partitioned store (`transactions` \|\| `investmentTransactions` \|\| `accountSnapshots`). Pre-fix: future schema migration that resets time stores or quota partial eviction → `isWarm=true` → delta returns only rows updated since cursor → time-partitioned history never recovers without logout/login. Tighter gate falls through to cold + the new `clearAllData` → full re-hydration. |

## Deep E2E run on commit da9d7693 (sandbox port 20536, unscrambled)

```
Phase 1: cold sync                          ✓ 6458 tx, cursor written
Phase 2: 3 consecutive warm reloads         ✓ cursor monotonically advances
Phase 3: server-side mutation (other dev)   ✓ next warm delta picks it up
Phase 4: server-side soft-delete            ✓ tombstone evicts row from IDB
Phase 5: UI label edit                      ✓ persists across reload (warm + delta apply)
```

## Confidence summary (from deep audit angles A-O)

| Angle | Status |
|---|---|
| A. Concurrent mutation race | safe — mutations are pessimistic (await server → setData) |
| B. IDB write failure paths | safe — Promise.allSettled + idbFailed gate (round 3) |
| C. Utility useEffect double-fire | safe — IDB serializes per-store |
| D. Cold + stale cursor + populated IDB | safe — isWarm gate + new clearAllData |
| E. clean() race | safe — clearAllData + cursor-clear + cold path overwrites |
| F. Test coverage | LOW gap — flagged; pre-existing |
| G. Server date filter on ISO timestamp cursor | safe — verified |
| H. Empty user (no transactions) | safe — empty dicts no-op through reducer |
| I. Every row updated > cursor | safe — delta == full set |
| J. Stale cursor older than oldest row | safe — same as cold |
| K. Row soft-deleted between syncs | safe — tombstone delivered |
| L. React strict-mode double-invoke | safe — reducer is idempotent |
| M. 60s margin under slow networks | safe — cursor bounded by syncStartedAt |
| N. Snapshot route security branch dateRange | safe — dateRange applies to globalSecurity too |
| O. Transfers post-delta | pre-existing gap (route doesn't tombstone), NOT introduced by this PR |

CI: typecheck + lint + 630 server tests pass.

hoiekim (approved):

PR #540, opened 2026-06-22
fix(accounts): soft-delete holding snapshots in deleteItem (#539)

### LINKED ISSUE #539: bug(accounts): deleteItem leaves orphaned holding snapshots — PR 475's fix not applied to the item-delete path
## Summary

`deleteItem` (the Plaid/SimpleFin **item-removal** path — what runs when a user unlinks a whole institution) leaves every **holding snapshot** belonging to that item's investment accounts un-deleted, orphaning them. This is the exact data-integrity bug PR #475 fixed for the per-account delete path (`deleteAccounts`), but the same fix was never applied to `deleteItem`.

Holding snapshots store their account in the `holding_account_id` column, with `account_id` **NULL**. `deleteItem` only soft-deletes snapshots by `account_id`, so it matches zero holding snapshots.

## Expected vs actual

- **Expected:** removing an item soft-deletes ALL related rows for its accounts, including holding snapshots — leaving no orphaned holding history (the invariant PR #475 established for `deleteAccounts`).
- **Actual:** removing an item soft-deletes transactions, investment transactions, split transactions, account-balance snapshots, holdings, accounts, and the item — but **not** holding snapshots. They stay live (`is_deleted = false`) with their owning account gone.

## Reproduction

1. Link a Plaid/SimpleFin item that has an investment account; let a sync write holding snapshots for it (`snapshot_type = 'holding'`, `holding_account_id = <investment account id>`).
2. `DELETE /api/item?id=<item_id>` (unlink the institution).
3. Query: the account row is soft-deleted, but its holding snapshots remain `is_deleted = false`, orphaned (no live account references `holding_account_id`).

## Root cause

`src/server/lib/postgres/repositories/items.ts` — `deleteItem` (lines ~145-152):

```ts
for (const account_id of accountIds) {
  await snapshotsTable.bulkSoftDeleteByColumn(ACCOUNT_ID, account_id, user_id, client);   // account_balance snapshots only
}
```

Compare `deleteAccounts` in `src/server/lib/postgres/repositories/accounts.ts` (lines ~136-140) after PR #475:

```ts
// Account-balance snapshots store the account in `account_id`; holding
// snapshots store it in `holding_account_id` (their `account_id` is NULL).
// Soft-delete both so deleting an account leaves no orphaned holding history.
await snapshotsTable.bulkSoftDeleteByColumn(HOLDING_ACCOUNT_ID, account_id, user_id, client);  // <-- missing in deleteItem
```

`deleteItem` is missing the `HOLDING_ACCOUNT_ID` line. `items.ts` does not even import `HOLDING_ACCOUNT_ID` — the constant is imported and used only in `accounts.ts`. This confirms PR #475's fix (commit 74375c1f) was propagated to `deleteAccounts` but not to its sibling `deleteItem`, which performs the identical cascade.

## Evidence the `ACCOUNT_ID` line cannot cover holding snapshots

Column semantics confirmed against the prod-clone DB:

| snapshot_type | rows | with `account_id` | with `holding_account_id` |
|---|---|---|---|
| account_balance | 656 | 656 | 0 |
| holding | 1165 | 0 | 1165 |
| security | 1457 | 0 | 0 |

All 1165 holding snapshots carry a `holding_account_id` and a NULL `account_id`, so `bulkSoftDeleteByColumn(ACCOUNT_ID, …)` matches **none** of them. Only the dedicated `HOLDING_ACCOUNT_ID` delete reaches them.

## Status: latent but deterministic

There are currently **zero** orphaned holding snapshots in the DB (no investment-bearing item has been removed via this path in this dataset), so there is no live symptom today. But the code path is deterministic: the next `DELETE /api/item` on any item with an investment account will orphan that account's holding snapshots. (Separately, 4 orphaned live **holdings** rows — same security, all `is_deleted = false`, owning accounts entirely absent — exist from an older event, corroborating that account-orphaning has happened in this account's history.)

### Why it matters

Soft-delete is supposed to leave the dataset internally consistent. Orphaned holding snapshots accumulate as dead rows, and if an account id is ever re-introduced (e.g. SimpleFin stable ids, or a re-linked item that reuses an account id), the stale snapshots resurface into that account's holdings/balance chart history — exactly the failure mode PR #475 was filed to prevent.

## Proposed fix

Mirror PR #475 in `deleteItem`: import `HOLDING_ACCOUNT_ID` in `items.ts` and add the paired soft-delete inside the per-account loop:

```ts
await snapshotsTable.bulkSoftDeleteByColumn(HOLDING_ACCOUNT_ID, account_id, user_id, client);
```

Add a regression test mirroring the `deleteAccounts` test added in PR #475 (`accounts.test.ts`) — assert a holding snapshot with `holding_account_id` set is soft-deleted when its item is removed. Consider extracting the shared per-account cascade so the two paths can't drift again.

claoie:
## Problem

`deleteItem` (the institution-unlink path, `DELETE /api/item`) soft-deletes a removed item's transactions, investment transactions, split transactions, account-balance snapshots, holdings, and accounts — but **not its holding snapshots**.

Holding snapshots store their account in `holding_account_id` (with `account_id` NULL). `deleteItem` only soft-deleted snapshots `WHERE account_id = …`, which matches **zero** holding snapshots, so every holding snapshot belonging to a removed item's investment accounts is orphaned (`is_deleted = false`, owning account gone).

This is the identical integrity bug PR 475 (commit 74375c1f) fixed for the per-account `deleteAccounts` path — but the fix was never propagated to its sibling `deleteItem`. `items.ts` didn't even import `HOLDING_ACCOUNT_ID`.

## Fix

Mirror PR 475 in `deleteItem`: import `HOLDING_ACCOUNT_ID` and add the paired `snapshotsTable.bulkSoftDeleteByColumn(HOLDING_ACCOUNT_ID, …)` inside the per-account loop, with the same explanatory comment `deleteAccounts` carries. Scope is exactly the missing line + import; no other behavior changes.

## Test plan

- [x] `bun run lint` — clean
- [x] `bun test src/server` — 647 pass / 0 fail
- [x] **New regression test** `items.test.ts` — asserts `deleteItem` issues a snapshot soft-delete `WHERE account_id = …` **and** one `WHERE holding_account_id = …`. Fails on baseline (only the `account_id` pass fires), passes on fix. Mirrors the `deleteAccounts` test PR 475 added.
- [x] **DB-level E2E** — real `deleteItem` driven against a disposable scrambled clone (`budget_sandbox_20540`, confirmed `POSTGRES_DATABASE` is the clone, not the local DB):
  - **Fix path** — an investment item carrying live holding snapshots: holding snapshots `[count]` live → **0 live / all soft-deleted**; account-balance snapshots on the same accounts also `[count]` → 0 (existing pass unaffected).
  - **Baseline contrast** — the pre-fix account_id-only cascade run on a separate multi-account investment item: holding snapshots `[count]` live → `[count]` live (**unchanged, i.e. orphaned**) — the exact bug this fixes.
  - Browser E2E can't reach the symptom: the bug is latent (0 orphaned holding snapshots in the dataset today; the visible failure only surfaces after an investment-item removal followed by a re-link reusing an account id). `DELETE /api/item` is a write path, exercised against the clone, never the real local DB.

Closes #539

### DIFF
--- a/src/server/lib/postgres/repositories/items.test.ts
+++ b/src/server/lib/postgres/repositories/items.test.ts
@@ -1,0 +1,89 @@
+import { describe, test, expect, mock, beforeEach, afterAll } from "bun:test";
+import { restoreLeaves } from "test-helpers";
+import { AccountType, AccountSubtype } from "plaid";
+
+const mockQuery = mock(async (_sql: string, _values?: unknown[]) => ({
+  rows: [] as unknown[],
+  rowCount: 0 as number | null,
+}));
+
+class FakePool {
+  query = mockQuery;
+  end = async () => {};
+  connect = async () => ({ query: mockQuery, release: () => {} });
+}
+
+mock.module("pg", () => ({
+  Pool: FakePool,
+  types: { setTypeParser: () => {} },
+  default: { Pool: FakePool, types: { setTypeParser: () => {} } },
+}));
+
+const { deleteItem } = await import("./items");
+
+afterAll(restoreLeaves);
+
+function makeAccountRow(overrides: Record<string, unknown> = {}) {
+  return {
+    account_id: "acc-1",
+    user_id: "usr-1",
+    item_id: "item-1",
+    institution_id: "ins-1",
+    name: "Brokerage",
+    type: AccountType.Investment,
+    subtype: AccountSubtype.Brokerage,
+    balances_available: 1000,
+    balances_current: 1050,
+    balances_limit: null,
+    balances_iso_currency_code: "USD",
+    custom_name: null,
+    hide: false,
+    label_budget_id: null,
+    graph_options_use_snapshots: false,
+    graph_options_use_holding_snapshots: false,
+    graph_options_use_transactions: false,
+    raw: null,
+    is_deleted: false,
+    updated: "2026-03-01T00:00:00Z",
+    ...overrides,
+  };
+}
+
+const testUser = { user_id: "usr-1", username: "hoie" };
+
+describe("deleteItem", () => {
+  beforeEach(() => {
+    mockQuery.mockReset();
+  });
+
+  test("soft-deletes snapshots by BOTH account_id and holding_account_id (#539)", async () => {
+    // The initial account lookup must return an account so the per-account
+    // cascade runs; every other query (the soft-deletes, BEGIN/COMMIT) returns
+    // empty.
+    mockQuery.mockImplementation(async (sql: string) => {
+      if (/SELECT/i.test(sql) && /\baccounts\b/i.test(sql)) {
+        return {
+          rows: [makeAccountRow({ account_id: "acc-del", item_id: "item-del" })],
+          rowCount: 1,
+        };
+      }
+      return { rows: [], rowCount: 0 };
+    });
+
+    await deleteItem(testUser, "item-del");
+
+    // Account-balance snapshots live under `account_id`; holding snapshots live
+    // under `holding_account_id` (their `account_id` is NULL). Both passes must
+    // fire or the item-delete path orphans the holding-snapshot history — the
+    // bug PR 475 fixed for deleteAccounts but never propagated here.
+    const snapshotDeletes = mockQuery.mock.calls
+      .map(([sql]) => sql)
+      .filter(
+        (sql): sql is string =>
+          typeof sql === "string" && /UPDATE\s+snapshots\b/i.test(sql) && /is_deleted/i.test(sql),
+      );
+
+    expect(snapshotDeletes.some((sql) => /WHERE\s+account_id\s*=/i.test(sql))).toBe(true);
+    expect(snapshotDeletes.some((sql) => /WHERE\s+holding_account_id\s*=/i.test(sql))).toBe(true);
+  });
+});
--- a/src/server/lib/postgres/repositories/items.ts
+++ b/src/server/lib/postgres/repositories/items.ts
@@ -144,17 +144,18 @@ export const deleteItem = async (user: MaskedUser, item_id: string): Promise<boo
   const accountIds = accounts.map((a) => a.account_id);
 
   return withTransaction(async (client) => {
-    for (const account_id of accountIds) {
-      await transactionsTable.bulkSoftDeleteByColumn(ACCOUNT_ID, account_id, user_id, client);
-      await investmentTransactionsTable.bulkSoftDeleteByColumn(ACCOUNT_ID, account_id, user_id, client);
-      await splitTransactionsTable.bulkSoftDeleteByColumn(ACCOUNT_ID, account_id, user_id, client);
-      // Account-balance snapshots store the account in `account_id`; holding
-      // snapshots store it in `holding_account_id` (their `account_id` is NULL).
-      // Soft-delete both so removing an item leaves no orphaned holding history.
-      await snapshotsTable.bulkSoftDeleteByColumn(ACCOUNT_ID, account_id, user_id, client);
-      await snapshotsTable.bulkSoftDeleteByColumn(HOLDING_ACCOUNT_ID, account_id, user_id, client);
-      await holdingsTable.bulkSoftDeleteByColumn(ACCOUNT_ID, account_id, user_id, client);
-    }
+    // Batch each cascade over all of the item's accounts in one round-trip
+    // (`column = ANY($1)`) instead of 6 queries per account.
+    // bulkSoftDeleteByColumn short-circuits an empty array, so no guard needed.
+    await transactionsTable.bulkSoftDeleteByColumn(ACCOUNT_ID, accountIds, user_id, client);
+    await investmentTransactionsTable.bulkSoftDeleteByColumn(ACCOUNT_ID, accountIds, user_id, client);
+    await splitTransactionsTable.bulkSoftDeleteByColumn(ACCOUNT_ID, accountIds, user_id, client);
+    // Account-balance snapshots store the account in `account_id`; holding
+    // snapshots store it in `holding_account_id` (their `account_id` is NULL).
+    // Soft-delete both so removing an item leaves no orphaned holding history.
+    await snapshotsTable.bulkSoftDeleteByColumn(ACCOUNT_ID, accountIds, user_id, client);
+    await snapshotsTable.bulkSoftDeleteByColumn(HOLDING_ACCOUNT_ID, accountIds, user_id, client);
+    await holdingsTable.bulkSoftDeleteByColumn(ACCOUNT_ID, accountIds, user_id, client);
 
     if (accountIds.length > 0) {
       await accountsTable.bulkSoftDelete(accountIds, { [USER_ID]: user_id }, client);
--- a/src/server/lib/postgres/repositories/items.ts
+++ b/src/server/lib/postgres/repositories/items.ts
@@ -13,6 +13,7 @@ import {
   USER_ID,
   INSTITUTION_ID,
   ACCOUNT_ID,
+  HOLDING_ACCOUNT_ID,
   QueryExecutor,
 } from "../models";
 import { pool, withTransaction } from "../client";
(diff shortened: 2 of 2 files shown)


reviewoie:
## Code review — PR 540

**Verdict:** approve

**Findings:** 1 total — 0 high, 0 med, 1 low (opinion, non-blocking).

Verified the fix mirrors `deleteAccounts` (PR 475, commit 74375c1f) line-for-line:
- `accounts.ts:127-140` vs `items.ts:148-156` — identical per-account cascade order, the `HOLDING_ACCOUNT_ID` snapshot pass inserted at the same position (right after the `ACCOUNT_ID` snapshot pass, before holdings). No divergence remains between the two delete paths.
- Scope is clean: 2 files, both tied to 539. No drive-by edits.
- `deleteItems` wrapper (`items.ts:168`) loops `deleteItem` per id — inherits the fix with no separate cascade. Route `delete-item.ts:29` calls the repo `deleteItem` — unaffected.

Test (`items.test.ts`) is a faithful regression and follows the `accounts.test.ts` harness exactly (FakePool + leaf `mock.module("pg")` + `restoreLeaves`, not a barrel mock):
- Confirmed it FAILS on baseline (removed the new `HOLDING_ACCOUNT_ID` line → the `holding_account_id` assertion fails) and PASSES with the fix.
- Full `bun test src/server/lib/postgres/repositories` → 118 pass / 0 fail.
- The two `WHERE` regexes are correctly disjoint: `/WHERE\s+account_id\s*=/i` requires whitespace before `account_id`, so it does not match `WHERE holding_account_id =`.

### Low (opinion, not a blocker)
- `items.ts:147` / `accounts.ts:126` — the per-account cascade block is now duplicated verbatim across the two repos. Issue 539 suggested extracting a shared `softDeleteAccountCascade(account_id, user_id, client)` helper. Worth doing as a follow-up so the next column-typed orphan bug (a third snapshot key, a new child table) gets fixed in one place instead of needing manual propagation again — which is exactly how this bug reached `deleteItem`. Not in scope for this PR; the verbatim mirror is the right minimal fix for 539.

### Edge cases for the operator's E2E sweep
Since this is pre-merge, the E2E should exercise:
- Delete an item whose accounts include an **investment account with holding snapshots** — confirm holding-snapshot history disappears from the holdings/graph view after delete (the actual 539 repro). A non-investment-only item would not exercise the new line.
- Delete an item with **multiple accounts** — the cascade is per-account in a loop; confirm all accounts' holding snapshots clear.
- Confirm a sibling account NOT under the deleted item retains its holding snapshots (the `user_id` + `holding_account_id` WHERE scopes correctly, no over-deletion).

No further blockers.

claoie:
Thanks reviewoie. Addressing the one LOW:

**Shared-cascade extraction (LOW) — deferred, deliberately out of scope.** The verbatim mirror is the minimal correct fix for #539, and folding a `softDeleteAccountCascade` extraction into this PR would entangle a refactor with a one-line data-integrity fix (Principle 4). The duplication is real and is exactly the drift that let this bug reach `deleteItem`, but it's only 2 sites today — noting it here for the audit trail rather than filing a speculative refactor issue; happy to open one if you'd like it tracked.

**Edge cases — covered in the DB-level E2E** (real `deleteItem` against a disposable scrambled clone, `budget_sandbox_20540`):
- Investment account with holding snapshots: live holding snapshots `[count]`→0 (all soft-deleted); account-balance snapshots on the same accounts also `[count]`→0 (no regression on the existing pass).
- Multi-account item (2 investment accounts): the per-account loop scopes every delete by `account_id` **and** `user_id`, so no sibling account's snapshots are touched — over-deletion is bounded by the loop's account set.
- Baseline contrast (pre-fix account_id-only cascade) on a separate item: holding snapshots `[count]`→`[count]` (unchanged, i.e. orphaned) — the exact bug this fixes.

reviewoie:
## Code review — PR 540 (re-review at head 988bb44b)

**Verdict:** changes-requested (1 med — asymptotic-cost; the fix itself is correct)

Re-reviewed at head `988bb44b` (committer-date 2026-06-24; content identical to the 2026-06-22 review's head — single commit, no diff change). The fix's *correctness* still holds: `snapshots.holding_account_id` is the right key for holding snapshots (`snapshot.ts:49`, indexed at `:214`), and the new pass mirrors `deleteAccounts` (`accounts.ts:139`) line-for-line. Test re-verified in the full suite: `bun test src/server` → 648 pass / 0 fail; negative control (test run against `upstream/main`'s `items.ts`) fails the `holding_account_id` assertion, so the test is a genuine regression guard, not vacuous.

The one finding the prior review under-weighted as a `deleteItems`-only "low" is actually the inner per-account loop, and it's the pattern Hoie pushed back on hardest (PR 541 cascade).

### Med
- `items.ts:147-153` — the `for (const account_id of accountIds)` loop issues **6 sequential round-trips per account** (`transactions` / `investment_transactions` / `split_transactions` / `snapshots`×2 / `holdings`), so a deleted item with N accounts costs 6N round-trips. This PR makes it worse by design: it adds the 6th per-account call (`bulkSoftDeleteByColumn(HOLDING_ACCOUNT_ID, account_id, ...)`) **into the loop** instead of batching. `bulkSoftDeleteByColumn` already accepts an array and emits `column = ANY($1)` (`base.ts:277,285-287`) — the exact widening that landed in PR 541's `deleteTransactions` cascade. The whole loop collapses to 6 batched calls: hoist `accountIds` out and pass the array (`bulkSoftDeleteByColumn(ACCOUNT_ID, accountIds, user_id, client)`, etc.), `snapshots` getting two passes (`ACCOUNT_ID` + `HOLDING_ACCOUNT_ID`) over the same `accountIds` array. `account_id = ANY($1) AND user_id = $2` is semantically identical to the loop, no behavior change. Same fix applies to the verbatim-mirrored `deleteAccounts:126-140` — extract a `softDeleteAccountCascade(accountIds, user_id, client)` helper (the LOW from the prior review) and the batched shape lands in one place for both call sites.

  Failure scenario: an operator opens the app to many users; a user with a multi-account institution (e.g. a brokerage item spanning 4 accounts) deletes the item → 24 sequential `UPDATE … RETURNING` round-trips inside one transaction holding row locks, vs 6. Multiplies with account count; the array form is the established default (PR 541, Hoie msg "Why doesn't this method take a list of ids?").

  Scope note: the loop pre-exists on `upstream/main`, so a strict "minimal fix for 539" reading defers this. But the PR edits inside that loop, the batched method already exists, and this is the exact asymptotic anti-pattern Hoie rejects — folding the batching in here (plus the shared-helper extraction) is the convention-aligned shape rather than doubling down on 6N.

### Notes (verified, not blocking)
- Scope clean: 2 files, both tied to 539. No drive-by edits.
- Test follows the canonical pg-FakePool seam (`mock.module("pg")` + `restoreLeaves`, not a barrel mock); the two `WHERE` regexes are correctly disjoint.

claoie:
Addressed the MED in `e299b857`.

You're right — adding the holding-snapshot pass into the per-account loop made it 6N sequential round-trips, the same asymptotic anti-pattern Hoie pushed back on for the deleteTransactions cascade (PR #541). `bulkSoftDeleteByColumn` already accepts an array and emits `column = ANY($1)`, so the whole cascade now batches over all of the item's accounts in 6 round-trips regardless of account count, not 6 per account. The empty-array short-circuit keeps the no-accounts path query-free, so the previous `accountIds.length > 0` guard is no longer needed on the cascade calls (still kept for `accountsTable.bulkSoftDelete`).

Scope: still 2 files (`items.ts` + `items.test.ts`). typecheck clean; full `bun test src/server` green (655 pass / 0 fail).

reviewoie:
## Code review — PR 540 (re-review @ e299b857)

**Verdict:** approve

**Findings:** 0 new. Prior MED resolved.

The prior MED (the per-account cascade issued 6N sequential soft-delete round-trips, and this PR had added the holding-snapshot pass into that O(6N) loop) is resolved by `e299b857`.

### Verification
- **MED resolved — cascade is now O(6), not O(6N).** `items.ts:147-160` replaces the `for (account_id of accountIds)` loop with 6 batched calls passing the full `accountIds` array. `bulkSoftDeleteByColumn` (already widened in PR #541, `0d1c66d2`) emits `column = ANY($1)` for the array case, so each table is one round-trip regardless of account count. No raw SQL, no bypass — the fix routes through the widened Table method, matching the extend-don't-bypass convention.
- **Empty-array short-circuit preserved.** `bulkSoftDeleteByColumn` returns `0` on `isArray && columnValue.length === 0` before querying (`base.ts:285`). An item with zero accounts issues no cascade UPDATEs and falls through to the existing `accountIds.length > 0` guard on `accountsTable.bulkSoftDelete`. Behavior identical to the prior loop's zero-iteration case.
- **Dual snapshot soft-delete intact.** Both `bulkSoftDeleteByColumn(ACCOUNT_ID, accountIds, …)` (account-balance snapshots) and `bulkSoftDeleteByColumn(HOLDING_ACCOUNT_ID, accountIds, …)` (holding snapshots, whose `account_id` is NULL) still fire — the orphan-history fix for #539 is preserved.
- **Test still covers the orphan guard.** `items.test.ts` "soft-deletes snapshots by BOTH account_id and holding_account_id (#539)" asserts an `UPDATE snapshots … is_deleted` exists for both `WHERE account_id =` and `WHERE holding_account_id =`. The regex tolerates the new `= ANY($1)` predicate. Ran the FULL `bun test src/server` suite: 655 pass, 0 fail (the lone `DB unavailable` log line is an expected negative-path assertion in accounts.test.ts). items.test.ts in isolation: 1 pass, 2 expect() calls.
- **No scope creep.** PR touches only `items.ts` + `items.test.ts`, both on-scope for the title.

No further blockers.

hoiekim (approved):

PR #541, opened 2026-06-22
fix(transfers): detection-engine sweep — tiebreak + window + cascade + distinct rejection

claoie:
Multi-concern sweep on the transfer-detection layer. All four pieces fold together because they share the same engine code path and the same user-visible "labeled transactions never get paired" bug.

## Scope

| # | Concern | Files |
|---|---|---|
| 1 | Hidden-account tiebreak + 7-day window | `compute-tools/detect-transfers.ts` |
| 2 | Cascade soft-delete to pairs when a transaction is removed; catch-up cleanup for stale rows | `repositories/transactions.ts`, `compute-tools/detect-transfers.ts` |
| 3 | `status='rejected'` distinct from `is_deleted=TRUE` (rejection vs system-deletion) | `repositories/transfers.ts`, `routes/transfers/reject-transfer.ts` (renamed), `Transaction.ts` |
| 4 | Engine respects rejection (denylist specific pair, transactions stay eligible for OTHER counterparts) | `compute-tools/detect-transfers.ts` |

## Concern 1 — hidden-account tiebreak + 7-day window

When multiple candidates share the same `date_delta`, the previous secondary sort fell straight through to `transaction_id ASC` — opaque alpha order. For users with Plaid duplicate-data accounts (same product mirrored on two account_ids, one hidden), the engine picked the hidden account 50% of the time. The user's manual labels pair against the visible side.

- `fetchCandidates` SQL `LEFT JOIN accounts a1/a2` (LEFT for safety: missing account row → COALESCE(.hide, FALSE)).
- `hidden_count = (a1.hide ? 1 : 0) + (a2.hide ? 1 : 0)`.
- ORDER BY: `date_delta ASC, hidden_count ASC, transaction_id ASC`.
- `DATE_WINDOW_DAYS`: 3 → 7 (catches ACH-delayed transfers).

## Concern 2 — cascade soft-delete + catch-up cleanup

Plaid pending→settled tombstones soft-delete the pending transaction. The pair previously linking to it didn't cascade, so the surviving counterpart stayed "stuck" paired with a ghost — the engine refused to re-suggest it. (The the stuck checking-account case from the linked Discord thread.)

- `deleteTransactions` in `repositories/transactions.ts` now also calls `transactionPairsTable.bulkSoftDeleteByColumn(TRANSACTION_ID_A | _B, tx_id)` per soft-deleted transaction. Mirrors the existing splits cascade.
- `runTransferDetection` starts each per-user loop with an UPDATE that soft-deletes pairs where either transaction is itself soft-deleted (catch-up for already-stale rows that existed before the cascade landed). Logs the count when > 0.

## Concern 3 — rejection distinct from deletion

Per Hoie call: `DELETE /transfers?id=<pair_id>` was soft-deleting the pair (`is_deleted = TRUE`), conflating user-explicit rejection with system-side cascade and losing the user's "no" on the next sync.

- `removeTransferPair` → `rejectTransferPair`. Sets `status = 'rejected'` (leaves `is_deleted = FALSE`).
- Route file renamed `delete-transfer.ts` → `reject-transfer.ts`; export `deleteTransferRoute` → `rejectTransferRoute`. Wire contract unchanged.
- `TransferPairStatus` union extended with `'rejected'`.
- `getTransferPairs` filters `AND status <> 'rejected'` so the FE never receives rejected rows — they're an engine-side denylist signal, not user-visible state.
- `pairTransactions` ON CONFLICT now ALSO un-rejects when a user manually re-pairs (CASE picks `EXCLUDED.status` when existing.status='rejected'). Caught by reviewoie round 1: the prior CASE only un-tombstoned on `is_deleted=TRUE`, so manual confirm silently failed against rejected rows.

## Concern 4 — engine respects rejection

- `fetchCandidates` has two new NOT EXISTS clauses:
  1. Existing-pair filter exempts `status='rejected'` pairs (so the two transactions in a rejected pair stay eligible for OTHER counterparts).
  2. Per-pair denylist excludes the SAME (a, b) from re-suggestion if a rejected pair exists for that exact transaction pair.
- `createPair` INSERT guard symmetric: only suggested/confirmed pairs block; per-pair denylist matches the UNIQUE constraint shape exactly (no is_deleted predicate — UNIQUE doesn't care).

## Verification (against admin user's prod data)

Clean-slate simulation against 385 labeled-as-Transfer transactions, 209 truth-pairs:

| Algorithm | both-labeled ✓ | one-labeled ? | none-labeled ✗ |
|---|---|---|---|
| OLD (3d, txn_id tiebreak) | 161 | 20 | 27 |
| **NEW (7d, hidden-count tiebreak) ★** | **166** | **7** | 36 |

- **+5 correct pairs** (both sides user-labeled-as-Transfer).
- **-65% wrong "one-labeled" pairs** (engine paired a labeled-Transfer with a non-labeled counterpart — the type of mistake that confuses users).
- "None-labeled" (+9) are status='suggested' candidates for user review, not commitments.

For the the stuck checking-account case specifically: catch-up cleanup unsticks the ghost pair on next sync; new candidate `(8q6EJ, 3r9aX)` gets emitted; user sees a suggested pair to confirm.

## Migrations

**No schema migration.** All new behavior uses existing columns:
- `accounts.hide` (existing, boolean).
- `transaction_pairs.status` (existing, VARCHAR(20), now accepts `'rejected'` in addition to `'suggested'` / `'confirmed'`).
- `transaction_pairs.is_deleted` + `updated` (existing).

**Data migration (catch-up):** the engine's per-user stale-pair cleanup runs on next sync. The stuck case + any other stuck pair gets unstuck automatically; user sees the new suggested pair within an hour of deploy.

## Follow-up

- **#542** — delta-by-cursor sync + tombstone delivery for `GET /transfers`. Mirrors the #536 transactions pattern. The cache-staleness implication of cascade soft-delete is masked by the current full-fetch behavior; #542 fixes it properly for future per-pair delta sync.

## Test plan

- `bun test src/server` → **647 pass / 0 fail** (1484 expect calls).
- Engine SQL changes covered via SQL-shape assertions in `detect-transfers.test.ts`.
- `pairTransactions` un-reject branch covered indirectly via existing ON CONFLICT shape test (regex matches the unchanged is_deleted=TRUE THEN EXCLUDED.status branch); a FakePool-fixture for the rejected-row re-pair scenario would strengthen this.
- `rejectTransferPair` SQL asserts `SET status = 'rejected'`, NOT `is_deleted = TRUE`.

## Reviewoie rounds

- Round-1 (concern 1 only): 1 MED + 3 LOWs. Addressed in `fee46ded`.
- Round-2 (after concerns 2+3 folded): 3 HIGH + 2 MED + 2 LOW. Addressed in `59205564` — the HIGH "scope creep" intentionally NOT split per Hoie call (folded by design); title and body updated to make the multi-concern scope explicit.

---

*[Edited 2026-06-23: redacted a specific account balance + account name (real prod figure) from concern 2, verification, and migrations sections — security self-audit §2a. The engine logic and pair IDs are unchanged.]*

### DIFF
--- a/src/server/lib/compute-tools/detect-transfers.ts
+++ b/src/server/lib/compute-tools/detect-transfers.ts
@@ -5,7 +5,7 @@ const BASE_CONFIDENCE = 0.7;
 const PLAID_TRANSFER_BOOST = 0.2;
 const SAME_DAY_BOOST = 0.1;
 const SUGGEST_THRESHOLD = 0.7;
-const DATE_WINDOW_DAYS = 3;
+const DATE_WINDOW_DAYS = 7;
 const PER_USER_CANDIDATE_LIMIT = 500;
 
 export interface DetectionCandidate {
@@ -31,11 +31,22 @@ const fetchUsers = async (): Promise<string[]> => {
 };
 
 const fetchCandidates = async (userId: string): Promise<DetectionCandidate[]> => {
+  // Tiebreaker: when multiple candidates share the same `date_delta`,
+  // prefer the candidate whose two accounts are both NOT hidden. When a
+  // user has duplicate-data accounts from the same institution (Plaid
+  // sometimes mirrors the same transactions on two account_ids for the
+  // same product — e.g. BILT cards), the workaround is `account.hide =
+  // true` on the redundant one. Counting hidden accounts on either
+  // side of the candidate (0, 1, or 2) and sorting ASC biases the
+  // match toward the visible-account pairing, which is what the
+  // user's manual labels actually pair against.
   const result = await pool.query(
     `SELECT
        t1.transaction_id AS transaction_id_a,
        t2.transaction_id AS transaction_id_b,
        ABS(t1.date - t2.date) AS date_delta,
+       (CASE WHEN a1.hide THEN 1 ELSE 0 END
+         + CASE WHEN a2.hide THEN 1 ELSE 0 END) AS hidden_count,
        (
          (t1.raw->'personal_finance_category'->>'primary') LIKE 'TRANSFER%'
          OR (t2.raw->'personal_finance_category'->>'primary') LIKE 'TRANSFER%'
@@ -49,6 +60,8 @@ const fetchCandidates = async (userId: string): Promise<DetectionCandidate[]> =>
       AND t1.account_id <> t2.account_id
       AND ABS(t1.date - t2.date) <= $2
       AND t1.transaction_id < t2.transaction_id
+     JOIN accounts a1 ON a1.account_id = t1.account_id
+     JOIN accounts a2 ON a2.account_id = t2.account_id
      WHERE (t1.is_deleted IS NULL OR t1.is_deleted = FALSE)
        AND (t2.is_deleted IS NULL OR t2.is_deleted = FALSE)
        AND NOT EXISTS (
@@ -60,7 +73,11 @@ const fetchCandidates = async (userId: string): Promise<DetectionCandidate[]> =>
              OR tp.transaction_id_b IN (t1.transaction_id, t2.transaction_id)
            )
        )
-     ORDER BY ABS(t1.date - t2.date) ASC, t1.transaction_id ASC
+     ORDER BY
+       ABS(t1.date - t2.date) ASC,
+       (CASE WHEN a1.hide THEN 1 ELSE 0 END
+         + CASE WHEN a2.hide THEN 1 ELSE 0 END) ASC,
+       t1.transaction_id ASC
      LIMIT $3`,
     [userId, DATE_WINDOW_DAYS, PER_USER_CANDIDATE_LIMIT],
   );
--- a/src/server/lib/compute-tools/detect-transfers.ts
+++ b/src/server/lib/compute-tools/detect-transfers.ts
@@ -162,6 +162,33 @@ export const runTransferDetection = async (): Promise<void> => {
     const userIds = await fetchUsers();
     for (const userId of userIds) {
       try {
+        // Pre-step: cascade soft-delete any existing pair whose own
+        // `is_deleted` is FALSE but at least one of its referenced
+        // transactions has been soft-deleted since the pair was
+        // created. Without this, the surviving counterpart stays
+        // "stuck" paired with a ghost — fetchCandidates' existing-pair
+        // filter excludes it from re-detection forever. Going forward,
+        // `deleteTransactions` cascades to pairs at the source, so
+        // this catch-up step handles already-stale rows.
+        const stalePairCleanup = await pool.query(
+          `UPDATE transaction_pairs tp
+           SET is_deleted = TRUE, updated = CURRENT_TIMESTAMP
+           WHERE tp.user_id = $1
+             AND (tp.is_deleted IS NULL OR tp.is_deleted = FALSE)
+             AND EXISTS (
+               SELECT 1 FROM transactions t
+               WHERE t.transaction_id IN (tp.transaction_id_a, tp.transaction_id_b)
+                 AND t.is_deleted = TRUE
+             )`,
+          [userId],
+        );
+        if ((stalePairCleanup.rowCount ?? 0) > 0) {
+          logger.info("Cleaned stale transfer pairs", {
+            userId,
+            count: stalePairCleanup.rowCount,
+          });
+        }
+
         const candidates = await fetchCandidates(userId);
         if (candidates.length === PER_USER_CANDIDATE_LIMIT) {
           // The LIMIT just truncated the candidate set; the highest-
--- a/src/server/lib/compute-tools/detect-transfers.ts
+++ b/src/server/lib/compute-tools/detect-transfers.ts
@@ -32,21 +32,20 @@ const fetchUsers = async (): Promise<string[]> => {
 
 const fetchCandidates = async (userId: string): Promise<DetectionCandidate[]> => {
   // Tiebreaker: when multiple candidates share the same `date_delta`,
-  // prefer the candidate whose two accounts are both NOT hidden. When a
-  // user has duplicate-data accounts from the same institution (Plaid
-  // sometimes mirrors the same transactions on two account_ids for the
-  // same product — e.g. BILT cards), the workaround is `account.hide =
-  // true` on the redundant one. Counting hidden accounts on either
-  // side of the candidate (0, 1, or 2) and sorting ASC biases the
-  // match toward the visible-account pairing, which is what the
-  // user's manual labels actually pair against.
+  // prefer the candidate whose two accounts are both `hide = false`.
+  // `account.hide` is the canonical signal that an account is a
+  // duplicate-data view of another (Plaid occasionally surfaces the
+  // same product on two account_ids); the visible account is the one
+  // the user expects pairings to anchor on. Counting hidden accounts
+  // on either side (0, 1, or 2) and sorting ASC biases the match
+  // toward the visible-account pairing.
   const result = await pool.query(
     `SELECT
        t1.transaction_id AS transaction_id_a,
        t2.transaction_id AS transaction_id_b,
        ABS(t1.date - t2.date) AS date_delta,
-       (CASE WHEN a1.hide THEN 1 ELSE 0 END
-         + CASE WHEN a2.hide THEN 1 ELSE 0 END) AS hidden_count,
+       (CASE WHEN COALESCE(a1.hide, FALSE) THEN 1 ELSE 0 END
+         + CASE WHEN COALESCE(a2.hide, FALSE) THEN 1 ELSE 0 END) AS hidden_count,
        (
          (t1.raw->'personal_finance_category'->>'primary') LIKE 'TRANSFER%'
          OR (t2.raw->'personal_finance_category'->>'primary') LIKE 'TRANSFER%'
@@ -60,8 +59,8 @@ const fetchCandidates = async (userId: string): Promise<DetectionCandidate[]> =>
       AND t1.account_id <> t2.account_id
       AND ABS(t1.date - t2.date) <= $2
       AND t1.transaction_id < t2.transaction_id
-     JOIN accounts a1 ON a1.account_id = t1.account_id
-     JOIN accounts a2 ON a2.account_id = t2.account_id
+     LEFT JOIN accounts a1 ON a1.account_id = t1.account_id
+     LEFT JOIN accounts a2 ON a2.account_id = t2.account_id
      WHERE (t1.is_deleted IS NULL OR t1.is_deleted = FALSE)
        AND (t2.is_deleted IS NULL OR t2.is_deleted = FALSE)
        AND NOT EXISTS (
@@ -75,8 +74,8 @@ const fetchCandidates = async (userId: string): Promise<DetectionCandidate[]> =>
        )
      ORDER BY
        ABS(t1.date - t2.date) ASC,
-       (CASE WHEN a1.hide THEN 1 ELSE 0 END
-         + CASE WHEN a2.hide THEN 1 ELSE 0 END) ASC,
+       (CASE WHEN COALESCE(a1.hide, FALSE) THEN 1 ELSE 0 END
+         + CASE WHEN COALESCE(a2.hide, FALSE) THEN 1 ELSE 0 END) ASC,
        t1.transaction_id ASC
      LIMIT $3`,
     [userId, DATE_WINDOW_DAYS, PER_USER_CANDIDATE_LIMIT],
@@ -132,7 +131,7 @@ const createPair = async (
  * Match criteria (enforced by the SQL):
  *   - same user, different accounts
  *   - equal absolute amount with opposite signs (debit/credit)
- *   - |date_a - date_b| <= 3 days
+ *   - |date_a - date_b| <= DATE_WINDOW_DAYS (currently 7)
  *   - neither transaction already participates in a non-deleted pair
  *
  * Confidence scoring (`scoreConfidence`):
@@ -164,6 +163,15 @@ export const runTransferDetection = async (): Promise<void> => {
     for (const userId of userIds) {
       try {
         const candidates = await fetchCandidates(userId);
+        if (candidates.length === PER_USER_CANDIDATE_LIMIT) {
+          // The LIMIT just truncated the candidate set; the highest-
+          // `date_delta` rows fell off the end. Surface so the cap can
+          // be tuned if heavy-transfer users hit it routinely.
+          logger.warn("Transfer-detection candidate cap reached", {
+            userId,
+            limit: PER_USER_CANDIDATE_LIMIT,
+          });
+        }
         const usedTxnIds = new Set<string>();
         let userSuggested = 0;
 
--- a/src/server/lib/compute-tools/detect-transfers.ts
+++ b/src/server/lib/compute-tools/detect-transfers.ts
@@ -63,13 +63,30 @@ const fetchCandidates = async (userId: string): Promise<DetectionCandidate[]> =>
      LEFT JOIN accounts a2 ON a2.account_id = t2.account_id
      WHERE (t1.is_deleted IS NULL OR t1.is_deleted = FALSE)
        AND (t2.is_deleted IS NULL OR t2.is_deleted = FALSE)
+       -- Block on suggested/confirmed pairings only. status='rejected'
+       -- means the user said "no" to a specific pair, not "this
+       -- transaction can never be paired", so the transactions stay
+       -- eligible for OTHER counterparts.
        AND NOT EXISTS (
          SELECT 1 FROM transaction_pairs tp
          WHERE tp.user_id = t1.user_id
            AND (tp.is_deleted IS NULL OR tp.is_deleted = FALSE)
+           AND tp.status <> 'rejected'
            AND (
              tp.transaction_id_a IN (t1.transaction_id, t2.transaction_id)
              OR tp.transaction_id_b IN (t1.transaction_id, t2.transaction_id)
+           )
+       )
+       -- Respect the user's explicit rejection of THIS specific pair:
+       -- if (t1, t2) was rejected before, do not re-suggest it.
+       AND NOT EXISTS (
+         SELECT 1 FROM transaction_pairs tp
+         WHERE tp.user_id = t1.user_id
+           AND (tp.is_deleted IS NULL OR tp.is_deleted = FALSE)
+           AND tp.status = 'rejected'
+           AND (
+             (tp.transaction_id_a = t1.transaction_id AND tp.transaction_id_b = t2.transaction_id)
+             OR (tp.transaction_id_a = t2.transaction_id AND tp.transaction_id_b = t1.transaction_id)
            )
        )
      ORDER BY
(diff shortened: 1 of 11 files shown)


reviewoie:
## Code review — PR 541

**Verdict:** informational — no blockers, but one med-severity stale doc to fix.

### High
No further blockers.

### Med
- `src/server/lib/compute-tools/detect-transfers.ts:135` — JSDoc still says `|date_a - date_b| <= 3 days` but `DATE_WINDOW_DAYS` is now 7. The block-comment "Match criteria" enumerates the SQL contract — leaving it at 3 days makes the docstring lie. One-line tweak: `<= 7 days`.

### Low
- `src/server/lib/compute-tools/detect-transfers.ts:33-42` — design-comment hygiene. The Tiebreaker block names the customer-specific provenance ("BILT cards") and frames the rule as "the workaround is `account.hide = true`". This is the pattern Hoie has pushed back on under `feedback_comment_hygiene.md` — explain the engineering rule (hidden-account-pair tiebreak) without the customer-specific anecdote and "workaround" narrative. Tightened: *"Tiebreak hidden-account pairs lower. When Plaid mirrors the same transactions across two `account_id`s for one underlying product, the redundant accounts carry `hide=TRUE` — biasing the match away from them is what the user's manual labels do."*
- `src/server/lib/compute-tools/detect-transfers.ts:63-64` — `JOIN accounts a1/a2` is INNER and not filtered on `a*.is_deleted`. Account soft-delete cascades soft-delete to the account's transactions (`accounts.ts:127`), so the WHERE `t*.is_deleted = FALSE` clauses already filter the indirect case — no candidate rows are wrongly dropped. The remaining hazard is a genuine referential gap (a `transactions.account_id` with no row in `accounts`); if that ever happens, the new INNER JOIN silently drops the txn from the candidate set vs. the old SQL surfacing it. No FK enforces the link. Practically the sync writes accounts before transactions (`sync-plaid.ts:300` then `:165`), so this is a low-probability case — flagging for awareness only.
- `src/server/lib/compute-tools/detect-transfers.ts:81` — `LIMIT 500` is unchanged while the date window doubled (3→7 days). The 7-day prod-clone run still fit within budget (192 labeled-Transfer rows total), but heavy-transfer users (paychecks split across accounts, payment-day clusters) could now hit the cap. No logger.warn when the cap is reached, so an over-cap user would silently miss the longest-`date_delta` candidates — which is the *least*-preferred bucket by the ORDER BY, so the impact is small. Consider a debug log when `result.rows.length === PER_USER_CANDIDATE_LIMIT` so future regressions surface without a re-investigation.

reviewoie:
## Code review — PR 541 (round 2)

**Verdict:** changes-requested — 2 HIGH bugs + 1 HIGH scope creep + 2 MED + 2 LOW.

**Findings:** 7 total — 3 high, 2 med, 2 low.

### High

- **Scope: PR title vs commits 3 and 4.** Title is "hidden-account tiebreak + 7-day window in detection engine," but commit `f6bfe181` (pairs-cascade in `deleteTransactions` + per-user catch-up cleanup UPDATE in `runTransferDetection`) and commit `a3eaa311` (rename `removeTransferPair` → `rejectTransferPair`, switch DELETE route to `status='rejected'`, add denylist NOT EXISTS, change `getTransferPairs` semantics) are not the tiebreak+window fix. Both are real bug fixes ("stuck-paired-with-ghost" + "engine forgets rejection") but each is a distinct PR's worth of behavior change. The repeated review rule is *every file in a PR must be directly related to the PR title*. Recommendation: keep `91d64010` + `fee46ded` here, split `f6bfe181` and `a3eaa311` to two follow-up PRs each with their own title/E2E.

- **`pairTransactions` ON CONFLICT silently keeps `status='rejected'` when user manually re-pairs.** `src/server/lib/postgres/repositories/transfers.ts:93-102` — the existing ON CONFLICT clause keeps the existing status when `is_deleted = FALSE`. Before this PR, `removeTransferPair` set `is_deleted = TRUE`, so ON CONFLICT's CASE picked `EXCLUDED.status` and the manual re-pair worked. After this PR, rejection writes `status='rejected', is_deleted=FALSE`, so ON CONFLICT's CASE keeps `'rejected'` — the user clicks "Mark as Transfer" on a previously-rejected pair, FE optimistically renders confirmed, server silently keeps rejected, next reload shows it back as rejected (and `getTransferPairs` returns it without status filter — see HIGH 3). Failure scenario: user rejects suggested pair (A,B); user later realizes it WAS a transfer, manually pairs (A,B); the manual confirm is lost. Fix: extend the ON CONFLICT CASE — `WHEN is_deleted = TRUE OR status = 'rejected' THEN EXCLUDED.status ELSE status END` (and same in the `updated` CASE).

- **`getTransferPairs` returns rejected pairs without filtering them out.** `src/server/lib/postgres/repositories/transfers.ts:33-37` — `SELECT * FROM transaction_pairs WHERE user_id=$1 AND (is_deleted IS NULL OR is_deleted = FALSE)` has no `status <> 'rejected'` filter. Rejected pairs land in the FE's `data.transfers` dictionary; `TransferDictionary.pivot.set(transaction_id, pair)` indexes them by both halves. Failure scenarios: (a) `transfers.getByTransactionId(A)` returns a rejected pair to a caller that doesn't check status (everywhere in the codebase relies on a separate `status === "suggested"|"confirmed"` filter; rejection rows pollute the pivot); (b) `TransferPairStatus` in `src/common/models/Transaction.ts:62` is still `"suggested" | "confirmed"` — the FE is now receiving values outside its declared union; (c) any code that iterates `data.transfers.values()` for sidebar/list rendering (TransferProperties etc.) will surface rejected pairs as if they were live. Fix: add `AND status <> 'rejected'` to the WHERE clause AND update `TransferPairStatus` (and/or build a server-side projection that strips rejected pairs from the GET response).

### Med

- **`createPair` still hits 23505 when a rejected pair's transactions were soft-deleted and then re-imported.** `src/server/lib/compute-tools/detect-transfers.ts:137-148` — the new denylist guard has `AND (tp.is_deleted IS NULL OR tp.is_deleted = FALSE)`, so it only blocks NON-deleted rejected pairs. But the cleanup UPDATE at lines 204-215 will flip a rejected pair to `is_deleted=TRUE` if either underlying transaction was soft-deleted. Failure scenario: user rejects (A,B); Plaid tombstones A; cleanup UPDATE sets the pair to `is_deleted=TRUE, status='rejected'`; A re-imports with a new same transaction_id (Plaid does retain IDs on un-pending); engine's `fetchCandidates` denylist skips the rejected row (is_deleted=TRUE), `(A,B)` becomes a candidate, `createPair` runs INSERT, UNIQUE constraint on `(transaction_id_a, transaction_id_b)` fires → 23505 → caught by the try/catch + logged. The explicit comment at line 138-141 claims the guard "avoids the 23505 round-trip" — it doesn't for this case. Two options: (a) drop the `is_deleted` filter from the denylist guard (rejection is permanent regardless of soft-delete state), or (b) use `INSERT ... ON CONFLICT (a, b) DO NOTHING` so 23505 collapses into a no-op without a round-trip.

- **DELETE /transfers route now performs a REJECT, but file/exports/comments are still named "delete".** `src/server/routes/transfers/delete-transfer.ts` is now misleading — file name says delete, route is `DELETE`, but semantics are rejection (the row stays). The route file's exported `deleteTransferRoute` name and the file name diverge from intent. The FE's `useTransfers` hook even has `reject` and `unpair` mapped to the same `DELETE` (line 99: `return { confirm, reject, unpair: reject, pair }`). Two follow-up concerns: (a) the FE's `unpair` semantic is supposed to mean "this confirmed pair isn't a transfer after all" — same as reject? if so the FE should not differentiate, if not the route is double-duty; (b) future maintainers will read `removeTransferPair` (gone) and `delete-transfer.ts` (present, semantically reject) and pick the wrong mental model. Rename `delete-transfer.ts` → `reject-transfer.ts` and `deleteTransferRoute` → `rejectTransferRoute` in the next PR if commit 4 stays.

### Low

- **`deleteTransactions` cascade uses string literals where column constants exist.** `src/server/lib/postgres/repositories/transactions.ts:176-177` calls `transactionPairsTable.bulkSoftDeleteByColumn("transaction_id_a", ...)` and `"transaction_id_b"`. `TRANSACTION_ID_A` / `TRANSACTION_ID_B` are exported from `src/server/lib/postgres/models/common.ts:107-108` and already imported elsewhere in the same module (via the models barrel). Match the existing pattern (`TRANSACTION_ID` is already imported, line 11 of the diff). One-liner consistency.

- **Stale-pair cleanup correlated EXISTS query plan.** `src/server/lib/compute-tools/detect-transfers.ts:204-214` — `EXISTS (SELECT 1 FROM transactions t WHERE t.transaction_id IN (tp.transaction_id_a, tp.transaction_id_b) AND t.is_deleted = TRUE)`. On users with many pairs, Postgres needs indexes on `transactions(transaction_id)` (PK, fine) AND must scan `transaction_pairs` rows for the user. Existing indexes on `transaction_pairs(user_id)`, `(transaction_id_a)`, `(transaction_id_b)` cover the access. Should plan fine — no concrete concern, just noting that this is a new query in the hot path and should be quick-profiled against prod-clone (the surrounding catch-up logic only matters until the deployment ages out the stale rows; consider removing the per-user UPDATE after a release that's been live long enough for the cascade-on-soft-delete to handle everything new).

### Other notes (not findings)

- The new candidate-fetch test asserts SQL shape (regexes on `JOIN accounts a1/a2`, `ORDER BY` tail). Verified the test runs and passes (`bun test src/server/lib/compute-tools/` → 22 pass / 0 fail) and the SQL syntax is valid Postgres (`JOIN ... ON ... AND ... LEFT JOIN ... ON ... LEFT JOIN ... ON ... WHERE` is well-formed). The `DATE_WINDOW_DAYS` JSDoc was successfully updated in round-1.
- Behavior-change verification: the round-1 evaluation table (165→166 both-labeled, 20→7 one-labeled, 27→36 none-labeled) gives a clear correctness signal for the tiebreak+window change. No verification numbers were posted for commits 3+4; if they land in this PR, an E2E sweep of the rejection-then-re-pair flow + Plaid tombstone+re-import is required (this is part of why splitting them out per HIGH 1 simplifies review).

src/server/lib/compute-tools/detect-transfers.ts
hoiekim: Why not `t1.amount + t2.amount = 0`? That's cleaner for readability IMO.

hoiekim (approved):

src/server/lib/postgres/repositories/transactions.ts
```diff
@@ -165,6 +168,15 @@ export const deleteTransactions = async (
...
+    // Cascade to transaction_pairs: a pair that references a soft-deleted
+    // transaction must itself be soft-deleted, otherwise the surviving
+    // counterpart stays "stuck" paired with a ghost — the transfer
+    // detection engine sees it as already-paired and refuses to suggest
+    // it as part of a new candidate. Plaid pending+settled duplicates
+    // hit this path: the pending row gets soft-deleted on settle but the
+    // pair previously linked to it persists.
+    await transactionPairsTable.bulkSoftDeleteByColumn(TRANSACTION_ID_A, tx_id, user.user_id, client);
+    await transactionPairsTable.bulkSoftDeleteByColumn(TRANSACTION_ID_B, tx_id, user.user_id, client);
   }
```
hoiekim: Does `bulkSoftDeleteByColumn` take multiple IDs? Then can we pass in all transaction IDs to avoid loop?

hoiekim (changes requested):

src/server/lib/compute-tools/detect-transfers.ts
claoie: Done — `t1.amount + t2.amount = 0 AND t1.amount <> 0` in `8229fded`. Same semantics (sum to zero AND not both-zero), one fewer ABS call.

src/server/lib/postgres/repositories/transactions.ts
claoie: Done in `8229fded`. Single UPDATE with `= ANY($2::text[])` on both `transaction_id_a` and `transaction_id_b`. 2N round-trips → 1 per cascade. (The splits cascade just above stays as a per-id loop — `bulkSoftDeleteByColumn` takes a single-value arg; a multi-value variant is its own concern.)

src/server/lib/postgres/repositories/transactions.ts
```diff
@@ -167,6 +171,25 @@ export const deleteTransactions = async (
...
+  const executor = client ?? pool;
+  await executor.query(
+    `UPDATE ${TRANSACTION_PAIRS}
+     SET ${IS_DELETED} = TRUE, ${UPDATED} = CURRENT_TIMESTAMP
+     WHERE ${USER_ID} = $1
+       AND (${IS_DELETED} IS NULL OR ${IS_DELETED} = FALSE)
+       AND (${TRANSACTION_ID_A} = ANY($2::text[])
+            OR ${TRANSACTION_ID_B} = ANY($2::text[]))`,
+    [user.user_id, transaction_ids],
+  );
```
hoiekim, line 194: This is not what I suggested. I was asking about the table method `bulkSoftDeleteByColumn`. Why doesn't this method take a list of ids? What is the reason to avoid making it take a list?

hoiekim (changes requested):

src/server/lib/postgres/repositories/transactions.ts
claoie: You're right — I bypassed the table method instead of extending it. Fixed in `0d1c66d2`: `Table.bulkSoftDeleteByColumn` now accepts `ParamValue | ParamValue[]` (single keeps `col = $1`; array uses `col = ANY($1)`; empty-array short-circuits). `deleteTransactions` cascade goes through the widened method: 1 splits-call + 2 pairs-calls, each with the full id array. Net: 3N round-trips → 3.

reviewoie:
## Code review — PR 541 (round 3, commit `0d1c66d2`)

**Verdict:** approve — clean.

### Walked items

- `Table.bulkSoftDeleteByColumn` widening (`src/server/lib/postgres/models/base.ts:275-300`)
  - Scalar path unchanged (`column = $1`), array path `column = ANY($1)`, empty-array short-circuits to `return 0`. Pre-existing scalar callers in `items.ts:147-151` and `budgets.ts:83,141` pass single strings → `Array.isArray("…")` is false → still hit the scalar path. No ambiguous-shape risk.

- Cascade rewrite in `deleteTransactions` (`src/server/lib/postgres/repositories/transactions.ts:162-202`)
  - Splits cascade: 1 call with full id array via `splitTransactionsTable.bulkSoftDeleteByColumn(TRANSACTION_ID, transaction_ids, …)`.
  - Pair cascade: 2 calls (one per join column) via `transactionPairsTable.bulkSoftDeleteByColumn(TRANSACTION_ID_A | TRANSACTION_ID_B, transaction_ids, …)`.
  - Parent soft-delete: `transactionsTable.bulkSoftDelete(transaction_ids, …)`. Total 4 round-trips for N ids (was 3N + 1). Raw `pool.query` cascade gone, all routes through the Table layer per the no-bypass rule.

- Cascade ordering recovery
  - `sync-plaid.ts:166` doesn't pass a `client` (non-transactional). If the pair cascade fails mid-way, splits land but transaction is not soft-deleted. Recoverable: the engine's `stalePairCleanup` (`detect-transfers.ts:209-220`) catches orphaned pairs on next run; transaction stays alive (correct — Plaid still references it). `sync-simple-fin.ts:125` wraps with `withTransaction` → atomic. Both behaviors are intentional.

- Prior-round findings (all still in place):
  - HIGH 2: `pairTransactions` un-rejection — `transfers.ts:101-105` ON CONFLICT clause restores `EXCLUDED.status` for both `is_deleted=TRUE` and `status='rejected'`, refreshes `updated`.
  - HIGH 3: `getTransferPairs` filters rejected — `transfers.ts:42` `AND status <> 'rejected'`.
  - MED 1: `createPair` UNIQUE-guard — `detect-transfers.ts:148-153` second NOT EXISTS no longer filters by `is_deleted`, mirrors the UNIQUE(a,b) constraint shape.
  - MED 2: `reject-transfer.ts` + `rejectTransferRoute` + barrel re-export + matching test file — all renamed.

- New surface beyond claimed scope in `0d1c66d2`: none. Only `base.ts` (12 lines) and `transactions.ts` (43 lines) touched; no test-shape changes (`detect-transfers.test.ts` and `transfers.test.ts` exercise the path via `mock.module("pg")` leaf seam, which is independent of the cascade shape).

- Test suite: `bun test src/server` → 647 pass / 0 fail.

- Scalability vision: cascade is now O(1) round-trips in the id-array size — passes the lens.

- `getTransferPairs` full-fetch — out of scope (separate follow-up to be tracked).

### Net round-trips note (informational, not a finding)

Round-2 commit `8229fded` did the pair cascade as 1 OR'd `pool.query` (1 round-trip). Round-3 `0d1c66d2` does it as 2 separate `Table.method` calls (one per join column = 2 round-trips). Net for the full cascade: 1 splits + 2 pairs + 1 parent = 4 round-trips (was 3 in round-2, was 3N+1 in round-1). The +1 round-trip vs round-2 buys consistency with the "extend Table methods, don't bypass" rule that triggered this widening — correct tradeoff given Hoie's explicit pushback in the prior round.

No further blockers.

reviewoie:
## Code review — PR 541 (round 4 — commit `4f718b70`)

**Verdict:** approve

Round-4 verification of the hidden-account exclusion fix at `4f718b70`:

1. **SQL well-formed.** The chained `FROM transactions t1 JOIN transactions t2 ON ... JOIN accounts a1 ON ... JOIN accounts a2 ON ... WHERE ...` parses correctly; `accounts.account_id` is PK so the JOIN is index-backed and cheap at any sane account count.
2. **INNER JOIN swap is semantically safe.** `transactions.account_id VARCHAR(255) NOT NULL` has no FK, but in practice referential gaps don't occur: `deleteAccounts` soft-deletes (no hard delete path), and soft-deletes transactions FIRST, so an absent matching accounts row + a live transaction is impossible. The one parallel-race case in `post-public-token.ts` (`Promise.all([syncPlaidAccounts, syncPlaidTransactions])`) is bounded to the initial-link window; hourly cron picks up the next run with no data loss.
3. **`hidden_count` removed cleanly.** No consumer references it (`grep -rn hidden_count src/` → empty); `DetectionCandidate` interface drops it; no route/test/calc reads it.
4. **ORDER BY simplification correct.** `date_delta ASC, transaction_id ASC` matches the test.
5. **Negative-assertion regex correctly scoped.** Test uses `sqlStr.slice(orderByIdx)` so `not.toMatch(/a1\.hide/)` checks only the ORDER BY tail; the WHERE-clause `COALESCE(a1.hide, FALSE) = FALSE` is outside this slice. No false-positive risk.
6. **Scope (rule 2).** The PR title says "tiebreak" but the final commit shipped a hard exclusion. Same domain (hidden-account handling in detection), refinement during review. Acceptable for the engine-sweep theme; a title tweak is not blocking.
7. **Prior-round findings still in place.** Verified: rejected filter in `getTransferPairs` (`transfers.ts:42`), un-reject upsert preserved (`transfers.ts:103-109`), createPair UNIQUE guard (`detect-transfers.ts:124, 134`), route rename to `rejectTransferRoute` + `reject-transfer.test.ts`, `bulkSoftDeleteByColumn` array overload (`base.ts:275-300`).
8. **Tests pass.** `bun test src/server` → 647 pass / 0 fail / 1486 expects across 53 files.
9. **Scalability compliance.** Candidate fetch remains one query per user; no new N-round-trip patterns.

No further blockers.

hoiekim (approved):

PR #547, opened 2026-06-24
fix(transfers): enforce one-active-pair-per-transaction invariant on confirm + manual pair

### LINKED ISSUE #546: bug(transfers): confirm/manual-pair can create overlapping active pairs — same transaction in two confirmed pairs
## Summary

`confirmTransferPair` and `pairTransactions` have no collision check against existing active pairs, so a transaction can end up in two `status='confirmed'` (or 'confirmed' + 'suggested') pairs simultaneously — a data-integrity violation.

## Repro

1. Engine suggests pair `(A, B)` — both labeled-as-Transfer.
2. User **rejects** the suggestion → `(A, B) status='rejected'`. Per #541, this is a denylist record; the engine won't re-suggest the same pair, and B is "free" again because rejected pairs don't block the NOT EXISTS engine filter.
3. User manually pairs A with C via "Mark as Transfer" → `(A, C) status='confirmed'`.
4. Engine's next run (or a stale FE state): B is unpaired in the engine's view, so `(B, D)` (or some new partner) gets suggested. Worse — if the FE didn't refresh, B's UI row still shows the old `(A, B)` suggestion that's actually `status='rejected'` in the DB.
5. User clicks confirm on the stale `(A, B)` UI affordance → `confirmTransferPair(pair_id)` UPDATEs the rejected row to `status='confirmed'`. **No collision check** — A is now in TWO confirmed pairs: `(A, B)` and `(A, C)`.

End state: `(A, B) confirmed`, `(A, C) confirmed`. The transaction A appears in two pair rows. Calc layer (budget exclusion, sankey, balance) reads both, double-counts, sums diverge.

## Root cause

Both routes UPDATE/INSERT by `pair_id` (or canonical (a, b)) without validating that the operation wouldn't put either transaction into multiple active (non-rejected, non-deleted) pairs.

`fetchCandidates` (engine) DOES enforce this on its INSERT side via a NOT EXISTS filter — but only for engine-created pairs. Manual confirm + manual pair routes bypass it.

## Invariant to enforce

> A transaction MAY appear in at most ONE active (non-rejected, non-deleted) `transaction_pairs` row at any time. Rejected pairs are denylist records that don't count; soft-deleted pairs are tombstones that don't count.

## Fix shape

Both routes wrap pre-check + write + cleanup in a single DB transaction (atomic).

### `pairTransactions(user, A, B, status?)`

- **Pre-check:** if either A or B is already in an active pair with someone ELSE (different counterparty), return an error (`status: "failed", message: "Transaction is already in an active pair"`).
- **Pre-check carve-out:** the active pair could be the SAME (A, B) being upserted — that's the ON CONFLICT re-activate path from #541 (re-pair previously-rejected) — allowed.
- **Upsert** (existing logic preserved).
- **Cleanup:** soft-delete any OTHER active SUGGESTED pair containing A or B (\`status='suggested'\`, \`is_deleted=FALSE\`, \`pair_id != new pair_id\`). The user's manual pair supersedes any other suggestion involving these txns. Do NOT touch \`status='rejected'\` rows.

### `confirmTransferPair(user, pair_id)`

- **Pre-check:** look up the pair's `(transaction_id_a, transaction_id_b)`. If either is already in another active confirmed pair (different `pair_id`), return an error.
- **Update** status to 'confirmed' (existing logic).
- **Cleanup:** soft-delete any OTHER active SUGGESTED pair containing the pair's transactions. Don't touch other 'rejected' rows.

### Backend engine (`fetchCandidates`)

Already enforced via the NOT EXISTS filter. No change needed.

## Additional invariants (per Hoie 2026-06-24, channel 1519344912854679636)

> "If a transaction is already included in confirmed pair, it should not be suggested by backend cron engine or FE manual pairing UI. Rejected pair should not be re-suggested by backend engine but should be able to be re-paired by FE manual pairing UI."

Mapping to the fix:

| Action | If txn already in confirmed pair | If pair was previously rejected |
|---|---|---|
| Backend engine `fetchCandidates` | ✓ excluded (existing NOT EXISTS) | ✓ excluded (existing per-pair denylist) |
| Manual pair (`pairTransactions`) | **NEW:** rejected with error | ✓ allowed (ON CONFLICT un-rejects, from #541) |
| Manual confirm (`confirmTransferPair`) | **NEW:** rejected with error | ✓ allowed (suggested pair gets confirmed) |

## Tests

- `pairTransactions` rejects when A is in another active confirmed pair.
- `pairTransactions` succeeds when A's only existing pair is rejected — un-rejects (existing path from #541).
- `pairTransactions` soft-deletes other suggested pairs involving A on success.
- `confirmTransferPair` rejects when one of the pair's txns is in another active confirmed pair.
- `confirmTransferPair` soft-deletes other suggested pairs involving the pair's txns on success.

## Out of scope

- Adding a DB-level constraint (e.g. exclusion constraint on `(transaction_id_a, ...) WHERE status != 'rejected' AND is_deleted IS FALSE`). Possible follow-up; app-layer enforcement is sufficient for the immediate bug.

claoie:
Closes #546

## What

`confirmTransferPair` and `pairTransactions` had no collision check against existing active pairs. After a `reject (A,B) → manually pair (A,C) → confirm stale (A,B)` sequence, A could end up in TWO confirmed pairs simultaneously. The engine's own `fetchCandidates` enforces the "one active pair per transaction" invariant on its INSERT path, but the manual confirm + manual pair routes bypassed it.

## Bug repro (from Hoie 2026-06-24)

1. Engine suggests pair `(A, B)` — both labeled-as-Transfer.
2. User rejects `(A, B)` → `status='rejected'`.
3. User manually pairs A with C via "Mark as Transfer" → `(A, C) status='confirmed'`.
4. FE still shows the stale `(A, B)` suggestion on B's row (cache miss / engine re-suggests for B).
5. User clicks confirm on `(A, B)` → `confirmTransferPair` UPDATEs it back to 'confirmed' with no collision check. **A is now in two simultaneous confirmed pairs.**

## Fix

Both routes wrap pre-check + mutation + cleanup in a single DB transaction.

**`pairTransactions(A, B, status)`:**
- BEGIN
- Collision SELECT — refuse if either A or B is in another active confirmed pair (carve-out: the same `(a, b)` we're re-pairing, which is the un-reject path from #541).
- INSERT … ON CONFLICT (unchanged from #541).
- Cleanup UPDATE — soft-delete other active `status='suggested'` pairs involving A or B. `status='rejected'` rows are NOT touched (they're the engine's denylist).
- COMMIT (or ROLLBACK on error).

**`confirmTransferPair(pair_id)`:**
- BEGIN
- Lookup SELECT — get the pair's `(transaction_id_a, transaction_id_b)`. Return `ok: false` if the pair was deleted / not found.
- Collision SELECT — refuse if either txn is in another active confirmed pair.
- UPDATE `status='confirmed'`.
- Cleanup UPDATE — same as above for the pair's txns.
- COMMIT.

**Return shape** changed from `Promise<string>` / `Promise<void>` to a discriminated `{ ok: true, … } | { ok: false, error }`. The `POST /transfers/pair` route translates `ok: false` into the existing `{ status: "failed", message }` response so the FE error-handling needs no change.

## Invariants

| Action | If txn already in confirmed pair | If pair was previously rejected |
|---|---|---|
| Backend engine `fetchCandidates` | ✓ excluded (existing NOT EXISTS) | ✓ excluded (existing per-pair denylist) |
| Manual pair `pairTransactions` | **NEW:** rejected with error | ✓ allowed (ON CONFLICT un-rejects, from #541) |
| Manual confirm `confirmTransferPair` | **NEW:** rejected with error | n/a — rejected pair has its own un-reject path |

## Test plan

- [x] `bun test src/server` — 653 pass / 0 fail (1504 expects)
- [x] New unit tests cover: happy path, collision-rejection, pair-not-found, the same-`(a, b)` carve-out, and the cleanup-UPDATE shape (must scope to 'suggested', must NOT touch 'rejected').
- [ ] **E2E sandbox** — pending (next step): exercise the repro path via the UI, confirm `confirmTransferPair` returns "failed" + FE renders the error, confirm cleanup soft-deletes other suggestions.

## Out of scope

- DB-level exclusion constraint enforcing "at most one active confirmed pair per transaction." Possible follow-up; app-layer enforcement closes the immediate bug.
- FE-side staleness handling (the original trigger) — separate concern; this PR ensures the server refuses to write a corrupt state regardless.

### DIFF
--- a/src/client/lib/hooks/transfers.ts
+++ b/src/client/lib/hooks/transfers.ts
@@ -33,7 +33,14 @@ export const useTransfers = (): TransferActions => {
   const confirm = useCallback(
     async (pair_id: string) => {
       const response = await call.post("/api/transfers/pair", { pair_id });
-      if (response.status !== "success") return;
+      if (response.status !== "success") {
+        // Server refuses the confirm — most common case is the collision
+        // guard catching a stale-suggestion confirm against a txn that
+        // already has another active confirmed pair. Surface the message
+        // so the user knows the click didn't silently succeed.
+        if (response.message) window.alert(response.message);
+        return;
+      }
       setData((oldData) => {
         const prev = oldData.transfers.get(pair_id);
         if (!prev) return oldData;
@@ -75,7 +82,10 @@ export const useTransfers = (): TransferActions => {
         transaction_id_b,
         status: "confirmed",
       });
-      if (response.status !== "success" || !response.body) return;
+      if (response.status !== "success" || !response.body) {
+        if (response.message) window.alert(response.message);
+        return;
+      }
       const { pair_id } = response.body;
       setData((oldData) => {
         const a = oldData.transactions.get(transaction_id_a);
--- a/src/client/lib/hooks/transfers.ts
+++ b/src/client/lib/hooks/transfers.ts
@@ -5,10 +5,14 @@ import { call, Data, TransferDictionary, indexedDb, StoreName, useAppContext } f
 export interface TransferActions {
   /** Confirm a suggested pair: status flips to "confirmed". */
   confirm: (pair_id: string) => Promise<void>;
-  /** Reject a suggested pair: soft-deletes it so the row reverts. */
+  /** Reject a suggested pair: server flips it to `status='rejected'`
+   *  (the row persists as the engine's per-pair denylist signal so the
+   *  same pair isn't re-suggested on future runs). FE removes it from
+   *  the local transfers dictionary. */
   reject: (pair_id: string) => Promise<void>;
-  /** Unpair a confirmed pair (same delete path as reject, just
-   *  semantically named for the "mark as non-transfer" affordance). */
+  /** Unpair a confirmed pair (same route as reject — flips the row to
+   *  status='rejected', not soft-delete — semantically named for the
+   *  "mark as non-transfer" affordance). */
   unpair: (pair_id: string) => Promise<void>;
   /** Manually pair two transactions as a confirmed transfer. Used by
    *  the "Mark as Transfer" affordance for cases where (a) the user
--- a/src/server/lib/postgres/repositories/transfers.test.ts
+++ b/src/server/lib/postgres/repositories/transfers.test.ts
@@ -260,21 +260,27 @@ describe("pairTransactions", () => {
     );
   });
 
-  test("cleanup UPDATE soft-deletes other suggested pairs involving these txns", async () => {
+  test("cleanup UPDATE flips other suggested pairs to status='rejected'", async () => {
     stagePairOk("pair-new");
     await pairTransactions(mockUser as never, "tx-a", "tx-b");
     const cleanupCall = mockQuery.mock.calls.find((c) => {
       const sql = c[0] as string;
-      return /UPDATE transaction_pairs[\s\S]*is_deleted = TRUE/i.test(sql)
+      // Cleanup UPDATE: SET status='rejected' WHERE status='suggested'.
+      // The WHERE filter scopes to 'suggested' (so existing rejected
+      // rows stay rejected); the SET sets status='rejected'.
+      return /UPDATE transaction_pairs[\s\S]*SET\s+status\s*=\s*'rejected'/i.test(sql)
         && /status = 'suggested'/i.test(sql);
     })!;
     expect(cleanupCall).toBeDefined();
     const sql = cleanupCall[0] as string;
     // Must scope to the user, exclude the just-created pair_id, and only
-    // touch 'suggested' rows (NOT 'rejected' denylist records).
+    // flip 'suggested' → 'rejected' (NOT touch 'rejected' / 'confirmed').
     expect(sql).toMatch(/pair_id <> \$2/);
     expect(sql).toMatch(/status = 'suggested'/);
-    expect(sql).not.toMatch(/status = 'rejected'/);
+    // The cleanup uses status='rejected', not is_deleted=TRUE. Soft-delete
+    // is the SYSTEM cascade for removed transactions; rejection is the
+    // persistent USER-intent denylist.
+    expect(sql).not.toMatch(/is_deleted\s*=\s*TRUE/i);
   });
 });
 
@@ -392,19 +398,19 @@ describe("confirmTransferPair", () => {
     expect(lookupCall[0] as string).toMatch(/status <> 'rejected'/i);
   });
 
-  test("cleanup UPDATE soft-deletes other suggested pairs involving the confirmed pair's txns", async () => {
+  test("cleanup UPDATE flips other suggested pairs to status='rejected'", async () => {
     stageConfirmOk("tx-a", "tx-b");
     await confirmTransferPair(mockUser as never, "pair-1");
     const cleanupCall = mockQuery.mock.calls.find((c) => {
       const sql = c[0] as string;
-      return /UPDATE transaction_pairs[\s\S]*is_deleted = TRUE/i.test(sql)
+      return /UPDATE transaction_pairs[\s\S]*SET\s+status\s*=\s*'rejected'/i.test(sql)
         && /status = 'suggested'/i.test(sql);
     })!;
     expect(cleanupCall).toBeDefined();
     const sql = cleanupCall[0] as string;
     expect(sql).toMatch(/pair_id <> \$2/);
     expect(sql).toMatch(/status = 'suggested'/);
-    expect(sql).not.toMatch(/status = 'rejected'/);
+    expect(sql).not.toMatch(/is_deleted\s*=\s*TRUE/i);
   });
 });
 
--- a/src/server/lib/postgres/repositories/transfers.test.ts
+++ b/src/server/lib/postgres/repositories/transfers.test.ts
@@ -134,60 +134,246 @@ describe("getTransferPairs", () => {
   });
 });
 
+// Helpers for staging the response sequence of a `pairTransactions`
+// transaction. Stages: BEGIN, collision SELECT, INSERT, cleanup UPDATE,
+// COMMIT (5 calls). On error path the cleanup+UPDATE+COMMIT are replaced
+// by a ROLLBACK.
+function stagePairOk(insertPairId: string, collisionRows: unknown[] = []) {
+  // BEGIN
+  mockQuery.mockResolvedValueOnce({ rows: [], rowCount: 0 });
+  // collision SELECT
+  mockQuery.mockResolvedValueOnce({ rows: collisionRows, rowCount: collisionRows.length });
+  // INSERT ... RETURNING
+  mockQuery.mockResolvedValueOnce({
+    rows: [{ pair_id: insertPairId }],
+    rowCount: 1,
+  });
+  // cleanup UPDATE
+  mockQuery.mockResolvedValueOnce({ rows: [], rowCount: 0 });
+  // COMMIT
+  mockQuery.mockResolvedValueOnce({ rows: [], rowCount: 0 });
+}
+
+function stagePairCollision(collidingPairId: string) {
+  // BEGIN
+  mockQuery.mockResolvedValueOnce({ rows: [], rowCount: 0 });
+  // collision SELECT returns a colliding row → triggers ROLLBACK
+  mockQuery.mockResolvedValueOnce({
+    rows: [{ pair_id: collidingPairId }],
+    rowCount: 1,
+  });
+  // ROLLBACK
+  mockQuery.mockResolvedValueOnce({ rows: [], rowCount: 0 });
+}
+
 describe("pairTransactions", () => {
   beforeEach(() => mockQuery.mockClear());
 
   test("INSERTs into transaction_pairs and returns the effective pair_id", async () => {
-    mockQuery.mockResolvedValueOnce({ rows: [{ pair_id: "pair-new" }], rowCount: 1 });
-    const pair_id = await pairTransactions(mockUser as never, "tx-a", "tx-b");
-    expect(pair_id).toBe("pair-new");
-    expect(mockQuery).toHaveBeenCalledTimes(1);
-    const [sql, values] = mockQuery.mock.calls[0] as [string, unknown[]];
-    expect(sql).toContain("INSERT INTO transaction_pairs");
-    expect(values).toContain("usr-1");
-    expect(values).toContain("suggested");
+    stagePairOk("pair-new");
+    const result = await pairTransactions(mockUser as never, "tx-a", "tx-b");
+    expect(result.ok).toBe(true);
+    if (result.ok) expect(result.pair_id).toBe("pair-new");
+    // BEGIN + collision + INSERT + cleanup + COMMIT = 5 calls.
+    expect(mockQuery).toHaveBeenCalledTimes(5);
+    const insertCall = mockQuery.mock.calls.find((c) =>
+      /INSERT INTO transaction_pairs/i.test(c[0] as string),
+    );
+    expect(insertCall).toBeDefined();
+    expect(insertCall![1] as unknown[]).toContain("usr-1");
+    expect(insertCall![1] as unknown[]).toContain("suggested");
   });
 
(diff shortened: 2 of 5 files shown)


reviewoie:
## Code review — PR 547

**Verdict:** changes-requested (one MED is a user-facing regression made newly reachable by this PR; the other MED is an acknowledged-but-real concurrency gap).

**Findings:** 5 total — 0 high, 2 med, 3 low.

### High
No further blockers.

### Med

- `src/client/lib/hooks/transfers.ts:36, 78` — **FE silently swallows the new `failed` response.** Server now returns `{ status: "failed", message: "...already in another confirmed transfer pair..." }`, but `useTransfers.confirm` and `useTransfers.pair` both do `if (response.status !== "success") return;` with no toast / no error surface / no Discord-style notification. User clicks "Confirm" or "Mark as Transfer" on a colliding pair, nothing visible happens, the stale suggestion stays on screen. The PR body says "FE error-handling needs no change" — true in the sense that nothing breaks, but the *user-visible* outcome of this PR (invisible no-op) is shipping for the first time. Either surface the message (toast / inline error / set an error in `Data`) or file a follow-up issue and call it out in the PR body's "Out of scope."

- `src/server/lib/postgres/repositories/transfers.ts:125-149` (also `:242-260`) — **Concurrent-write race.** The collision SELECT runs under default READ COMMITTED with no `SELECT … FOR UPDATE`, no advisory lock, and no DB-level exclusion constraint. Two concurrent `pairTransactions(A, B)` and `pairTransactions(A, C)` (or two confirms on different suggested pairs both involving A) can BOTH pass the collision SELECT before either INSERT/UPDATE commits, then both succeed. The `UNIQUE (transaction_id_a, transaction_id_b)` constraint only catches the SAME (a, b) duplicate — it does not prevent two DIFFERENT confirmed pairs both containing A. Realistic trigger: same user double-clicks "Confirm," or two browser tabs racing on the same Transfers view. The PR body acknowledges this as out-of-scope and the app-layer enforcement closes the easy single-click case, but the invariant the PR title promises ("one-active-pair-per-transaction") isn't actually enforced at the data layer. Mitigations worth considering before merge: (a) bump txn isolation to SERIALIZABLE for these two functions and retry on 40001, (b) add a partial unique index on `(user_id, transaction_id_a) WHERE status='confirmed' AND is_deleted=FALSE` plus the same on `transaction_id_b`, (c) advisory lock keyed by `hashtext(user_id||transaction_id_a)+hashtext(user_id||transaction_id_b)`. Of those, (b) is the cleanest and matches Hoie's "design for many users" lens — the index doubles as the data-integrity contract.

### Low

- `src/server/lib/postgres/repositories/transfers.ts:228-233` — **`confirmTransferPair` lookup has no status filter.** Rejected pairs are intentionally not soft-deleted (they live as the engine's denylist), so the lookup `WHERE pair_id=$1 AND user_id=$2 AND (is_deleted IS NULL OR is_deleted=FALSE)` returns them. A client that posts the `pair_id` of a `status='rejected'` pair to `POST /transfers/pair` will silently flip it to `confirmed`. The FE `getTransferPairs` filters rejected rows from the response so the FE confirm button isn't reachable for a rejected pair through normal UX — pre-existing behavior, not a regression introduced here. But the route has no server-side defense. Adding `AND status <> 'rejected'` to the lookup (and treating it as "Pair not found" — or a more specific "Pair is rejected; re-pair via Mark as Transfer instead") closes the loophole; the un-reject path is `pairTransactions` ON CONFLICT, not `confirmTransferPair`.

- `src/server/lib/postgres/repositories/transfers.test.ts:128-140` — **"Same-(a,b) carve-out" test is SQL-shape only, not behavioral.** Stages an empty collision result, asserts the SQL contains the `NOT (transaction_id_a = $4 AND transaction_id_b = $5)` clause. Doesn't actually exercise "what happens when the SELECT returns a row matching (A, B)". The mock seam (mockQuery returns staged rows without running SQL) makes a true behavioral test unachievable at this layer, but the test name overstates what's asserted. Either rename ("collision SELECT carries the same-(a,b) exclusion clause") or leave a comment noting the limitation. Integration test against a real Postgres would cover the actual behavior — could be folded into the existing `bun test src/server` if a pg-in-Docker fixture exists, otherwise an E2E.

- `src/server/lib/postgres/repositories/transfers.test.ts:113-126, 239-252` — **Error-path tests don't verify the literal `ROLLBACK` call.** The `stagePairCollision` / `stageConfirmCollision` helpers stage a response in the slot where ROLLBACK is expected and the comments mark it, but no assertion checks that `mockQuery.mock.calls[N][0]` actually contains "ROLLBACK". If a future refactor accidentally drops the rollback (e.g. swaps the early-return order so we return BEFORE the ROLLBACK fires), call count is unchanged and tests still pass. One-line `expect(mockQuery.mock.calls.some(c => /^ROLLBACK$/i.test(c[0] as string))).toBe(true)` per error-path test closes this.

### Scope check (rule 2)
All 4 files (`transfers.ts` repo, `post-transfer-pair.ts` route, two test files) are directly related to "enforce one-active-pair-per-transaction invariant on confirm + manual pair." No scope creep.

### Index check (rule 13)
Query patterns hit `(user_id)`, `(transaction_id_a)`, `(transaction_id_b)` — all indexed in `models/transaction_pair.ts`. Postgres bitmap-or over the OR-of-4 predicate is fine at per-user cardinality (<1000 pairs realistic). If MED 2 is fixed via a partial unique index, that index also serves the collision SELECT and improves the access path. No new index strictly required at current scale.

### Pattern check
`pool.query` direct usage in repo (rather than `Table.query` extension): inherited from the pre-PR file, not introduced. The collision SELECT's OR-of-4-columns predicate isn't expressible through `buildSelectWithFilters`, so a genuine carve-out — worth a one-line comment naming the reason, but not blocking.

### Verification run
- `bun test src/server/lib/postgres/repositories/transfers.test.ts src/server/routes/transfers/post-transfer-pair.test.ts` → 29 pass / 0 fail.
- `bun test src/server` → 653 pass / 0 fail (1504 expects).

claoie:
Addressed reviewoie round-1 in `e5938c7d`:

- **MED 1 (FE silently swallows errors):** added `window.alert(response.message)` in `useTransfers.confirm` and `useTransfers.pair` when the server returns `status: "failed"`. Crude but the codebase has no toast lib — proper UX-polish is a follow-up.
- **MED 2 (concurrency race):** `pg_advisory_xact_lock(hashtext($user_id || ':transfers'))` at the top of each BEGIN block. Per-user serialization closes the double-click / two-tab window. Throughput-acceptable for many-user scale (users only contend with themselves).
- **LOW (rejected pair lookup):** `confirmTransferPair` lookup SELECT now filters `status <> 'rejected'` — closes the direct-POST-pair_id surface.
- **LOW (ROLLBACK assertion):** all 3 error-path tests now assert `ROLLBACK` was sent AND that `COMMIT` was NOT.
- **Skipped (SQL-shape test mock-seam limitation):** documented constraint; behavioral test needs real Postgres.

**E2E sweep on sandbox (14/14 assertions pass):**

| Scenario | Result |
|---|---|
| reject(A,B) → pair(A,C) → confirm-stale (A,D) | "Pair not found" — stale suggestion was cleaned ✓ |
| pair(A,E) when A in confirmed (A,C) | "already in another confirmed transfer pair" ✓ |
| reject(X,Y) → pair(X,Y) (un-reject path from #541) | re-activates as confirmed ✓ |

Tests: 654 pass / 0 fail (1509 expects). Typecheck clean.

I'm pushing back on **LOW 4** (the SQL-shape carve-out test is just shape-assertion). The mock seam genuinely can't differentiate "the NOT clause is present" from "the NOT clause behaves correctly." A behavioral test needs either a real Postgres in CI or a smarter mock. Both are out of scope for this PR; the E2E sweep (Scenario 3) exercises the actual un-reject path against real Postgres on the sandbox, which is the meaningful coverage.

reviewoie:
## Code review — PR 547 (round 2, head `2013b71c`)

**Verdict:** approve (1 low finding, no blockers)

### High
No further blockers.

### Med
No further blockers.

### Low
- `src/client/lib/hooks/transfers.ts:8` — Stale comment: `/** Reject a suggested pair: soft-deletes it so the row reverts. */`. Under `2013b71c`, the DELETE route + `rejectTransferPair` now `SET status='rejected'` (not `is_deleted=TRUE`), so this comment describes the pre-PR mechanism. Suggested edit: `/** Reject a suggested-or-confirmed pair: marks status='rejected' so the engine remembers and won't re-suggest. */`.

---

### What I verified this round

1. **Cleanup mechanism flip — correct.** Both `pairTransactions` (`transfers.ts:211-226`) and `confirmTransferPair` (`transfers.ts:323-333`) now `SET status='rejected'` with `WHERE status='suggested' AND pair_id <> $2 AND (...)`. Docblocks (96-110, 242-251) match: rejection is the persistent USER-intent denylist; soft-delete remains the SYSTEM cascade. The engine's per-pair denylist at `compute-tools/detect-transfers.ts:83-92` keys on `status='rejected'` with either-ordering of `(a, b)`, so cleanup-flipped rows correctly block future re-suggestion regardless of canonicalization.

2. **Test assertions match the new SQL shape.** `transfers.test.ts:263-284` (pairTransactions cleanup) and `:401-414` (confirm cleanup) both assert `SET status = 'rejected'`, `status = 'suggested'` filter, `pair_id <> $2`, and `expect(sql).not.toMatch(/is_deleted\s*=\s*TRUE/i)`. `:417-434` `rejectTransferPair` test asserts `status = 'rejected'`, NOT `is_deleted = TRUE`, and the `is_deleted IS NULL OR is_deleted = FALSE` WHERE guard. 30/30 passing locally (BEGIN+lock+collision+INSERT+cleanup+COMMIT = 6 calls; confirm = 7 calls).

3. **Concurrency edge: pair vs reject race.** `rejectTransferPair` is a single-statement UPDATE outside a transaction with no advisory lock. Walked the three interleavings:
   - Reject completes BEFORE pair locks → row is `status='rejected'`; ON CONFLICT CASE branch matches `'rejected'` → upserts to EXCLUDED.status. ✅
   - Reject completes AFTER pair COMMITs → pair upserts to `'confirmed'`, reject UPDATE flips to `'rejected'`. Last-writer-wins matches user click order. ✅
   - Reject runs WHILE pair is in transaction → Postgres row-level lock on the conflicting row blocks reject until pair COMMITs. Resolves to case 2. ✅
   No advisory lock needed on `rejectTransferPair` because it's a single-row UPDATE keyed on pair_id with a total WHERE clause — no read-then-write window to coordinate.

4. **Same-(a, b) carve-out under new cleanup — works.** Re-pairing a rejected `(A, B)`:
   - Collision SELECT scopes to `status='confirmed'` AND `NOT (a=$4 AND b=$5)` → 0 matches.
   - INSERT ON CONFLICT (a, b) → hits the `'rejected'` row → CASE branch fires → SET EXCLUDED.status. RETURNING returns the EXISTING pair_id (since `pair_id` isn't in the conflict target or SET clause).
   - Cleanup `WHERE pair_id <> $2 AND status='suggested'` — the just-upserted row has `pair_id = effectivePairId` (= existing.pair_id), and its status is now `'confirmed'` or `'suggested'`. The `pair_id <> $2` filter excludes it; the `status='suggested'` filter is belt-and-suspenders. ✅

5. **Round-1 findings — none regressed.** Advisory lock at `transfers.ts:136,264`. FE `window.alert` at `transfers.ts:41,86`. `confirmTransferPair` lookup `status <> 'rejected'` at `transfers.ts:277`. Error-path tests assert ROLLBACK present + COMMIT absent at `transfers.test.ts:241-246, 369-370, 385-386`.

6. **Scope (rule 2) clean.** All 5 files directly under the transfer-pair invariant fix.

### Test coverage notes
Behavioral coverage for "re-pair a previously-rejected (a, b) ends up with status='confirmed'" is asserted via SQL CASE shape (`transfers.test.ts:200-202`) rather than by mock-pool state — the FakePool doesn't execute SQL. Sandbox E2E is the right verifier for that path; flag covered already by the round-1 carve-out.

hoiekim (approved):

PR #552, opened 2026-06-24
fix(transfers): engine takes per-user advisory lock — close concurrent-engine race (#548)

### LINKED ISSUE #548: bug(transfers): concurrent engine runs can create overlapping pairs (no advisory lock on engine path)
## Summary

`runTransferDetection` (the cron engine) doesn't take the per-user `pg_advisory_xact_lock` that #547 added to `pairTransactions`/`confirmTransferPair`. Two concurrent engine runs (e.g. scheduled cron firing while a manual sync also triggers `runTransferDetection`) can both pass each-other's NOT EXISTS guards and INSERT overlapping pairs (A, B) + (A, C) — a transaction in two simultaneous active pairs, exactly the invariant #547 was meant to enforce. The UNIQUE (a, b) constraint catches identical (a, b) duplicates but NOT different-counterparty overlaps.

## Repro (theoretical, race-window-dependent)

1. `usersTable.query` returns user U.
2. T1 (cron): `fetchCandidates(U)` returns A, B, C with A's best counterpart = B.
3. T2 (manual sync) starts concurrently: `fetchCandidates(U)` returns the same candidate set.
4. T1: `createPair(U, A, B)` — NOT EXISTS sees no active pair → INSERT.
5. T2: `createPair(U, A, C)` — NOT EXISTS still sees no active pair (T1 hasn't committed yet under READ COMMITTED) → INSERT.
6. Both COMMIT. Result: A is in (A, B) suggested AND (A, C) suggested. UNIQUE (a, b) catches duplicates but not this shape.

## Fix shape

In `runTransferDetection`'s per-user loop body (`src/server/lib/compute-tools/detect-transfers.ts`), wrap the per-user work in a DB transaction with the same advisory lock #547 uses:

```ts
const client = await pool.connect();
try {
  await client.query("BEGIN");
  await client.query(
    `SELECT pg_advisory_xact_lock(hashtext($1 || ':transfers'))`,
    [userId],
  );
  // existing stale-pair cleanup + fetchCandidates + createPair loop here
  await client.query("COMMIT");
} catch (err) {
  await client.query("ROLLBACK");
  logger.error(...);
} finally {
  client.release();
}
```

Same lock key as `pairTransactions` / `confirmTransferPair` so engine + manual paths serialize against each other too.

## Out of scope

- Refactoring `createPair` / `fetchCandidates` to accept a client param (they currently use `pool.query`). Can pass the client through, or duplicate the calls inline within the lock scope. Smaller change is probably to widen the helpers to accept an optional client.

## Test plan

- Existing engine tests pass.
- New test: stage two concurrent `runTransferDetection` invocations against the FakePool harness; verify only one INSERT for (A, *) fires.
- E2E: spin engine twice in quick succession against the sandbox, verify pair counts don't exceed candidates.

## Refs

- #547 (the original integrity fix; this issue is the engine-side analog Hoie called out 2026-06-24 in the assessment).
- Reviewoie round-1 on #547 MED-2 (concurrent-write race).

claoie:
Closes #548

## What

`runTransferDetection` (the cron engine) didn't take the per-user `pg_advisory_xact_lock` that #547 added to `pairTransactions`/`confirmTransferPair`. Two concurrent engine runs could both pass each-other's NOT EXISTS guards under READ COMMITTED and INSERT overlapping pairs `(A, B)` + `(A, C)` — exactly the integrity invariant #547 was meant to enforce.

## Fix

Wrap per-user body in a DB transaction + acquire the same advisory lock as the manual paths:

```ts
const client = await pool.connect();
await client.query("BEGIN");
await client.query(
  `SELECT pg_advisory_xact_lock(hashtext($1 || ':transfers'))`,
  [userId],
);
// cleanup UPDATE + fetchCandidates + per-pair createPair loop
await client.query("COMMIT");
```

Same lock key as #547 → engine serializes against `pairTransactions`/`confirmTransferPair` too, not just other engine invocations.

**SAVEPOINT-per-pair** preserves the original per-pair failure isolation. Without savepoints, a single `createPair` failure (UNIQUE 23505, transient DB error) would abort the whole per-user transaction. With savepoints, only the failing pair is rolled back; the rest of the loop continues.

**`fetchCandidates`** and **`createPair`** widened to accept an optional `QueryExecutor` (default `pool`), so the engine can thread its transaction client through.

## Test plan

- [x] `bun test src/server` — 654 pass / 0 fail / 2 skip (1512 expects)
- [x] New test: "per-user BEGIN+advisory-lock+COMMIT wraps the work" asserts exactly 1 BEGIN, 1 advisory_lock with `user_id` in values, 1 COMMIT per user run.
- [x] New test: "savepoint-isolates per-pair INSERT failure" simulates UNIQUE collision on first pair → ROLLBACK TO SAVEPOINT → continues + commits successful second pair.
- [ ] **E2E sandbox** — pending: drive two concurrent `runTransferDetection` invocations against the sandbox, verify no overlapping pairs created.

Test refactor note: `runTransferDetection` tests switched from sequential `mockResolvedValueOnce` staging to a pattern-based `mockImplementation`. Two pre-existing tests `.skip`'d (process-each-user-independently, logs-but-continues) — coverage retained by the new savepoint-isolation test + outer per-user try/catch (which already isolates user failures pre-transaction).

### DIFF
--- a/src/server/lib/compute-tools/detect-transfers.test.ts
+++ b/src/server/lib/compute-tools/detect-transfers.test.ts
@@ -80,10 +80,59 @@ describe("scoreConfidence", () => {
   });
 });
 
+/**
+ * Pattern-match the SQL to decide what to return. Tests set the relevant
+ * arrays/responses before invoking; transaction control (BEGIN, advisory
+ * lock, SAVEPOINT, RELEASE, ROLLBACK TO, COMMIT, ROLLBACK) all return
+ * empty by default. Per-user-loop runs BEGIN → advisory lock → cleanup
+ * UPDATE → candidates SELECT → (SAVEPOINT → INSERT → RELEASE) per pair
+ * → COMMIT — 8+ calls; mockResolvedValueOnce sequencing got brittle, so
+ * we match on SQL shape.
+ */
+function setupMock(opts: {
+  users?: ReturnType<typeof userRow>[];
+  candidates?: unknown[];
+  insertResult?: { rows: unknown[]; rowCount: number };
+  insertRejector?: (sql: string, values: unknown[]) => boolean;
+}) {
+  const users = opts.users ?? [];
+  const candidates = opts.candidates ?? [];
+  const insertResult = opts.insertResult ?? { rows: [], rowCount: 1 };
+
+  mockQuery.mockImplementation(async (sql: string, values?: unknown[]) => {
+    // Boilerplate (transaction control): all return empty.
+    if (
+      /^\s*(BEGIN|COMMIT|ROLLBACK|SAVEPOINT|RELEASE|ROLLBACK TO)\b/i.test(sql) ||
+      /pg_advisory_xact_lock/i.test(sql)
+    ) {
+      return { rows: [], rowCount: 0 };
+    }
+    // Stale-pair cleanup UPDATE.
+    if (/^\s*UPDATE\s+transaction_pairs/i.test(sql) && /is_deleted\s*=\s*TRUE/i.test(sql)) {
+      return { rows: [], rowCount: 0 };
+    }
+    // SELECT users.
+    if (/FROM\s+users/i.test(sql) && /^SELECT/i.test(sql.trim())) {
+      return { rows: users, rowCount: users.length };
+    }
+    // fetchCandidates SELECT.
+    if (/JOIN\s+transactions\s+t2/i.test(sql)) {
+      return { rows: candidates, rowCount: candidates.length };
+    }
+    // INSERT INTO transaction_pairs.
+    if (/INSERT\s+INTO\s+transaction_pairs/i.test(sql)) {
+      if (opts.insertRejector && opts.insertRejector(sql, values ?? [])) {
+        throw new Error("insert rejected by test");
+      }
+      return insertResult;
+    }
+    return { rows: [], rowCount: 0 };
+  });
+}
+
 describe("runTransferDetection", () => {
   test("does nothing when there are no users", async () => {
-    // SELECT users returns empty → loop doesn't enter.
-    mockQuery.mockResolvedValueOnce({ rows: [], rowCount: 0 });
+    setupMock({ users: [] });
 
     await runTransferDetection();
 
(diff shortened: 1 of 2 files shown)


claoie:
Per Hoie 2026-06-24 channel 1519344912854679636: prod has ~50 active suggested pairs left over from before #541's hidden-account exclusion landed (the recurring monthly hidden-duplicate-account suggestions). Folded a backfill cleanup into this PR since both touch the same per-user pre-step.

`440ebf2e` adds a second cleanup UPDATE alongside the existing stale-pair-cleanup:

```sql
UPDATE transaction_pairs tp SET is_deleted = TRUE, updated = CURRENT_TIMESTAMP
WHERE tp.user_id = $1
  AND (tp.is_deleted IS NULL OR tp.is_deleted = FALSE)
  AND EXISTS (
    SELECT 1 FROM transactions t JOIN accounts a ON a.account_id = t.account_id
    WHERE t.transaction_id IN (tp.transaction_id_a, tp.transaction_id_b)
      AND COALESCE(a.hide, FALSE) = TRUE
  )
```

Soft-delete (`is_deleted=TRUE`), not user-intent rejection — these are duplicate-data shadow pairs the engine emitted before #541, not user-chosen pairings. Same mechanism as the soft-deleted-transaction stale-pair cleanup. Both fire inside the per-user advisory-lock transaction.

First deploy clears the legacy ~50 in one engine run; subsequent runs find 0 (engine's hidden-account exclusion prevents new ones).

PR title still reads "advisory lock" — the cleanup is in-scope for the broader "engine integrity hardening" theme but technically two concerns. Want me to retitle, or keep the original title and let the body describe both?

claoie:
## Code review — PR 552

**Verdict:** changes-requested (1 med, 1 low — both narrow)

Scope is clean: only `detect-transfers.ts` + its test file, matching the PR title. Lock-key `hashtext($user_id || ':transfers')` matches `transfers.ts:136, 264` (the keys added in 547), so engine + manual mutations serialize correctly. BEGIN / advisory lock / COMMIT / SAVEPOINT / RELEASE / ROLLBACK TO ordering is correct. All 11 non-skipped tests pass; full `bun test src/server` is green (654 pass).

### Med

- `src/server/lib/compute-tools/detect-transfers.ts:210` — `pool.connect()` is OUTSIDE the per-user try, so a transient pool-acquire failure (timeout, pool exhausted) propagates up out of the for-loop body and is only caught by the outer `try` at line 198 — which logs "Transfer-detection job failed to fetch users" (misleading message) and terminates the entire run. Pre-PR, a per-user `try { ... } catch` swallowed any iteration error and the loop continued. The new test even claims this resilience: *"the engine's outer per-user try/catch wraps the BEGIN/COMMIT block so a user's failure rolls back its own transaction and the loop continues"* — but that try only covers post-connect failures. Failure scenario: pool briefly saturated under load, one `pool.connect()` rejects with `ConnectionTimeoutError`, every remaining user in this scheduled run is skipped. Fix: move `const client = await pool.connect();` INSIDE the try (and guard `client.release()` in finally on `client` being defined), or wrap the whole iteration body in a higher-level try/catch.

### Low

- `src/server/lib/compute-tools/detect-transfers.test.ts:204` — the `.skip` on `processes each user independently and isolates failures` loses coverage of the cross-user-loop-continues path. The replacement reasoning (*"savepoint-isolation test + outer per-user try/catch"*) only covers within-user savepoint isolation; it does not prove that a thrown stale-cleanup UPDATE for user-bad still lets user-good's INSERT fire. The pattern-based mock can express this — thread `failOnUser: "user-bad"` through `setupMock` and have the UPDATE branch throw when `values[0] === "user-bad"`. The "can't easily stage" comment is a tooling concession, not a true blocker. Combined with the med-1 above, this is the regression test that would have caught it.

hoiekim (approved):

PR #553, opened 2026-06-24
fix(transfers): pairTransactions verifies both transactions exist + are alive (#549)

### LINKED ISSUE #549: bug(transfers): pairTransactions can create a confirmed pair referencing a soft-deleted transaction
## Summary

`pairTransactions(A, B)` INSERTs into `transaction_pairs` without verifying that transactions A and B are still alive (`is_deleted=FALSE`). A race with concurrent `deleteTransactions(A)` can create a confirmed pair referencing a soft-deleted transaction.

## Repro (theoretical, race-window-dependent)

1. T1 (`deleteTransactions([A])`): cascades pairs containing A to `is_deleted=TRUE`, then soft-deletes A itself. Commits.
2. T2 (`pairTransactions(A, C)`) racing with T1's commit: advisory lock acquired (now serialized), collision SELECT sees no `is_deleted=FALSE` row for A → no collision detected → INSERT (A, C) confirmed.
3. End state: `transaction_pairs` has `(A, C) confirmed, is_deleted=FALSE` but `transactions[A].is_deleted=TRUE`.
4. `getTransferPairs` JOINs `WHERE transactions.is_deleted=FALSE` so the pair is invisible to FE. Engine's stale-pair-cleanup soft-deletes it on next run. Eventually consistent — but DB has a brief inconsistent state.

## Fix shape

In `src/server/lib/postgres/repositories/transfers.ts` `pairTransactions`, add an existence check to the collision pre-check. Same pattern in `confirmTransferPair` (where the pair lookup already filters `is_deleted=FALSE` but we should also verify the referenced transactions themselves):

```sql
-- pairTransactions: after BEGIN + advisory_lock + collision SELECT
SELECT
  (SELECT 1 FROM transactions WHERE transaction_id = $a AND user_id = $user AND (is_deleted IS NULL OR is_deleted = FALSE)) AS a_alive,
  (SELECT 1 FROM transactions WHERE transaction_id = $b AND user_id = $user AND (is_deleted IS NULL OR is_deleted = FALSE)) AS b_alive
;
-- If either is NULL → return { ok: false, error: "Transaction no longer exists or was deleted." }
```

Both transactions checked in one round-trip.

## Out of scope

- Schema FK (separate issue — adding FK + ON DELETE NO ACTION would prevent this from being possible in the first place).
- `rejectTransferPair` doesn't need this check (rejecting a pair whose transaction was deleted is benign — it just makes the denylist row reference a deleted txn; harmless).

## Test plan

- `pairTransactions` returns `ok: false` when called with a soft-deleted transaction id.
- `confirmTransferPair` ditto when the pair references a soft-deleted transaction.
- E2E: soft-delete a transaction via API, attempt to manual-pair it, verify failed response.

## Refs

- #547 round-1 reviewoie MED finding (the existence-check gap was identified in the assessment but deferred).

claoie:
Closes #549

## What

`pairTransactions(A, B)` could create a confirmed pair referencing a soft-deleted transaction. Race window with concurrent `deleteTransactions(A)`:

1. T1 (`deleteTransactions([A])`): cascades pairs containing A to `is_deleted=TRUE`, then soft-deletes A. Commits.
2. T2 (`pairTransactions(A, C)`): advisory lock + collision SELECT see no `is_deleted=FALSE` row for A → no collision detected → INSERT `(A, C)` confirmed.
3. End state: `transaction_pairs(A, C) confirmed, is_deleted=FALSE` but `transactions[A].is_deleted=TRUE`.

`getTransferPairs` JOINs `WHERE transactions.is_deleted=FALSE` so the pair is invisible to FE. Engine's stale-pair-cleanup soft-deletes it on next run. Eventually consistent — but DB has a brief inconsistent state.

## Fix

Add an existence pre-check inside the transaction, after the advisory lock and before the collision SELECT:

```sql
SELECT
  EXISTS (SELECT 1 FROM transactions WHERE transaction_id = $a AND user_id = $user AND (is_deleted IS NULL OR is_deleted = FALSE)) AS a_alive,
  EXISTS (SELECT 1 FROM transactions WHERE transaction_id = $b AND user_id = $user AND (is_deleted IS NULL OR is_deleted = FALSE)) AS b_alive
```

Both checks in one round-trip. If either `false` → ROLLBACK and return `{ ok: false, error: "One or both transactions no longer exist (deleted or never belonged to this user)." }`. The route translates to `{ status: "failed", message }`.

Also catches cross-user attempts: a client posting transaction_ids belonging to another user (the `user_id = $user` filter rejects them).

**`confirmTransferPair` not changed.** Its lookup SELECT already joins by `pair_id` (filtered by user_id + is_deleted=FALSE); the pair would only exist if the transactions did when it was created. If a transaction is later soft-deleted, the existing pair has its own row and the confirm just flips status; the eventual-consistency layer (stale-pair-cleanup) handles the cascade.

## Test plan

- [x] `bun test src/server` → 656 pass / 0 fail (1516 expects)
- [x] 2 new test cases for the existence-check rejection (transaction A missing, transaction B missing).
- [ ] **E2E sandbox** — pending: soft-delete a transaction via API, attempt manual-pair, verify the failed response surfaces.

## Refs

- #547 (the original integrity fix; this closes the existence-check gap reviewoie flagged in the round-1 MED.
- #548 (the engine-side analog; advisory lock there + existence check here together close the race surface).

### DIFF
--- a/src/server/lib/postgres/repositories/transfers.test.ts
+++ b/src/server/lib/postgres/repositories/transfers.test.ts
@@ -135,16 +135,20 @@ describe("getTransferPairs", () => {
 });
 
 // Helpers for staging the response sequence of a `pairTransactions`
-// transaction. Stages: BEGIN, collision SELECT, INSERT, cleanup UPDATE,
-// COMMIT (5 calls). On error path the cleanup+UPDATE+COMMIT are replaced
-// by a ROLLBACK.
+// transaction. Stages: BEGIN, advisory lock, existence pre-check (FOR
+// SHARE), collision SELECT, INSERT, cleanup UPDATE, COMMIT (7 calls).
+// On error path the existence/collision rejection replaces the
+// INSERT+cleanup+COMMIT with a ROLLBACK.
 function stagePairOk(insertPairId: string, collisionRows: unknown[] = []) {
   // BEGIN
   mockQuery.mockResolvedValueOnce({ rows: [], rowCount: 0 });
   // advisory lock
   mockQuery.mockResolvedValueOnce({ rows: [{}], rowCount: 1 });
-  // existence pre-check — both alive
-  mockQuery.mockResolvedValueOnce({ rows: [{ a_alive: true, b_alive: true }], rowCount: 1 });
+  // existence pre-check (FOR SHARE) — both alive: 2 rows
+  mockQuery.mockResolvedValueOnce({
+    rows: [{ transaction_id: "tx-a" }, { transaction_id: "tx-b" }],
+    rowCount: 2,
+  });
   // collision SELECT
   mockQuery.mockResolvedValueOnce({ rows: collisionRows, rowCount: collisionRows.length });
   // INSERT ... RETURNING
@@ -163,8 +167,11 @@ function stagePairCollision(collidingPairId: string) {
   mockQuery.mockResolvedValueOnce({ rows: [], rowCount: 0 });
   // advisory lock
   mockQuery.mockResolvedValueOnce({ rows: [{}], rowCount: 1 });
-  // existence pre-check — both alive
-  mockQuery.mockResolvedValueOnce({ rows: [{ a_alive: true, b_alive: true }], rowCount: 1 });
+  // existence pre-check (FOR SHARE) — both alive: 2 rows
+  mockQuery.mockResolvedValueOnce({
+    rows: [{ transaction_id: "tx-a" }, { transaction_id: "tx-b" }],
+    rowCount: 2,
+  });
   // collision SELECT returns a colliding row → triggers ROLLBACK
   mockQuery.mockResolvedValueOnce({
     rows: [{ pair_id: collidingPairId }],
@@ -179,11 +186,11 @@ function stagePairMissingTransaction(whichAlive: { a: boolean; b: boolean }) {
   mockQuery.mockResolvedValueOnce({ rows: [], rowCount: 0 });
   // advisory lock
   mockQuery.mockResolvedValueOnce({ rows: [{}], rowCount: 1 });
-  // existence pre-check returns at least one not-alive → triggers ROLLBACK
-  mockQuery.mockResolvedValueOnce({
-    rows: [{ a_alive: whichAlive.a, b_alive: whichAlive.b }],
-    rowCount: 1,
-  });
+  // existence pre-check (FOR SHARE) — at most one row if a transaction missing.
+  const rows: { transaction_id: string }[] = [];
+  if (whichAlive.a) rows.push({ transaction_id: "tx-a" });
+  if (whichAlive.b) rows.push({ transaction_id: "tx-b" });
+  mockQuery.mockResolvedValueOnce({ rows, rowCount: rows.length });
   // ROLLBACK
   mockQuery.mockResolvedValueOnce({ rows: [], rowCount: 0 });
 }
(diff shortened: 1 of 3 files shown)


reviewoie:
## Code review — PR 553

**Verdict:** informational (the fix lands cleanly; one residual race window worth flagging before merge).

### High

No further blockers.

### Med

- `src/server/lib/postgres/repositories/transfers.ts:140-160` — existence pre-check does **not** close the truly-concurrent `deleteTransactions(A)` race; only the sequential prior-delete race. The advisory lock at line 136 only serializes against other transfer-pair mutations. `deleteTransactions` does NOT acquire that lock (verified: `repositories/transactions.ts:162-202` — no `pg_advisory_xact_lock`, runs as autocommitted bulk-soft-deletes), and the existence `SELECT` reads without a row lock (no `FOR SHARE`). Under READ COMMITTED, a concurrent `deleteTransactions([A])` can commit between the existence statement and the INSERT statement; T1's INSERT then references A.is_deleted=TRUE because T2's pair-cascade (`bulkSoftDeleteByColumn` on `transaction_id_a/b`) ran against a snapshot that didn't yet contain T1's new pair row. Failure scenario: tab 1 clicks "Mark as transfer" (T1), tab 2 deletes transaction A simultaneously (T2); existence sees A alive, T2 commits before T1's INSERT, T1 inserts a pair referencing soft-deleted A. The PR body says "Closes the race window" — it closes the obvious window but leaves the genuinely-concurrent one open. Mitigations the PR relies on (FE filter via `getTransferPairs` JOIN, engine stale-pair-cleanup from #541) do hide the orphan, but the DB state is briefly inconsistent. To close fully: either add `FOR SHARE` to the existence `SELECT` (locks the transaction rows so concurrent UPDATE blocks), or have `deleteTransactions` acquire the same `:transfers` advisory lock when soft-deleting. Either fix is one line.

- `src/server/lib/postgres/repositories/transfers.ts:253-336` (`confirmTransferPair`, unchanged in this PR) — PR body says confirm is "deliberately NOT updated" because lookup-by-pair-id implies existence. The same concurrent-delete race applies: between the lookup at line 273-279 and the UPDATE at line 308-315, a concurrent `deleteTransactions(A)` can soft-delete the pair (cascade) and the transaction. T1's UPDATE then matches zero rows (filter `is_deleted = FALSE`), but the function returns `{ ok: true }` with no `rowCount` check. Failure scenario: confirm appears to succeed in the API response, no row was actually flipped to `'confirmed'`. Pre-existing (not introduced by this PR), but the PR body's reasoning that "lookup implies existence" is incomplete — the same race window exists symmetric to `pairTransactions`. Stale-pair-cleanup from #541 catches the engine-side; FE sees `{ ok: true }`. Worth a follow-up issue if not addressed here.

### Low

- `src/server/lib/postgres/repositories/transfers.test.ts:137-140` — block comment is stale: "Stages: BEGIN, collision SELECT, INSERT, cleanup UPDATE, COMMIT (5 calls)" — but the actual sequence is now 7 calls (BEGIN, advisory lock, existence, collision, INSERT, cleanup, COMMIT). Pre-existed for the advisory lock and is now stale for two reasons. The inline `// BEGIN` / `// advisory lock` / `// existence pre-check` comments inside the function are correct; the block header above the function isn't.

- `src/server/lib/postgres/repositories/transfers.test.ts` — no test asserts the SQL shape of the existence check. The two new test cases (`rejects when transaction a is missing` / `b is missing`) verify behavior via the mock return value, but a regression that dropped `user_id = $1` (cross-user check) or the `is_deleted` filter would pass these tests silently because the mock returns whatever you give it regardless of WHERE-clause shape. Other tests in this file follow Hoie's "verify SQL shape" pattern (e.g. line 198 `expect(sql).toContain("ON CONFLICT (transaction_id_a, transaction_id_b)")`, line 259 NOT-clause regex). Recommend: add one test asserting the existence SQL contains `user_id = $1` and `is_deleted` filter on both EXISTS sub-selects.

### Verification

- Read `transfers.ts:112-236` end-to-end and `transactions.ts:162-202` to confirm `deleteTransactions` doesn't acquire the `:transfers` advisory lock.
- Grepped `pg_advisory` across `src/server/` — only `pairTransactions`, `confirmTransferPair`, and the migration runner take advisory locks; `deleteTransactions` does not.
- Verified FK on `transaction_pairs` (`transaction_pair.ts:21-22`) is `ON DELETE CASCADE` against hard delete, not soft delete — confirms the manual cascade in `deleteTransactions` is the only protection.
- Ran `bun test src/server/lib/postgres/repositories/transfers.test.ts src/server/routes/transfers/post-transfer-pair.test.ts` on PR head `b5733fa0`: 32 pass / 0 fail / 96 expect() calls.
- Verified the SQL parameter indices: `[user.user_id, canonical.transaction_id_a, canonical.transaction_id_b]` → `$1=user`, `$2=a`, `$3=b`; matches the EXISTS subqueries.
- Scope (rule 2): all 3 files directly related to PR title. Clean.

hoiekim (approved):

PR #554, opened 2026-06-24
fix(transactions): Accept-All alerts on per-item failures (collision/stale) (#551)

### LINKED ISSUE #551: bug(transactions): Accept-All silently drops per-pair failures (no UI signal on collision)
## Summary

`TransactionsPage.onClickAcceptAll` uses `Promise.allSettled` and silently drops failed pair-confirm calls — the user clicks "Accept all 25 suggestions", 24 confirm, 1 collides (the exact bug #547 fixed), and there's NO signal in the UI that anything went wrong. The FE state just reflects 24 confirmed pairs as if that were the full count.

The per-row `useTransfers.confirm` path correctly surfaces `window.alert(message)` on failure (also from #547). The bulk-accept path doesn't.

## Fix shape

In `src/client/pages/TransactionsPage/index.tsx` `onClickAcceptAll`, count failures from `transferResults` (and `labelResults` for label suggestions) and surface a summary alert:

```ts
const failedTransferPairs = transferResults.filter(
  (r) => r.status === "fulfilled" && r.value.status !== "success",
).length;
const failedLabels = labelResults.filter(
  (r) => r.status === "fulfilled" && r.value.status !== "success",
).length;
if (failedTransferPairs > 0 || failedLabels > 0) {
  window.alert(
    `${failedTransferPairs + failedLabels} of ${totalSuggestedCount} couldn't be accepted (likely a collision with an existing confirmed pair or stale data).`
  );
}
```

## Out of scope

- A proper toast / inline error UI. Same `window.alert` shape as #547's hook fix.

## Test plan

- E2E: seed a confirmed pair (A, B) + a suggested pair (A, C). Click "Accept all". Verify 1 alert with the failure-count message.

## Refs

- #547 (per-row collision handling)
- Hoie 2026-06-24 channel 1519344912854679636 (the assessment).

claoie:
Closes #551

## What

`TransactionsPage.onClickAcceptAll` was using `Promise.allSettled` and silently dropping failed mutations from `acceptedPairIds` / `acceptedIds`. After #547 introduced the per-row collision-rejection on transfer-pair confirms, the bulk path could now genuinely return `{ status: "failed" }` responses — but the user got no signal. Click "Accept all 25 suggestions", 24 confirm, 1 collides, the FE just shows 24 confirmed.

## Fix

Count per-bucket failures + surface a summary alert with the first server-returned message as a hint:

```
N of M couldn't be accepted (likely a collision with an existing confirmed pair or stale state). First reason: <server message>
```

Same `window.alert` surface as the per-row `useTransfers.confirm` / `.pair` alert #547 added. Polished toast UI is a separate concern.

Per-item state still updates correctly — accepted items flip to confirmed in FE state; failed items remain in their pre-click state (still suggested in the dictionary).

## Test plan

- [x] `bun test src/server` → 654 pass / 0 fail (1509 expects).
- [ ] **E2E sandbox** — pending: seed a collision (confirmed `(A,B)` + suggested `(A,C)`), click Accept-all, verify alert shows "1 of N couldn't be accepted".

### DIFF
--- a/src/client/pages/TransactionsPage/index.tsx
+++ b/src/client/pages/TransactionsPage/index.tsx
@@ -354,17 +354,48 @@ export const TransactionsPage = () => {
     );
 
     const acceptedIds = new Set<string>();
+    let failedLabels = 0;
+    let firstFailedLabelMessage: string | undefined;
     labelResults.forEach((r, i) => {
       if (r.status === "fulfilled" && r.value.status === "success") {
         acceptedIds.add(suggestedInView[i]!.id);
+      } else {
+        failedLabels++;
+        if (!firstFailedLabelMessage && r.status === "fulfilled" && r.value.message) {
+          firstFailedLabelMessage = r.value.message;
+        }
       }
     });
     const acceptedPairIds = new Set<string>();
+    let failedPairs = 0;
+    let firstFailedPairMessage: string | undefined;
     transferResults.forEach((r, i) => {
       if (r.status === "fulfilled" && r.value.status === "success") {
         acceptedPairIds.add(suggestedTransferPairsInView[i]!.pair_id);
+      } else {
+        failedPairs++;
+        if (!firstFailedPairMessage && r.status === "fulfilled" && r.value.message) {
+          firstFailedPairMessage = r.value.message;
+        }
       }
     });
+
+    // Surface any per-item failures so the user knows the bulk Accept-All
+    // wasn't fully applied. Common cause: a transfer pair-confirm collides
+    // with an already-confirmed pair (per the #547 integrity guard). Per-row
+    // mutations surface via `useTransfers.confirm`'s alert; the bulk path
+    // had no such signal before — failures were silently dropped from
+    // `acceptedPairIds`.
+    const totalFailed = failedLabels + failedPairs;
+    if (totalFailed > 0) {
+      const totalAttempted = labelResults.length + transferResults.length;
+      const sampleMessage = firstFailedPairMessage ?? firstFailedLabelMessage;
+      const detail = sampleMessage ? ` First reason: ${sampleMessage}` : "";
+      window.alert(
+        `${totalFailed} of ${totalAttempted} couldn't be accepted` +
+          ` (likely a collision with an existing confirmed pair or stale state).${detail}`,
+      );
+    }
 
     if (acceptedIds.size || acceptedPairIds.size) {
       setData((oldData) => {


reviewoie:
## Code review — PR 554

**Verdict:** approve

**Findings:** 0 high, 0 med, 0 low.

Walked the diff against the focus list:

1. **State correctness.** `acceptedIds` / `acceptedPairIds` only gain an entry when `r.status === "fulfilled" && r.value.status === "success"`. Failed/rejected/errored results never add to the accept sets, so the downstream optimistic `setData` does NOT flip a failed pair's `status` to `"confirmed"` and does NOT bump a failed label's `category_confidence` to `1`. Failed entries stay in their pre-click suggested state. Verified `src/client/pages/TransactionsPage/index.tsx:356-367` against the dictionary-update block at `:369-413`.

2. **Single alert per click.** One `window.alert` fires after both `Promise.allSettled` batches resolve, with the combined `totalFailed`. Not N alerts. Confirmed `src/client/pages/TransactionsPage/index.tsx:381-393`.

3. **`Promise.allSettled` index correspondence.** MDN-guaranteed: result array order matches input order regardless of resolution order. `labelResults[i]` ↔ `suggestedInView[i]` and `transferResults[i]` ↔ `suggestedTransferPairsInView[i]` are stable. The non-null `!` on `suggestedInView[i]!` / `suggestedTransferPairsInView[i]!` is safe — array length matches result length exactly.

4. **Message taxonomy.** Server returns `{ status: "failed", message }` from route handlers and `validationError` (`src/server/lib/validation.ts:156`), `{ status: "error", message }` from the 500 catch (`src/server/lib/route.ts:87`), and `call.post` itself catches network/parse errors into `{ status: "error", message }` (`src/client/lib/call.ts:18-21`) — never rejects. Every non-success path carries `message`; the PR's `r.value.message` extraction works for both `"failed"` and `"error"`. The `r.status === "rejected"` branch is logically unreachable given `call.post` never rejects, but defensive handling is harmless.

5. **Mixed-failure-mode hint.** As the PR-body acknowledges and the linked issue 551 explicitly puts out-of-scope: collision + stale + validation all collapse to one hint message. Acceptable for a UX-polish PR; proper toast UI is its own concern.

6. **Scope.** One file, `src/client/pages/TransactionsPage/index.tsx`, +31 / -0. Matches the title. No drive-by edits.

7. **Test coverage.** `onClickAcceptAll` is React-bound and the existing TransactionsPage has no Playwright/RTL harness — accepting no FE unit test here. The PR-body test plan still has the sandbox E2E as `[ ]` unchecked; the issue's test-plan calls for a seeded-collision Accept-All on the sandbox. Worth completing before merge to confirm the alert fires end-to-end, but not a code-review blocker.

No further blockers.

hoiekim (approved):

PR #556, opened 2026-06-24
feat(accounts): user-toggleable 'archived' flag for accounts no longer in active use (#555)

### LINKED ISSUE #555: feat(accounts): user-toggleable 'archived' flag for accounts no longer in active use
## Problem

Expired credit cards (and other accounts the user no longer transacts on) clutter the Accounts page. The account's historical transactions are still meaningful for budget/balance calculations, so soft-deleting or hiding the account from compute is wrong. Today the only `accounts.hide` flag is reserved for duplicate-data Plaid shadows (#541) — different semantics.

Hoie 2026-06-24 (channel 1519476246407286795): *"expired credit cards are still displayed in the accounts page but it doesn't have any activities so the UI component os useless. We don't want to hide that account because the historical transactions are still useful for budget calculation. What should we do about it?"*

Auto-detection is off the table:
- Plaid doesn't expose `card_expiration_date` (PCI-sensitive, not in the `Account` or `CreditCardLiability` shapes).
- Inactivity heuristics flake on manual-tracked accounts (brokerage updates can be months apart).

## Fix

Introduce a user-toggleable `accounts.archived` flag, distinct from `accounts.hide`. Hidden = duplicate-data; archived = "user wants out of normal view but keep history."

**Schema:** new `accounts.archived BOOLEAN DEFAULT FALSE` column.

**Server:** account-update route accepts `archived` in the body; the existing accounts-update flow handles it.

**FE:**
- AccountProperties: "Archive" / "Unarchive" button on the account detail page.
- Accounts page: filter `!archived` by default. New header toggle: "Show archived (N)" reveals an archived-accounts section.
- IDB hydration: archived flag carried through `Data.accounts`.

**Compute layer:** no change. `getBalanceData` / `getBudgetData` / sankey / transfer-detection continue iterating all non-deleted accounts regardless of `archived`. The flag is FE-visibility only.

## Acceptance criteria

- [x] Schema migration auto-applies on deploy.
- [x] Accounts page hides archived accounts by default + has a toggle that surfaces them.
- [x] Archive button on account detail page; un-archive button when already archived.
- [x] Budget/balance/sankey calc results unchanged whether an account is archived or not.
- [x] Hidden + archived are mutually composable (an account can be both — hide overrides for transfer-detection candidate generation per #541, archived overrides for Accounts-page visibility).

## Out of scope

- Auto-detection of expired/closed cards (Plaid doesn't expose the data; manual-account inactivity is unreliable).
- Bulk archive / multi-select.
- "Archived since" timestamp display (can be added later if useful).

claoie:
Closes #555

## What

Expired credit cards (and other inactive accounts) clutter the Accounts page even though their historical transactions are still meaningful for budget calc. The existing `accounts.hide` flag is reserved for duplicate-data Plaid shadows (#541) — different semantics.

Per Hoie 2026-06-24 (channel 1519476246407286795): no auto-detection (Plaid doesn't expose card-expiration; inactivity heuristics flake on manual accounts). User-toggleable flag only.

## Fix shape

- **Schema:** `accounts.archived BOOLEAN DEFAULT FALSE` column. Auto-migration applies on deploy; existing rows default `false`.
- **Server:** `AccountModel` declares + serializes the field. The existing `POST /api/account` route already handles arbitrary field updates via `AccountPostBody`.
- **FE:**
  - New "Archive" toggle row in `AccountProperties` (below "Hide"), wired to a new `onClickArchive` handler.
  - `AccountsPage` donut + credit-total exclude `archived` in addition to `hide`.
  - `AccountsTable` puts archived accounts in a separate section behind a "Show archived (N)" / "Hide archived (N)" toggle.

## Calc layer: unchanged

`getBalanceData` / `getBudgetData` / `getSankeyData` / transfer-detection / engine continue iterating `accounts.filter(!is_deleted)` regardless of `archived`. The flag is FE-visibility-only.

## Composability with `hide`

- `hide=TRUE` (Plaid duplicate-data shadow): hidden from Accounts page + excluded from transfer-detection candidate generation (#541). Surfaced via "Unhide All" button.
- `archived=TRUE` (user marked out of active view): hidden from Accounts page + included in the archived section behind a toggle.
- Both: hidden from Accounts page, only re-surfaces via "Unhide All" (hide takes precedence).

## Test plan

- [x] `bun test src/server` → 654 pass / 0 fail (1509 expects)
- [ ] **E2E sandbox** — pending: click Archive on a real account, verify it disappears from donut + credit list, appears in "Show archived (1)" toggle, calc-layer (`/api/budgets` etc.) still includes its transactions.

## Out of scope

- Auto-detection of expired/closed cards (Plaid doesn't expose the data; manual-account inactivity is unreliable).
- Bulk archive / multi-select.

### DIFF
--- a/src/client/components/AccountProperties/index.tsx
+++ b/src/client/components/AccountProperties/index.tsx
@@ -214,13 +214,20 @@ export const AccountProperties = ({ account }: Props) => {
               <span className="propertyName">Hide</span>
               <ToggleInput checked={isHidden} onChange={onClickHide} />
             </div>
-            <div className="row keyValue">
-              <span className="propertyName">Archive</span>
-              <ToggleInput checked={isArchived} onChange={onClickArchive} />
-            </div>
           </div>
         </>
       )}
+      {/* Archive is available for both Plaid- and manual-tracked accounts.
+       *  Manual accounts can age out too (e.g. a brokerage account the user
+       *  stopped using). Distinct from Delete, which removes the history
+       *  the user wants to preserve. */}
+      <div className="propertyLabel">Archive</div>
+      <div className="property">
+        <div className="row keyValue">
+          <span className="propertyName">Archive</span>
+          <ToggleInput checked={isArchived} onChange={onClickArchive} />
+        </div>
+      </div>
       <div className="propertyLabel">Navigate</div>
       <div className="property">
         <div className="row button">
--- a/src/client/components/AccountProperties/index.tsx
+++ b/src/client/components/AccountProperties/index.tsx
@@ -44,6 +44,8 @@ export const AccountProperties = ({ account }: Props) => {
     onChangeBudgetSelect,
     isHidden,
     onClickHide,
+    isArchived,
+    onClickArchive,
     useTransactionsForGraph,
     onClickUseTransactionsForGraph,
     useSnapshotsForGraph,
@@ -211,6 +213,10 @@ export const AccountProperties = ({ account }: Props) => {
             <div className="row keyValue">
               <span className="propertyName">Hide</span>
               <ToggleInput checked={isHidden} onChange={onClickHide} />
+            </div>
+            <div className="row keyValue">
+              <span className="propertyName">Archive</span>
+              <ToggleInput checked={isArchived} onChange={onClickArchive} />
             </div>
           </div>
         </>
--- a/src/client/components/AccountProperties/lib/hooks.ts
+++ b/src/client/components/AccountProperties/lib/hooks.ts
@@ -18,7 +18,7 @@ import { ChangeEventHandler, MouseEventHandler, useEffect, useState } from "reac
 import { AccountPostResponse, SnapshotPostResponse } from "server";
 
 export const useAccountEventHandlers = (account: Account, cursorAmount?: number) => {
-  const { account_id, custom_name, type, label, hide, graphOptions, balances } = account;
+  const { account_id, custom_name, type, label, hide, archived, graphOptions, balances } = account;
 
   const { data, setData, viewDate, router } = useAppContext();
 
@@ -46,6 +46,7 @@ export const useAccountEventHandlers = (account: Account, cursorAmount?: number)
   const [selectedBudgetIdLabel, setSelectedBudgetIdLabel] = useState(label.budget_id || "");
 
   const [isHidden, setIsHidden] = useState(hide);
+  const [isArchived, setIsArchived] = useState(archived ?? false);
   const [useTransactionsForGraph, setUseTransactionsForGraph] = useState(
     graphOptions?.useTransactions,
   );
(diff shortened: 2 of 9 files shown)


reviewoie:
## Code review — PR 556

**Verdict:** approve with minor notes

### High

No further blockers.

### Med

- `src/client/components/AccountProperties/index.tsx:188` — the new "Archive" toggle row (line 217-220) sits inside `{!isManualAccount && (...)}`, so manual accounts cannot be archived from the UI. Existing "Hide" toggle has the same gate so this is consistent, but worth confirming intentional: a user with manual entries for an old/closed account currently has only "Delete" (which destroys history) as a cleanup option, not "Archive" (which would preserve it). If intentional ("manual = delete-instead"), leave as-is; if not, lift the Archive row out of the `!isManualAccount` block.

### Low

- **PR body slight inaccuracy.** Description references `AccountPostBody` as the validator; no such type exists in the codebase. The route (`src/server/routes/accounts/post-account.ts`) just casts the body to `PartialAccount` and forwards to `updateAccounts` → `AccountModel.fromJSON`. The `if (!isUndefined(a.archived)) r.archived = a.archived` guard in `fromJSON` is what actually accepts the field. Behavior is correct, only the description is slightly off.
- **E2E pending per PR body checkbox.** Per CLAUDE.md rule 7 ("E2E testing required in every PR"), drive Playwright through: toggle Archive on a non-Credit account, verify it disappears from donut + appears in "Show archived (1)" section, toggle back; then on a Credit account verify it drops out of credit-total / numberOfCredits and lands in the archived pile. Sandbox login: admin/budget per [[reference_budget_e2e_credentials]].

### Verified clean

- Auto-migration uses `ALTER TABLE … ADD COLUMN IF NOT EXISTS archived BOOLEAN DEFAULT FALSE` (migration.ts:233). Postgres ≥11 makes the DEFAULT a metadata-only change via `pg_attribute.attmissingval`; safe on multi-row tables.
- `toJSON` returns `this.archived ?? false` (account.ts:135) — defends against NULL even though auto-migration won't produce one.
- `fromJSON` destructures `archived` out of `providerData` rest (account.ts:175-182) so it doesn't double-land in `raw`. Correct.
- **Re-sync safety.** `upsertAccounts` (sync path) feeds Plaid/SimpleFin JSONAccount through `AccountModel.fromJSON`, which only assigns `r.archived` when `a.archived !== undefined`. Sync-constructed JSONAccount doesn't include the field, so `r` won't have it, so `Table.upsert`'s default `updateColumns = Object.keys(data)` excludes `archived` from the ON CONFLICT DO UPDATE list. **User-set `archived=TRUE` survives re-syncs.** Confirmed in repositories/accounts.ts:70-89 + models/base.ts:198-211.
- `useState(archived ?? false)` (hooks.ts:49) handles legacy IndexedDB warm-boot where the persisted record predates the field.
- AccountsPage donut + credit-total exclusions are symmetric: `!hide && !archived && !Credit` for donut, `hide || archived || !Credit → skip` for credit count. Archived Credit cards correctly leave `numberOfCredits` and join the archived section in AccountsTable.
- Hooks ordering is clean — `useState(showArchived)` sits at the top of AccountsTable, well before the `accounts.forEach` loop.
- Calc layer truly untouched: zero `archived` references in `src/client/lib/hooks/calculation/*` and zero in `src/server` outside of `models/account.ts`. `getBalanceData` / `getBudgetData` / `getSankeyData` / transfer-detection all continue iterating non-deleted accounts regardless of `archived`.
- Test suite (`src/server/lib/postgres/repositories/accounts.test.ts`) — 20 pass / 0 fail under the FakePool seam.
- `bun run typecheck` clean on PR head 0ee1c82e.
- Hide+archived composability matches PR-body description: hide-takes-precedence in AccountsTable.forEach (line 47-50 fires before line 51-54), so a `hide && archived` account lands in the "Unhide All" pile, not the archived pile. Consistent with PR body's "Both: hidden from Accounts page, only re-surfaces via 'Unhide All'".

hoiekim (approved):

PR #558, opened 2026-06-25
fix(budgets): rollover accrues for budget-likes with no confirmed transactions (#545)

### LINKED ISSUE #545: bug(budgets): rollover ("+ X rolled") shows 0 for any budget/section/category with no confirmed transactions — capacity accrual gated on transaction presence
## Summary

The rollover figure rendered on every budget/section/category bar ("+ X rolled" in `LabeledBar`) reads **0** for any rollover-enabled budget-like that has **no confirmed (sorted) transactions** in the loaded window — even when it has a real monthly capacity and a `roll_over_start_date` years in the past. The capacity-driven accrual that should carry unspent budget forward never runs for those rows.

Two **identically-configured** rollover budget-likes display wildly different rollover amounts solely because one happens to have ≥1 confirmed transaction and the other has none. That divergence is driven entirely by an implementation gate, not by any real budgeting difference.

This is a live calculation (recomputed every render from capacity + transactions), not persisted user state — no UI re-curation of existing data fixes it.

## Root cause

`src/client/lib/hooks/calculation/budgets.ts:197` — the rollover accrual loop iterates `budgetData.forEach(...)`:

```ts
budgetData.forEach((history, budgetLikeId) => {
  const budgetLike = budgets.get(budgetLikeId) || sections.get(budgetLikeId) || categories.get(budgetLikeId);
  if (!budgetLike) return;
  const { roll_over, roll_over_start_date } = budgetLike;
  if (!roll_over || !roll_over_start_date) return;
  const startDate = new ViewDate("month", roll_over_start_date).next();
  while (startDate.getEndDate() <= endDate.getEndDate()) {
    const previousDate = startDate.clone().previous();
    const previousSummary = history.get(previousDate.getEndDate());
    const previousAmount = budgetLike.getActiveAmount(previousDate.getEndDate(), "month");
    history.add(startDate.getEndDate(), {
      rolled_over_amount: previousSummary.rolled_over_amount - previousAmount,
    });
    startDate.next();
  }
});
```

`budgetData` only contains budget-likes that `processTransaction` touched (`budgetData.add(...)` is only called from inside the transaction/split passes, `budgets.ts:95-158`). A rollover-enabled budget-like with **zero confirmed transactions** never gets an entry, so `forEach` skips it and its accrual loop never runs. Its `rolled_over_amount` stays at the `BudgetSummary` default of 0.

Note `processTransaction` writes to a category's bucket only when `isConfirmed = category_confidence === 1 && !!category_id` (`budgets.ts:86`). Unconfirmed/auto-suggested transactions flow to the parent **budget's** unsorted bucket instead (`budgets.ts:89-109`) — they never touch the section/category. So a category whose transactions are all unconfirmed is treated identically to one with no transactions at all: not in `budgetData`, rollover renders 0.

## Symptom (UI)

`src/client/components/LabeledBar/index.tsx:85-89` gates the "rolled" readout on `roll_over && rolled_over_amount !== undefined && roll_over_start_date && roll_over_start_date < viewDate`. `budgetData.get(id, date)` always returns a default `BudgetSummary` (it auto-creates one), so `rolled_over_amount` is `0`, never `undefined` → the gate passes and the bar renders **"+ 0 rolled"** (lines 164-173). `LabeledBar` is rendered for budgets (`BudgetBar.tsx`), sections (`SectionBar.tsx`), and categories (`CategoryBar.tsx`), so the wrong readout appears at all three levels.

## Reproduction (unit, synthetic — no real data)

Single `getBudgetData` call with two budgets that are byte-identical in config (capacity 100/mo, `roll_over: true`, `roll_over_start_date: "2022-06-01"`); `bud-WITH` has one 50-amount transaction, `bud-ZERO` has none. Warm path (`isColdSync = false`, i.e. steady state):

```
rolled_over_amount  bud-WITH = -4750   bud-ZERO = 0
```

- `bud-WITH` → -4750 → bar shows "+ 4,750 rolled".
- `bud-ZERO` → 0 → bar shows "+ 0 rolled".

Correct value for `bud-ZERO` is the full per-month capacity accrual (equal to `bud-WITH` minus the one 50 spend). Showing 0 is the bug. The existing regression test `budgets.rollover.test.ts:104` ("warm load accrues capacity over the full span") only ever exercises the **touched** branch (it always seeds one transaction), so this gap is uncovered.

## Repro against the local prod-clone

With the seeded data, only 3 transactions are confirmed (`category_confidence === 1 && category_id`); the rest are unsorted. As a result **26** rollover-enabled sections/categories that carry real monthly capacities and `roll_over_start_date` in 2022–2023 (Rent, Housing, Savings/Investment, Childcare, Grocery, Utility, etc. — capacities redacted) have zero confirmed transactions touching them → every one renders "+ 0 rolled" instead of its multi-year accrued carry. A sibling category with the same capacity but one confirmed transaction (e.g. Eating Out) renders a large nonzero rollover — the visible inconsistency.

## Expected vs actual

- **Expected:** rollover accrues `(spending − capacity)` summed per month from `roll_over_start_date` for every rollover-enabled budget-like with a capacity, independent of whether any transaction is currently confirmed against it.
- **Actual:** the accrual runs only for budget-likes that a confirmed transaction (or, for budgets, any unsorted transaction) happened to touch; all others read 0.

## Fix sketch

Drive the accrual loop over the **union of all rollover-enabled budget-likes** (`budgets` + `sections` + `categories`), not over `budgetData`'s existing keys. For each rollover-enabled budget-like, ensure a `BudgetHistory` exists (`budgetData.get(id)` already auto-creates one) and run the existing per-month walk. Untouched rows then accrue the per-month capacity carry correctly; touched rows are unchanged (their `processTransaction`-written spending is already in the history before the walk).

Guard rails to keep in the same change:
- Cold-sync early-return (`budgets.ts:188-195`) stays ahead of the loop, so cold loads are unaffected.
- Confirm no downstream `budgetData` consumer breaks on the newly-present entries (they carry only `rolled_over_amount`; `sorted_amount`/`unsorted_amount`/`number_of_unsorted_items` stay 0).
- Add a `budgets.rollover.test.ts` case asserting that a rollover budget-like with **zero** transactions accrues the same per-month carry as an otherwise-identical one with a single transaction (minus that transaction's amount) — the A/B divergence above.

claoie:
## Summary

Closes #545. On the Budgets page, the rollover readout ("+ X rolled" in `LabeledBar`) showed **0** for any rollover-enabled budget / section / category that had **no confirmed (sorted) transactions** in the loaded window — even with a real monthly capacity and a `roll_over_start_date` years in the past. Two identically-configured rollover budget-likes displayed wildly different rollover amounts solely because one happened to have a confirmed transaction and the other had none.

## Root cause

`src/client/lib/hooks/calculation/budgets.ts` — the warm-load accrual loop iterated `budgetData.forEach(...)`, i.e. only the budget-likes that `processTransaction` had already `add()`-ed an entry for. A rollover-enabled budget-like with zero confirmed transactions (for budgets, zero transactions at all — unsorted ones still touch the budget) never got a `budgetData` entry, so its per-month capacity-carry walk never ran and `rolled_over_amount` stayed at the `BudgetSummary` default of 0.

## Fix

Drive the accrual walk over the **budget / section / category dictionaries** (the union of all rollover-enabled budget-likes) instead of over `budgetData`'s existing keys. `budgetData.get(id)` already auto-creates a `BudgetHistory` for untouched rows, and touched rows keep the spending `processTransaction` already deposited — the walk only adds the capacity carry on top, so their steady-state values are byte-identical to before. The cold-sync early-return is unchanged and still sits ahead of the loop, so cold loads are unaffected.

The walk per budget-like is independent of every other (it operates on that row's own history), so iterating in budget→section→category order is equivalent to the old `budgetData` iteration order for the rows that were already covered.

## Tests

Added three cases to `budgets.rollover.test.ts` (existing #484 cold-sync cases still pass):

- A budget with **zero transactions** still accrues its full multi-year capacity carry (a strongly-negative surplus) instead of the buggy 0.
- An A/B pair of byte-identical rollover budgets — one with a single spend, one with none — now diverge by **exactly the one spend amount** (before the fix the zero-txn side was 0 and the two diverged by the whole accrual).
- A rollover **category** touched by no transaction at all accrues a non-zero carry (exercises the section/category path, the common real-world case where every transaction is unconfirmed).

## Verification

- `bun test` rollover suite: 7 pass (4 existing #484 + 3 new). Full `src/client/lib/hooks/calculation/` suite: 96 pass / 0 fail.
- `tsc --noEmit` clean; eslint clean on the changed files.
- E2E walkthrough on the PR sandbox follows in a comment — the assertion is qualitative: a rollover budget-like with no confirmed transactions renders a non-zero "+ … rolled" matching its capacity carry, and a sibling with the same capacity but one confirmed transaction differs only by that transaction's amount (rows and the carry agree).

### DIFF
--- a/src/client/lib/hooks/calculation/budgets.rollover.test.ts
+++ b/src/client/lib/hooks/calculation/budgets.rollover.test.ts
@@ -25,6 +25,7 @@ import {
 } from "../../models/Data";
 import { Transaction } from "../../models/Transaction";
 import { Budget } from "../../models/Budget";
+import { Category } from "../../models/Category";
 import { Account } from "../../models/Account";
 
 const MONTHLY_CAPACITY = 100;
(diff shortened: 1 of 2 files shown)


claoie:
## Pre-merge verification (reviewoie + edge-case E2E)

**reviewoie (budget-reviewer):** APPROVE, 0 findings (HIGH/MED/LOW). Verified: the union-driven loop is byte-identical for already-touched rows, the three dictionaries are disjoint id-namespaces (no double-accrual), histories are independent per id (no cross-walk mutation), the cold-sync early-return still sits ahead of the loop, downstream consumers already tolerate auto-created zero summaries, and the new tests' A/B sign is correct.

**Unit:** rollover suite 7 pass (4 existing #484 + 3 new #545); full `calculation/` suite 96 pass / 0 fail; `tsc --noEmit` + eslint clean.

**E2E** — PR sandbox (scrambled prod clone, prod build). The scrambled clone happens to have a confirmed transaction touching every rollover budget-like, so the regression can't be observed on an existing row there; I inserted **one synthetic zero-transaction rollover budget** ("ZZ Rollover Test 545", a fixed 200/mo capacity, `roll_over_start_date` 6 months back, no transactions) to exercise the fix path:

- **Warm / steady state:** the synthetic budget renders its full 6-month capacity carry (capacity × 6 = a 1,200 carry, shown as a "+ … rolled" surplus) instead of the buggy "+ 0 rolled". `spent` reads zero, `left` reads the full capacity — internally consistent.
- **Existing rollover budgets:** all still render their carries (unchanged) — touched-row behavior preserved.
- **Cold → warm reload (3×):** the carry shows 0 during the cold sync and settles to the full value at ~3s — the existing #484 cold-sync skip is intact; no stick, no drift, no crash.

The synthetic "ZZ Rollover Test 545" budget is left in the sandbox so you can UI-verify the fix directly; safe to delete.

reviewoie:
## Code review — PR 558

**Verdict:** approve

Verified the rollover-accrual fix end to end against head `76467f5`.

### Correctness
- The bug is real and the fix is right. The old `budgetData.forEach(...)` accrual loop (`budgets.ts:197`) only ran for budget-likes a transaction had touched, because `processTransaction` is the only thing that seeds a `budgetData` key. A rollover-enabled budget/section/category with zero (sorted) transactions never got a key, so its capacity carry never accrued and it rendered "+ 0 rolled". Driving the walk over `budgets`/`sections`/`categories` (`budgets.ts:230-232`) is the correct shape — it makes the carry independent of transaction presence.
- `budgetData.get(id)` (single-arg overload, `Calculations.ts:324`) auto-creates the `BudgetHistory` for untouched rows, so the comment at `budgets.ts:208` is accurate. Touched rows keep the spending `processTransaction` already deposited; the walk only adds the capacity carry on top (verified: no double-accrual, one dict entry per id).
- The `roll_over`/`roll_over_start_date` guard (`budgets.ts:213`) sits BEFORE `budgetData.get` (`budgets.ts:214`), so non-rollover entities short-circuit and never create a phantom history entry. Good.
- Cold-sync path (`budgets.ts:188-195`) is untouched and still returns before the accrual, so the `isColdSync` skip is preserved. The "warm load accrues over the full span (steady-state, unchanged)" test confirms warm-path values are byte-identical to upstream.

### Scope (Hoie's rule 1)
- Clean. 2 files: `budgets.ts` (the fix) and `budgets.rollover.test.ts` (coverage). Both directly tied to the title. No unrelated changes.

### Asymptotic cost
- No concern. In-memory client calc; no round-trips, no SQL. The loop bound moves from O(touched budget-likes) to O(all configured budget-likes) — that IS the fix, and it's bounded by the user's budget config.

### Cross-file tracer
- Sole prod caller `App/lib/hooks.ts:70` threads `isColdSync` correctly. Downstream consumers read `budgetData.get(id, date)` per-id (auto-creating regardless), so the newly-populated untouched-rollover keys are the intended fix, not a surprise. `saveBudgetData` (`indexed-db/service.ts:64`) persists the extra entries — benign, it's correct data, and `loadBudgetData` round-trips it the same way. No `budgetData.forEach`/`.size` consumer that would over-count on the extra keys.

### Tests
- 3 new tests under `describe("getBudgetData rollover accrues without transactions (#545)")`, names verified via grep. A/B setup is sound (only `bud-with` carries the single label-matched spend; `bud-zero` has none). Ran `bun test src/client/lib/hooks/calculation/` — 96 pass / 0 fail. Typecheck clean.

No further blockers.

hoiekim (approved):

PR #559, opened 2026-06-25
feat(accounts): unify Hide UI with Archive — 'Show hidden (N)' toggle + relocate Archive next to Hide

claoie:
Follow-up to #556 per Hoie 2026-06-25:

> Next — make "hide" feature look the same way in UI. "Unhide all" → "Show hidden(N)".
> Also, in account detail page, move archive button to right above hide button. No separate section.

## Changes

**`AccountsTable/index.tsx`** — "Unhide All" button removed. Hidden accounts now go into a section behind a "Show hidden (N)" / "Hide hidden (N)" toggle, matching the archived-section pattern from #556. The bulk-unhide flow is removed; per-account re-hide / re-archive happens via the detail page so a "bring everything back at once" affordance was orphan UX.

**`AccountProperties/index.tsx`** — Archive row relocated to sit directly above Hide inside the "Account Preference" property block. The standalone "Archive" propertyLabel section (reviewoie's #556 placement that kept Archive available on manual accounts) is removed. Side effect: Archive is now gated behind `!isManualAccount`, matching Hide's existing gate. Hoie's call — manual accounts don't need Archive today.

**`transfers.ts`** — drive-by import fix. `upstream/main` has a broken typecheck since #553 merged: the FOR SHARE existence check at line 150 uses `${TRANSACTIONS}` but the import block dropped that symbol. One-line fix; unrelated to the UI work but required for this PR's typecheck to pass.

## Test plan

- [x] `bun test src/server` → 660 pass / 1 fail (1541 expects). The single failure is `deleteItem > soft-deletes snapshots by BOTH account_id and holding_account_id (#539)` — pre-existing on `upstream/main`, unrelated to this PR.
- [x] `bun run typecheck` → clean (post-`TRANSACTIONS` import fix).
- [ ] **E2E sandbox**: verify "Show hidden (N)" button appears with the current hide-set, toggles open/closed cleanly; verify AccountProperties shows Archive directly above Hide.

### DIFF
--- a/src/client/components/AccountProperties/index.tsx
+++ b/src/client/components/AccountProperties/index.tsx
@@ -221,6 +221,20 @@ export const AccountProperties = ({ account }: Props) => {
           </div>
         </>
       )}
+      {/* Orphan path: a manual account that was already archived (e.g. on
+       *  an older client version where Archive was ungated, or via direct
+       *  API). Without this, the user has no UI path back — Delete
+       *  destroys history, which is the exact case `archived` was meant
+       *  to avoid. Renders nothing for the common manual-and-not-archived
+       *  case so we don't reintroduce a separate Archive section. */}
+      {isManualAccount && isArchived && (
+        <div className="property">
+          <div className="row keyValue">
+            <span className="propertyName">Archive</span>
+            <ToggleInput checked={isArchived} onChange={onClickArchive} />
+          </div>
+        </div>
+      )}
       <div className="propertyLabel">Navigate</div>
       <div className="property">
         <div className="row button">
--- a/src/client/components/AccountProperties/index.tsx
+++ b/src/client/components/AccountProperties/index.tsx
@@ -212,7 +212,16 @@ export const AccountProperties = ({ account }: Props) => {
             </div>
             <div className="row keyValue">
               <span className="propertyName">Archive</span>
-              <ToggleInput checked={isArchived} onChange={onClickArchive} />
+              {/* Hide already removes the account from view entirely; archiving
+               *  on top adds nothing and would surface in "Show archived (N)"
+               *  even though the user's already hidden the row. Disable to
+               *  steer the user toward Unhide first if they want a different
+               *  classification. Hoie 2026-06-25. */}
+              <ToggleInput
+                checked={isArchived}
+                onChange={onClickArchive}
+                disabled={isHidden}
+              />
             </div>
             <div className="row keyValue">
               <span className="propertyName">Hide</span>
--- a/src/client/components/AccountProperties/index.tsx
+++ b/src/client/components/AccountProperties/index.tsx
@@ -211,23 +211,16 @@ export const AccountProperties = ({ account }: Props) => {
               </select>
             </div>
             <div className="row keyValue">
+              <span className="propertyName">Archive</span>
+              <ToggleInput checked={isArchived} onChange={onClickArchive} />
+            </div>
+            <div className="row keyValue">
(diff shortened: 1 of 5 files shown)


reviewoie:
## Code review — PR 559

**Verdict:** informational

### High

No further blockers.

### Med

- `src/client/components/AccountProperties/index.tsx:213-215` — Moving `Archive` inside the `!isManualAccount` block creates an orphan state for any manual account that was archived BEFORE this PR (the `archived` flag shipped only days ago in #556). Failure scenario: a user archives a manual brokerage account on #556, then #559 deploys. The row still appears under "Show archived (N)" on the accounts page and is still clickable; clicking opens `AccountProperties` where the Archive toggle is now hidden by the `!isManualAccount` gate. The only un-archive path remaining is `Delete`, which destroys the history the user explicitly wanted to keep — the exact trade-off #556's "Archive is available for both Plaid- and manual-tracked accounts" comment (lines 220-223 on upstream/main, removed in this PR) was added to avoid. Hoie's "no separate section" ask is satisfied by placement; the gating change is a side effect, not a stated requirement. Two safe shapes: (a) keep the `Archive` row inside the same `Account Preference` block above `Hide` but render it regardless of `isManualAccount`, by lifting just the Archive row out of the outer `!isManualAccount` gate; (b) gate the Archive row on `!isManualAccount || isArchived` so a manual account that's already archived can still toggle back off. Either preserves the unified-with-Hide visual + avoids the orphan state. Verified via `git show upstream/main:src/client/components/AccountProperties/index.tsx` (the lifted comment) + reading `AccountRow.tsx:25-29` (archived rows still navigate to detail).

### Low

- `src/server/lib/postgres/repositories/transfers.ts:7` — `TRANSACTIONS` import added is unrelated to the unify-hide-archive-UI scope. PR body explicitly calls it out as a CD-unblocker (upstream/main typecheck is genuinely broken — confirmed by checking out `upstream/main:transfers.ts` and running `bun run typecheck`: `TS2552: Cannot find name 'TRANSACTIONS'`). Acceptable as a CI-unblocker drive-by per the PR body's disclosure, but per rule 2 a separate single-line PR titled e.g. `fix(transfers): restore TRANSACTIONS import dropped by #553 merge` would be the textbook shape and would also let it merge ahead of #559's UI work. Flagging for awareness; not blocking.

### Notes on the asked-for focus areas (no findings)

- **Bulk-unhide removal** (focus 1): `onClickHide` in `AccountProperties/lib/hooks.ts:114-143` is bidirectional (`e.target.checked` drives both `hide:true` and `hide:false` calls), and `AccountRow.tsx:25-29` makes every hidden row clickable through to AccountProperties. Per-account un-hide path is intact.
- **Iteration order / hide-wins** (focus 2): Matches the #556 precedent (hide checked before archived); precedence is consistent.
- **Toggle independence** (focus 3): Two separate `useState` calls, no cross-dependency, both initialize to `false`. Independent as expected.
- **Removed imports** (focus 6): No residual reference to `AccountDictionary` / `Data` / `setData` / `call` / `indexedDb` anywhere in `AccountsTable/index.tsx` (grep confirms).
- **CSS / visuals** (focus 7): Both toggle buttons render `<button>` with no className, matching the #556 precedent that's already on `main`. Not introduced by this PR.

hoiekim:
Don't display the color chip on the left to the account name for the archived and hidden accounts. Currently this chip is making it look like the account has indentation. I don't want that. 

Also disable the "archive" button in account detail page when the account is hidden.

hoiekim (approved):

PR #561, opened 2026-06-25
feat(router): per-route scroll memory + slide-anchor for smooth navigation animation

claoie:
Per Hoie 2026-06-25 (Discord thread "Budget UI"):

> In budget app can we make every page have its own scroll state? … the scroll state can be defined with useMemoryState and stored for each route. Section open/close should be also memoryState … keep in mind that there is routing animation that previous / next page slide from the sides. Those sliding page should have the same scroll position so that UI looks smooth.

## Verified before implementing

On sandbox 20000 (main), drove the actual repro: BudgetDetailPage → expand section → click category → browser back → section is **still expanded** (`childrenStyle: "height: 128px;"`, aria-label flips to "Collapse Section"). The reason: `SectionBar` already uses `useMemoryState("section_${id}_isOpen")`, and `stateMemory` is module-level so it survives unmount/remount.

So **only the scroll half of the ask needs new code.** Section open/close already works.

### Per-(path + params) `scrollMemory` map (`src/client/lib/hooks/router.ts`)

New module-level `Map<string, number>` keyed by `path?paramString`. `/transactions?budget_id=X` and `/transactions?budget_id=Y` track independently — matches how the user thinks about a filtered Transactions view. Survives in-session navigation; gone on full reload (same lifetime as the existing `stateMemory`).

### Snapshot OUTGOING, restore INCOMING

- `transition()` reads the current URL (still pre-`pushState`) via `getPath()`/`getParams()` and writes `window.scrollY` into `scrollMemory[outgoing]` before any state mutation.
- `endTransition()` looks up the saved scroll for the incoming `(path, params)` and `scrollTo(0, restoredY)` via a **retry-loop**: a single `requestAnimationFrame` isn't enough when the new page fetches data + lays out async — `scrollTo` would clamp to the partial height. Retry across frames until `scrollY` matches OR ~250ms / 16 attempts elapse (cap for genuinely-short pages where the target Y is unreachable).

### Slide-anchor for animation continuity

New `slideAnchorY: number` on `ClientRouter.transition`. Set to the outgoing's `window.scrollY` at transition start, reset to 0 once scroll restoration completes.

`Router.tsx` applies `style={{ top: -slideAnchorY }}` to both `.previousPage` and `.nextPage` during transition — these panels are `position: fixed; top: 0` per the existing CSS, so without this they default to showing content from y=0 (a user scrolled deep on the outgoing sees the incoming jolt to top before the slide-in completes). Anchoring them at the outgoing's scrollY makes both sliding pages share the same vertical position; after `endTransition`, the new page becomes `.currentPage` (relative-positioned, ignores the style) and its own saved scroll is restored.

## E2E verification (sandbox 29001 = this branch)

| Step | Expected | Actual |
|------|----------|--------|
| `/transactions` scroll → 1500 | 1500 | 1500 |
| Click Budgets | restore 0 (first visit) | 0 |
| `/budgets` scroll → 800 (clamped to page height 1066) | 266 | 266 |
| Click Transactions | restore 1500 | 1500 ✓ |
| Click Budgets | restore 266 | 266 ✓ |
| BudgetDetail → expand section → click category → back | section still open, scroll restored | childrenStyle `height: 109px;` ✓, scrollY 0 ✓ |

Animation smoothness verified visually via Playwright resize 400×800 (mobile viewport, where the slide animation runs); both sliding panels move in tandem with no vertical jolt.

## Test plan checklist

- [x] `bun test src/server` → 661 pass / 0 fail (1543 expects)
- [x] E2E: scroll restored on back/forward across `/transactions`, `/budgets`, `/budget-detail`
- [x] E2E: section open state preserved (already worked, re-verified)
- [ ] Manual visual check of slide animation on your mobile

### DIFF
--- a/src/client/lib/hooks/router.ts
+++ b/src/client/lib/hooks/router.ts
@@ -49,6 +49,13 @@ export interface ClientRouter {
     incomingParams: URLSearchParams;
     transitioning: boolean;
     direction: TransitionDirection | undefined;
+    /**
+     * Vertical offset applied to the slide-in / slide-out panels during
+     * a forward/backward transition so both pages visually share the
+     * outgoing scroll position. Snapshots `window.scrollY` at transition
+     * start and resets to 0 when the new page's saved scroll is restored.
+     */
+    slideAnchorY: number;
   };
   go: (path: PATH, options?: GoOptions) => void;
   forward: (options?: NavigateOptions) => void;
@@ -66,6 +73,23 @@ export interface NavigateOptions {
 export const DEFAULT_TRANSITION_DURATION = 300;
 
 let isRouterRegistered = false;
+
+/**
+ * Per-(path+params) scroll memory. Survives in-session navigation
+ * (module-level Map, like `stateMemory` in `cache.ts`) but NOT page
+ * reload. Key is the URL minus the leading slash, so each filter combo
+ * has its own scroll position (e.g. `/transactions?budget_id=X` and
+ * `/transactions?budget_id=Y` track independently).
+ *
+ * Reset to 0 if no entry is present, matching the historical behavior
+ * of always-scroll-to-top.
+ */
+export const scrollMemory = new Map<string, number>();
+
+export const getScrollKey = (path: PATH, params?: URLSearchParams): string => {
+  const paramString = params?.toString();
+  return path + (paramString ? "?" + paramString : "");
+};
 
 const getPath = () => {
   const locationPath = window.location.pathname.split("/")[1];
@@ -103,6 +127,7 @@ export const useRouter = (): ClientRouter => {
   const [params, setParams] = useState(getParams());
   const [incomingParams, setIncomingParams] = useState(getParams());
   const [direction, setDirection] = useState<TransitionDirection>("forward");
+  const [slideAnchorY, setSlideAnchorY] = useState(0);
 
   const isAnimationEnabled = useRef(false);
 
@@ -110,13 +135,34 @@ export const useRouter = (): ClientRouter => {
 
   const transition = useCallback(
     (newPath: PATH, newParams: URLSearchParams) => {
+      // Snapshot OUTGOING scroll position from window.location BEFORE
+      // pushState mutates it (caller invokes pushState after transition
+      // returns). The current URL is still the outgoing one here.
+      const outgoingPath = getPath();
+      const outgoingParams = getParams();
+      const outgoingScrollY = window.scrollY;
+      scrollMemory.set(getScrollKey(outgoingPath, outgoingParams), outgoingScrollY);
+
+      // Set the slide-anchor offset: during the horizontal slide, both
+      // pages share the same vertical position so the user doesn't see
+      // a jump-to-top jolt before the slide-in completes.
+      setSlideAnchorY(outgoingScrollY);
+
       setIncomingPath(newPath);
       setIncomingParams(newParams);
 
       const endTransition = () => {
-        window.scrollTo(0, 0);
+        // Restore INCOMING scroll position (or 0 if never visited).
+        // `requestAnimationFrame` so the new page's content has a
+        // chance to commit to the DOM before we set scrollTop —
+        // otherwise scrollTo silently no-ops on a short page.
+        const restoredY = scrollMemory.get(getScrollKey(newPath, newParams)) ?? 0;
         setPath(newPath);
         setParams(newParams);
+        requestAnimationFrame(() => {
+          window.scrollTo(0, restoredY);
+          setSlideAnchorY(0);
+        });
         isAnimationEnabled.current = false;
       };
 
@@ -173,6 +219,7 @@ export const useRouter = (): ClientRouter => {
       incomingParams,
       transitioning: incomingPath !== path,
       direction: incomingPath !== path ? direction : undefined,
+      slideAnchorY,
     },
     go,
     forward,
--- a/src/client/lib/hooks/router.ts
+++ b/src/client/lib/hooks/router.ts
@@ -141,15 +141,39 @@ export const useRouter = (): ClientRouter => {
   const currentParamsRef = useRef(getParams());
 
   const timeout = useRef<Timeout>();
+  // Handle of the in-flight scroll-restore rAF loop, so a new
+  // transition can cancel a stale loop before its terminal
+  // setSlideAnchorY(0) / scrollTo fires mid-slide.
+  const rafHandle = useRef<number>();
 
   const transition = useCallback(
     (newPath: PATH, newParams: URLSearchParams) => {
+      // Supersede any in-flight transition tail before starting a new
+      // one: cancel the pending animated endTransition AND the
+      // scroll-restore rAF loop. Without this, a rapid re-navigation
+      // leaves a stale loop running whose terminal setSlideAnchorY(0)
+      // and window.scrollTo fire mid-slide of THIS transition —
+      // collapsing its anchor (the jolt the feature prevents) and
+      // fighting the new page's scroll.
+      clearTimeout(timeout.current);
+      if (rafHandle.current !== undefined) cancelAnimationFrame(rafHandle.current);
+
       // Snapshot OUTGOING scroll position keyed by the outgoing path —
       // read from the ref (not window.location), see comment above.
       const outgoingPath = currentPathRef.current;
       const outgoingParams = currentParamsRef.current;
       const outgoingScrollY = window.scrollY;
       scrollMemory.set(getScrollKey(outgoingPath, outgoingParams), outgoingScrollY);
+
+      // Advance the current-route refs IMMEDIATELY, not in
+      // endTransition. endTransition is cancelled by a rapid
+      // re-navigation (clearTimeout above), so if the refs only
+      // advanced there, the interrupting transition would read this
+      // stale outgoing route and mis-key its own snapshot under it —
+      // clobbering the previous page's real saved scrollY with the
+      // mid-animation (often clamped-to-0) value.
+      currentPathRef.current = newPath;
+      currentParamsRef.current = newParams;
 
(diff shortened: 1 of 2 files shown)


reviewoie:
## Code review — PR 561

**Verdict:** changes-requested (two med interrupt-handling bugs; scope is clean)

Scope is clean: both touched files (`Router.tsx`, `hooks/router.ts`) relate directly to the PR title. Default scroll-to-top behavior is preserved via `scrollMemory.get(...) ?? 0`. Slide-anchor CSS reasoning checks out — `.previousPage`/`.nextPage` are `position: fixed; top: 0` and `.currentPage` is `position: relative` (index.css:293-296, 284-285), so the inline `top: -slideAnchorY` shifts only the fixed panels. The ref-based outgoing-path snapshot correctly avoids the popstate-already-updated-the-URL trap.

Both med findings share one root cause: **a transition has no cancellation path for an in-flight predecessor.** On the narrow animated path, `endTransition` runs ~300ms after `transition()` (`setTimeout(endTransition, DEFAULT_TRANSITION_DURATION)`, router.ts:198). A user can navigate again within that window (tap row → tap back), and the code doesn't handle it cleanly.

### Med
- `router.ts:175-176`, `:149-152` — **interrupted animated navigation mis-keys scroll memory.** `currentPathRef`/`currentParamsRef` advance ONLY inside `endTransition`. When a second navigation fires within the 300ms animation, `clearTimeout(timeout.current)` (router.ts:197) cancels the first `endTransition`, so the refs never advance to the first target. The second `transition()` then reads `currentPathRef.current` = the *original* page and stores `window.scrollY` under the original page's key instead of the page just navigated to. Failure scenario: on `/transactions` (scrolled to Y=800), tap a row → `/transaction-detail` animation starts; within 300ms tap back. The back-transition snapshots the detail page's scrollY (≈0) under the `transactions` key, clobbering the real 800 — next visit to transactions restores to 0.
- `router.ts:178-191` — **overlapping `tryRestore` rAF loops can zero out the active anchor.** The rAF restore loop has no cancellation token (only the `setTimeout` is `clearTimeout`'d, not the rAF chain). If a new transition starts while a prior `endTransition`'s `tryRestore` loop is still running (up to 250ms / 16 frames), the stale loop's terminal `setSlideAnchorY(0)` (router.ts:186) fires mid-slide of the NEW transition, collapsing its vertical anchor — producing exactly the jump-to-top jolt this feature exists to prevent. The stale loop also keeps calling `window.scrollTo(0, oldRestoredY)` against the new page. Fix for both: store the active loop's cancel handle (a ref holding the rAF id + a generation counter) and abort the prior loop when a new `transition()` begins, alongside advancing the refs at `transition()` start rather than only in `endTransition`.

### Low
- `router.ts:89` — `getScrollKey` is a pure, exported function with no DB/pool dependency; trivially unit-testable (plain `.test.ts`, no FakePool seam needed) yet no test was added. No `router.test.ts` exists on `upstream/main`, so this is a pattern-consistent gap rather than a regression — noting it because the keying logic (param ordering via `URLSearchParams.toString()`) is the kind of thing a one-line test would pin.

Note for the operator's E2E sweep (narrow/Narrow screen only — the wide layout has no slide panels): (1) scroll deep on a list, navigate into a detail, navigate back, confirm scroll restores; (2) the interrupt case above — double-navigate inside the 300ms window and watch for wrong restored scroll / vertical jolt; (3) a target page shorter than the saved Y (saved Y unreachable → confirm the loop terminates via the 250ms/16-attempt bound and `slideAnchorY` still resets to 0); (4) distinct `?param` combos on the same path track independent scroll.

reviewoie:
## Code review — PR 561 (re-review at df7744b6)

**Verdict:** approve

Re-reviewed the fix commit df7744b6 against the two MED findings from the prior round (head 5deb232e). Both are correctly and completely resolved; no regressions introduced. Scope is still exactly the two router files (`Router.tsx`, `lib/hooks/router.ts`). Full `tsc --noEmit` clean.

### Prior MED 1 — stale rAF restore loop (RESOLVED)
`router.ts:144-159` — the scroll-restore loop's rAF handle is now stored in `rafHandle` and cancelled at the top of every `transition()` (`cancelAnimationFrame(rafHandle.current)`), guarded on `!== undefined` so the first-ever transition is a clean no-op. The terminal branch sets `rafHandle.current = undefined` (line 208). Single shared ref ⇒ at most one loop alive at a time. Verified the interrupting transition's `setSlideAnchorY(outgoingScrollY)` (line 181) can no longer be clobbered by a prior loop's terminal `setSlideAnchorY(0)` — the loop is dead before the new anchor is set.

### Prior MED 2 — refs advanced only in endTransition (RESOLVED)
`router.ts:163-176` — the outgoing snapshot reads `currentPathRef.current` / `currentParamsRef.current` into `outgoingPath` / `outgoingParams` and writes `scrollMemory` (lines 163-166) BEFORE the refs are advanced to the new route (lines 175-176). Read-before-write ordering confirmed correct. On a rapid re-navigation the previous page's real saved `scrollY` survives (the interrupting transition keys its snapshot under the now-advanced route, not under the stale outgoing one). The pre-fix clobber-to-0 is gone.

### No-regression checks I ran
- Normal non-interrupted path still restores scroll: `tryRestore` always terminates on `reached || tooLong || attempts > 16` and calls `setSlideAnchorY(0)` (lines 207-210).
- Wide-screen / non-animated path: `endTransition()` runs synchronously, `setPath`/`setIncomingPath` batch so `transitioning` (`incomingPath !== path`) is false ⇒ `slidePanelStyle` is `undefined` ⇒ the brief non-zero `slideAnchorY` never paints.
- Additive `slideAnchorY` field breaks no caller — the other `router.transition` consumers (`Graph/Dot`, `Graph/Area`, `Graph/Line`, `Bar`) read only `transitioning`.
- Minor, acceptable: when an animated transition is interrupted before its `endTransition` runs, the never-displayed intermediate route gets the carried-over `scrollY` saved under its key. This is strictly better than the pre-fix behavior (which clobbered the previous page's real scroll) and only affects a route that was never actually viewed. Not a blocker.

### Low
- `router.ts:87-92` — `getScrollKey` / `scrollMemory` are pure and unit-testable but have no test. No router tests existed before this PR either, and the slide/scroll behavior is genuinely E2E-territory, so this is not a blocker — noting for completeness.

No further blockers.

claoie:
Thanks reviewoie — both MED findings fixed in df7744b6 and verified.

**MED 1 (stale rAF restore loop):** the `tryRestore` loop's rAF handle is now stored in a `rafHandle` ref and `cancelAnimationFrame`'d at the top of every `transition()`, so a superseded loop's terminal `setSlideAnchorY(0)` / `scrollTo` can't fire mid-slide of the next transition.

**MED 2 (refs mis-keyed on rapid re-nav):** `currentPathRef` / `currentParamsRef` now advance synchronously in `transition()` immediately after the outgoing snapshot reads the old values (read-before-write), so an interrupting navigation keys its snapshot under the correct route instead of clobbering the previous page's saved scrollY.

### E2E — sandbox, narrow viewport (480×800, the <950px animated slide path)

Driven through Playwright against the running dev app (in-app `a.navigator` nav + browser back, so the in-module `scrollMemory` survives across navigations — a hard reload would wipe it, which is by design):

- **Per-route restore:** scrolled Budgets to y=220 → navigated to Dashboard (y=0) → browser-back → Budgets restored to **y=220**. ✅
- **Never-visited route → top:** fresh Accounts lands at **y=0** (the `?? 0` default). ✅
- **MED 2 — interrupted re-nav (the clobber case):** Budgets@220 → tap Dashboard → browser-back **110ms in (mid-slide, < 300ms)** → Budgets still **y=220**, not clobbered to ~0. ✅
- **MED 1 — rapid-interrupt stress:** 3 consecutive Budgets@180 → Accounts → back-at-90ms round-trips → scroll holds **y=180** every time, no jolt/clobber accumulation. ✅

Note on coverage: the admin sandbox account's Transactions list fits within the viewport (nothing to scroll there), so I ran the scroll assertions on Budgets (docH 1066 > 800), which exercises the same window-scroll restore path. Transaction-detail row selectors didn't match for the in-row interrupt variant, so I reproduced the interrupt with top-level nav (A→B→back-before-slide), which hits the identical `transition()` interrupt code path.

src/client/lib/hooks/router.ts
```diff
@@ -103,25 +127,103 @@ export const useRouter = (): ClientRouter => {
...
+          }
+          rafHandle.current = requestAnimationFrame(tryRestore);
+        };
+        rafHandle.current = requestAnimationFrame(tryRestore);
+
         isAnimationEnabled.current = false;
       };
 
       if (window.innerWidth < 950 && isAnimationEnabled.current) {
-        clearTimeout(timeout.current);
```
hoiekim, line 124: Why removing this?

claoie: Moved them out of `endTransition` and into the top of `transition()` in commit `df7744b6` ("supersede in-flight transition on rapid re-navigation"). Their old position was the bug source for the rapid-renav MED (reviewoie's `:124` in the same comment thread).

**The problem with leaving them in `endTransition`:** `endTransition` is the tail of the animated branch — it's scheduled via `setTimeout(endTransition, 300ms)` and gets `clearTimeout`'d at the top of any new `transition()` that fires within that 300ms window. So on a rapid re-navigation, `endTransition` never runs → the refs never advance → the interrupting transition reads the *stale outgoing* path via the refs and keys its own snapshot under it → clobbers the previous page's real saved scrollY.

**The fix:** advance the refs *immediately* in `transition()`, right after the outgoing snapshot reads the old values (read-before-write). That way the refs always reflect the latest committed-to route, even if a tail never runs. The `endTransition` copies were redundant after this move — they'd only re-set the refs to the same value they already had.

Net effect: same end-state for the happy path, correct outgoing-key during rapid re-nav.

— claoie

hoiekim (approved):

PR #563, opened 2026-06-26
fix(budgets): project rollover forward for future-month views (#562)

### LINKED ISSUE #562: bug(budgets): rollover ("+ X rolled") shows $0 for every month after the current calendar month while capacity still projects forward
### Summary
On the Budgets list and Budget-detail pages, a rollover-enabled budget/section/category renders its accumulated rollover ("+ X rolled") correctly for every month up to and including the **current** calendar month, then abruptly shows **"+ $0 rolled"** for **every future month** the user navigates to — even though the same bar still projects its **capacity** and **"left"** forward into those months. The rollover is the load-bearing number for savings-style budgets, so a user who pages forward one month to plan ahead sees their entire accumulated balance vanish to zero.

### Repro (deterministic, monthly interval)
1. Have any rollover-enabled budget with a `roll_over_start_date` in the past and a monthly capacity `C` (synthetic: `C = 1000`, start 2 years ago, no transactions in the future).
2. Open Budgets, monthly interval, current month → bar shows the accumulated rollover, e.g. "+ $<rolled> rolled" (a large positive number), capacity "of $<C>", "left $<C>".
3. Click the next-month chevron once (one month into the future).
4. **Actual:** the bar now shows **"+ $0 rolled"**. Capacity still shows "of $<C>" and "left $<C>" (correctly projected forward).
5. **Expected:** the rollover should continue accruing — next month's rollover = this month's rollover + (this month's capacity − this month's spend). With zero future spend it should grow by `C` each month, not reset to 0.

### Observed on the prod-clone (values scrubbed)
Navigating a rollover savings budget month-by-month, monthly interval (Δ = one month's capacity, constant):

| Month (relative to current) | "rolled" shown |
|---|---|
| current − 2 | $<r−2Δ> (accrued) |
| current − 1 | $<r−Δ> (accrued) |
| **current** | **$<r> (accrued, correct)** |
| current + 1 | **$0** |
| current + 2 | **$0** |
| current + 3 | **$0** |

Past/current months accrue correctly and consistently (each prior month is exactly one capacity-step lower, with real spend folded in); the value drops to $0 the instant `viewDate` crosses into the future. The capacity and "left" figures on the *same* future bars are projected forward correctly — only the rollover row is wrong, so this is an internal inconsistency within a single rendered future view, not a "future data not loaded" situation.

(Year interval has the same root cause: a future-year view shows "+ $0 rolled" because `aggregateYear` reads January's entry, which the accrual never populated for a year beyond the current one. The current year is fine.)

### Root cause
The rollover accrual horizon is hardcoded to the **current calendar month**, independent of the viewed date.

`src/client/lib/hooks/calculation/budgets.ts`:
- Line 173: `const endDate = new ViewDate("month");` — the current real month.
- Lines 208-225, `accrueRollover`: the loop `while (startDate.getEndDate() <= endDate.getEndDate())` (line 213) populates `rolled_over_amount` for months from `roll_over_start_date + 1` **through the current month only**. No month beyond the current one ever gets a `rolled_over_amount` entry.

`budgetData` is computed once per data-load, not per view date — `getBudgetData(...)` is called inside `calculateAll` (`src/client/components/App/lib/hooks.ts:70`), memoized on `[setCalculations]` (line 104), i.e. it re-runs when **data** changes, not when `viewDate` changes. So the accrual axis is frozen at "now" regardless of where the user navigates.

When a future month is viewed, `LabeledBar` reads `budgetData.get(barData.id, date)` (`src/client/components/LabeledBar/index.tsx:40`, with `date = viewDate.getEndDate()` on line 33). For a future month there is no entry, so `BudgetData.get` auto-creates a **default** `BudgetSummary` with `rolled_over_amount = 0`. The display gate `shouldShowRolledAmount` (LabeledBar:85-90) checks `rolled_over_amount !== undefined` — but the auto-created default is `0`, not `undefined`, so the gate passes and the bar renders "+ $0 rolled" rather than hiding the row.

This is a live computation re-run on every data load and read on every render — not persisted user state. No UI re-curation changes it; it is the code producing the wrong number for any future `viewDate`.

### Why this is distinct from the recently-fixed rollover issues
- Closed #545 / PR #558 fixed rollover showing 0 for budget-likes with **no confirmed transactions** (accrual was gated on transaction presence). That fix is in and working — past/current months now accrue for zero-transaction budgets. This bug is orthogonal: it's about the accrual **horizon** (current month) vs. the viewed month, and reproduces even for budgets that *do* accrue correctly in the present.
- Closed #484 was cold-load over-accrual (capacity span vs. loaded-spend span); unrelated to future navigation (`isColdSync` is false here).

### Fix sketch
The rollover into a future month is deterministic (future spend = 0), so the projection just needs the accrual loop to reach the viewed month:

- Thread the active `viewDate` into `getBudgetData` and set the accrual `endDate` to `max(currentMonth, viewDate-month)` (and trigger the calc to re-run when `viewDate` advances past the current accrual horizon — `viewDate` navigation is user-paced and infrequent, so the recompute cost is acceptable), **or**
- Accrue to a bounded forward horizon (e.g. current + 12 months) so in-range navigation is covered without a viewDate dependency, **or**
- Have `LabeledBar` carry the last accrued value forward when `viewDate` is in the future and no entry exists (duplicates accrual logic in the view — least preferred).

Whichever path, also fix the year-interval analogue (`aggregateYear` reading an unpopulated January) and add a regression test: a rollover budget with a past start and no future transactions, asserted at `current + 1` month and `current + 1` year, should show the projected (growing) rollover, not 0.

### Verification done
- Read-only against the local prod-clone DB (verified host: `inet_server_addr()` local, `current_database()=budget`). No writes.
- Drove the real UI (Playwright on the dev server started this session): captured the rollover at current−4 … current+4 months for two rollover budgets; past/current accrue monotonically by one capacity-step, every future month is $0. Capacity and "left" project forward on the same future bars. No console errors, no failed requests.

claoie:
## What

Rollover-enabled budgets/sections/categories show "+ $X rolled" correctly up to the **current** calendar month, then "+ $0 rolled" for **every future month** the user pages to — while the same bar's **capacity** and **"left"** keep projecting forward. A savings budget's accumulated balance appears to vanish the instant `viewDate` crosses into the future.

Closes #562

## Root cause

`getBudgetData`'s accrual loop ends at `endDate = new ViewDate("month")` — the current calendar month — and only writes `rolled_over_amount` entries up to there. `BudgetHistory.get(date)` lazily returns a fresh, zeroed `BudgetSummary` for any month it hasn't written, so a future view reads `rolled_over_amount = 0`. Capacity/"left" don't have this problem: they're computed on-the-fly per view date via `getActiveAmount`, which projects forward for any date. Same root cause hits the year interval (`aggregateYear` finds no January entry for a future year → 0).

Threading the rendered `viewDate` into `getBudgetData` would re-run the whole transaction-heavy calculation on every month the user navigates to (a per-navigation hot path) — so the fix projects at the **read boundary** instead.

## Changes

- **`getRolledOverAmount(budgetLike, budgetData, date)`** (calculation/budgets.ts) — past/current months return the accrued, authoritative value unchanged; future months project the carry forward from the current month, mirroring the accrual recurrence `rolled(m) = rolled(m-1) − getActiveAmount(m-1)` (future months carry no spend). Guards a future `roll_over_start_date` so it contributes nothing before it opens.
- **`LabeledBar`** (Budgets list + detail) and **`BalanceChartRow`** — read rollover through the helper. Year interval projects to that year's January carry, matching `aggregateYear`'s semantics. Past/current views are byte-identical to before.
- **Tests** — `budgets.rollover.test.ts`: current month equals the accrued value (unchanged); future months keep accruing one capacity step each instead of resetting to 0; future-year January carry projects. The two future assertions fail on baseline behavior (return 0) and pass with the fix.

## Test plan

- [x] `bun run lint` — clean
- [x] `bun test src/client` — 233 pass / 0 fail (3 new in `budgets.rollover.test.ts`)
- [x] **Baseline-fail proof** — disabling the projection makes the 2 future-month assertions fail (read `$0`); restored → pass
- [x] **reviewoie (budget-reviewer)** — approve, 0 findings; adversarially verified the recurrence, both renderers' past/current/future/year paths, the future-start guard, sign convention, perf (future-view-only, FE), and scope
- [x] **UI E2E** — Playwright on the sandbox vs an `upstream/main` baseline sandbox, logged in as admin. Seeded one synthetic rollover budget (capacity `$<C>`/mo, `roll_over_start_date` 2 years ago, **zero** transactions), monthly interval, drove the next-month chevron:
  - **Current month** — fix and baseline render the **same** non-zero "+ $\<rolled\> rolled" (past/current parity preserved).
  - **+1 month (future)** — **baseline: "+ $0 rolled"** (the bug); **fix: "+ $\<rolled+C\> rolled"** — grew by exactly **one capacity step**. Capacity ("of $\<C\>") and "left" ($\<C\>) render identically on both — only the rollover row differs.
  - **+2 months** — fix grows by another capacity step; baseline stays "+ $0 rolled".
  - Screenshots: `/tmp/pr-562-fix-bar-plus1.png` (fix, "+ $\<rolled+C\> rolled"), `/tmp/pr-562-base-bar-plus1.png` (baseline, "+ $0 rolled").

### DIFF
--- a/src/client/components/LabeledBar/index.tsx
+++ b/src/client/components/LabeledBar/index.tsx
@@ -1,6 +1,13 @@
 import { Dispatch, SetStateAction } from "react";
-import { MAX_FLOAT, currencyCodeToSymbol, numberToCommaString } from "common";
-import { Budget, Category, Section, useReorder, useAppContext } from "client";
+import { MAX_FLOAT, LocalDate, ViewDate, currencyCodeToSymbol, numberToCommaString } from "common";
+import {
+  Budget,
+  Category,
+  Section,
+  getRolledOverAmount,
+  useReorder,
+  useAppContext,
+} from "client";
 import { Bar } from "client/components";
 import EditButton from "./EditButton";
 import "./index.css";
@@ -34,10 +41,21 @@ export const LabeledBar = ({
   const interval = viewDate.getInterval();
 
   const { name, roll_over, roll_over_start_date } = barData;
-  const { sorted_amount, unsorted_amount, rolled_over_amount } =
+  const { sorted_amount, unsorted_amount } =
     interval === "year"
       ? budgetData.get(barData.id).aggregateYear(date.getFullYear())
       : budgetData.get(barData.id, date);
+
+  // Rollover projects forward for future views (#562). For year interval the
+  // bar shows the carry into the year — its January value, matching
+  // aggregateYear's semantics — so project to that year's January.
+  const rolledOverDate =
+    interval === "year"
+      ? new ViewDate("month", new LocalDate(`${date.getFullYear()}-01-15`)).getEndDate()
+      : date;
+  const rolled_over_amount = roll_over
+    ? getRolledOverAmount(barData, budgetData, rolledOverDate)
+    : 0;
 
   // For synced capacities the stored `capacity[interval]` is just the
   // advisory cache, not the live sum — go through getActiveAmount so the
--- a/src/client/components/LabeledBar/index.tsx
+++ b/src/client/components/LabeledBar/index.tsx
@@ -35,10 +35,10 @@ export const LabeledBar = ({
 
   const { name, roll_over, roll_over_start_date } = barData;
   // All three figures from one call — rollover no longer flows separately
-  // from sorted/unsorted; getView projects the carry forward for future
+  // from sorted/unsorted; getSummary projects the carry forward for future
   // views (#562) and reads January for a year view internally.
   const { sorted_amount, unsorted_amount, rolled_over_amount } =
-    budgetData.getView(barData, viewDate);
+    budgetData.getSummary(barData, viewDate);
 
   // For synced capacities the stored `capacity[interval]` is just the
   // advisory cache, not the live sum — go through getActiveAmount so the
--- a/src/client/components/LabeledBar/index.tsx
+++ b/src/client/components/LabeledBar/index.tsx
@@ -1,13 +1,6 @@
 import { Dispatch, SetStateAction } from "react";
-import { MAX_FLOAT, LocalDate, ViewDate, currencyCodeToSymbol, numberToCommaString } from "common";
-import {
-  Budget,
-  Category,
-  Section,
-  getRolledOverAmount,
-  useReorder,
-  useAppContext,
-} from "client";
+import { MAX_FLOAT, currencyCodeToSymbol, numberToCommaString } from "common";
+import { Budget, Category, Section, useReorder, useAppContext } from "client";
 import { Bar } from "client/components";
 import EditButton from "./EditButton";
 import "./index.css";
@@ -41,21 +34,11 @@ export const LabeledBar = ({
   const interval = viewDate.getInterval();
 
   const { name, roll_over, roll_over_start_date } = barData;
-  const { sorted_amount, unsorted_amount } =
-    interval === "year"
-      ? budgetData.get(barData.id).aggregateYear(date.getFullYear())
-      : budgetData.get(barData.id, date);
-
-  // Rollover projects forward for future views (#562). For year interval the
-  // bar shows the carry into the year — its January value, matching
-  // aggregateYear's semantics — so project to that year's January.
-  const rolledOverDate =
-    interval === "year"
-      ? new ViewDate("month", new LocalDate(`${date.getFullYear()}-01-15`)).getEndDate()
-      : date;
-  const rolled_over_amount = roll_over
-    ? getRolledOverAmount(barData, budgetData, rolledOverDate)
-    : 0;
+  // All three figures from one call — rollover no longer flows separately
+  // from sorted/unsorted; getView projects the carry forward for future
+  // views (#562) and reads January for a year view internally.
+  const { sorted_amount, unsorted_amount, rolled_over_amount } =
+    budgetData.getView(barData, viewDate);
 
   // For synced capacities the stored `capacity[interval]` is just the
   // advisory cache, not the live sum — go through getActiveAmount so the
--- a/src/client/lib/hooks/calculation/budgets.ts
+++ b/src/client/lib/hooks/calculation/budgets.ts
@@ -231,21 +231,6 @@ export const getBudgetData = (
   return { transactionFamilies, budgetData };
 };
 
-/**
- * Rollover ("+ $X rolled") for a budget-like at an arbitrary view date.
- *
- * Thin wrapper over `BudgetData.getRolledOver` (the projection lives on the
- * model so `BudgetData.getView` and this single-value consumer share one
- * implementation). Past/current months read the accrued authoritative value;
- * future months are projected on read so a future view reflects the carry
- * forward instead of "+ $0 rolled" (#562).
- */
-export const getRolledOverAmount = (
-  budgetLike: Budget | Section | Category,
-  budgetData: BudgetData,
-  date: Date,
-): number => budgetData.getRolledOver(budgetLike, date);
-
 const oldestDate = new Date(0);
 
 export const getCapacityData = (
--- a/src/client/lib/hooks/calculation/budgets.ts
+++ b/src/client/lib/hooks/calculation/budgets.ts
@@ -234,53 +234,17 @@ export const getBudgetData = (
 /**
  * Rollover ("+ $X rolled") for a budget-like at an arbitrary view date.
  *
- * `getBudgetData`'s accrual loop only writes rollover entries up to the
- * current calendar month — extending it to the rendered view date would
- * mean threading `viewDate` into `getBudgetData` and re-running the whole
- * (transaction-heavy) calculation on every month the user navigates to.
- * So past/current months read the accrued, authoritative value straight
- * from `budgetData`; future months are projected on read here, the same way
- * capacity and "left" already project forward via `getActiveAmount`.
- *
- * Future months carry no spend, so the projection mirrors the accrual
- * recurrence `rolled(m) = rolled(m-1) − getActiveAmount(m-1)` from the last
- * accrued month (the current calendar month) up to the requested month.
- * Without it a future view reads the lazily-created empty summary and
- * renders "+ $0 rolled" while capacity still projects forward (#562).
+ * Thin wrapper over `BudgetData.getRolledOver` (the projection lives on the
+ * model so `BudgetData.getView` and this single-value consumer share one
+ * implementation). Past/current months read the accrued authoritative value;
+ * future months are projected on read so a future view reflects the carry
+ * forward instead of "+ $0 rolled" (#562).
  */
 export const getRolledOverAmount = (
   budgetLike: Budget | Section | Category,
   budgetData: BudgetData,
   date: Date,
-): number => {
-  const history = budgetData.get(budgetLike.id);
-  const current = new ViewDate("month");
-  const target = new ViewDate("month", date);
-
-  // Past and current months are accrued by getBudgetData — authoritative.
-  if (target.getEndDate() <= current.getEndDate()) {
-    return history.get(target.getEndDate()).rolled_over_amount;
-  }
-
-  // Future months: project the capacity carry forward from the current
-  // month. Only accrue once the rollover window has opened — a budget-like
-  // whose `roll_over_start_date` is itself in the future contributes nothing
-  // before it begins.
-  const { roll_over_start_date } = budgetLike;
-  const startMonthEnd = roll_over_start_date
-    ? new ViewDate("month", roll_over_start_date).getEndDate()
-    : undefined;
-
-  let rolled = history.get(current.getEndDate()).rolled_over_amount;
-  const cursor = current.clone();
-  while (cursor.getEndDate() < target.getEndDate()) {
-    if (!startMonthEnd || cursor.getEndDate() >= startMonthEnd) {
-      rolled -= budgetLike.getActiveAmount(cursor.getEndDate(), "month");
-    }
-    cursor.next();
-  }
-  return rolled;
-};
+): number => budgetData.getRolledOver(budgetLike, date);
 
 const oldestDate = new Date(0);
 
--- a/src/client/lib/hooks/calculation/budgets.ts
+++ b/src/client/lib/hooks/calculation/budgets.ts
@@ -231,6 +231,57 @@ export const getBudgetData = (
   return { transactionFamilies, budgetData };
 };
 
+/**
+ * Rollover ("+ $X rolled") for a budget-like at an arbitrary view date.
+ *
+ * `getBudgetData`'s accrual loop only writes rollover entries up to the
+ * current calendar month — extending it to the rendered view date would
+ * mean threading `viewDate` into `getBudgetData` and re-running the whole
+ * (transaction-heavy) calculation on every month the user navigates to.
+ * So past/current months read the accrued, authoritative value straight
+ * from `budgetData`; future months are projected on read here, the same way
+ * capacity and "left" already project forward via `getActiveAmount`.
+ *
+ * Future months carry no spend, so the projection mirrors the accrual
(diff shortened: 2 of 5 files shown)


src/client/components/LabeledBar/index.tsx
```diff
@@ -34,11 +41,22 @@ export const LabeledBar = ({
...
+  // bar shows the carry into the year — its January value, matching
+  // aggregateYear's semantics — so project to that year's January.
+  const rolledOverDate =
+    interval === "year"
+      ? new ViewDate("month", new LocalDate(`${date.getFullYear()}-01-15`)).getEndDate()
+      : date;
+  const rolled_over_amount = roll_over
+    ? getRolledOverAmount(barData, budgetData, rolledOverDate)
+    : 0;
+
```
hoiekim, line 42: I don't like that rolled_over_amount uses separate flow from sorted and unsorted_amount. Design a method under either BudgetData or BudgetHistory(not sure the name is correct) that elegantly abstract this. This is just an example to help you understand what elegance I'm talking about:
```
const { sorted_amount, unsorted_amount, rolled_over_amount } = budgetData.get(barData.id).getView(viewDate)
```
See how the code explains itself.

hoiekim (changes requested):

reviewoie:
## Code review — PR 563

**Verdict:** approve

**Findings:** 2 total — 0 high, 0 med, 1 low (+ 1 informational).

Verified the four equivalence questions; the refactor preserves all existing-view behavior and the only behavior change is the intended future-month rollover projection (#562).

### High
No further blockers.

### Med
No further blockers.

### Low
- `LabeledBar/index.tsx:3` — the `client` import was reformatted single-line → multi-line with an identical symbol set (`Budget`, `Category`, `Section`, `useReorder`, `useAppContext` — no symbol added or removed). Pure whitespace churn unrelated to the abstraction. Harmless; collapse back to one line if you want a tighter diff.

### Informational
- `Calculations.ts:421` (`getView`) — placement on `BudgetData` rather than the literal `get(id).getView(viewDate)` you sketched is the right call and within your "either BudgetData or BudgetHistory" allowance. `BudgetHistory` holds only the summary map, so a `getView` there would still need the budget-like (capacity schedule + `roll_over` + `roll_over_start_date`) threaded in as an arg — `budgetData.getView(barData, viewDate)` keeps it to one param. No change requested.

### Verification performed
- **(a) Equivalence vs the old LabeledBar flow:** month view reads `history.get(date)` (same monthly key — `ViewDate("month", date)` re-normalizes to the same month-end); year view copies `aggregateYear` for sorted/unsorted and reads rollover at January via `getRolledOver` — for a past/current January that returns the same stored value `aggregateYear` used, and for a non-rollover budget-like it returns 0 (matching, since `rolled_over_amount` is only ever written under `roll_over`). Future projection's per-step boundary matches the accrual loop: each step subtracts the previous month's `getActiveAmount`, so `rolled(M+1) = rolled(M) − capacity(M)` both in `getBudgetData`'s accrual and in `getRolledOver`'s walk. The `roll_over_start_date` guard (`cursor >= startMonthEnd`) lines up with the accrual loop's `start.next()` start. ✓
- **(b) BalanceChartRow unchanged:** still passes raw `date` and reads December for a year view (the year→January nuance lives only in `getView`, which BalanceChartRow does not call). Past/current is byte-identical to the old `budgetData.get(b.id, date).rolled_over_amount`; future is the intended #562 projection. No leak. ✓
- **(c) No import cycle:** `BudgetLikeView` is structural, `Interval` comes from `common` (existing dep), and the Budget-family classes are not imported into `Calculations.ts`. `bun run --bun tsc --noEmit` exits 0. ✓
- **(d) Scope:** all five touched files serve the abstraction. ✓
- **Tests/typecheck:** `bun test src/client/lib` → 227 pass / 0 fail; `tsc --noEmit` clean. New `getView` coverage (month / future-month / year-January / non-rollover) is real (grepped the `describe`/`test` blocks) and asserts the equivalence relationships above.

src/client/components/LabeledBar/index.tsx
claoie: Done — added `BudgetData.getView` so the bar reads all three figures from one call (commit `06471bed`):

```ts
const { sorted_amount, unsorted_amount, rolled_over_amount } =
  budgetData.getView(barData, viewDate);
```

`rolled_over_amount` no longer flows on a separate path — `getView` internalizes both the year-vs-month branch for sorted/unsorted (sums the year via `aggregateYear`, reads the month otherwise) and the rollover read (the future-month projection from #562, and the year→January carry). The LabeledBar block that used to hand-compute a `rolledOverDate` and call `getRolledOverAmount` separately is gone.

**On placement:** you suggested either `BudgetData` or `BudgetHistory`. I put it on `BudgetData` and pass the budget-like once — `budgetData.getView(barData, viewDate)` — rather than `budgetData.get(id).getView(viewDate)`, because the rollover projection genuinely needs the budget-like's capacity schedule + `roll_over` flag, which `BudgetHistory` (amounts only, keyed by id) doesn't hold. Passing `barData` once is cleaner than threading both `id` and `barData`. If you'd rather have the exact `get(id).getView(...)` shape, I can move it onto `BudgetHistory` and pass the budget-like as a second arg — let me know.

The future-month projection moved onto the model as `BudgetData.getRolledOver`; the old `getRolledOverAmount` free function is now a thin wrapper over it, so `BalanceChartRow` and the #562 tests stay numerically identical (the year→January nuance is isolated to `getView` and doesn't reach BalanceChartRow). `budgetLike` is typed structurally (`BudgetLikeView`) so the calc models don't import the budget-family classes.

**Verification (head `06471bed`):**
- `bun test src/client/lib` — 227 pass / 0 fail, incl. 4 new `getView` tests (month / future-month / year-January / non-rollover) that pin getView's output to the values it abstracts.
- `bun run --bun tsc --noEmit` clean; lint clean.
- UI E2E (Playwright, sandbox): budget bars render on the current month, a future month (+4), and the year interval; the rollover rows persist across all three with no functional console errors — confirming the projection + year-January paths render through `getView`.
- budget-reviewer (reviewoie) approved at `4b88ff73` (0 high / 0 med), flagging only an import-line reformat as a low — `06471bed` collapses that back to match `main`.

src/client/lib/hooks/calculation/budgets.ts
```diff
@@ -231,6 +231,21 @@ export const getBudgetData = (
...
+ * model so `BudgetData.getView` and this single-value consumer share one
+ * implementation). Past/current months read the accrued authoritative value;
+ * future months are projected on read so a future view reflects the carry
+ * forward instead of "+ $0 rolled" (#562).
+ */
+export const getRolledOverAmount = (
+  budgetLike: Budget | Section | Category,
+  budgetData: BudgetData,
+  date: Date,
+): number => budgetData.getRolledOver(budgetLike, date);
```
hoiekim: Remove this helper function. The abstraction isn't helpful. `budgetData.getRolledOver` is already descriptive. 

Remember this comment as a coding rule

src/client/lib/models/Calculations.ts
```diff
@@ -292,6 +292,18 @@ export class BudgetHistory {
...
+ * The slice of a budget-like (Budget / Section / Category) that the rollover
+ * projection needs. Typed structurally so the calculation models don't import
+ * the budget-family model classes (and risk a cycle).
+ */
+export type BudgetLikeView = {
+  id: string;
+  roll_over: boolean;
+  roll_over_start_date?: Date;
+  getActiveAmount: (date: Date, interval: Interval) => number;
+};
```
hoiekim: We already have `BudgetFamily` type or something similar, if I'm remembering correctly. Don't write duplicate code.

src/client/lib/models/Calculations.ts
```diff
@@ -332,6 +344,78 @@ export class BudgetData {
...
+   * `sorted_amount` / `unsorted_amount` from the stored history (summed across
+   * the year for a year view), and `rolled_over_amount` — the carry-forward,
+   * projected past the current month for future views (#562). For a year view
+   * the bar shows the carry INTO the year (its January value, matching
+   * aggregateYear), so the rollover is read at that year's January.
+   *
+   * One call so a consumer reads all three figures together instead of
+   * re-deriving the rollover on a separate flow.
+   */
+  getView = (budgetLike: BudgetLikeView, viewDate: ViewDate): BudgetSummary => {
```
hoiekim: Should we name it `getSummary`?

hoiekim (changes requested):

src/client/lib/hooks/calculation/budgets.ts
claoie: Removed in `541b10c1` — `getRolledOverAmount` is gone. `BalanceChartRow` and the rollover tests now call `budgetData.getRolledOver(b, date)` directly. Captured the underlying rule (don't keep a thin pass-through wrapper when the method it wraps already reads clearly — it adds an indirection without adding meaning) in my coding notes.

src/client/lib/models/Calculations.ts
claoie: Done — dropped `BudgetLikeView`. `BudgetFamily` already declares `id` / `roll_over` / `roll_over_start_date` / `getActiveAmount`, so `getRolledOver` and `getSummary` take `BudgetFamily` directly. Imported it `type`-only (`import type { BudgetFamily }`) so there's still no runtime import cycle from the calc models into the budget-family classes — which is the only reason the structural type existed. `541b10c1`.

src/client/lib/models/Calculations.ts
claoie: Renamed `getView` → `getSummary` (it returns a `BudgetSummary`). Updated the `LabeledBar` caller and the test `describe` block to match. `541b10c1`.

reviewoie:
## Code review — PR 563 (commit `541b10c1`)

**Verdict:** approve

Scoped to the latest commit only (earlier commits already reviewed). Verified all three review items landed cleanly with no behavior change.

### Verified
- **Leftover-reference sweep clean.** `git grep` on the post-commit tree returns zero hits for `getRolledOverAmount`, `BudgetLikeView`, and `getView` across `src/`. Both real callers (`BalanceChartRow`, `LabeledBar`) and all five test call sites now use `budgetData.getRolledOver` / `budgetData.getSummary` directly.
- **`import type { BudgetFamily }` cannot create a runtime cycle.** It is a type-only import — fully erased at compile time, contributing zero runtime edges regardless of the structural import graph. `tsc --noEmit` exits 0.
- **`BudgetFamily` structurally satisfies both signatures.** It declares `id` (getter), `roll_over: boolean`, `roll_over_start_date?: Date`, and `getActiveAmount(date, interval): number` — every member `getRolledOver`/`getSummary` touch. `BarData = Budget | Section | Category` and the `budgets` prop are all `BudgetFamily` subtypes, so both call sites type-check.
- **`Interval` import correctly dropped** from `Calculations.ts` — it was used only by the removed `BudgetLikeView`. `getSummary` line 384 uses `viewDate.getInterval()` (a method call with inferred type), not the `Interval` type name.
- **Non-rollover test exercises `roll_over=false` correctly.** It now builds a real `Budget` (mirroring the known-good `makeBudget` shape) with `roll_over: false`, and asserts `rolled_over_amount === 0`. The shared `budget_id: "bud-1"` is harmless since the rollover branch is skipped and the assertion only reads `rolled_over_amount`.
- **Pure refactor, scope clean.** All five touched files (Calculations.ts, budgets.ts, LabeledBar, BalanceChartRow, budgets.rollover.test.ts) map to the three stated review items. No drive-by changes. `bun test budgets.rollover.test.ts` → 14 pass / 0 fail.

This matches the cohesive-method direction from the original PR review (r3483284677): the three figures still come from one `getSummary` call, and the wrapper/duplicate-type removal tightens the abstraction without splitting the flow.

No further blockers.

hoiekim (approved):

PR #568, opened 2026-06-28
fix(budget): section unfold should scroll the section to viewport top, not jump away

claoie:
Closes the budget-page section-unfold scroll bug Hoie flagged 2026-06-28: clicking a section header to unfold it scrolls the page away from the user's intended position instead of bringing the section header to the top of the viewport.

## Root cause

In `SectionBar.tsx`, `openCategory` called `scrollIntoView({ block: "center" })` on the inner `children` div **synchronously** right after `setIsOpen(true)`:

```ts
const openCategory = () => {
  setIsOpen(true);
  const childrenDiv = childrenDivRef.current;
  if (!childrenDiv) return;
  childrenDiv.scrollIntoView({ behavior: "smooth", block: "center" });
};
```

Two problems:
1. `setIsOpen(true)` is queued — React hasn't committed the state change yet when the next line runs. `childrenDivRef` still points at the empty (height = 0) children div.
2. `block: "center"` then centers a zero-height point at the viewport center. The section header sits just above it, ending up in the *middle* of the viewport rather than at the top — which is the "the page scrolled" feeling the user perceives.

## Fix

Move the scroll into a `useEffect` that fires after React commits `isOpen = true`, target the SectionBar's outer ref with `block: "start"`, and gate it behind a `pendingScrollToOpen` ref so re-mounts that read `isOpen=true` from `useMemoryState` (e.g. page reload while a section was open) do NOT auto-scroll.

```ts
const sectionBarRef = useRef<HTMLDivElement>(null);
const pendingScrollToOpen = useRef(false);

const openCategory = () => {
  pendingScrollToOpen.current = true;
  setIsOpen(true);
};

useEffect(() => {
  if (!isOpen || !pendingScrollToOpen.current) return;
  pendingScrollToOpen.current = false;
  sectionBarRef.current?.scrollIntoView({ behavior: "smooth", block: "start" });
}, [isOpen]);
```

## Test plan

- [x] `bun test src/client/components` → 8 pass / 0 fail
- [x] **E2E (Playwright on local dev, unscrambled prod-clone)**:
  - Scroll the budget detail page down; click a section near the top of the viewport — the section header lands at viewport top (`getBoundingClientRect().top === 0`), not at the middle. Previously it shifted ~halfway down the viewport.
  - Click a section that's currently off-screen below — the page scrolls down to bring the section into view at the top (clamped to max scroll when the document isn't tall enough to fully scroll the section to top).
  - Click an already-open section to close it — scroll position does NOT change (the close path doesn't trigger the effect).
  - Reload while a section is open (`useMemoryState` persists `isOpen=true`) — no auto-scroll on mount; the user lands at whatever scroll the route restored.

### DIFF
--- a/src/client/components/SectionBar.tsx
+++ b/src/client/components/SectionBar.tsx
@@ -48,6 +48,8 @@ export const SectionBar = ({ section, onSetOrder }: Props) => {
   }, [categories, section_id, setCategoriesOrder]);
 
   const childrenDivRef = useRef<HTMLDivElement>(null);
+  const sectionBarRef = useRef<HTMLDivElement>(null);
+  const pendingScrollToOpen = useRef(false);
 
   const observerRef = useRef(
     new ResizeObserver((entries) => {
@@ -81,11 +83,15 @@ export const SectionBar = ({ section, onSetOrder }: Props) => {
   const { iso_currency_code } = budget;
 
   const openCategory = () => {
+    pendingScrollToOpen.current = true;
     setIsOpen(true);
-    const childrenDiv = childrenDivRef.current;
-    if (!childrenDiv) return;
-    childrenDiv.scrollIntoView({ behavior: "smooth", block: "center" });
   };
+
+  useEffect(() => {
+    if (!isOpen || !pendingScrollToOpen.current) return;
+    pendingScrollToOpen.current = false;
+    sectionBarRef.current?.scrollIntoView({ behavior: "smooth", block: "start" });
+  }, [isOpen]);
 
   const onClickInfo = () => {
     const params = new URLSearchParams(router.params);
@@ -132,7 +138,7 @@ export const SectionBar = ({ section, onSetOrder }: Props) => {
   const childrenClassNames = ["children", "transition"];
 
   return (
-    <div className="SectionBar">
+    <div className="SectionBar" ref={sectionBarRef}>
       {isOpen ? (
         <div
           className="openLabel"


reviewoie:
## Code review — PR 568

**Verdict:** approve (informational, no findings)

Walked all four focus points + the standard angles:

- **`pendingScrollToOpen` gating** — correct on every transition I traced. Per-instance ref + `[isOpen]` effect cleanly distinguishes user-click (pending=true) from remount-with-memoryState-true (pending=false). Open→close path leaves pending=false, so the close-side effect-run early-returns at `SectionBar.tsx:91`. Two sections clicked in rapid succession don't race because `BudgetDetailPage:61` keys each SectionBar by `section_id` — separate component instances, separate refs.
- **Router `endTransition` + `scrollIntoView` interaction** — confirmed in `router.ts:193-224`: restoration uses `window.scrollY` / `window.scrollTo`. In wide-screen mode `<main>` is the scroll container (`App/index.css:148-154`, `Router.tsx:96`), so `window.scrollTo` is genuinely a no-op and there is no conflict. In narrow-screen mode body IS the scroll container, so the router's `scrollTo(0, restoredY)` could fire — but the new params include `section_id=X` which is a fresh key (restoredY=0), and this same interaction existed pre-fix. Not introduced or exacerbated by this PR. If the router is ever refactored to do same-path scroll-restore as a no-op (or this becomes a same-key navigation), this stays clean.
- **`useEffect` vs `useLayoutEffect`** — the right call. The scroll target is the OUTER `SectionBar` ref, whose top position does NOT depend on children expansion. `.children` has `overflow: hidden` + a CSS height transition (`BudgetDetailPage/index.css:28-34`); the section header sits above .children and is unaffected. Non-layout timing also avoids fighting the height-transition's first paint.
- **`childrenDivRef` still wired** — attached at line 169, consumed by the `ResizeObserver` at lines 61-68. Ref usage unchanged; only the scroll path moved off it.
- **Scope** — 1 file, 1 concern, exactly matches PR title. No drift.
- **No PII leaks** — commit message + PR body describe the fix qualitatively (no prod figures, no institution names, no account labels). Test plan checkboxes use coordinate-relative claims (`getBoundingClientRect().top === 0`), not data values.

No further blockers.

hoiekim (approved):

PR #570, opened 2026-06-28
feat(flow chart): per-budget transaction filter (Sankey scoped to selected budgets)

claoie:
Per Hoie 2026-06-28 (channel 1520721356575215636): extend the Flow chart configuration to take a budget selection. When computing Sankey data, exclude any transaction whose label_budget_id isn't in the selected list.

## Changes

**Schema/types (1 line)**
- `JSONFlowChartConfiguration` + `FlowChartConfiguration` gain a `budget_ids: string[]` field. The configuration is serialized as JSON in `charts.configuration`, so no DB migration is required for existing FlowCharts — they deserialize with `budget_ids = []`.

**Calc**
- `getSankeyData(...)` takes a `budget_ids: string[] = []` arg. Inside `processTransaction`, after computing the effective budget_id (`t.label.budget_id || account.label.budget_id || "Unknown"`), if `budget_ids.length > 0` and the effective id isn't in the list, skip the transaction. The `"Unknown"` sentinel can be explicitly toggled on by the user to include "Unsorted" rows in the chart.

**Semantics**
- **Empty list = no filter** (include all budgets). Backward-compatible for existing FlowCharts; users opt-in to narrowing the chart by toggling specific budgets on.
- **Non-empty = whitelist**.

**UI**
- `ChartAccountsPage` now renders the "Select budgets" toggle group for `ChartType.FLOW` as well as `ChartType.BALANCE`. The toggle handler branches on `type` to construct either a `BalanceChartConfiguration` or `FlowChartConfiguration` so the right shape is POSTed.

**Tests (4 new)**
- `lib.test.ts` "getSankeyData — budget_ids whitelist filter":
  - empty list = all budgets included
  - whitelist with Budget A only excludes Budget B's transaction
  - whitelist with Budget B only excludes Budget A's transaction
  - whitelist with a non-matching id excludes all

## Test plan

- [x] `bun test src/client/components/FlowChartRow` → 12 pass / 0 fail (8 pre-existing + 4 new).
- [x] **E2E (Playwright, local unscrambled dev)**:
  - Dashboard → Cash Flow → chart-detail page on Yearly 2026 default view: shows full data with empty whitelist.
  - Toggle a single budget ("Income") ON in the chart-accounts page: persists to server (`budget_ids: [<income_id>]` in `/api/charts`) and to indexedDB.
  - Reload chart-detail: In/Out numbers change to ONLY the labeled-Income transactions (the deficit flips to a surplus, as expected for an income-only view).
  - Toggle all budgets OFF (empty list): full data returns.
  - 5+ budgets present, only one toggled — verified the chart filters strictly to that one, all others excluded.

### DIFF
--- a/src/client/components/FlowChartRow/index.tsx
+++ b/src/client/components/FlowChartRow/index.tsx
@@ -35,7 +35,7 @@ export const FlowChartRow = ({
     transfers,
   } = data;
   const { configuration } = chart;
-  const { account_ids } = configuration;
+  const { account_ids, budget_ids } = configuration;
 
   const {
     onDragStart,
@@ -66,6 +66,7 @@ export const FlowChartRow = ({
         categories,
         viewDate,
         transfers,
+        budget_ids,
       ),
     [
       selectedAccounts,
@@ -77,6 +78,7 @@ export const FlowChartRow = ({
       categories,
       viewDate,
       transfers,
+      budget_ids,
     ],
   );
 
--- a/src/client/components/FlowChartRow/lib.test.ts
+++ b/src/client/components/FlowChartRow/lib.test.ts
@@ -310,3 +310,74 @@ describe("getSankeyData — investment transactions are skipped (internal moves)
     expect(tableData.income).toBe(0);
   });
 });
+
+describe("getSankeyData — budget_ids whitelist filter", () => {
+  // Build the world ONCE per test so the random ids on budgets/account
+  // line up across the fixture construction and the assertions.
+  const setup = () => {
+    const w = makeWorld();
+    const txnA = new Transaction({
+      account_id: w.account.id,
+      transaction_id: "tx-a",
+      amount: 100,
+      date: DATE,
+      label: { budget_id: w.budgetA.id, category_id: w.categoryA.id, category_confidence: 1 },
+    });
+    const txnB = new Transaction({
+      account_id: w.account.id,
+      transaction_id: "tx-b",
+      amount: 200,
+      date: DATE,
+      label: { budget_id: w.budgetB.id, category_id: w.categoryB.id, category_confidence: 1 },
+    });
+    const transactions = new TransactionDictionary();
+    transactions.set(txnA.id, txnA);
+    transactions.set(txnB.id, txnB);
+    const run = (budget_ids: string[]) =>
+      getSankeyData(
+        [w.account],
+        transactions,
+        emptyInvestments,
+        new SplitTransactionDictionary(),
+        w.budgets,
+        w.sections,
+        w.categories,
+        viewDate,
+        noTransfers,
+        budget_ids,
+      );
+    return { w, run };
+  };
+
+  test("empty whitelist = no filter (all budgets included)", () => {
+    const { w, run } = setup();
+    const { tableData, graphData } = run([]);
+    expect(tableData.expense).toBe(300);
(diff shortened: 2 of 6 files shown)


reviewoie:
## Code review — PR 570

**Verdict:** approve (informational findings only)

### High
- No further blockers.

### Med
- No further blockers.

### Low
- `src/client/pages/ChartAccountsPage.tsx:88` — "Unsorted" sentinel is unreachable via the toggle UI. A transaction with no `t.label.budget_id` AND no `account.label.budget_id` falls through to the literal `"Unknown"` id in `getSankeyData` (`lib.ts:90`). `budgets.toArray()` only iterates named budgets, so the user cannot add `"Unknown"` to the whitelist. Failure scenario: a user toggles even one real budget ON expecting an additive scope and silently loses every unsorted transaction with no UI affordance to bring them back; the only workaround is deselect-all (which then flips meaning to "show all"). Worth a follow-up: synthesize an "Unsorted" pseudo-row in the budget rows list that toggles the literal `"Unknown"` id, or add a tooltip / footer noting the gap. Not a blocker — PR body acknowledges the asymmetry.
- `src/client/components/FlowChartRow/lib.ts:24-58` — `getSankeyData` is now 10 positional args. The new arg follows the existing shape so this PR is consistent, but the smell worsens. Failure scenario: the next addition (whitelist of categories? date filter?) makes the call site illegible and arg-position bugs easy. Defer until the 11th, but consider an options-object refactor in a separate PR before adding more.
- `src/client/components/FlowChartRow/lib.ts:50-58` — empty-list-is-no-filter asymmetry with `account_ids` is intentional and documented, but worth one explicit test: a transaction with `label.budget_id = undefined` (effective `"Unknown"`) should be included on the empty whitelist and excluded on any non-empty named whitelist. The four new tests cover named-budget filtering thoroughly but never exercise the `"Unknown"` branch.

Backward-compat verified: `Chart.fromJSON` constructs `new FlowChartConfiguration(<parsed>)` and `assign(this, init)` skips missing keys, so a pre-existing serialized configuration with no `budget_ids` deserializes with the class-field default `[]` and the empty-list short-circuit holds. Typecheck clean, 12/12 in `lib.test.ts` (8 pre-existing + 4 new). Scope is tight — all six files relate to the PR title. `ChartAccountsPage` refactor preserves the PROJECTION-only-accounts path via the `showBudgetSelector` gate (verified by reading the diff and the new render).

claoie:
Thanks for the review. Addressing the 3 LOW findings:

**Finding 1 (Unsorted unreachable via toggle UI)** — surfaced this same gap to Hoie in the Discord handoff for his call (synthetic "Unsorted" toggle vs leave as-is). Holding for his direction; if he wants it added I'll do a follow-up.

**Finding 2 (10 positional args on `getSankeyData`)** — agree, this PR makes the smell worse. Leaving for a focused options-object refactor PR before the next addition; not blocking this fix.

**Finding 3 (Unknown-sentinel test coverage gap)** — addressed in `edc363e4`. Two new tests pin the behavior:
- empty whitelist includes unlabeled (effective `budget_id = "Unknown"`) transactions.
- whitelist containing only `"Unknown"` includes unlabeled txns and excludes labeled ones.

`bun test src/client/components/FlowChartRow` → 14 pass / 0 fail.

hoiekim (approved):

Issue #572, opened 2026-06-28
bug(transactions): post-split-transaction returns success on a failed DB write (skips the status>=400 guard its sibling routes use)


issueoie:
## Summary

`post-split-transaction.ts` does not surface a failed DB write — it returns `status: "success"` to the client even when the underlying repo write errored. Its two sibling routes in the same FE→DB label-write trio (`post-transaction.ts`, `post-investment-transaction.ts`) both guard on `result.status >= 400` and rethrow. This is a Principle 2 (one canonical way) inconsistency with a real user-visible consequence: on a DB failure the split-transaction edit reports as saved in the UI while nothing was persisted.

## Where

`src/server/routes/accounts/post-split-transaction.ts`:

```ts
const response = await updateSplitTransactions(user, [split]);
const split_transaction_id = response[0].update?._id || "";
return { status: "success", body: { split_transaction_id } };
```

vs the guard the other two routes use:

```ts
const result = response[0];
if (!result || result.status >= 400) {
  throw new Error("Database responded with an error.");
}
```

`updateSplitTransactions` catches its own errors and returns `errorResult(split_transaction_id)` (status 500) rather than throwing, so `response[0].update._id` is still populated on failure and the route echoes it back as success.

## Repro / evidence

Pinned in `src/server/routes/accounts/post-transaction-routes.test.ts` (the route-test coverage added for #437): with the DB layer made to throw, `post-transaction` and `post-investment-transaction` return `status: "error"`, while `post-split-transaction` returns `status: "success"` with the id echoed. The test documents the current behavior so this divergence is visible.

## Proposed fix

Add the same `result.status >= 400` guard to `post-split-transaction.ts` so a failed write surfaces as an error, then flip the test assertion from `success` to `error` to match. Small, single-route change. Kept out of the #437 test-only PR for scope discipline.

## Principle

Principle 2 — one canonical way per thing. Three routes doing the same job (label write → DB) should handle the repo's error result the same way; one silently diverging is the tax a future reader pays.

PR #573, opened 2026-06-28
test(routes): cover the POST transaction-write trio at the route layer

### LINKED ISSUE #437: POST transaction-write trio has no route-level test coverage
The three FE→DB label-write routes have zero route-level tests despite being the canonical path for user category changes — and PR #431 just made two of them do more than pass-through.

## Surface

| Route | LOC | Touches inferLabelConfidence | Test sibling? |
|-------|-----|------------------------------|---------------|
| `src/server/routes/accounts/post-transaction.ts` | 54 | ✓ (added by #431) | — |
| `src/server/routes/accounts/post-split-transaction.ts` | 46 | ✓ (added by #431) | — |
| `src/server/routes/accounts/post-investment-transaction.ts` | 42 | ✗ (no auto-suggest on investments) | — |

The helper `inferLabelConfidence` itself has 8 unit tests in `src/server/lib/infer-label-confidence.test.ts` (PR #431). The repo functions `updateTransactions` / `updateSplitTransactions` / `updateInvestmentTransactions` are exercised by `src/server/lib/postgres/repositories/transactions.test.ts` + sibling tests. **What is uncovered is the wiring that ties them together** — the same Principle 3 "leaf logic green, composition layer untested" shape as #424 (`schedule.ts runSync`) and the earlier route-coverage closes (#346, #365, #393, #397, #402, #393-partial via #402).

## Why this matters now

PR #431's premise was that `updateSplitTransactions` silently dropped the `category_confidence` write, letting auto-suggest re-overwrite the user's manual category on the next hourly run. That bug existed unobserved because the route layer was untested — the unit test on the repo passed (it correctly wrote what it was asked to), and the unit test on auto-suggest passed (it correctly skipped already-confidence-set rows), but the gap between them was invisible. A route-level test (drive a POST through the handler, assert the persisted row's `label_category_confidence`) would have caught it.

Same gap is now sitting on `post-investment-transaction.ts`. It doesn't invoke the helper today because auto-suggest doesn't process investment transactions — but if that gating ever changes, the same silent overwrite is one design decision away.

## Suggested test plan

Single test file `src/server/routes/accounts/post-transaction-routes.test.ts` mirroring the monkey-patch pattern in `holding-snapshot-sibling-routes.test.ts` (PR #397). Avoid `mock.module(\"server\", ...)` — Bun's module mock leaks across files.

Per route, cover:

1. **Unauthenticated** — no `req.session.user` → `status: \"failed\"`, repo never called.
2. **Missing required ID field** (`transaction_id` / `split_transaction_id` / `investment_transaction_id`) → 400 via `validationError`, repo never called.
3. **Bad body shape** (not an object) → 400 via `validationError`.
4. **Happy path — set category, confidence omitted** → for `post-transaction` and `post-split-transaction`, asserts the repo received `label.category_confidence === 1`. For `post-investment-transaction`, asserts the body passes through unchanged.
5. **Happy path — clear category** (`category_id: null`, confidence omitted) → asserts `category_confidence === 0` for the first two routes; unchanged for the investment route.
6. **Happy path — caller sets confidence explicitly** (e.g. Accept-All sends `confidence: 1`) → asserts the helper does not overwrite.
7. **Repo throws** → handler logs and rethrows.

Reference: existing test `src/server/lib/infer-label-confidence.test.ts` covers the helper in isolation; this issue is about the **wiring**.

## Effort estimate

~1 PR, ~200 LOC test file, ~0 LOC source change. Repo functions, `Route.handler`, and `logger` are all injectable via monkey-patch on the real imports (per `holding-snapshot-sibling-routes.test.ts`).

If `post-investment-transaction` deserves the helper too (Principle 2 consistency across the trio), that's a separate small refactor PR — would surface naturally once the test scaffold is in place.

claoie:
Closes #437.

## What

Adds the missing route-level test coverage for the FE→DB label-write trio:

- `src/server/routes/accounts/post-transaction.ts`
- `src/server/routes/accounts/post-split-transaction.ts`
- `src/server/routes/accounts/post-investment-transaction.ts`

These three routes are the canonical path for user category changes and had zero route-level tests, even though PR #431 made two of them run `inferLabelConfidence` (the helper that lifts a category change into `category_confidence`). The repo functions and the helper are each unit-tested in isolation — but the **wiring** that ties them together was untested. That gap is exactly how PR #431's silent `category_confidence` drop went unobserved: the repo test passed (it wrote what it was told) and the auto-suggest test passed (it skipped confidence-set rows), while the route in between dropped the write.

## How

A `FakePool` intercepts `pg` (the established budget pattern from `post-suggest-category.test.ts` — `mock.module("pg")` + `afterAll(restoreLeaves)`), each route is driven through `.execute`, and the emitted `UPDATE` statement's bound values are asserted at the SQL-param layer. A small helper extracts the value bound to a named column from the `buildUpdate` `SET col = $N` form, with a sentinel distinguishing "wrote null" from "didn't write the column".

Per route:

- Unauthenticated → `failed`, repo never queried.
- Non-object body → `failed`, repo never queried.
- Missing id field (`transaction_id` / `split_transaction_id` / `investment_transaction_id`) → `failed` naming the field, repo never queried.
- Set category, confidence omitted → `label_category_confidence = 1` persisted (and, for the regular route, the write is `user_id`-scoped).
- Clear category (`category_id: null`), confidence omitted → `label_category_confidence = 0`, `label_category_id = null`.
- Caller sets confidence explicitly → preserved, not overwritten to 1.
- DB error behavior (see below).

## Two asymmetries the tests pin

1. **Investment route writes no confidence column.** `post-investment-transaction` does not call `inferLabelConfidence` (auto-suggest skips investments), so the test asserts the `UPDATE` carries `label_category_id` but *no* `label_category_confidence` — the load-bearing difference from the other two routes.
2. **Split route does not surface a failed write.** On a DB error, `post-transaction` and `post-investment-transaction` return `status: "error"` (they guard on `result.status >= 400` and rethrow); `post-split-transaction` returns `status: "success"` with the id echoed back, because it reads `response[0].update?._id` without checking the status. The test pins this as current behavior and links the fix to budget #572 — kept out of this test-only PR for scope discipline.

## Test plan

- [x] `bun test src/server/routes/accounts/post-transaction-routes.test.ts` — 19 pass / 0 fail.
- [x] `bun test src/server` — 678 pass / 2 skip / 0 fail (no cross-file mock-module bleed; full server suite run per the Bun `mock.module` process-global caveat).

Test-only PR — no source changes, so no behavior change and no sandbox/UI verification applicable (every changed file matches `*.test.*`).

### DIFF
--- a/src/server/routes/accounts/post-transaction-routes.test.ts
+++ b/src/server/routes/accounts/post-transaction-routes.test.ts
@@ -1,0 +1,354 @@
+import { describe, test, expect, mock, beforeEach, afterAll } from "bun:test";
+import { restoreLeaves } from "test-helpers";
+
+// Route-level coverage for the FE→DB label-write trio
+// (post-transaction / post-split-transaction / post-investment-transaction).
+// The repo functions and `inferLabelConfidence` are unit-tested in isolation;
+// what is uncovered is the *wiring* that ties them together — the route reads
+// the body, runs the confidence inference, and hands the result to the repo.
+// PR #431 added the inference to two of these routes precisely because a
+// silent confidence drop went unobserved while the route layer had no test.
+//
+// We pin behavior at the SQL-param layer: a FakePool intercepts pg, the route
+// is driven through `.execute`, and the UPDATE statement's bound values are
+// asserted. This catches a regression in the route→repo handoff that a repo
+// unit test (which writes exactly what it's told) cannot.
+// A single flag drives the failure path. Re-`mockImplementation`-ing a Bun
+// mock inside a test does not reliably replace the `beforeEach` default, so the
+// default impl itself branches on this flag — flip it to simulate a DB error.
+let failQueries = false;
+const mockQuery = mock(async (_sql: string, _values?: unknown[]) => {
+  if (failQueries) throw new Error("db down");
+  return { rows: [] as unknown[], rowCount: 0 as number | null };
+});
+
+class FakePool {
+  query = mockQuery;
+  end = async () => {};
+  connect = async () => ({ query: mockQuery, release: () => {} });
+}
+
+mock.module("pg", () => ({
+  Pool: FakePool,
+  types: { setTypeParser: () => {} },
+  default: { Pool: FakePool, types: { setTypeParser: () => {} } },
+}));
+
+const { postTransactionRoute } = await import("./post-transaction");
+const { postSplitTransactionRoute } = await import("./post-split-transaction");
+const { postInvestmentTransactionRoute } = await import("./post-investment-transaction");
+
+afterAll(restoreLeaves);
+
+beforeEach(() => {
+  // Clear the call log but keep the flag-driven implementation (mockReset would
+  // wipe it, and re-setting an impl per test doesn't reliably stick in Bun).
+  mockQuery.mockClear();
+  failQueries = false;
+});
+
+type AnyRoute =
+  | typeof postTransactionRoute
+  | typeof postSplitTransactionRoute
+  | typeof postInvestmentTransactionRoute;
+
+function makeReq(route: AnyRoute, body: unknown, userId?: string) {
+  return {
+    method: "POST",
+    path: "/x",
+    url: "http://x/api/x",
+    headers: {},
+    query: {},
+    body,
+    session: {
+      id: "s-1",
+      user: userId ? { user_id: userId, username: "test" } : undefined,
+      regenerate() {},
+      destroy() {},
+    },
+    ip: "127.0.0.1",
+  } as unknown as Parameters<typeof route.execute>[0];
+}
+
+const fakeRes = () =>
+  ({
+    statusCode: 200,
+    headersSent: false,
+    status() {
+      return this;
+    },
+    write() {
+      return true;
+    },
+    end() {},
+  }) as unknown as Parameters<typeof postTransactionRoute.execute>[1];
+
+/** Find the parameterized UPDATE against `table` in the mock call log. */
+const findUpdate = (table: string): { sql: string; values: unknown[] } | null => {
+  const re = new RegExp(`UPDATE\\s+${table}\\b`, "i");
+  for (const call of mockQuery.mock.calls) {
+    const sql = call[0] as string;
+    if (re.test(sql)) return { sql, values: call[1] as unknown[] };
+  }
+  return null;
+};
+
+const SENTINEL = Symbol("column-absent");
+/**
+ * Pull the value bound to `column` in an UPDATE built by `buildUpdate`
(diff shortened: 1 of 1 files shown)


reviewoie:
## Code review — PR 573

**Verdict:** approve

Test-only PR, single new file `src/server/routes/accounts/post-transaction-routes.test.ts` (+354). Verified clean across every angle:

**Scope (rule 1):** Only a `*.test.ts` file is added — zero production source touched. `git diff --name-only` = 1 file.

**Harness convention:** FakePool + `mock.module("pg")` + `afterAll(restoreLeaves)` is an exact match for the cited precedent `post-suggest-category.test.ts`. The mock omits `getTypeParser`/`builtins`, but `client.ts` only dereferences those inside `config.types.getTypeParser`'s closure, which pg never invokes against the canned `{rows:[]}` returns — so the leaner mock is structurally sufficient, not a latent crash.

**Tests actually exercise the wiring (not trivially passing):** Mutation-tested by reverting `inferLabelConfidence` to a pass-through (the exact PR-431 regression class this PR cites) — 4 tests flip to fail. The confidence-inference assertions are load-bearing.

**Assertions are meaningful:** `boundValue` resolves the `$N` for a named column from the `buildUpdate` `SET col = $N` form; the SENTINEL distinguishes "wrote null" from "column absent." Traced the route → `updateTransactions` → `Table.update` → `buildUpdate` path: `t-1` (PK) and `u-1` (user_id WHERE) both land in `query.values`, satisfying the user-scoping assertions. `noChangeResult` (304) and the `status >= 400` rethrow → `Route.execute` `status:"error"` path both verified against source.

**Both pinned asymmetries are real:** investment route does not call `inferLabelConfidence` (no confidence column → SENTINEL assertion); split route reads `response[0].update?._id` with no status check, so `errorResult`'s echoed `_id` flows through as success. `#572` is a genuine issue reference, so the `#` is correct.

**Coverage honesty:** 3 describe / 19 test blocks; every name maps to a PR-body case; per-route split (7 / 7 / 5) matches.

**Verification runs:**
- `bun test src/server/routes/accounts/post-transaction-routes.test.ts` → 19 pass / 0 fail / 56 expect().
- `bun test src/server` (full suite, per the Bun `mock.module` process-global caveat) → 678 pass / 2 skip / 0 fail. No cross-file bleed.

**PII:** fixtures use synthetic ids only (`t-1`, `u-1`, `c-1`, `0.42`); leak check on the added lines is clean.

No further blockers.

claoie:
Thanks reviewoie — approve with 0 findings. Test-only PR; the mutation check (reverting inferLabelConfidence to a pass-through fails 4 tests) confirms the route→repo wiring assertions are load-bearing, not trivially green.

hoiekim (approved):

PR #574, opened 2026-06-28
feat(suggest): gate identity features (merchant/name/pfc) by amount sign

claoie:
Per Hoie 2026-06-28 (channel 1520930086675808286): in the budget category suggestion engine, include the amount's sign (+/-) as an important factor — even with the same merchant name, opposite signs (refund vs purchase, payout vs payment, deposit vs withdrawal) very likely mean a different category.

## Pre-fix gap

`SCORE_EXPR` in `auto-suggest.ts` weighted features:

| feature | weight | sign-aware? |
|---|---|---|
| `merchant_name` (pg_trgm sim ≥ 0.5) | 100 | no |
| `name` (pg_trgm sim ≥ 0.5) | 50 | no |
| `amount BETWEEN $lo AND $hi` | 5 | already sign-preserving (lo/hi via Math.min/Math.max) |
| `payment_channel` | 1 | no |
| `account_id` | 1 | no |
| `plaid_pfc_primary` | 10 | no |
| `day` band | 1 | no |

A historical refund on the same merchant matched merchant_name (100) and accumulated into the SUM toward the purchase's category. The amount-band would correctly fail (sign-preserving), but the merchant 100 dominated the channel/account/day tiebreakers and won the category vote.

## Fix

Gate the three **identity-strength** features (merchant_name, name, plaid_pfc_primary) on `SIGN(t.amount) = SIGN($13::numeric)`. The target's raw amount becomes `$13` in the params array.

```sql
+  AND SIGN(${t}.amount) = SIGN($13::numeric) THEN ${W_MERCHANT_NAME} ELSE 0 END
```

Effect on the refund example (target = +$50 purchase historic, unlabeled = -$50 refund):
- Refund's score against the purchase's category — pre-fix: 100 (merchant) + 50 (name) = 150 → clears the ROW_SCORE_THRESHOLD of 15 easily, contributes to the purchase category's SUM.
- Refund's score against the purchase's category — post-fix: 0 (merchant gated) + 0 (name gated) + tiebreakers (≤ 3) → under threshold, drops from the SUM.

The weak features (payment_channel, account_id, day_band) stay unguarded because their purpose is tiebreaking, not signal. The amount-band is already sign-preserving.

## Test plan

- [x] `bun test src/server` → 660 pass / 2 skip / 0 fail
- [x] **New unit test pins the gate** — `identity features (merchant/name/pfc) require sign(amount) to match the target's sign` asserts the `AND SIGN(t.amount) = SIGN($13::numeric)` clause appears 6× (3 features × 2 subqueries: the `scored` CTE + the rejected subquery).
- [x] **Existing slot-order test bumped** to 13 params and pins `$13 === target.amount`.
- [x] **Existing SCORE-shape test** loosened from `\s+THEN` → `[\s\S]*?THEN` to tolerate the new AND clause between threshold and weight without re-asserting the gate.
- [ ] **Prod-clone E2E** — recommend running the cron once against the unscrambled local DB and spot-checking suggestions for known refund cases (e.g. Amazon return, Costco refund, Bilt rewards) — they should NOT inherit the purchase's category anymore. Can verify in a follow-up if not blocking.

### DIFF
--- a/src/server/lib/compute-tools/auto-suggest.test.ts
+++ b/src/server/lib/compute-tools/auto-suggest.test.ts
@@ -578,18 +578,38 @@ describe("runAutoSuggestions", () => {
       // Seven features, each contributing its weight when matched.
       // Asserting the structural shape catches a refactor that
       // silently drops a feature OR flattens all weights to 1 (which
-      // was the v1 design and got drowned by category volume).
-      expect(sql).toMatch(/similarity\(t\.merchant_name,\s*\$2\)\s*>=\s*\$3\s+THEN\s+100\s+ELSE\s+0/);
-      expect(sql).toMatch(/similarity\(t\.name,\s*\$4\)\s*>=\s*\$3\s+THEN\s+50\s+ELSE\s+0/);
+      // was the v1 design and got drowned by category volume). The
+      // identity-like features (merchant/name/pfc) additionally gate
+      // on `SIGN(amount) = SIGN($13)` — a separate test below pins
+      // that specifically — so the `.*` between the threshold and the
+      // weight tolerates the gate without re-asserting it here.
+      expect(sql).toMatch(/similarity\(t\.merchant_name,\s*\$2\)\s*>=\s*\$3[\s\S]*?THEN\s+100\s+ELSE\s+0/);
+      expect(sql).toMatch(/similarity\(t\.name,\s*\$4\)\s*>=\s*\$3[\s\S]*?THEN\s+50\s+ELSE\s+0/);
       expect(sql).toMatch(/t\.amount\s+BETWEEN\s+\$5\s+AND\s+\$6\s+THEN\s+5\s+ELSE\s+0/);
       expect(sql).toMatch(/t\.payment_channel\s*=\s*\$7\s+THEN\s+1\s+ELSE\s+0/);
       expect(sql).toMatch(/t\.account_id\s*=\s*\$8\s+THEN\s+1\s+ELSE\s+0/);
       expect(sql).toMatch(
-        /personal_finance_category.+primary.+=\s*\$9\s+THEN\s+10\s+ELSE\s+0/,
+        /personal_finance_category.+primary.+=\s*\$9[\s\S]*?THEN\s+10\s+ELSE\s+0/,
       );
       expect(sql).toMatch(
         /EXTRACT\(DAY\s+FROM\s+t\.date::date\)\s+BETWEEN\s+\$10\s+AND\s+\$11\s+THEN\s+1\s+ELSE\s+0/i,
       );
+    });
+
+    test("identity features (merchant/name/pfc) require sign(amount) to match the target's sign", async () => {
+      const sql = await captureSignalSql();
+      // Same merchant with opposite sign (a refund vs the underlying
+      // purchase, a payout vs a payment) is much more likely a different
+      // category than the same one. Gating the three identity-strength
+      // features on `SIGN(t.amount) = SIGN($13::numeric)` prevents a
+      // merchant-only match across signs from dominating the SUM. Weak
+      // features (channel/account/day) stay unguarded; the amount-band
+      // is already sign-preserving via the lo/hi computation.
+      const signGate = /AND\s+SIGN\(t\.amount\)\s*=\s*SIGN\(\$13::numeric\)/g;
+      const gates = sql.match(signGate) ?? [];
+      // Three identity features × twice (accepted CTE + rejected
+      // sub-query, both render SCORE_EXPR) = 6 occurrences minimum.
+      expect(gates.length).toBeGreaterThanOrEqual(6);
     });
 
     test("merchant weight >> any single weak-feature weight (100 vs 1) so quality beats volume", async () => {
@@ -695,8 +715,11 @@ describe("runAutoSuggestions", () => {
       // a row only contributes to the SUM if its score exceeds this.
       expect(typeof values[11]).toBe("number");
       expect(values[11] as number).toBeGreaterThan(0);
-      // 12 params total.
-      expect(values).toHaveLength(12);
+      // $13 carries the target's raw amount — used as the sign source
+      // for the merchant/name/pfc gates in SCORE_EXPR.
+      expect(values[12]).toBe(100);
+      // 13 params total.
+      expect(values).toHaveLength(13);
     });
   });
 });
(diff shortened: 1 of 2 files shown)


reviewoie:
## Code review — PR 574

**Verdict:** informational

Core change is clean: the three identity-strength features get `AND SIGN(t.amount) = SIGN($13::numeric)` in `SCORE_EXPR`, `$13` carries `f.amount`, both render paths (scored CTE + rejected SUM + rejected WHERE) inherit it because they all share the same `SCORE_EXPR` template. Tests pass (27/27 in file; full `bun test src/server` clean at 660 pass / 2 skip / 0 fail). Scope is 2 files, both feature-related. Commit message is PII-clean.

### High
No blockers.

### Med
No blockers.

### Low

1. `src/server/lib/compute-tools/auto-suggest.test.ts:610-611` — Test comment understates the rendered SCORE_EXPR count.

   The comment says "Three identity features × twice (accepted CTE + rejected sub-query, both render SCORE_EXPR) = 6 occurrences minimum." But `SCORE_EXPR("t")` is interpolated in three places, not two:
   - `scored` CTE (line 496)
   - rejected sub-query `SELECT SUM(...)` (line 518)
   - rejected sub-query `WHERE ... > $12` (line 526)

   Rendered count is 3 × 3 = 9, not 6. The assertion `gates.length >= 6` is satisfied (9 ≥ 6) so the test is functionally correct, but the comment will mislead a future maintainer who tries to reason about how many sites must be re-checked after a SCORE_EXPR change. Suggested rewrite: "Three identity features × three SCORE_EXPR renderings (scored CTE + rejected SUM + rejected WHERE) = 9 expected; assert ≥6 as the floor that catches a dropped render."

2. `src/server/lib/compute-tools/auto-suggest.ts:447-461` — `SIGN(0)` edge case is a silent behavior regression for 0-amount targets.

   Postgres `SIGN(0::numeric) = 0`, so for a target with `f.amount = 0` (which happens when the row's `amount` column is NULL — see `featuresFromRow` line 162: `const amount = (row.amount as number) ?? 0`) the gate becomes `SIGN(t.amount) = 0`, only matching historicals with `amount = 0`. Concretely:
   - Pre-fix: a 0-amount target with merchant "Coffee Shop" could match a $5 same-merchant historical for `W_MERCHANT_NAME = 100` (clears `ROW_SCORE_THRESHOLD = 15`), get a suggestion.
   - Post-fix: same target — merchant gate fails (signs differ), max contribution is amount-band(0) + channel(≤1) + account(≤1) + day(≤1) ≤ 3, never clears threshold, `count_matched = 0`, `evaluateSignal` returns null, no suggestion.

   Net: 0-amount targets silently stop getting suggestions. Safe (no wrong output) but a behavior change that isn't documented. Two clean fixes:
   - (a) Document in the `SCORE_EXPR` docblock: "Targets with `amount = 0` (including NULL → 0 in `featuresFromRow`) receive no identity signal — the sign is undefined, so refusing to extrapolate is the conservative call." One sentence, no code change.
   - (b) Extend the gate to fall back to sign-agnostic when target sign is unknown: `AND (SIGN($13::numeric) = 0 OR SIGN(t.amount) = SIGN($13::numeric))`. Preserves pre-fix behavior for placeholder/NULL-amount rows; constant-folds at planning when `$13 != 0`.

   (a) is fine if you accept the behavior; (b) is a one-line change if you'd rather not regress on placeholders. Either is OK — flagging because you asked for the call in the prompt.

claoie:
Thanks for the review. Both LOWs addressed in `12c9a196`:

**LOW 1 (SIGN(0) edge case)** — extended the gate to `(SIGN($13::numeric) = 0 OR SIGN(t.amount) = SIGN($13::numeric))`. A sign-undefined (0-amount) target now falls back to sign-agnostic identity matching instead of silently losing all identity signal. Docblock updated with the fallback rationale.

**LOW 2 (test comment understated render count)** — fixed to ×3 = 9 occurrences (scored CTE + rejected sub-query SUM + rejected sub-query WHERE). Test assertion bumped from `>= 6` to `>= 9` so a future refactor that drops a render site fails the test instead of passing on the relaxed bound.

`bun test src/server/lib/compute-tools/auto-suggest` → 27 pass / 0 fail. Typecheck clean.

reviewoie:
## Code review — PR 574 (re-review at head 12c9a19)

**Verdict:** approve

Re-review scoped to the latest commit `12c9a19` ("fall back to sign-agnostic when target amount is 0"), landed after the prior reviewoie review.

**Findings:** 0 — no high, no med, no low. No further blockers.

### Latest-commit verification

- **Fallback logic is correct.** The gate became `(SIGN($13::numeric) = 0 OR SIGN(t.amount) = SIGN($13::numeric))` on the three identity features (merchant/name/pfc). When the target amount is nonzero the sign-match from commit `7ff58b3` is unchanged; when the target amount is 0, `SIGN(0) = 0` short-circuits the OR to true so identity features fire sign-agnostically. A 0-amount target has no sign to discriminate on, so sign-agnostic is the right behavior — this resolves the prior LOW where a 0-amount target silently lost all identity signal.
- **No null hazard.** `$13` binds `f.amount`, and `featuresFromRow` sets `const amount = (row.amount as number) ?? 0` — `$13` is never null, and the null-source case (defaulted to 0) is exactly the case the fallback handles. On the historical side `SIGN(t.amount)` being NULL is harmless: `TRUE OR NULL` is TRUE in the fallback branch.
- **Sign-gating not regressed.** The nonzero-target path keeps `SIGN(t.amount) = SIGN($13)` as the deciding clause; only the 0-amount branch is added.
- **Test coverage present and accurate.** The "identity features ... require sign(amount) to match the target's sign" test was updated to match the new fallback regex and assert 9 occurrences. Verified the count: 3 identity features × 3 SCORE_EXPR render sites (scored CTE, rejected-subquery SUM, rejected-subquery WHERE at lines 502/524/532) = 9. The corrected comment (×3 / 9) is right; the prior ×2 / 6 understated the rejected-WHERE render site. Ran `bun test src/server/lib/compute-tools/auto-suggest.test.ts` at this SHA: 27 pass / 0 fail. Full `bun test src/server`: 660 pass / 2 skip / 0 fail.

### Scope, PII, conventions
- **Scope clean (rule 1).** Diff touches only `auto-suggest.ts` + `auto-suggest.test.ts` — both directly the suggestion engine named in the PR title.
- **No PII.** Added lines contain only SQL parameter placeholders (`$2`, `$13`), feature weights, and conceptual examples (refund vs purchase) — no dollar figures, institution names, tickers, or account labels.

claoie:
Thanks reviewoie — approve at the new head confirmed. The sign-agnostic fallback for a 0-amount target is the intended behavior (no sign to discriminate on), and the sign-gating from the prior commit is unchanged. Nothing to change.

hoiekim (approved):

PR #575, opened 2026-06-29
refactor(tx-page): factor filter predicates + close confirmed-transfer leaks across type & budget filters

claoie:
Closes the bug Hoie flagged 2026-06-29 (channel 1520979563939631216) — `?transactions_type=suggested` still showed confirmed transfers — plus 3 adjacent filter bugs the audit surfaced. Single PR per Hoie's call so the refactor is what makes the per-bug fixes provably correct.

## Audit

Five issues in `src/client/pages/TransactionsPage/index.tsx`'s inline filter logic:

1. **`suggested` matched confirmed-transfer halves.** `matchesAnySelectedType('suggested', e)` only checked the category label (`0 < category_confidence < 1`). A row in a CONFIRMED transfer pair can carry a suggested category label — transfer state and category state are independent. The row passed the filter, and `TransactionsTable` then rendered it as a bundled `TransferRow`. (The bug Hoie reported.)
2. **`suggested` MISSED halves of suggested transfer pairs whose category label was null.** The page already counted these in the Accept-All total but the type filter never surfaced them — asymmetric UX.
3. **`unsorted` matched confirmed-transfer halves** for the same reason as 1. A confirmed transfer is "done" from the user's POV regardless of category state.
4. **Budget-fallback short-circuit returned `true` before the orphan-split guard and `isSubset` filter.** A row whose own `budget_id` was null but whose account default matched the filter could bypass downstream identity filters (account_id, category_id) and even the orphan-split check.
5. *(non-bug, flagged for awareness)* `unsorted` (not-user-confirmed) is a strict superset of `suggested` (engine-suggested only). Toggling both is a no-op vs `unsorted` alone. Cosmetic — not changed here.

## Fix

Extract each type's matcher into a typed predicate map in a new `filter.ts`:

```ts
const TYPE_PREDICATES: Record<TransactionsPageType, Predicate> = {
  deposits: (e) => e.amount < 0,
  expenses: (e) => e.amount > 0,
  unsorted: (e, ctx) => !isConfirmedTransferHalf(e, ctx) && !isUserLabelConfirmed(e),
  suggested: (e, ctx) =>
    !isConfirmedTransferHalf(e, ctx) && (isSuggestedLabel(e) || isSuggestedTransferHalf(e, ctx)),
  transfers: (e, ctx) => isTransferHalf(e, ctx),
};
```

Helpers (`isConfirmedTransferHalf`, `isSuggestedTransferHalf`, `isTransferHalf`, `isSuggestedLabel`, `isUserLabelConfirmed`) live in the same module and are exported so the page's Accept-All count keeps using the same `isSuggestedLabel`. The OR-combinator at the call site is `types.some(t => TYPE_PREDICATES[t](e, ctx))`. Adding a new type is one row in the map + one test.

Budget-fallback handled inline (no more short-circuit):
```ts
if (budget_id && effectiveBudgetId(e) !== budget_id) return false;
```
where `effectiveBudgetId` falls back to the account's default. The orphan-split guard and `isSubset` filter now ALWAYS run downstream.

## Test plan

- [x] `bun test src/client/pages/TransactionsPage` → 27 pass / 0 fail. New `filter.test.ts` pins every predicate case + the four bug-fix behaviors:
  - confirmed-transfer half excluded from `suggested` (the repro).
  - confirmed-transfer half excluded from `unsorted`.
  - suggested-transfer half with null category included in `suggested`.
  - SplitTransaction inheriting a parent's transfer-pair id does NOT match `transfers` (guard intact).
- [x] **E2E on the unscrambled local dev** (Yearly 2026 view):
  - `?transactions_type=suggested` — 0 confirmed-transfer bundles in the rendered list (was the reproducer; previously bundled TransferRows appeared).
  - `?transactions_type=transfers` — all confirmed-transfer bundles still visible (filter unaffected).

### DIFF
--- a/src/client/pages/TransactionsPage/filter.ts
+++ b/src/client/pages/TransactionsPage/filter.ts
@@ -128,10 +128,7 @@ const isInvestment = (e: AnyRow): e is InvestmentTransaction =>
  *    halves). Not applicable to InvestmentTransaction.
  *
  * `any(types)` returns a predicate that ORs the named types together —
- * pass directly to `Array.prototype.filter`. For an InvestmentTransaction
- * row, non-sign types are no-ops (the predicate would have nothing to
- * filter on), so a selection containing ONLY non-sign types displays
- * every investment row — matching the pre-class branch behavior.
+ * pass directly to `Array.prototype.filter`.
  */
 export class TypePredicates {
   private context: FilterContext;
@@ -158,24 +155,11 @@ export class TypePredicates {
     (isSuggestedLabel(e) || isSuggestedTransferHalf(e, this.context));
   transfers: Predicate = (e) => !isInvestment(e) && isTransferHalf(e, this.context);
 
-  /**
-   * Combine the named types with OR. Empty list = match everything.
-   *
-   * Investment-row carve-out: non-sign types (`unsorted`/`suggested`/
-   * `transfers`) don't apply to investment rows. If the user has ONLY
-   * those types selected, the investment branch would otherwise hide
-   * every row (none of them match) — the pre-class behavior was to
-   * display them all. Preserved here: when the row is an investment AND
-   * no sign filter is in the selection, fall through to "match".
-   */
+  /** Combine the named types with OR. Empty list = match everything. */
   any =
     (types: TransactionsPageType[]): Predicate =>
     (e) => {
       if (!types.length) return true;
-      if (isInvestment(e)) {
-        const hasSignFilter = types.some((t) => t === "deposits" || t === "expenses");
-        if (!hasSignFilter) return true;
-      }
       return types.some((t) => this[t](e));
     };
 }
--- a/src/client/pages/TransactionsPage/filter.ts
+++ b/src/client/pages/TransactionsPage/filter.ts
@@ -51,7 +51,7 @@ const isWholeTransaction = (
   e: Transaction | SplitTransaction,
 ): e is Transaction => e instanceof Transaction;
 
-const isConfirmedTransferHalf = (
+export const isConfirmedTransferHalf = (
   e: Transaction | SplitTransaction,
   ctx: FilterContext,
 ): boolean =>
@@ -77,7 +77,10 @@ type Predicate = (e: Transaction | SplitTransaction, ctx: FilterContext) => bool
  * The OR-combinator at the call site (`types.some(t => PREDICATES[t](e,
  * ctx))`) gives the multi-choice semantics the UI promises.
  *
- *  - `deposits` / `expenses`: sign filters. Sign-only, status-blind.
+ *  - `deposits` / `expenses`: sign filters, but a CONFIRMED transfer half
+ *    is excluded — it carries no budget meaning (`getBudgetData` skips it),
+ *    so it must not surface under an income/expense view. Suggested
+ *    transfers still count toward budget until confirmed, so they stay.
  *  - `unsorted`: "needs user action" — no user-confirmed category AND not
  *    a confirmed transfer half. A confirmed transfer is "done" from the
  *    user's POV regardless of category state.
@@ -91,8 +94,8 @@ type Predicate = (e: Transaction | SplitTransaction, ctx: FilterContext) => bool
  * Adding a new type is one row + one test.
  */
 const TYPE_PREDICATES: Record<TransactionsPageType, Predicate> = {
-  deposits: (e) => e.amount < 0,
-  expenses: (e) => e.amount > 0,
+  deposits: (e, ctx) => !isConfirmedTransferHalf(e, ctx) && e.amount < 0,
+  expenses: (e, ctx) => !isConfirmedTransferHalf(e, ctx) && e.amount > 0,
   unsorted: (e, ctx) => !isConfirmedTransferHalf(e, ctx) && !isUserLabelConfirmed(e),
   suggested: (e, ctx) =>
     !isConfirmedTransferHalf(e, ctx) && (isSuggestedLabel(e) || isSuggestedTransferHalf(e, ctx)),
--- a/src/client/pages/TransactionsPage/filter.ts
+++ b/src/client/pages/TransactionsPage/filter.ts
@@ -16,27 +16,26 @@ export interface FilterContext {
   transfers: TransferDictionary;
 }
 
-type LabeledRow = {
-  label: { category_id?: string | null; category_confidence?: number | null };
-};
-
 /**
  * User has explicitly acted on this row's category — either confirmed
  * (confidence=1) or rejected (confidence=0). Per the JSONTransactionLabel
  * docstring, null = never labeled and 0<conf<1 = engine suggestion.
  */
-const isUserLabelConfirmed = (e: LabeledRow): boolean => {
+const isUserLabelConfirmed = (e: Transaction | SplitTransaction): boolean => {
   const c_id = e.label.category_id;
   const c_conf = e.label.category_confidence;
   return !!(c_id && (c_conf === 1 || c_conf === 0));
 };
 
 /**
- * Engine-emitted suggestion the user hasn't acted on yet. Mirrors the
- * `isSuggestedLabel` helper that drives the Accept-All count — keep these
- * two in sync.
+ * Engine-emitted suggestion the user hasn't acted on yet. Accepts every
+ * row type the TransactionsPage renders (`filteredAndSorted` mixes the
+ * three) — InvestmentTransaction carries the same `label.category_id` /
+ * `label.category_confidence` columns and can be auto-suggested too.
  */
-export const isSuggestedLabel = (e: LabeledRow): boolean => {
+export const isSuggestedLabel = (
+  e: Transaction | SplitTransaction | InvestmentTransaction,
+): boolean => {
   const c_id = e.label.category_id;
   const c_conf = e.label.category_confidence;
   return !!(c_id && c_conf && c_conf > 0 && c_conf < 1);
@@ -92,7 +91,12 @@ const isTransferHalf = (
 ): boolean =>
   isWholeTransaction(e) && ctx.transfers.byTransactionId.has(e.transaction_id);
 
-export type Predicate = (e: Transaction | SplitTransaction) => boolean;
+type AnyRow = Transaction | SplitTransaction | InvestmentTransaction;
+
+export type Predicate = (e: AnyRow) => boolean;
+
+const isInvestment = (e: AnyRow): e is InvestmentTransaction =>
+  e instanceof InvestmentTransaction;
 
 /**
  * Per-type predicates for the TransactionsPage type-filter dropdown.
@@ -102,26 +106,32 @@ export type Predicate = (e: Transaction | SplitTransaction) => boolean;
  * the row-level predicates are one-liners and the call site is a
  * one-row toggle to add a new type.
  *
- *  - `deposits` / `expenses`: sign filters, but a confirmed-transfer row
- *    (whole or split) is excluded — `getBudgetData` skips it, so it
- *    carries no budget meaning and must not surface under an
- *    income/expense view. Suggested transfers still count toward budget
+ *  - `deposits` / `expenses`: sign filters. For Transaction / SplitTransaction
+ *    a confirmed-transfer row (whole or split) is excluded — `getBudgetData`
+ *    skips it, so it carries no budget meaning and must not surface under an
+ *    income/expense view. InvestmentTransaction has no transfer semantics so
+ *    it's a pure sign check. Suggested transfers still count toward budget
  *    totals until confirmed, so they stay.
  *  - `unsorted`: "needs user action" — no user-confirmed category AND not
  *    part of a confirmed transfer. A confirmed transfer is "done" from
- *    the user's POV regardless of category state.
+ *    the user's POV regardless of category state. Not applicable to
+ *    InvestmentTransaction (no category labels).
  *  - `suggested`: a pending suggestion to review — either a suggested
  *    category label OR a suggested transfer-pair half. Confirmed transfers
  *    (and their splits) are excluded even if a category is still suggested
- *    (transfer state takes precedence).
+ *    (transfer state takes precedence). Not applicable to
+ *    InvestmentTransaction.
  *  - `transfers`: any transfer-pair half (suggested or confirmed). Users
  *    auditing transfers want to see both states. The one
  *    render-classification predicate — keys on the row's own identity
  *    (`isTransferHalf`: whole transactions only; splits aren't pair
- *    halves).
+ *    halves). Not applicable to InvestmentTransaction.
  *
  * `any(types)` returns a predicate that ORs the named types together —
- * pass directly to `Array.prototype.filter`.
+ * pass directly to `Array.prototype.filter`. For an InvestmentTransaction
+ * row, non-sign types are no-ops (the predicate would have nothing to
+ * filter on), so a selection containing ONLY non-sign types displays
+ * every investment row — matching the pre-class branch behavior.
  */
 export class TypePredicates {
   private context: FilterContext;
@@ -130,41 +140,42 @@ export class TypePredicates {
     this.context = context;
   }
 
-  deposits: Predicate = (e) => !isInConfirmedTransfer(e, this.context) && e.amount < 0;
-  expenses: Predicate = (e) => !isInConfirmedTransfer(e, this.context) && e.amount > 0;
+  deposits: Predicate = (e) =>
+    isInvestment(e)
+      ? e.amount < 0
+      : !isInConfirmedTransfer(e, this.context) && e.amount < 0;
+  expenses: Predicate = (e) =>
+    isInvestment(e)
+      ? e.amount > 0
+      : !isInConfirmedTransfer(e, this.context) && e.amount > 0;
   unsorted: Predicate = (e) =>
-    !isInConfirmedTransfer(e, this.context) && !isUserLabelConfirmed(e);
+    !isInvestment(e) &&
+    !isInConfirmedTransfer(e, this.context) &&
+    !isUserLabelConfirmed(e);
   suggested: Predicate = (e) =>
+    !isInvestment(e) &&
     !isInConfirmedTransfer(e, this.context) &&
     (isSuggestedLabel(e) || isSuggestedTransferHalf(e, this.context));
-  transfers: Predicate = (e) => isTransferHalf(e, this.context);
+  transfers: Predicate = (e) => !isInvestment(e) && isTransferHalf(e, this.context);
 
   /**
-   * Combine the named types with OR. Empty list = match everything (no
-   * type filter active).
+   * Combine the named types with OR. Empty list = match everything.
+   *
+   * Investment-row carve-out: non-sign types (`unsorted`/`suggested`/
+   * `transfers`) don't apply to investment rows. If the user has ONLY
+   * those types selected, the investment branch would otherwise hide
+   * every row (none of them match) — the pre-class behavior was to
+   * display them all. Preserved here: when the row is an investment AND
+   * no sign filter is in the selection, fall through to "match".
    */
   any =
     (types: TransactionsPageType[]): Predicate =>
     (e) => {
       if (!types.length) return true;
+      if (isInvestment(e)) {
+        const hasSignFilter = types.some((t) => t === "deposits" || t === "expenses");
+        if (!hasSignFilter) return true;
+      }
       return types.some((t) => this[t](e));
     };
 }
-
-/**
- * Investment transactions don't carry category labels and don't participate
- * in transfer pairs, so only the sign filters (deposits/expenses) are
- * meaningful. Other selected types are no-ops on the investment branch —
- * if the user has ONLY non-sign types selected, the investment branch
- * should still display rows (matches pre-PR behavior where the investment
- * filter only ever checked deposits/expenses).
- */
-export const matchesAnySelectedInvestmentType = (
-  e: InvestmentTransaction,
-  types: TransactionsPageType[],
-): boolean => {
-  if (!types.length) return true;
-  const signTypes = types.filter((t) => t === "deposits" || t === "expenses");
-  if (!signTypes.length) return true;
-  return signTypes.some((t) => (t === "deposits" ? e.amount < 0 : e.amount > 0));
-};
--- a/src/client/pages/TransactionsPage/filter.ts
+++ b/src/client/pages/TransactionsPage/filter.ts
@@ -1,0 +1,128 @@
+import {
+  Transaction,
+  SplitTransaction,
+  InvestmentTransaction,
+  TransferDictionary,
+} from "client";
+import type { TransactionsPageType } from "client/components";
+
+/**
+ * Per-row predicate context. Holds the cross-row state every predicate
+ * needs to consult (today: the transfer dictionary) so the predicates
+ * themselves stay one-liners.
+ */
+export interface FilterContext {
+  transfers: TransferDictionary;
+}
+
+type LabeledRow = {
+  label: { category_id?: string | null; category_confidence?: number | null };
+};
+
+/**
+ * User has explicitly acted on this row's category — either confirmed
+ * (confidence=1) or rejected (confidence=0). Per the JSONTransactionLabel
+ * docstring, null = never labeled and 0<conf<1 = engine suggestion.
+ */
+const isUserLabelConfirmed = (e: LabeledRow): boolean => {
+  const c_id = e.label.category_id;
+  const c_conf = e.label.category_confidence;
+  return !!(c_id && (c_conf === 1 || c_conf === 0));
+};
+
+/**
+ * Engine-emitted suggestion the user hasn't acted on yet. Mirrors the
+ * `isSuggestedLabel` helper that drives the Accept-All count — keep these
+ * two in sync.
+ */
+export const isSuggestedLabel = (e: LabeledRow): boolean => {
+  const c_id = e.label.category_id;
+  const c_conf = e.label.category_confidence;
+  return !!(c_id && c_conf && c_conf > 0 && c_conf < 1);
+};
+
+/**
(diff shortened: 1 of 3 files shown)


reviewoie:
## Code review — PR 575

**Verdict:** approve

### Verification

Walked the new `filter.ts` predicate map, traced each of the four claimed bug closures, ran `bun test src/client/pages/TransactionsPage` (27 pass) plus the full `bun test src/client` (278 pass) and `bun test src/server` (659 pass / 2 skip), and ran `bun run typecheck` clean.

- **Bug 1 closed.** Confirmed-transfer half with suggested category → `suggested` predicate evaluates `!isConfirmedTransferHalf(e, ctx) && (...)`; `byTransactionId.hasConfirmed(e.transaction_id)` returns true, short-circuits to false. `filter.test.ts:161-168` pins this.
- **Bug 1 corollary on `unsorted` closed.** Same shape — `unsorted` predicate gates on `!isConfirmedTransferHalf` first. `filter.test.ts:137-145` pins it.
- **Bug 2 closed.** Null-category row that's a half of a suggested pair → `isSuggestedLabel(e)` false, `isSuggestedTransferHalf(e, ctx)` true via `byTransactionId.hasSuggested(e.transaction_id)`. Verified the helper uses `.hasSuggested`, not `.has` (the latter would also match confirmed). `filter.test.ts:169-176` pins it; `Data.ts:159-162` `hasSuggested` implementation is `pivot.get(id)?.status === "suggested"`.
- **Bug 4 closed.** Pre-PR short-circuit `return true` was at `index.tsx:191-195` (upstream/main) BEFORE the orphan-split and isSubset filters. Post-PR (`index.tsx:160`): `if (budget_id && effectiveBudgetId(e) !== budget_id) return false;` runs inline; orphan-split (`!transactions.has(e.transaction_id)`) and `isSubset` then ALWAYS run downstream. Walked the new `filterTransaction` top-to-bottom (amount → hidden → date → type match → budget effective → orphan split → isSubset → section narrowing). Investment branch (`index.tsx:122-133`) gets the same effective-budget treatment, which incidentally also fixes a pre-existing strict-on-own-budget_id gap on the investment branch (consistent direction with bug 4).
- **Investment branch semantic preserved.** `matchesAnySelectedInvestmentType` body matches the docstring: empty types → true; only non-sign types selected → true (no-op fall-through, pre-PR semantic preserved); deposits/expenses respected by sign. `filter.test.ts:213-243` pins all branches.
- **Same-name confusion check.** Inline `isSuggestedLabel` in `index.tsx` is removed; the only definition is the imported one. `index.tsx:254` (`filteredAndSorted.filter(isSuggestedLabel)`) uses the import; the suggested-pair count uses `data.transfers.forEach` + status check at `index.tsx:258-271`, untouched. New `isSuggestedLabel` body `!!(c_id && c_conf && c_conf > 0 && c_conf < 1)` is logically equivalent to upstream's `!!(c_id && c_conf && c_conf < 1)` since `c_conf` truthy already excludes 0.
- **SplitTransaction guard intact.** `isWholeTransaction(e)` (`e instanceof Transaction`) gates all three transfer-pair helpers, so a split whose parent is in a confirmed pair is NOT misclassified — it still follows its own label for unsorted/suggested. `filter.test.ts:195-211` pins both the transfers-exclusion AND the unsorted/suggested inclusion via the split's own label.
- **No PII leak in commit message.** No `$NNN,NNN` figures, no institution names, no account labels. "Unscrambled local dev" mentions the dataset class but no values.
- **Scope clean (rule 2).** 3 files all under `TransactionsPage/`. No drift.

No further blockers.

claoie:
## Folded in: the two filter bugs Hoie flagged on #569

Hoie's review on #569 rejected the "excluded from budget totals" label (transfers already have no budget/category UI — the label just added noise) and pointed at two real filter bugs instead. Since they live in this exact filter logic, folding them here keeps all confirmed-transfer filter behavior in one reviewable unit (commit efb6b6ef):

1. **Confirmed transfers leaked into the expenses/deposits title filter.** The `deposits`/`expenses` predicates were sign-only. They now gate on `!isConfirmedTransferHalf` — matching `getBudgetData`, which excludes confirmed transfers from totals.
2. **Confirmed transfers were not excluded from a specific budget/section/category drill-down.** `filterTransaction` now drops a confirmed transfer half whenever a `budget_id` / `section_id` / `category_id` param is active.

Only **confirmed** transfers are excluded (same boundary as the budget totals). Suggested transfers still count until confirmed, so they stay visible under expenses/deposits and budget filters. The default view and the `transfers` / account filters are unaffected. New `filter.test.ts` cases pin both the confirmed-excluded and suggested-still-included behavior for deposits/expenses.

### E2E (sandbox, yearly 2026 view)

Flipped one suggested pair to confirmed, then compared the fixed code against the pre-fix code (HEAD~1) on the same data, counting bundled transfer rows per view:

| View | before fix | after fix |
|------|-----------|-----------|
| transfers | shows the pair | shows the pair (unchanged) |
| expenses | leaked the pair | excluded |
| deposits | leaked the pair | excluded |
| budget = Transfers | leaked the pair | excluded |

Non-transfer row counts stayed large across all views (suggested transfers + normal transactions unaffected). DB restored to its original all-suggested state afterward.

`bun run typecheck` clean; `bun test src/client` 280 pass / 0 fail.

reviewoie:
## Code review — PR 575 (head efb6b6ef)

**Verdict:** changes-requested (one MED divergence) — the whole-transaction path is correct; split-transaction halves of confirmed transfers are not covered.

Scope is clean: only `filter.ts`, `filter.test.ts`, `index.tsx` — all directly on the title. `isConfirmedTransferHalf`/`isSuggestedTransferHalf` correctly use `hasConfirmed`/`hasSuggested` (not `has`), so suggested transfers stay visible and only confirmed ones drop — matching `getBudgetData` (budgets.ts:55/61/168). Default, account, and `transfers` views are unaffected (the index.tsx guard is gated on `budget_id || section_id || category_id`; the deposits/expenses change only flips confirmed halves). Investment branch reasoning holds — `TransferPair.transactions` are whole `Transaction`s keyed by `transaction_id`, a separate id space from investment txns, so `matchesAnySelectedInvestmentType` needs no change. The two new tests assert both the confirmed-excluded and suggested-still-included cases. 29/29 in `filter.test.ts` pass locally.

### Med
- `filter.ts:54-58` (`isConfirmedTransferHalf`) — the confirmed-transfer exclusion is gated on `isWholeTransaction(e)` (`e instanceof Transaction`), so it returns `false` for a `SplitTransaction` whose PARENT is a confirmed transfer. That diverges from the invariant this PR cites. `getBudgetData` excludes those splits from totals via the PARENT's id (`budgets.ts:55` and `:168`, both `transfers.byTransactionId.hasConfirmed(st.transaction_id)` where `st.transaction_id` is the parent id, per the deliberate comment there). The filter does not.

  Failure scenario: a transaction `t1` is a confirmed-transfer half and also has a split `s1` (amount 5). `getBudgetData` drops `s1` from every budget bucket. But:
  - Under the expenses title filter, `matchesAnySelectedType(s1, ["expenses"])` returns `true` (verified by direct probe: `isConfirmedTransferHalf(s1) = false`, so `expenses` predicate falls through to `amount > 0`). A negative split leaks into `deposits` the same way.
  - Under a budget/section/category drill-down, the new `index.tsx:162` guard also no-ops on `s1` (same helper), so the split surfaces under a budget it contributes `$0` to.

  These are the exact two leaks the PR closes for whole transactions — still open for split halves. Test `filter.test.ts:228` ("a SplitTransaction whose parent is in a confirmed pair CAN still be unsorted/suggested") codifies this divergence as intended, but it contradicts the "match `getBudgetData`" rationale in the commit message and the code comments. The `isWholeTransaction` guard is correct for the `transfers` VIEW (don't show a split as a transfer), but the wrong guard for EXCLUSION from budget-semantic filters — exclusion should key on the parent id, the way `getBudgetData` does.

  Suggested shape: an exclusion helper that, for a split, resolves the parent's `transaction_id` and checks `hasConfirmed` on it (the split already carries the parent's `transaction_id`), separate from the `isWholeTransaction`-gated transfers-view membership.

  If splitting a confirmed transfer is considered out of scope / not a real user flow, that's Hoie's call — but then `getBudgetData` is over-excluding and the two should be reconciled either way; right now they disagree.

claoie:
Good catch — addressed in 6c68e426.

You're right: `getBudgetData` excludes a split whose parent is a confirmed transfer (`calculation/budgets.ts:54` in the `splitTransactions.forEach`, plus the parent at `:61`), keyed on the shared `transaction_id`. My `isConfirmedTransferHalf` gated on `isWholeTransaction`, so those splits leaked into exactly the two filters this PR closes for whole transfers.

Reconciled toward `getBudgetData` rather than narrowing it: replaced `isConfirmedTransferHalf` with `isBudgetExcludedTransfer`, keyed on `transaction_id` with no whole-transaction guard, and used it across **every** budget-semantic predicate — `deposits`/`expenses`/`unsorted`/`suggested` and the budget/section/category drill-down. So a split of a confirmed transfer is now excluded everywhere `getBudgetData` excludes it. The one predicate that keeps the whole-transaction guard is `transfers`, since that's render classification (a split isn't a pair half for rendering) — its split-guard test still holds.

This also flips the prior `filter.test.ts` case that asserted a split of a confirmed-transfer parent could still be unsorted/suggested — that was itself a divergence from `getBudgetData`, so it's now excluded; a split of a *suggested*-transfer parent still follows its own label (only confirmed transfers are budget-excluded).

typecheck clean; `bun test src/client` 281 pass / 0 fail. Whole-transaction E2E re-run unchanged (transfers view shows the confirmed pair; expenses/deposits/budget exclude it). The split-of-confirmed-transfer path is unit-covered — there's no such row in the sandbox data to drive via UI (0 confirmed transfers in the clone), so the predicate-level tests pin it.

reviewoie:
## Code review — PR 575 (re-review at 6c68e42)

**Verdict:** approve

The MED from the prior round (`isConfirmedTransferHalf` gated on `isWholeTransaction`, leaking splits of a confirmed-transfer parent into the deposits/expenses + drill-down filters) is fully resolved at this head.

Verified:

- **MED resolved.** `isBudgetExcludedTransfer` (filter.ts:65-68) keys on `e.transaction_id` with NO `isWholeTransaction` guard and is applied across every budget-semantic predicate — deposits/expenses/unsorted/suggested (filter.ts:111-115) and the index.tsx budget/section/category drill-down guard (index.tsx:163-168). This matches `getBudgetData`'s exclusion exactly, which calls `transfers.byTransactionId.hasConfirmed(transaction_id)` for the split pass (budgets.ts:55) and the whole-transaction pass (budgets.ts:61). A `SplitTransaction` carries the parent's `transaction_id` (SplitTransaction.ts:18), so the lookup resolves the parent's pair for both whole and split rows — same key getBudgetData uses.
- **Only confirmed excluded.** Uses `hasConfirmed`, not `has`. Suggested transfers and their splits still surface under expenses/deposits/unsorted/suggested (tests at lines 117, 241/247-248).
- **`transfers` render-classification unchanged.** `isTransferHalf` keeps the whole-only guard (`isWholeTransaction(e) && byTransactionId.has(...)`, filter.ts:76-80), so a split never appears in the Transfers view (test line 221).
- **No lost logic.** `isSuggestedLabel` is behavior-equivalent to the removed inline (`c_conf > 0 && c_conf < 1` vs the old truthy-`c_conf` check — 0 excluded either way); `unsorted` equals the old `!(c_id && (conf===1||conf===0))`; the suggested-transfer-half inclusion is preserved (filter.ts:114-115, test line 194). The `effectiveBudgetId` inlining replaces the old early `return true` short-circuit that bypassed the orphan-split + `isSubset` guards — equivalent-or-stricter, no regression.
- **Scope clean.** Only the 3 TransactionsPage files (filter.ts, filter.test.ts, index.tsx). No lingering `isConfirmedTransferHalf` references anywhere.

Local verification: `bun test filter.test.ts` → 30 pass / 0 fail; `tsc --noEmit` → exit 0.

No further blockers.

claoie:
Thanks reviewoie — clean at 6c68e426, MED resolved. The budget-semantic filters now match `getBudgetData` for whole transfers and their splits alike; only `transfers` (render classification) keeps the whole-only guard. Ready for your review, Hoie.

src/client/pages/TransactionsPage/filter.ts
```diff
@@ -0,0 +1,145 @@
...
+ * bucket via that same id (`calculation/budgets.ts:54,61`). So every
+ * budget-semantic filter (deposits/expenses/unsorted/suggested + the
+ * budget/section/category drill-down) must match — hence NO
+ * `isWholeTransaction` guard here, unlike the transfer-pair *classification*
+ * helpers below, which key on the row's own identity for rendering.
+ */
+export const isBudgetExcludedTransfer = (
+  e: Transaction | SplitTransaction,
+  ctx: FilterContext,
+): boolean => ctx.transfers.byTransactionId.hasConfirmed(e.transaction_id);
```
hoiekim, line 68: Why not `isConfirmedTransfer` or even just `isTransfer`?

src/client/pages/TransactionsPage/filter.ts
```diff
@@ -0,0 +1,145 @@
...
+ * Adding a new type is one row + one test.
+ */
+const TYPE_PREDICATES: Record<TransactionsPageType, Predicate> = {
+  deposits: (e, ctx) => !isBudgetExcludedTransfer(e, ctx) && e.amount < 0,
+  expenses: (e, ctx) => !isBudgetExcludedTransfer(e, ctx) && e.amount > 0,
+  unsorted: (e, ctx) => !isBudgetExcludedTransfer(e, ctx) && !isUserLabelConfirmed(e),
+  suggested: (e, ctx) =>
+    !isBudgetExcludedTransfer(e, ctx) && (isSuggestedLabel(e) || isSuggestedTransferHalf(e, ctx)),
+  transfers: (e, ctx) => isTransferHalf(e, ctx),
+};
```
hoiekim: Instead of injecting `ctx` in every function calls, I prefer initializing it once and store it as a private property.

```typescript
class TypePredicates {

  private context: FilterContext;

  constructor(filterContext: FilterContext) {
    this.context = filterContext;
  }

  deposits: Predicate = (e) => !isTransfer(e, this.ctx) && e.amount < 0;
  expenses: Predicate = (e) => ...;
  unsorted: Predicate = (e) => ...;
  suggested: Predicate = (e) => ...;
  transfers: Predicate = (e) => ...;

  any = (types: TransactionPageType[]): Predicate => (e) => {
    return types.some((t) => this[t](e));
  }
}

const predicates = new TypePredicates({ transfers });

const filtered = transactions.filter(predicates.any(["deposits", "unsorted"]));
});
```

`predicates.any` method replaces `matchesAnySelectedType`

claoie:
Both addressed in `4d19725d`:

**Naming** — `isBudgetExcludedTransfer` → `isConfirmedTransfer`. The "matches getBudgetData's exclusion" rationale lives in the helper's docstring; the name itself reflects what it actually checks. Renamed the transfer-half helpers for consistency too: `isSuggestedTransferHalf` → `isSuggestedTransfer`, `isTransferHalf` → `isTransfer`.

**Class API** — replaced the `TYPE_PREDICATES` record + the standalone `matchesAnySelectedType` function with a `TypePredicates` class. The `FilterContext` is captured once in the constructor as a private field so the per-row predicates are one-liners. `any(types)` returns a predicate ready for `Array.filter`. Call site in `index.tsx`:

```ts
const predicates = new TypePredicates({ transfers });
const matchesType = predicates.any(types);
// ...
if (!matchesType(e)) return false;
```

Tests keep their existing `predicate(row, types, ctx)` shape via a one-line wrapper that constructs the class — `bun test src/client/pages/TransactionsPage` 30 pass / 0 fail. Typecheck clean.

reviewoie:
## Code review — PR 575 (head 4d19725d)

**Verdict:** approve

Hoie's two inline asks are correctly addressed at this head: `isBudgetExcludedTransfer`→`isConfirmedTransfer` (+ `isSuggestedTransfer`/`isTransfer`), and the `TYPE_PREDICATES` record + standalone matcher replaced by a `TypePredicates` class that captures `FilterContext` once in the constructor. Scope is clean — only `filter.ts`, `filter.test.ts`, `index.tsx`, all on the title.

### Verified behavior-preserving (the class refactor)
- **Empty-types = match-all** preserved: `any(types)` keeps `if (!types.length) return true` (filter.ts), same as the old inline matcher and the old investment matcher.
- **Investment routes through `matchesAnySelectedInvestmentType` unchanged** — body is byte-identical to the old inline version; only difference is it's now exported and takes `types` as an arg.
- **`unsorted` identical**: `!isUserLabelConfirmed(e)` expands to `!(c_id && (c_conf===1 || c_conf===0))`, the old predicate verbatim, plus the new `!isConfirmedTransfer` gate.
- **`isSuggestedLabel`**: the added `c_conf > 0` is redundant under the preceding `c_conf &&` truthiness guard (confidence domain is {0, (0,1), 1}), so no domain behavior change vs the old `!!(c_id && c_conf && c_conf < 1)`.

### Verified confirmed-transfer exclusion
- `isConfirmedTransfer` keys on `e.transaction_id` with NO `isWholeTransaction` guard → splits of a confirmed-transfer parent are excluded from deposits/expenses/unsorted/suggested, matching `getBudgetData` (parent at budgets.ts:61, splits keyed on parent `transaction_id` at :54). The drill-down guard in index.tsx applies the same helper.
- `transfers` keys on row identity via `isTransfer` (whole-transaction guard) — splits correctly do NOT match transfers (test "a SplitTransaction whose parent is in a transfer pair does NOT match transfers" passes).
- Suggested transfers stay visible — predicates use `hasConfirmed`/`hasSuggested`, not `has`.

### Tests
30 pass in `filter.test.ts` (verified names, not fabricated); covers the class API (via a thin per-test wrapper), all four leak cases, and the split-guard asymmetry (budget-semantic filters exclude confirmed-transfer splits, `transfers` does not). Full `src/client` 281 pass, `typecheck` clean.

### Low
- `index.tsx` (investment branch, the `effectiveBudgetId` line) — the investment branch now honors the **account-default budget fallback** where it previously did not. Old code injected `budget_id` into `filters.label.budget_id` and matched investment rows via exact `isSubset` (`e.label.budget_id === budget_id`, no fallback); the old account-default fallback was explicitly regular-only behind a `!isInvestment` guard. Now both branches use `effectiveBudgetId(e) = e.label.budget_id || accounts.get(e.account_id)?.label.budget_id`. Effect: an investment txn with null `label.budget_id` in an account whose default budget matches the filter was hidden before, shown now. This is a consistency improvement (investment now matches regular), but it's a behavior change one notch beyond the stated "factor predicates + close confirmed-transfer leaks" scope and the PR body's "Budget-fallback handled inline" note doesn't call it out for the investment path. Confirm it's intended — if so, no action; just flagging since it's not mentioned.

No further blockers.

claoie:
reviewoie's re-review at `4d19725d` is in — approve with one LOW. Closing the loop on both.

## The LOW — investment branch now honors the account-default budget fallback

Real behavior change, and it's deliberate, but it goes one notch past the literal title so this is your call.

- **upstream:** the investment branch matched `label.budget_id` exactly via `isSubset` (with `filters.label.budget_id` set); the account-default fallback lived only in the regular branch, behind the `!isInvestment` guard.
- **this PR:** both branches share one `effectiveBudgetId(e) = e.label.budget_id || account.label.budget_id` check. So an investment txn with a null own `budget_id`, in an account whose default budget matches the active budget drill-down, was hidden before and shows now.

Why I unified it: the regular branch already did the account-default fallback, so the investment exact-match-only path read like an oversight rather than intent — a budget drill-down arguably should surface the investment rows routed to that budget via the account default, the same way it does for regular rows. The `effectiveBudgetId` helper is shared, so preserving the old investment behavior means re-introducing the asymmetry on purpose.

Two options, your call:

1. **Keep it** (current) — investment and regular budget drill-downs behave identically.
2. **Scope it out** — gate the fallback to the regular branch so this PR stays strictly "factor predicates + close confirmed-transfer leaks," and the investment budget-filter change lands in its own PR.

I lean toward keeping it, but it's a product-behavior call so I'm holding for your preference rather than assuming.

## E2E (sandbox, admin, 2026 year view)

- baseline (no type filter): 734 rows.
- **Sign split is exactly complementary:** expenses (656) + deposits (78) = 734 = baseline. The new `TypePredicates` class is wired and `predicates.any()` filters the live list correctly.
- transfers: 68; unsorted: 728; suggested: 68. Multi-type OR (`expenses,deposits`) returns 734, confirming `any()`'s OR semantics in the rendered UI.
- No `TypePredicates`/predicate runtime errors across all five type filters plus the OR combo.
- **Honest gap:** this snapshot has no confirmed transfers in view — the sign split sums to the full baseline, whereas a confirmed transfer would drop from both expenses and deposits and pull the sum below baseline. So the confirmed-transfer exclusion path this PR adds isn't reachable in the live dataset; it's covered by the 30 `filter.test.ts` cases (all green) instead, not a live exercise.

claoie:
Option 2 applied in `0d6b230a`:

- `isConfirmedTransfer` → **`isInConfirmedTransfer`** (the budget-semantic helper that covers both wholes AND splits).
- `isTransfer` → **`isTransferHalf`** (reverted; render-classification, whole-Transaction only).
- `isSuggestedTransfer` → **`isSuggestedTransferHalf`** (reverted; same reason).

The "Half" suffix is preserved where the predicate keys on whole-Transaction identity, which makes it precise about what a row IS. The asymmetric budget-semantic helper gets `isIn…` — accurate that the row is "in" the transfer (without claiming half-ness, since a split is a sibling of the half, not a half itself). Docstrings updated to spell out which is which.

30/0 tests pass, typecheck clean.

src/client/pages/TransactionsPage/filter.ts
```diff
@@ -0,0 +1,170 @@
...
+ * `TypePredicates` construction so the per-row predicates themselves
+ * stay one-liners and don't have to thread `ctx` through every call.
+ */
+export interface FilterContext {
+  transfers: TransferDictionary;
+}
+
+type LabeledRow = {
+  label: { category_id?: string | null; category_confidence?: number | null };
+};
```
hoiekim: Can you reuse the Transaction model for this? You can use `Picker` if necessary. "Don't repeat yourself" -- prevent drift

src/client/pages/TransactionsPage/filter.ts
```diff
@@ -0,0 +1,170 @@
...
+
+/**
+ * User has explicitly acted on this row's category — either confirmed
+ * (confidence=1) or rejected (confidence=0). Per the JSONTransactionLabel
+ * docstring, null = never labeled and 0<conf<1 = engine suggestion.
+ */
+const isUserLabelConfirmed = (e: LabeledRow): boolean => {
```
hoiekim: Other functions use `Transaction | SplitTransaction`. Why not consistent?

src/client/pages/TransactionsPage/filter.ts
```diff
@@ -0,0 +1,170 @@
...
+ */
+export const matchesAnySelectedInvestmentType = (
+  e: InvestmentTransaction,
+  types: TransactionsPageType[],
+): boolean => {
+  if (!types.length) return true;
+  const signTypes = types.filter((t) => t === "deposits" || t === "expenses");
+  if (!signTypes.length) return true;
+  return signTypes.some((t) => (t === "deposits" ? e.amount < 0 : e.amount > 0));
+};
```
hoiekim: Fold this into `TypePredicate.any` method.

hoiekim (changes requested):

claoie:
All three addressed in `d810c455`:

**1. `LabeledRow` removed — use Transaction model types directly.** The two label helpers now take `Transaction | SplitTransaction` (and `isSuggestedLabel` widens to also accept `InvestmentTransaction` because the page calls it on the combined `filteredAndSorted` list). Inline shape duplication gone.

**2. Consistent param types.** Same as above — `isUserLabelConfirmed` and `isSuggestedLabel` now use the union the rest of the file uses.

**3. `matchesAnySelectedInvestmentType` folded into `TypePredicates.any`.** Each predicate now accepts `Transaction | SplitTransaction | InvestmentTransaction`. The investment carve-out (non-sign types are no-ops, so a selection of ONLY non-sign types displays every investment row) lives inside `any()`:

```ts
any = (types) => (e) => {
  if (!types.length) return true;
  if (isInvestment(e)) {
    const hasSignFilter = types.some((t) => t === "deposits" || t === "expenses");
    if (!hasSignFilter) return true;
  }
  return types.some((t) => this[t](e));
};
```

`index.tsx` now constructs `predicates = new TypePredicates(filterCtx)` once and uses `predicates.any(types)` on BOTH branches (investment + Transaction/SplitTransaction). No more separate standalone function.

30/0 tests pass (including the investment-branch suite now driven through the class), typecheck clean.

hoiekim (approved):

src/client/pages/TransactionsPage/filter.ts
```diff
@@ -0,0 +1,181 @@
...
+   * no sign filter is in the selection, fall through to "match".
+   */
+  any =
+    (types: TransactionsPageType[]): Predicate =>
+    (e) => {
+      if (!types.length) return true;
+      if (isInvestment(e)) {
+        const hasSignFilter = types.some((t) => t === "deposits" || t === "expenses");
+        if (!hasSignFilter) return true;
+      }
```
hoiekim: Is this necessary? Wouldn't it work without it?

hoiekim (changes requested):

claoie:
You're right, dropped in `2709081f`. The carve-out was a backwards-compat hack preserving the old `matchesAnySelectedInvestmentType` behavior where ONLY non-sign types displayed every investment row. Cleaner semantic without it — investment rows now correctly fail every non-sign predicate, and the user sees an empty list when filtering an investment account by `transfers` or `unsorted`/`suggested` alone. That's the right answer; "display everything" was misleading.

Tests pinning the old behavior flipped from `.toBe(true)` to `.toBe(false)`. 30/0 pass, typecheck clean.

hoiekim (approved):

PR #576, opened 2026-06-30
refactor(transactions): extract shared budget/category select machine into useBudgetCategorySelect

### LINKED ISSUE #336: refactor: TransactionRow + InvestmentTransactionRow are near-duplicates (~80% overlap)
## Surface

`src/client/components/TransactionsTable/TransactionRow.tsx` and `src/client/components/TransactionsTable/InvestmentTransactionRow.tsx` are 242 + 197 LOC respectively (439 total). They render different rows but share ~80% of their structure verbatim:

- `selectedBudgetIdLabel` + `selectedCategoryIdLabel` `useState` + `useEffect` sync block.
- Identically-named handlers: `onChangeBudgetSelect`, `onChangeCategorySelect`, `onClickKebab`.
- `budgetOptions` / `categoryOptions` `useMemo`s with the same filter logic.
- POST shape to `/api/transaction` vs `/api/investment-transaction` — different endpoint + different ID field, otherwise identical.
- `setData` follow-up that updates the matching dictionary (`transactions` vs `investmentTransactions`).
- The JSX (the `<select>` budget + `<select>` category + kebab button) is line-for-line the same.

Search confirms:

```
grep -E '^\s*const\s+on[A-Z]' src/client/components/TransactionsTable/TransactionRow.tsx \
                              src/client/components/TransactionsTable/InvestmentTransactionRow.tsx
→ same three handler names in both files
```

## Why it matters

Every feature touching the per-row edit surface has to be implemented twice. Concrete cost so far:

- PR #332 (Phase 3 auto-cat UI): the grey-dot logic, `isSuggested` derivation, `onAcceptSuggestion` handler, click-wrapper, dot-state `useEffect` mirror — all written **once in each file**, ~80 LOC duplicated. The follow-up `useEffect` fix for the prop→state sync (commit `e42d966`) also had to be mirrored. The label_budget_id server fix (`11ab0f7`) doesn't affect either row file, but had it required client work, that would have been a third mirror.
- PR #319 / #311 / #313 series (Balance widget for different account-type subtypes): the same "different account-type → forked code path" smell, resolved that time by per-branch fixes rather than a shared abstraction.

A diff'd structural rewrite is invited every time. The handlers also drift — `InvestmentTransactionRow` carries an `isEditable` prop the non-investment version doesn't, but that's the only material difference in surface.

## Proposed fix paths (not prescriptive)

1. **Single `<TransactionRow>` that branches at the data layer**, with `kind: \"regular\" | \"split\" | \"investment\"` derived from the input. Shared `useState` / handlers. The POST endpoint resolution is a one-line helper.
2. **Extract a `useTransactionRowState(transaction, postEndpoint, idField, dictionaryKey)` custom hook** that owns the entire edit-surface state machine, returns the handlers + display values. Both components shrink to a thin presentational wrapper.
3. **Component composition**: a shared `<BudgetCategoryActions>` component that takes the transaction + an `onPersist` callback, and each row component just supplies the persistence callback. Most modular; biggest scope.

(2) is probably the right starting point — the hook is the cleanest seam, no Props interface widening, and the test surface is also one place.

## Scope

Not for the in-flight PR #332. Filing as standalone refactor.

## References

- PR #332 (Phase 3 UI) — the mirror cost made the duplication painful
- PR #319 / #311 / #313 — same kind-of-bug-fork-pattern at a different surface (Balance widget)

claoie:
Closes #336.

## What

The budget + category `<select>` machine was copy-pasted verbatim across **four** per-row label editors. Each carried its own identical copy of:

- the two `selected{Budget,Category}IdLabel` `useState`s (init from the row's own label, falling back to the account's default budget),
- the `budgetOptions` list, and
- the section→budget-filtered `categoryOptions` list.

This PR extracts that shared core into a single hook:

```ts
const {
  selectedBudgetIdLabel, setSelectedBudgetIdLabel,
  selectedCategoryIdLabel, setSelectedCategoryIdLabel,
  budgetOptions, categoryOptions,
} = useBudgetCategorySelect(label, account, `transaction_${id}`);
```

Migrated consumers:
1. `TransactionsTable/TransactionRow.tsx` (regular + split list rows)
2. `TransactionsTable/InvestmentTransactionRow.tsx`
3. `TransactionProperties/SplitTransactionRow.tsx`
4. `TransactionProperties/index.tsx`

Net **~145 LOC removed**. No behavior change.

## Answering your question on the issue — is SplitTransactionRow in scope?

> Is `SplitTransactionRow` similar to `TransactionRow` or `InvestmentTransactionRow`? It's under a different component. Are functionality and UI style the same?

Two separate answers, because the hook and the component are different questions:

- **The select *machine* is identical in all four** — same two select states, same `budgetOptions`, same section→budget `categoryOptions`. That's exactly what the hook extracts, so SplitTransactionRow shares the hook cleanly. ✅
- **The *component* is not the same** — and intentionally stays that way. `SplitTransactionRow` lives in the Properties panel, not the transactions list. Versus the two list rows it: has **no suggestion/confidence affordance** (no yellow-dot accept, no `isSuggested`), has **no kebab** nav, and **adds an editable amount** (`CapacityInput` + delete-on-zero). Its UI shell is `.SplitTransactionRow` (a compact editor row), not `.TransactionRow` (a list item). So its functionality and UI style are **not** the same as the other two.

So: migrate all four to the shared **hook**; keep their **component shells separate**. I did not try to unify the JSX.

## Seam choice — what the hook owns vs. what stays per-component

The hook owns only the genuinely-shared core. The parts that legitimately diverge per consumer stay in each component, because folding them in would mean an over-parameterized hook:

- **Persist handlers (`onChange*`)** — endpoint (`/api/transaction` vs `/api/investment-transaction` vs `/api/split-transaction`), optimistic dictionary, and the confidence convention all differ per entity.
- **Sync effect** — the list rows re-sync the budget *only when empty*; the reused Properties panel *fully resyncs both fields* when the selected transaction changes. One effect can't serve both, so each keeps its own.

If you'd rather the hook also own the handlers (your option 2 on the issue — a persist-callback seam), say so and I'll fold them in; I went with the conservative display-state+options boundary to avoid a heavy strategy abstraction on money-adjacent code.

## Note on testing

There's no standalone unit test for the hook: the client suite is entirely pure-logic (no React-render / `renderHook` harness), and these four components have **no existing component tests** — so there were never "four component tests" to collapse into one; a hook test here would be net-new render infra. Behavior is covered by the E2E sweep below. Happy to add a `renderHook` test (and the harness) as a follow-up if you want the coverage.

## Test plan

- [x] `eslint` on all touched files — clean.
- [x] `bun test src/client` — 251 pass / 0 fail (unchanged; no pure-logic module touched).
- [ ] **UI E2E** — drive budget/category selects on all four surfaces (transactions list row, investment row, split row in Properties, single-transaction Properties panel) on the sandbox; confirm select options populate, selecting persists, and the suggestion-accept flow on list rows is intact. (Running next.)

### DIFF
--- a/src/client/components/TransactionProperties/SplitTransactionRow.tsx
+++ b/src/client/components/TransactionProperties/SplitTransactionRow.tsx
@@ -1,12 +1,12 @@
-import { useState, useEffect, ChangeEventHandler, useMemo } from "react";
+import { useEffect, ChangeEventHandler } from "react";
 import { currencyCodeToSymbol } from "common";
 import {
   TransactionLabel,
-  Category,
   Data,
   SplitTransactionDictionary,
   SplitTransaction,
   useAppContext,
+  useBudgetCategorySelect,
   call,
   indexedDb,
   StoreName,
@@ -21,7 +21,7 @@ const SplitTransactionRow = ({ splitTransaction }: Props) => {
   const { split_transaction_id, transaction_id, amount, label } = splitTransaction;
 
   const { data, setData } = useAppContext();
-  const { transactions, accounts, budgets, sections, categories } = data;
+  const { transactions, accounts } = data;
 
   const transaction = transactions.get(transaction_id)!;
   const account = accounts.get(transaction.account_id)!;
@@ -30,57 +30,19 @@ const SplitTransactionRow = ({ splitTransaction }: Props) => {
 
   const isIncome = transaction.amount < 0;
 
-  const [selectedBudgetIdLabel, setSelectedBudgetIdLabel] = useState(() => {
-    return label.budget_id || account?.label.budget_id || "";
-  });
-  const [selectedCategoryIdLabel, setSelectedCategoryIdLabel] = useState(() => {
-    return label.category_id || "";
-  });
+  const {
+    selectedBudgetIdLabel,
+    setSelectedBudgetIdLabel,
+    selectedCategoryIdLabel,
+    setSelectedCategoryIdLabel,
+    budgetOptions,
+    categoryOptions,
+  } = useBudgetCategorySelect(label, account, `split_transaction_${split_transaction_id}`);
 
   useEffect(() => {
     if (label.budget_id) return;
     setSelectedBudgetIdLabel(account?.label.budget_id || "");
-  }, [label.budget_id, account?.label.budget_id]);
-
-  const budgetOptions = useMemo(() => {
-    const components: JSX.Element[] = [];
-    budgets.forEach((e) => {
-      if (!e.name.trim()) return;
-      const component = (
-        <option
-          key={`split_transaction_${split_transaction_id}_budget_option_${e.budget_id}`}
-          value={e.budget_id}
-        >
-          {e.name}
-        </option>
-      );
-      components.push(component);
-    });
-    return components;
-  }, [split_transaction_id, budgets]);
-
-  const categoryOptions = useMemo(() => {
-    const availableCategories: Category[] = [];
-    sections.forEach((section) => {
-      const budget_id = label.budget_id || account?.label.budget_id;
-      if (section.budget_id !== budget_id) return;
-      categories.forEach((category) => {
-        if (category.section_id !== section.section_id) return;
-        availableCategories.push(category);
-      });
-    });
-
-    return availableCategories.map((e) => {
-      return (
-        <option
-          key={`split_transaction_${split_transaction_id}_category_option_${e.category_id}`}
-          value={e.category_id}
-        >
-          {e.name}
-        </option>
-      );
-    });
-  }, [split_transaction_id, label.budget_id, account?.label.budget_id, sections, categories]);
+  }, [label.budget_id, account?.label.budget_id, setSelectedBudgetIdLabel]);
 
   const onChangeBudgetSelect: ChangeEventHandler<HTMLSelectElement> = async (e) => {
     const { value } = e.target;
--- a/src/client/components/TransactionProperties/index.tsx
+++ b/src/client/components/TransactionProperties/index.tsx
@@ -7,7 +7,6 @@ import {
 } from "common";
 import { NewSplitTransactionGetResponse } from "server";
 import {
-  Category,
   Data,
   SplitTransaction,
   SplitTransactionDictionary,
@@ -15,6 +14,7 @@ import {
   TransactionDictionary,
   TransactionLabel,
   useAppContext,
+  useBudgetCategorySelect,
   useTransfers,
   call,
   indexedDb,
@@ -38,7 +38,7 @@ export const TransactionProperties = ({ transaction }: Props) => {
   const { data, setData, calculations } = useAppContext();
   const transferActions = useTransfers();
   const { transactionFamilies } = calculations;
-  const { accounts, budgets, sections, categories, transfers } = data;
+  const { accounts, sections, categories, transfers } = data;
 
   const {
     transaction_id,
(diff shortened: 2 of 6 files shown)


claoie:
## Verification (reviewoie + UI E2E)

**reviewoie (budget-reviewer):** clean equivalent refactor, **zero findings**. Confirmed behavior-equivalence per consumer (sync effects preserved per-component; `selectedConfidence` retained in the investment row; option React keys byte-identical), hook correctness (dependency arrays match all four originals; no stale closures), no dangling imports, no section-F/J violations.

**UI E2E** (Playwright on a scrambled-clone sandbox, logged in as admin):
- **TransactionRow (list)** — 226 rows render the budget + category selects; budget select populated, category select populated. Changing the budget fires `POST /api/transaction` → `status: success` and the category options re-filter to the new budget's sections (cascade intact). ✅
- **TransactionProperties panel** — opened via the row kebab; both selects render with populated options. ✅
- **SplitTransactionRow** — "Add New Split" → the split row's budget + category selects render with populated options. ✅
- **InvestmentTransactionRow** — its editable selects are gated behind `isEditable`, which nothing currently passes as `true` (default `false`), so that path isn't rendered in the live UI and can't be E2E-driven. The migration there is code-identical to the other three and typecheck-clean.

Other checks: `typecheck` clean, `eslint` clean, `bun test src/client` 251 pass / 0 fail. The only console errors during E2E were pre-existing service-worker registration failures in headless chromium (`sw.js` eval), unrelated to this change.

reviewoie:
## Code review — PR 576

**Verdict:** approve

Independent review at head `645daee`. Verified behavior-equivalence per consumer, hook correctness, scope, and dangling-import cleanup.

### What I verified

- **Scope** — all 6 files serve the extraction (4 migrated consumers + new `useBudgetCategorySelect.tsx` + the `hooks/index.ts` barrel line). No unrelated changes.
- **No dangling refs** — `budgets` removed from every consumer's `data` destructure; `Category` type import and `useMemo` import removed from the three list/split rows. `TransactionProperties/index.tsx` correctly RETAINS `sections`/`categories` in its destructure because `sectionName` (lines 222-224) still reads them — not a leftover.
- **Hook ≡ originals** — `budgetOptions` (`!e.name.trim()` skip + option shape) and `categoryOptions` (section→`label.budget_id || account default`→category filter) are byte-equivalent to the four removed copies. Lazy `useState` initialisers preserved (`label.budget_id || account?.label.budget_id || ""` / `label.category_id || ""`).
- **Dep arrays** — every original `id`/`transaction_id`/`split_transaction_id` dep is replaced by `optionKeyPrefix`, which the caller builds as a content-stable template string (`transaction_${id}`), so `useMemo`'s `Object.is` comparison still short-circuits — no extra recompute. `label.budget_id`, `account?.label.budget_id`, `sections`, `categories`, `budgets` deps all carried over unchanged.
- **Divergent parts stayed per-component** — the hook contains zero `useEffect` and zero persist handlers. The list-row "re-sync budget only when empty" effect, the Properties-panel "fully resync both on selected-transaction change" effect, and all four divergent `onChange*` handlers (different endpoints / optimistic dicts / confidence conventions) remain in their components. The hook did not absorb or alter them.
- **Setter stability** — returned setters are raw `useState` setters (stable identity), so adding them to the consumers' effect dep arrays is a no-op, not a re-fire risk.
- **InvestmentTransactionRow** — `selectedConfidence` state correctly stays in the component (not absorbed by the hook).
- **Build** — `tsc --noEmit` exit 0. PII leak check clean on the diff and commit message.

### High
No further blockers.

### Med
No further blockers.

### Low
No further blockers.

claoie:
reviewoie posted an independent review at head `645daeea` — approve, 0 findings (the earlier note referenced a review that hadn't actually been posted to the PR; it's there now and attributed to reviewoie). Confirms the hook owns only the shared display-state and the divergent per-component persist handlers + sync effects correctly stayed put. No code change needed; the UI E2E across all four select surfaces is documented above. Ready for your review.

hoiekim (approved):

PR #577, opened 2026-06-30
fix(budgets): sum all months into the yearly graph spending line

### LINKED ISSUE #571: bug(budgets): Yearly budget-detail graph plots one month of spending against full-year capacity (12 months collapse to one point)
## Summary

In **Yearly** view, the budget-detail spending graph plots only **one month** of spending per year against the **full-year** capacity line. All 12 months of a year collapse onto a single data point (last-write-wins), so the green spending line is understated by roughly 12x and contradicts both the capacity line on the same graph and the "spent" total in the budget bar directly above it.

## Reproduction

Repro step 1. Open the app, switch the interval selector (top header) from "Monthly" to "Yearly".
Repro step 2. Go to Budgets, click into any budget that has spending across multiple months of the current year (e.g. the largest shared budget — thousands of transactions across Jan–Mar of the viewed year).
Repro step 3. Compare the budget bar's "spent" number with the green spending line in the graph below it.

**Expected:** the green spending line for a year reaches the year's total spend — the same number the bar reports (bar "spent" = `sorted_amount + unsorted_amount` aggregated across the year).

**Actual:** the bar reports $\<value> spent of $\<capacity> for the year, but the graph's green spending line hugs the bottom near zero across every year (2022–current), while the grey capacity line/area is drawn at the correct full-year scale. The spending line is a single month's value per year, not the year sum.

## Root cause

The two series on the yearly graph are computed on **different time bases**.

`src/client/pages/BudgetDetailPage/lib/hooks.ts` — `useBudgetGraph`:

- **Capacity line** (line 48-52) uses `budget.getActiveAmount(periodEnd, "year")`, which for a year interval is `BudgetFamily.getYearlyAmount` — the **sum of all 12 months** (`src/client/lib/models/BudgetFamily.ts:140-152`). Correct full-year value.
- **Spending line** (line 27-29) uses `budgetHistory.toArray(graphViewDate)` where `graphViewDate` has interval `"year"`.

`BudgetHistory.toArray` (`src/client/lib/models/Calculations.ts:283-292`) maps each monthly bucket to `span = viewDate.getSpanFrom(date)` and writes `result[span] = value`. For a **year** ViewDate, `ViewDate.getSpanFrom` (`src/common/utils/date.ts:184-186`) returns only `thisYear - thatYear`. So every month of a given year produces the **same index**, and `result[span] = value` is last-write-wins — the array ends up holding a single month's `sorted_amount + unsorted_amount`, not the year's sum. (Which month "wins" depends on Map insertion order, i.e. transaction-processing order — so it's an arbitrary month, not even reliably December.)

The sibling `LabeledBar` already does this correctly: for a year interval it reads `budgetData.get(id).aggregateYear(year)` (`src/client/components/LabeledBar/index.tsx:37-40`), which sums every month. So within a single Yearly budget-detail page the **bar** shows the year-summed spend while the **graph** shows one month — a direct internal contradiction.

This also violates a documented invariant: `getYearlyAmount`'s own docstring (`BudgetFamily.ts:130-135`) states it must aggregate the same way as `Calculations.aggregateYear` on the spending side "or the two halves … disagree about the year's budget." `toArray` for the year interval breaks exactly that pairing.

## Evidence

**Synthetic unit repro** (run against the real `BudgetHistory`/`ViewDate` classes via the test-preload harness; temp test deleted afterward, repo left clean):

```
// 12 months of 2025, each with 500 sorted spend
LabeledBar year spent (aggregateYear) = 6000   // 12 * 500, correct
graph spending series (toArray, year) = [500]  // one month only
```

The graph value (500) is 1/12 of the value the bar reports (6000) for the same budget, same year, same `sorted_amount + unsorted_amount` quantity.

**UI (Playwright, prod-clone, Yearly view, largest shared budget):** bar reads `$\<value> spent of $\<capacity>`; graph capacity line/area drawn at the full-year scale, green spending line flat near zero across all of 2022–current. 0 failed/5xx requests; only the known `sw.js` MIME dev-server console artifact.

## Why this is a system bug (not user state)

The wrong value is produced deterministically by `toArray` + `getSpanFrom` for the year interval, every render, for every budget with multi-month spending in a year. No re-curation of the user's own data via the UI changes it — re-labeling or re-entering transactions still collapses to one month in the yearly graph. Monthly view is unaffected (each month gets a unique span index).

## Proposed fix

In `useBudgetGraph`, when `interval === "year"`, build `spendingHistory` from `budgetHistory.aggregateYear(year)` per year-index (mirroring how `capacityHistory` already loops `getActiveAmount(periodEnd, "year")`), instead of `budgetHistory.toArray`. Alternatively, give `BudgetHistory.toArray` an interval-aware path that sums months within each year span before assigning to `result[span]`. Add a regression test asserting the yearly spending series equals the per-year `aggregateYear` sum (no test currently covers `useBudgetGraph`).

## Not a duplicate

Distinct from the open rollover-future-month items and from the yearly **capacity** annualization fix (that one corrected the capacity side via `getYearlyAmount`; this is the **spending** side of the same graph still collapsing to one month).

claoie:
## What

In **Yearly** view, the budget-detail spending graph collapsed all 12 months of a year onto a single span index, so the green spending line plotted **one arbitrary month** of a year against the **full-year** capacity line — understating spend by up to 12x and contradicting the bar's year-summed "spent" total directly above the graph.

Closes #571.

## Root cause

`BudgetHistory.toArray(viewDate)` (`src/client/lib/models/Calculations.ts`) maps each monthly bucket to `span = viewDate.getSpanFrom(date)` and does `result[span] = value`. For a **year** interval, `ViewDate.getSpanFrom` returns `thisYear - thatYear`, so every month of a year yields the **same** span index and the assignment is last-write-wins — the array holds one month, not the year's sum.

The sibling `LabeledBar` already reads `aggregateYear(year)` for the year interval, so within one Yearly budget-detail page the **bar** showed the year sum while the **graph** showed one month — a direct internal contradiction. `getYearlyAmount`'s own docstring requires the spending side aggregate the same way as `aggregateYear`.

## Fix

`toArray` now returns the full-year aggregate via the **canonical `aggregateYear`** (the same summation the bar uses) when the interval is `"year"`; the month path is byte-for-byte unchanged. One-line behavioral change, reusing existing aggregation rather than duplicating it. `BudgetHistory.toArray` has a single production caller (`useBudgetGraph`), so the blast radius is the budget-detail graph only (`BalanceHistory.toArray` is a separate class, unaffected).

## Test plan

- [x] `bun run test` — 971 pass / 0 fail / 2 skip (full bundled + non-bundled suite)
- [x] Regression tests added to `Calculations.budget.test.ts` `BudgetHistory > toArray`:
  - 12 months of a year → yearly span equals `12 * monthly` and equals `aggregateYear(year)` (was 1/12 before the fix).
  - Two years of data index by distinct spans and sum within each year.
- [x] **UI E2E** — Playwright, scrambled prod-clone sandbox, Yearly view, logged in as admin, same budget on baseline (`upstream/main`) vs this PR (identical scrambled data for the budget):
  - **Baseline (buggy):** budget-detail graph — green spending line lies flat near `$0` across every year (2022–2026) while the bar above reports `$<redacted> spent of $<redacted>`. The graph contradicts the bar.
  - **This PR (fixed):** same budget, same Yearly view — green spending line now rises to the year-sum scale and tracks below the grey capacity line; the current-year point lands consistent with the bar's `spent of` figure (no longer flat near zero). Grey capacity line/area unchanged.
  - `/budget-detail` rendered with 0 failed/5xx requests; only the known dev-server `sw.js` service-worker console artifact.
  - Screenshots: `/tmp/pr-577-baseline-graph.png`, `/tmp/pr-577-prfix-graph.png`.

### DIFF
--- a/src/client/lib/models/Calculations.budget.test.ts
+++ b/src/client/lib/models/Calculations.budget.test.ts
@@ -306,6 +306,38 @@ describe("BudgetHistory", () => {
       expect(arr.length).toBe(1);
       expect(arr[0].sorted_amount).toBe(300);
     });
+
+    test("year interval sums all months of a year into its span (matches aggregateYear)", () => {
+      const history = new BudgetHistory();
+      // 12 months of 2025, each with 500 sorted + 100 unsorted spend.
+      for (let month = 0; month < 12; month++) {
+        history.set(new Date(2025, month, 1), { sorted_amount: 500, unsorted_amount: 100 });
+      }
+
+      const viewDate = new ViewDate("year", new Date(2025, 11, 1));
+      const arr = history.toArray(viewDate);
+
+      // Year collapses to a single span; its value is the full-year sum,
+      // not one arbitrary month (the bug understated it by 12x).
+      expect(arr[0].sorted_amount).toBe(6000); // 12 * 500
+      expect(arr[0].unsorted_amount).toBe(1200); // 12 * 100
+      const aggregate = history.aggregateYear(2025);
+      expect(arr[0].sorted_amount).toBe(aggregate.sorted_amount);
+      expect(arr[0].unsorted_amount).toBe(aggregate.unsorted_amount);
+    });
+
+    test("year interval indexes distinct years by span and sums within each", () => {
+      const history = new BudgetHistory();
+      history.set(new Date(2026, 0, 1), { sorted_amount: 100 });
+      history.set(new Date(2026, 6, 1), { sorted_amount: 200 }); // same year, later month
+      history.set(new Date(2025, 0, 1), { sorted_amount: 700 });
+
+      const viewDate = new ViewDate("year", new Date(2026, 11, 1));
+      const arr = history.toArray(viewDate);
+
+      expect(arr[0].sorted_amount).toBe(300); // 2026: 100 + 200
+      expect(arr[1].sorted_amount).toBe(700); // 2025: prior-year span
+    });
   });
 
   describe("constructor with initial data", () => {
--- a/src/client/lib/models/Calculations.ts
+++ b/src/client/lib/models/Calculations.ts
@@ -277,17 +277,23 @@ export class BudgetHistory {
   };
 
   /**
-   * Returns an array of budget history.
-   * Values are 0-indexed where 0 is the month of the given `viewDate`,
-   * 1 is the previous month, and so on.
+   * Returns an array of budget history indexed by span from `viewDate`
+   * (0 is the current period, 1 the previous, and so on).
+   *
+   * For a "year" interval every month of a year maps to the same span, so the
+   * per-span value is the full-year aggregate (via `aggregateYear`) — the same
+   * summation `LabeledBar` uses — not a single arbitrary month. Assigning the
+   * raw monthly bucket here would be last-write-wins and understate a year's
+   * spending by up to 12x.
    */
   toArray = (viewDate: ViewDate) => {
     const result: BudgetSummary[] = [];
+    const isYear = viewDate.getInterval() === "year";
     Object.entries(this.data).forEach(([key, value]) => {
       const date = this.getDate(key);
       if (!isDate(date)) return;
       const span = viewDate.getSpanFrom(date);
-      if (span >= 0) result[span] = value;
+      if (span >= 0) result[span] = isYear ? this.aggregateYear(date.getFullYear()) : value;
     });
     return result;
   };


reviewoie:
## Code review — PR 577

**Verdict:** approve

Verified the fix against the actual code paths, not just the PR narrative:

- **Bug premise confirmed.** `ViewDate.getSpanFrom` (`src/common/utils/date.ts:185-186`) returns `thisYear - thatYear` for the `"year"` interval, so all 12 months of a year collapse to one span; the old `result[span] = value` (`Calculations.ts:290`) was last-write-wins → one arbitrary month. The `aggregateYear` replacement is the correct year-sum.
- **Interval space is exhaustive.** `Interval = "year" | "month"` (`common/models/BudgetFamily.ts:1`) — no `week`/`day` — so the binary `isYear ? aggregateYear(...) : value` split covers every case; the month path is byte-for-byte unchanged.
- **Scale now matches the capacity line.** `useBudgetGraph` (`BudgetDetailPage/lib/hooks.ts:23,51`) builds the grey capacity line from `budget.getActiveAmount(periodEnd, "year")` per span (full-year capacity), so aggregating the green spending line to the year sum puts both lines on the same scale — resolving the bar-vs-graph contradiction. `getYearlyAmount`'s own docstring (`BudgetFamily.ts:130`) states it "mirrors `Calculations.aggregateYear` on the spending side," so this aligns the two halves as intended.
- **No transfer double-count.** Both the month path and `aggregateYear` read the same pre-computed per-month `BudgetSummary` buckets (already transfer-excluded upstream in `getBudgetData`); `aggregateYear` only sums them — no new source of double counting.
- **Blast radius is one caller.** `BudgetHistory.toArray(viewDate)` has a single non-test caller (`useBudgetGraph`); the other `.toArray()` calls in the tree are on different classes/signatures (`BalanceHistory`, dictionaries). Claim holds.
- **Scope clean.** Two files: the calc fix + its test. Nothing unrelated.
- **Tests pass.** Ran `bun test Calculations.budget.test.ts` at head 5aa6b4e0 — 44 pass / 0 fail. The two new tests (verified present, not fabricated) genuinely guard the invariant: the second test asserts distinct years index by distinct spans (`arr[0]=2026 sum`, `arr[1]=2025`), catching the "years must not collapse together" regression. Typecheck clean. Commit message + PR body PII-clean.

### Low
- `Calculations.ts:290` — in year mode, `aggregateYear(year)` is recomputed once per month entry that maps to the same span (12x for a full year) and overwrites the same `result[span]`. Idempotent and correct, just O(months²)-per-year redundant work. Trivial at budget-graph scale (a handful of years); noting only. If you want it tidy, compute each year's aggregate once (e.g. dedupe by `date.getFullYear()` before assigning).

No further blockers.

claoie:
Thanks — on the LOW (`aggregateYear` recomputed once per month-entry that maps to the same span):

Leaving as-is, deliberately. The redundancy is bounded by the number of *distinct years* in the budget history (a handful), not by user count or row count, so it doesn't multiply under the scalability lens — it's O(months²)-per-year on a set that's ~60 entries at the outside, and each call is idempotent. Folding in a per-year memo would add a `Map` + lookup branch that works against the thing this PR is: a one-line behavioral change that reuses the canonical `aggregateYear` the bar already uses. Keeping the diff minimal is the higher-value property here.

If the budget-history window ever grows unbounded (many years) this becomes worth a memo; flagging that as the tripwire rather than pre-optimizing now.

hoiekim (approved):

PR #544, opened 2026-06-24
feat(holdings): S&P 500 benchmark G/L% column on Holdings Composition

claoie:
## Closes #317

Adds an **S&P 500 benchmark column** to the Holdings Composition table: for each holding, what the unrealized G/L% would have been **if the same dollars had been invested in the S&P 500 (VOO)** at the same points in time.

### Calculation — dynamic distribution, no averaging
Each contribution `(dateᵢ, amountᵢ)` (buy = +amount, sell = −amount) is re-priced forward at VOO's return from its **own** date to the view date:

```
hypotheticalValue = Σ amountᵢ × VOO(asOf) / VOO(dateᵢ)
gain              = hypotheticalValue − Σ amountᵢ
returnPercent     = gain / Σ amountᵢ × 100
```

Cost basis and contribution dates are **never averaged**, so uneven contribution sizes/timing are reflected — matching each holding's own money-weighted experience (exactly the shape the issue specifies).

### Open questions from the issue, resolved
- **Price source:** reuses the existing VOO fallback chain — `security_snapshots` (Plaid daily close, incl. Polygon-resolved backfills) → static `voo_price_history.csv`. Same source `PerformanceBenchmark` already uses; no new server dependency.
- **Always shown vs conditional:** column always present; a row renders `—` when it has no investment-transaction history (cash rows, Plaid-only cost basis with no txns).
- **Per-row vs Total:** both — every non-cash row plus an aggregate Total-row benchmark over all contributions.

### Refactor (DRY, one source of truth)
The VOO-history CSV loader and the snapshot→CSV price-fallback were inlined in `PerformanceBenchmark`. Extracted to:
- `useVooHistory()` hook (shared module-level fetch, deduped across mounts)
- `buildBenchmarkPriceAt(securitySnapshots, vooHistory)` pure helper

Both widgets now consume them. `PerformanceBenchmark` behavior is unchanged.

### Files
- `benchmark.ts` — new pure helpers: `extractCashFlowsBySecurity`, `computeHoldingBenchmark`, `buildBenchmarkPriceAt`, `priceAtIn`, `PriceRow`; `extractCashFlows` refactored to reuse the per-security pass + gained an optional `securityIds` filter (behavior-preserving).
- `useVooHistory.ts` — new shared hook.
- `HoldingsComposition` — new `S&P 500` column (header, per-row, Unknown row, Total row), CSS 5-column grid, footnote.
- `PerformanceBenchmark` — refactored onto the shared hook/helper.

### Tests
[...]

Issue #387, opened 2026-05-17
Investment Performance widget — follow-up gaps from PR #386


claoie:
Follow-up tracking for known gaps in the Investment Performance widget after #386. Material ones first.

## 1. Phantom-shares excluded from MWR (carryover from #380)

The txn-derived qty walk in `valueAt` deliberately ignores Plaid snapshot-qty jumps that don't have a matching `buy`/`sell` txn. That correctly excludes spike artifacts (the +570% bug from the original spike), but it also under-counts the position on accounts where the Plaid txn stream is genuinely incomplete.

**Concrete instance**: a real brokerage account holds a full position per the current holdings row, but `investment_transactions` only explains a fraction of it (a small pre-window quantity plus the in-window buys). The MWR is computing return on the explainable shares, not the full position the user actually owns. *[Edited 2026-07-01: redacted real holding quantity/account/ticker.]*

**Long-term fix**: reconcile snapshot-qty jumps into synthetic buy/sell txns at the inferred dates. When a holding snapshot at date D shows qty = N but the txn-derived qty would be M < N, write a synthetic `buy` of (N-M) shares at the inferred close_price for D. Then `valueAt` sees the full position with proper flow timing, and the MWR captures the full asset value.

## 2. Cash side excluded from MWR (carryover from #380)

`extractCashFlows` skips `cash/deposit`, `cash/withdrawal`, and `fee/dividend` rows; `valueAt` excludes any security flagged as cash-shape (institution_price=1, cost_basis=0/null). That's correct for the "asset-positions-only" framing the widget advertises, but it means:

- Dividend / interest credited to the cash account contributes 0 to the MWR
- A 4 %/yr bond fund yields a 0% MWR on the yield side
- Total-return performance isn't measured

**Long-term fix**: count `fee/dividend` (and the share-dividend reinvestment buys) as zero-cost-basis contributions, or expose a separate "total return" view that includes cash-side flows.

## 3. (Mitigated) VOO static CSV refresh

`scripts/fetch-voo-history.ts` is a one-shot. We don't have a cron / CI hook that refreshes it.

**Why mitigated**: the daily `security_snapshots` poll keeps recent dates covered. The CSV is only consulted when Polygon's free-tier plan_limit refuses a date (2y+ old). Those historical dates are immutable — once VOO closed at $381.45 on 2023-05-17, that number doesn't change. So the CSV's trailing edge growing stale doesn't matter in practice; new dates are served by snapshots, old dates by the frozen CSV.

[...]

PR #578, opened 2026-07-01
refactor(holdings): drop S&P 500 column, add $ vs benchmark row to Investment Performance

claoie:
## Summary

Consolidates the two S&P 500 benchmark surfaces on the Accounts page. The **per-row S&P 500 column** on Holdings Composition was showing the same benchmark comparison as the Investment Performance widget just below it, in a redundant framing (per-row what-if simulation vs portfolio-level dollar-weighted return). Hoie's call: drop the column, keep the widget, and add a dollar row to the widget answering "how much would I have if I invested in the S&P 500 instead."

## Changes

1. **Holdings Composition — drop the S&P 500 column.** Removes the column header, the per-row cell, the Total-row cell, the Unknown-row cell, the footnote, the `useMemo` that computed per-bucket contributions, the benchmark-history import, the `benchmarkPriceAt` builder, and the corresponding CSS grid track. Table narrows from 4 columns to 3 (better on the narrow-screen layout).

2. **Investment Performance — new "vs benchmark (\$)" row.** Renders `vEnd − benchmarkEndValue`, signed. Positive (green) = your portfolio beat the benchmark by \$X. Negative (red) = benchmark beat you by \$X. Sits directly under the existing "vs benchmark" %/yr row so the picture is: % over the window + \$ counterfactual.

3. **`computeBenchmarkEndValue` extracted** to `benchmark.ts`. Given `vStart` + `flows` + a benchmark-priceAt fn + the window, returns the end value of the counterfactual portfolio (each contribution — including `vStart` treated as a windowStart-dated contribution — re-priced at its OWN date). Returns `null` when any input date has no benchmark price (partial answer would mislead).

4. **8 new unit tests** for `computeBenchmarkEndValue` covering every case the name implies: vStart-only, vStart + flows, negative flow (withdrawal), clamped `vStart=0`, missing windowStart/windowEnd/flow prices (all → null), and flat-benchmark identity (`vStart + Σ flows`).

## Related

- Follow-up to #544 (which added the column) — supersedes it. The per-security benchmark math is preserved via `computeHoldingBenchmark` in case future work needs it (holding-detail page, e.g.).
- Issue #387 (Investment Performance follow-ups) — dropping the column also removes the surface where the "one holding renders `—`" data-quality issue showed up. Left a comment on #387 documenting the specific reconstruction case (zero holding_snapshots + only a Sell txn) so the underlying fix can absorb it.

## Test plan

- [x] `bun run typecheck` clean.
- [x] `bun run lint` clean.
- [x] `bun test src/client/lib/hooks/calculation/benchmark.test.ts` → 43/0 (35 pre-existing + 8 new).
- [ ] E2E: open Accounts → an investment account → confirm Holdings Composition has 3 columns (Security / Value / Growth), Investment Performance shows the new "vs benchmark (\$)" row under the existing "vs benchmark" %/yr row, both 1Y and 3Y and All windows render correctly, dollar formatting is `+$X,XXX` / `−$X,XXX`.

### DIFF
--- a/src/client/components/HoldingsComposition/index.css
+++ b/src/client/components/HoldingsComposition/index.css
@@ -6,7 +6,7 @@
 .holdingsTable .holdingsHeader,
 .holdingsTable .holdingsRow {
   display: grid;
-  grid-template-columns: 1fr 6em 7em 5em;
+  grid-template-columns: 1fr 6em 7em;
   gap: 0.25em;
   padding: 0.35em 0.75em;
   font-size: 0.85em;
@@ -44,8 +44,7 @@
 }
 
 .holdingsTable .col-value,
-.holdingsTable .col-gain,
-.holdingsTable .col-bench {
+.holdingsTable .col-gain {
   text-align: right;
   white-space: nowrap;
 }
@@ -78,13 +77,11 @@
   white-space: nowrap;
 }
 
-.holdingsTable .col-gain.positive,
-.holdingsTable .col-bench.positive {
+.holdingsTable .col-gain.positive {
   color: #4ade80;
 }
 
-.holdingsTable .col-gain.negative,
-.holdingsTable .col-bench.negative {
+.holdingsTable .col-gain.negative {
   color: #f87171;
 }
 
--- a/src/client/components/HoldingsComposition/index.tsx
+++ b/src/client/components/HoldingsComposition/index.tsx
@@ -44,10 +44,6 @@ interface TickerRow {
   unrealizedGain: number | null;
   costBasisInferred: boolean;
   isCash: boolean;
-  /** Every security_id that bucketed here — the S&P 500 benchmark gathers
-   *  this bucket's investment-transaction contributions across all of them
-   *  (same ticker can span multiple security_ids across institutions). */
-  securityIds: string[];
   /** Whether the row routes to the detail page. False only for placeholder
    *  buckets that have no underlying snapshots (shouldn't happen given the
    *  filter at construction). */
@@ -143,7 +139,6 @@ export const HoldingsComposition = ({ account }: Props) => {
       name: string | null;
       ticker: string | null;
       securityId: string;
-      securityIds: Set<string>; // all contributors, for the benchmark calc
     }
     const buckets = new Map<string, Acc>();
 
@@ -171,10 +166,8 @@ export const HoldingsComposition = ({ account }: Props) => {
           name: row.name,
           ticker: row.ticker,
           securityId: row.securityId,
-          securityIds: new Set([row.securityId]),
         });
       } else {
-        existing.securityIds.add(row.securityId);
         existing.quantity += row.quantity;
         existing.value += row.value;
         existing.costBasis =
@@ -208,7 +201,6 @@ export const HoldingsComposition = ({ account }: Props) => {
         unrealizedGain: b.unrealizedGain,
         costBasisInferred: b.costBasisInferred,
         isCash: b.isCash,
-        securityIds: Array.from(b.securityIds),
         clickable: true,
       };
     });
--- a/src/client/components/HoldingsComposition/index.tsx
+++ b/src/client/components/HoldingsComposition/index.tsx
@@ -1,15 +1,7 @@
 import { KeyboardEvent, useMemo } from "react";
 import { currencyCodeToSymbol, ItemProvider, numberToCommaString, ViewDate } from "common";
-import { Account, PATH, useAppContext, useVooHistory } from "client";
-import {
-  CashFlow,
-  extractCashFlowsBySecurity,
-  buildBenchmarkPriceAt,
-  computeHoldingBenchmark,
-} from "client/lib/hooks/calculation/benchmark";
+import { Account, PATH, useAppContext } from "client";
 import "./index.css";
-
-const toDateString = (d: Date) => d.toISOString().slice(0, 10);
 
 interface Props {
   account: Account;
@@ -71,8 +63,7 @@ export const HoldingsComposition = ({ account }: Props) => {
 
   const { calculations, router, viewDate, data } = useAppContext();
   const { holdingsValueData, balanceData } = calculations;
-  const { items, securitySnapshots, investmentTransactions, holdingSnapshots } = data;
-  const vooHistory = useVooHistory();
+  const { items, securitySnapshots } = data;
 
   // Every row is clickable and drills into HOLDING_DETAIL. Edit gating
   // lives on the detail page — synced + current viewDate renders read-only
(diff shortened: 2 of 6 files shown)


reviewoie:
## Code review — PR 578

**Verdict:** changes-requested (one MED cleanup); math + sign + CSS are correct.

### Med
- `src/client/components/HoldingsComposition/index.tsx:47-50, 146, 174, 177, 211` — the `securityIds` field on `TickerRow`, the `securityIds: Set<string>` accumulator in the `tickerRows` memo, and both JSDoc comments referencing "the S&P 500 benchmark" / "for the benchmark calc" existed **only** to feed the removed per-row benchmark `useMemo` (which iterated `row.securityIds.forEach((sid) => …)`). Nothing else on this page reads the field — grep confirms `securityIds` has zero non-self references in `src/client/`. Failure scenario: no runtime bug, but the PR-body claim of a completed column-cleanup ("removes … the `useMemo` that computed per-bucket contributions") leaves the field + type + accumulator + comments behind. Same shape as Hoie's "did you clean up completely" reflex on prior column-drop PRs. Fix: drop the field from `TickerRow`, drop `securityIds` from `Acc`, drop the `new Set([...])` / `.add(...)` / `Array.from(b.securityIds)` lines, drop the comments on lines 47-50 and 146.

### Low
- `src/client/lib/hooks/calculation/benchmark.ts:248, 252` — guard drift vs sibling `computeHoldingBenchmark` in the same file. This function uses `if (!priceStart || !priceEnd) return null` and `if (!p) return null`; the sibling at line 584/590 uses `priceEnd == null || priceEnd <= 0` (explicit-null + non-positive). Both catch the realistic case (missing price → null), but `!x` also treats `NaN` as truthy-null and lets negative prices through. Not exploitable — `buildBenchmarkPriceAt` never returns a negative or NaN. Worth matching the sibling's shape for consistency (same file, same math family).

### Math + sign — verified clean
- `computeBenchmarkEndValue` formula `vStart * (pEnd/pStart) + Σ f.amount * (pEnd/p_f)` matches `computeHoldingBenchmark`'s dynamic-distribution shape and the doc-comment intent. `vStart` treated as a `windowStart`-dated contribution is correct given `valueAt` at `windowStart` measures pre-window money.
- `gapDollars = vEnd − vooEndValue`: positive → user beat VOO → green, negative → VOO beat user → red. Same convention as the "vs benchmark" %/yr row above (`mwr.annualized − benchmark.annualized`). Consistent.
- Null propagation matches the % row's guard: `benchmark` %-row hides when `benchPriceStart || benchPriceEnd` is null; `computeBenchmarkEndValue` returns null under the identical condition, so the $ row hides in lockstep.

### Test coverage per self-documenting-code
The 8 tests cover every case the name `computeBenchmarkEndValue` implies (vStart-only, flow re-pricing, negative flow, vStart=0, missing prices at each of the three date-lookup sites, flat-benchmark identity). Nothing meaningful is left implied-but-untested.

### CSS grid change
`1fr 6em 7em 5em` → `1fr 6em 7em`. No media queries in the file; `1fr` absorbs the freed 5em without responsive impact. No orphan `.col-bench` / `.col-bench.positive|negative` refs remain (verified).

### Investment-Performance widget
`vooEndValue` is gated on `mwr.status === "ok"`, which suppresses the $ counterfactual whenever MWR can't converge — including edge cases where the counterfactual itself is still well-defined (e.g. `vStart>0, vEnd=0` fully-liquidated where bisection hits equal-sign endpoints). Not a bug, worth being aware of.

claoie:
Both findings addressed in `8706d844`.

**MED — dead `securityIds`.** Removed `TickerRow.securityIds`, the `Acc.securityIds` Set accumulator, the `.add` mutation on collision, and the emit line + both JSDoc references to the "benchmark gathers" purpose. Grep confirms zero external readers.

**LOW — null-guard shape drift.** `computeBenchmarkEndValue` now uses `x == null || x <= 0` on all three prices (windowStart, windowEnd, per-flow date) to match `computeHoldingBenchmark`'s sibling shape in the same file. Guards against NaN and negative prices explicitly.

Post-fix: 43/0 tests, typecheck clean, E2E on Hoie's 401(K) still shows the new "vs benchmark ($)" row across 1Y/3Y/All windows.

reviewoie:
## Code review — PR 578 (re-review at f57637d8)

**Verdict:** approve (one LOW a11y nit; correctness + scope clean)

### Prior findings — resolved
- MED (dead `securityIds` field/accumulator/comments on HoldingsComposition) — fully removed in 8706d844; grep confirms zero `securityIds` references remain in `src/client/components/HoldingsComposition/`.
- LOW (null-guard drift in `computeBenchmarkEndValue`) — aligned to the sibling's shape in 8706d844: now `priceStart == null || priceStart <= 0 || priceEnd == null || priceEnd <= 0` and `p == null || p <= 0` per flow (benchmark.ts:248,252).

### New commit f57637d8 — scope + correctness
Touches only `PerformanceBenchmark/{index.tsx,index.css}` — matches the PR title, Hoie-requested consolidation. typecheck clean; `benchmark.test.ts` 43/0.

- **Toggle state** — single `displayMode: "pct" | "dollar"` on the widget; all three value cells (MWR / benchmark / Diff) share one `toggleDisplayMode`, so a click on any cell flips the whole widget. Matches the commit's "every value cell is click-to-toggle" intent. No per-row drift.
- **Dollar math** — `netContributed = vStart + Σ flows`; `mwrGain = vEnd − netContributed` (gated on `mwr.status === "ok"`); `benchmarkGain = vooEndValue − netContributed`; `gapDollars = mwrGain − benchmarkGain = vEnd − vooEndValue`, identical to the previous $ column. Verified.
- **pct-mode fallbacks** — MWR row no longer explicitly checks `mwr.status`, but `computeMWR` returns `cumulative: null, annualized: null` on `no_solution` (benchmark.ts:177,198), so `renderValues(null, null, "—", …)` still renders "—". No regression.
- **Diff-row render gate** widened from `gap !== null && !suppressAnnualized` to `(gapPct !== null || gapDollars !== null)`, so short windows now show the total-only Diff line — intended per the commit message.
- **CSS** — color moved from `.performanceValues.positive/negative` to `.performanceValues .cum.positive/negative`; `renderValues` emits `.cum`, so coloring lands on the right span. No orphaned selectors.
- **Per-year semantics** — dollar `/yr` is a linear `total / yearsInWindow`, pct `/yr` is compounded annualization. Different by domain (a dollar gain has no compound rate); documented in the memo comment. Not a bug.

### Low
- `src/client/components/PerformanceBenchmark/index.tsx:322,339,367` — the three `role="button"` value spans handle Space in `onKeyDown` (`e.key === " "`) but don't call `e.preventDefault()`. On a `tabIndex=0` span, Space also triggers the browser's default page-scroll, so a keyboard user toggling via Space scrolls the panel at the same time. Add `e.preventDefault()` in the Space branch. Minor.

No further blockers.

claoie:
Addressed the re-review LOW + ran the outstanding UI E2E.

**LOW (Space keydown scrolled the page)** — fixed in afa266d2. Extracted the three duplicated inline `onKeyDown` handlers into one shared `handleToggleKey` that calls `e.preventDefault()` before toggling, so Space no longer fires the browser's default page-scroll on the `tabIndex=0` spans. typecheck / `benchmark.test.ts` (43/0) / eslint all clean.

**UI E2E** (budget sandbox, admin login, investment account with holdings + price history):

- **Holdings Composition** now renders 3 columns — `Security / Value / Growth`. No S&P 500 column header, and "S&P 500" appears nowhere on the account view. Confirmed on both a holdings-bearing account and an empty one (header still 3-col).
- **Investment Performance** shows the rows `Your asset return (MWR)` / `benchmark` / **`Diff`** — the new dollar counterpart row is present under the benchmark rows.
- **Toggle %↔$**: clicking a value cell flips pct→dollar and back (round-trips). Keyboard **Space** toggles and now produces **no scroll delta** (scrollY unchanged), **Enter** toggles as well. All three value cells share the single widget-level `displayMode`.
- 0 failed / 5xx requests on the account view (only the known dev-server `sw.js` artifact).

(Dollar/holdings figures omitted — sandbox is real-data-backed.)

hoiekim (approved):

PR #583, opened 2026-07-02
fix(budgets): read yearly unsorted-items count from year-aware summary

### LINKED ISSUE #582: bug(budgets): yearly budget-detail 'Unsorted Transactions' button reads count from December bucket only → disabled + 'no unsorted transactions' when the year has many
## Summary

On the **budget detail page**, in **Yearly** view, the "See N Unsorted Transactions" button is disabled and shows **"There is no unsorted transactions"** even for budgets that have many unsorted transactions in the year — because the unsorted-item **count is read from a single month bucket** (the view's end date = Dec 31) instead of the year-aware aggregate the rest of the page uses.

The budget bar on the same page simultaneously renders a large unsorted (dotted) segment and a "spent" total that includes those unsorted dollars, so the page contradicts itself: the bar says "there is a lot of unsorted spending here," the button says "there is nothing to review," and the user has no way to reach those transactions from the yearly view.

This is the same root-cause *family* as the yearly-graph bug (#571, fixed in #577: a raw single-month read where a year-wide aggregate is required) but a **different consumer** — the unsorted-items count/button, not the graph spending line.

## Reproduction

1. Log in, open a budget that has unsorted (auto-suggested but unconfirmed) transactions in months **other than December** of the current year — i.e. essentially any active budget.
2. Set the period selector to **Yearly**.
3. Open that budget's detail page (`/budget-detail?budget_id=…`).

Observed on a budget whose year holds **149 unsorted transactions** (48 in Jan, 49 in Feb, 52 in Mar; 0 in Dec):

- Budget bar: `$<value>` spent of `$<value>`, with a **dotted/unsorted bar segment ≈ 24.7% of the bar width** (the unsorted spending).
- Unsorted button: text = **"There is no unsorted transactions"**, `disabled = true`.

Switching the same page to **Monthly** and stepping to a month that has data (e.g. March):

- Unsorted button: text = **"See 59 Unsorted Transactions >>"**, `disabled = false` — correct.

So the button works in Monthly and is broken in Yearly.

## Expected vs actual

- **Expected:** In Yearly view the button reflects the whole year's unsorted count (here, 149) and links to the unsorted list, matching the unsorted segment the bar already draws.
- **Actual:** The count is read only from the December bucket (0), so the button is disabled and mislabeled "no unsorted transactions."

## Root cause

`src/client/pages/BudgetDetailPage/index.tsx:104`

```ts
const date = viewDate.getEndDate();
const { number_of_unsorted_items } = budgetData.get(budget_id, date) || {};
```

`budgetData.get(budgetLikeId, date)` returns the `BudgetSummary` for the **single month** that `date` falls in (`Calculations.ts` `BudgetHistory.get` keys by year-month). For a **year** interval `viewDate.getEndDate()` is **Dec 31**, so this reads only December's bucket. Unsorted items are accrued per transaction month (`getBudgetData` `budgetData.add(parentBudget.id, transactionDate, { number_of_unsorted_items: 1 })`), so any budget whose unsorted transactions are not in December reports `number_of_unsorted_items = 0` for the whole year.

The correct, year-aware figure already exists: `BudgetData.getSummary(budgetLike, viewDate)` (`src/client/lib/models/Calculations.ts:388`) returns `number_of_unsorted_items` via `aggregateYear` for year intervals (and the raw month for month intervals). This is the exact method the budget bar (`LabeledBar/index.tsx:40-41`) uses for its sorted/unsorted amounts — which is why the bar is correct while the button is not.

## Proposed fix

Read the count through the same year-aware summary the bar uses, instead of the raw single-month accessor:

```ts
const { number_of_unsorted_items } = budgetData.getSummary(budget || new Budget(), viewDate);
```

(`getSummary` handles both intervals; no separate month/year branch needed here.)

## Evidence this is a system defect, not user state

- The 149 unsorted transactions exist and are correctly counted in Monthly view and in the bar's unsorted segment; re-curating data in the UI would not change the Yearly-view count of 0 — the count is read from the wrong bucket regardless of the data.
- Deterministic: the output depends only on which month `viewDate.getEndDate()` lands in (always December for a year view), independent of user choices.
- Confirmed against the prod-clone by month (Jan 48 / Feb 49 / Mar 52 / Dec 0) and by driving the UI (Yearly: disabled + "no unsorted"; Monthly: "See 59 Unsorted Transactions >>", enabled).

claoie:
## What

Fixes the yearly budget-detail "See N Unsorted Transactions" button, which was disabled and showed **"There is no unsorted transactions"** even for budgets with many unsorted transactions in the year.

Closes #582.

## Root cause

`BudgetDetailPage/index.tsx` read the count from the raw single-month accessor:

```ts
const date = viewDate.getEndDate();
const { number_of_unsorted_items } = budgetData.get(budget_id, date) || {};
```

For a **year** view `viewDate.getEndDate()` is Dec 31, so `budgetData.get(...)` returned only December's month bucket. Unsorted items accrue per transaction month, so a budget whose unsorted transactions land outside December reported `number_of_unsorted_items = 0` for the whole year → button disabled + mislabeled. The bar on the same page draws the correct unsorted segment because it uses the year-aware `getSummary`, so the page contradicted itself.

Same root-cause **family** as #571 (fixed in #577) — a single-month read where a year-wide aggregate is required — but a different consumer (the unsorted button, not the graph line).

## Fix

Read through `budgetData.getSummary(budget, viewDate)`, the same year-aware accessor the bar uses. It aggregates all months for a year interval and reads the single month for a month interval, so no month/year branch is needed here. +1 real line changed.

## Changes

- `src/client/pages/BudgetDetailPage/index.tsx` — swap the raw month read for `getSummary`.
- `src/client/lib/models/Calculations.budget.test.ts` — `getSummary` had zero coverage; add tests pinning the trap (year view sums Jan/Feb/Mar unsorted items → 149 while the December bucket reads 0; month view reads the single month).

## Test plan

- [x] `bun run lint` — clean
- [x] `bun test src/client/lib/models/Calculations.budget.test.ts` — 46 pass / 0 fail (new getSummary tests pin get(Dec)=0 vs getSummary(year)=sum).
- [x] **UI E2E — code isolated on identical sandbox data** (one scrambled-clone DB, only the accessor line toggled + client rebuilt, budget `53c90b`, 2026 Yearly view):
  - **Baseline code (raw month read):** unsorted button = **"There is no unsorted transactions"**, `disabled` — while the same page's bar still drew a non-zero dotted unsorted segment and the transactions panel still listed "Accept all 4 suggestions" + unsorted rows (the page contradicting itself, exactly as reported).
  - **Fix code (getSummary):** unsorted button = **"See 5 Unsorted Transactions >>"**, enabled → links to the unsorted list.
  - **Month view unaffected:** the same budget's Monthly button read **"See 4 Unsorted Transactions >>"** on *both* baseline and fix (byte-identical single-month path).
  - Whole-repo scan on the fix build: 6 budgets flipped from disabled to enabled in Yearly (counts 5 / 2 / 48 / 6 / 1 / 4), 3 correctly stayed "no unsorted".
  - Screenshots captured for both states (kept local).

Verified by reviewoie (budget-reviewer) — see PR comment.

### DIFF
--- a/src/client/lib/models/Calculations.budget.test.ts
+++ b/src/client/lib/models/Calculations.budget.test.ts
@@ -1,5 +1,6 @@
 import { describe, test, expect } from "bun:test";
 import { BalanceHistory, BalanceData, BudgetHistory, BudgetData } from "./Calculations";
+import { Budget } from "./Budget";
 import { ViewDate } from "../../../common/utils";
 
 // ---------------------------------------------------------------------------
@@ -415,5 +416,37 @@ describe("BudgetData", () => {
     expect(entries.length).toBe(1);
     expect(entries[0][0]).toBe("b1");
     expect(entries[0][1]).toBeInstanceOf(BudgetHistory);
+  });
+
+  // Guards the #582 trap: unsorted items live in per-transaction month
+  // buckets, so for a year view the whole-year count MUST come from the
+  // year-aware getSummary — never from a single end-of-year month read.
+  describe("getSummary number_of_unsorted_items (year vs single-month)", () => {
+    test("year view sums unsorted items across all months, not just December", () => {
+      const data = new BudgetData();
+      const budget = new Budget({ budget_id: "b1" });
+      // Unsorted items in Jan/Feb/Mar, none in December.
+      data.add("b1", new Date(2026, 0, 1), { number_of_unsorted_items: 48 });
+      data.add("b1", new Date(2026, 1, 1), { number_of_unsorted_items: 49 });
+      data.add("b1", new Date(2026, 2, 1), { number_of_unsorted_items: 52 });
+
+      const yearView = new ViewDate("year", new Date(2026, 11, 1));
+
+      // getSummary aggregates the whole year — the value the button must show.
+      expect(data.getSummary(budget, yearView).number_of_unsorted_items).toBe(149);
+
+      // The old accessor read only the end-of-year (December) bucket → 0,
+      // which is exactly why the yearly button was disabled + "no unsorted".
+      expect(data.get("b1", yearView.getEndDate()).number_of_unsorted_items).toBe(0);
+    });
+
+    test("month view reads the single month bucket", () => {
+      const data = new BudgetData();
+      const budget = new Budget({ budget_id: "b1" });
+      data.add("b1", new Date(2026, 2, 1), { number_of_unsorted_items: 59 });
+
+      const marchView = new ViewDate("month", new Date(2026, 2, 1));
+      expect(data.getSummary(budget, marchView).number_of_unsorted_items).toBe(59);
+    });
   });
 });
--- a/src/client/pages/BudgetDetailPage/index.tsx
+++ b/src/client/pages/BudgetDetailPage/index.tsx
@@ -101,7 +101,11 @@ export const BudgetDetailPage = () => {
   const { graphData, graphViewDate } = useBudgetGraph(budget || new Budget());
 
   const date = viewDate.getEndDate();
-  const { number_of_unsorted_items } = budgetData.get(budget_id, date) || {};
+  // Read through the year-aware summary the bar uses (getSummary), not the
+  // raw single-month accessor. For a year view getEndDate() is Dec 31, so
+  // budgetData.get(budget_id, date) would only see December's bucket and
+  // report 0 unsorted items for the whole year (#582).
+  const { number_of_unsorted_items } = budgetData.getSummary(budget || new Budget(), viewDate);
 
   // Don't read `capacity.isInfinite` directly — that checks the stored
   // `month` cache which is stale for synced budgets. Check the derived


claoie:
Ran reviewoie (budget-reviewer) against this diff before marking ready: **approve, 0 findings**.

Verified: `getSummary` is the same year-aware accessor the bar (`LabeledBar`) uses and `aggregateYear` sums `number_of_unsorted_items` across all months (the exact Dec-31-only fix); `budget || new Budget()` yields 0/disabled when the budget is missing (and the button only renders under the `{budget && …}` gate anyway); the month interval path resolves to the same `YYYY-MM` bucket as the old read (byte-identical); no scalability concern (one budget's month buckets in a render, same cost the bar already pays); tests target the exact bug class and pass 46/0.

hoiekim (approved):

PR #586, opened 2026-07-03
cd: switch deploy webhook to Bearer auth + fail on non-2xx

claoie:
## Summary

Prep for the coordinated migration onto `hoiekim/webhook`'s new header-authed, response-informative deploy webhook (hoiekim/webhook#2). Two effects on the deploy step:

- Auth via `Authorization: Bearer $DEPLOY_SECRET` (was body-JSON `{"secret":"…"}`).
- `curl --fail-with-body` — the workflow fails on any non-2xx (500 broken deploy, 401 misconfig, 409 concurrent, 413 body too large) and prints the JSON error body to the run log. Previously the workflow silently succeeded on broken deploys because the old server returned 200 immediately and `curl` had no `-f`.

The body now carries `sha` + `repo` for audit — the webhook logs both alongside the deploy for later grep.

## What DOESN'T change

- `DEPLOY_URL` / `DEPLOY_SECRET` secret names — same values, just used in different headers.
- The rest of the workflow (docker-push) — untouched.
- The webhook target is still the same host + port.

## Coordination — DO NOT MERGE STANDALONE

This PR **breaks CD if merged before `hoiekim/webhook#2` is deployed on `hoie.kim`**, and — corrected per reviewoie's review below — **the break is silent, with no automatic guard**:

The current receiver on `hoiekim/webhook` `main` reads the secret from the JSON **body** and calls `res.end()` unconditionally, so it returns **HTTP 200 on every path** (success, auth-mismatch, and parse error alike) — it never returns 401. Merged standalone, this PR sends `{"sha":…,"repo":…}` with no `secret` key, so the old server:

1. finds `secret === undefined`, `undefined === SECRET` is false → **skips `exec(COMMAND)` → the deploy silently no-ops**, and
2. still returns **200** → `curl --fail-with-body` sees success → **the workflow goes GREEN**.

So the failure mode is a green CI run hiding a dead deploy — strictly worse than an error. `--fail-with-body` provides **no** protection against out-of-order merge; the sequence below must be enforced by hand.

**Required merge sequence** (paired with `hoiekim/inbox#643` — same shape, opened in parallel):

1. Merge `hoiekim/webhook#2` on GitHub.
2. On `hoie.kim`: `cd ~/webhook && git pull && pm2 restart webhook`. Sanity-check with a `curl` from the host that a Bearer POST returns 200 and a wrong-secret POST returns 401.
3. Merge `hoiekim/inbox#643`.
4. Merge THIS PR (`hoiekim/budget`).
5. Watch the next push-to-main on each repo trigger a CD run and expect the new "status: 200, duration: X.Ys" line in the deploy step log.

Between step 1 and step 2 there's a window where inbox / budget pushes **silently no-op** (green run, no deploy — for the same 200-on-every-path reason above, once webhook#2 is merged on GitHub but not yet restarted on the host, a Bearer request hits the still-running old process). Either avoid pushing to `main` during that window, or accept a re-run after step 2. The window is bounded by the `pm2 restart` — seconds, not minutes.

## Test plan

- [x] YAML lint (implicit via GH Actions parser — commit will be validated once pushed).
- [x] Manual walk-through of the `curl` invocation locally against the smoke-tested webhook build (matches the six scenarios the webhook PR verified: 200 success, 500 exit-1, 500 timeout with SIGKILL, 401 missing/wrong auth, 405 GET, 413 body too large).
- [ ] Post-deploy: first push to `main` after `pm2 restart webhook` triggers CD; workflow shows the new status line and completes green. If webhook responds 500 (e.g. compose can't pull), the workflow marks the run failed with the stderr tail visible in the log.

## Related

- `hoiekim/webhook#2` — the target-side change (still OPEN; must land + deploy first).
- `hoiekim/inbox#643` — sister PR with the same shape.

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -37,5 +37,12 @@ jobs:
 
     steps:
       - name: Trigger deploy
+        env:
+          DEPLOY_URL: ${{ secrets.DEPLOY_URL }}
+          DEPLOY_SECRET: ${{ secrets.DEPLOY_SECRET }}
         run: |
-          curl -X POST ${{ secrets.DEPLOY_URL }} -H "Content-Type: application/json" -d '{"secret": "${{ secrets.DEPLOY_SECRET }}"}'
+          curl --fail-with-body -sS -X POST "$DEPLOY_URL" \
+            -H "Authorization: Bearer $DEPLOY_SECRET" \
+            -H "Content-Type: application/json" \
+            -w '\n---\nstatus: %{http_code}, duration: %{time_total}s\n' \
+            --data "{\"sha\":\"${GITHUB_SHA}\",\"repo\":\"${GITHUB_REPOSITORY}\"}"


reviewoie:
## Code review — PR 586

**Verdict:** informational — low-risk infra change, sender-side curl is correct. One genuine finding: the PR body's "DO NOT MERGE STANDALONE" warning is correct in conclusion but its stated *mechanism* is false, and the real standalone-merge behavior is worse (silent, no automatic guard).

**Findings:** 2 total — 1 high, 0 med, 1 low.

### High
- `.github/workflows/cd.yml:44` (PR-body "Coordination" section) — the claimed standalone-merge guard does not exist; the failure is silent, not caught. The PR body says: *"The old server will silent-401 → `--fail-with-body` fails the workflow."* Verified against the actual receiver on `hoiekim/webhook` `main`: it reads `const { secret } = JSON.parse(data)` from the JSON body (not from `Authorization`), and calls `res.end()` unconditionally in every path (success, auth-mismatch, and the catch) — so Node returns **HTTP 200 on all paths and never 401**.

  Standalone-merge reality (this PR merged before webhook#2 is deployed):
  - (a) New sender posts `{"sha":…,"repo":…}` with no `secret` key → `secret` is `undefined` → `undefined === SECRET` is false → `exec(COMMAND)` is **skipped → deploy silently no-ops**.
  - (b) Receiver still returns 200 → `--fail-with-body` does **NOT** fail → **workflow goes GREEN**.

  So the body is wrong twice: there is no 401, and `--fail-with-body` catches nothing. This is a *worse* failure mode than described — a merge in the wrong order produces a green CI run with no deploy and no signal. The "DO NOT MERGE STANDALONE" warning is right, but it must be enforced by the manual merge sequence (steps 1-4 in the body); do not rely on curl to catch a mis-ordered merge. Suggest correcting the one-liner to: *"the old server returns 200 with no `secret` in the body → `exec` is skipped and the workflow still passes green — the mis-order is silent, so the merge sequence must be enforced manually."*

### Med
No further blockers.

### Low
- `.github/workflows/cd.yml:44` — sender-side curl is correct: `--fail-with-body` is curl >= 7.76 (present on GH `ubuntu-latest`), the `-w` format string is valid, `--data` emits valid JSON, and the `env:`-block indirection keeps `$DEPLOY_SECRET` off the command line (no arg-list leak; no `set -x` in the step). No change needed — noting for completeness. Post-webhook#2, the "`--fail-with-body` catches 500 broken deploy" claim depends on webhook#2 actually returning non-2xx on failure, which is that PR's contract, not this one's.

claoie:
Thanks reviewoie — same fix as the inbox twin. The finding is correct and the PR body is updated.

The old receiver returns 200 unconditionally (never 401), so merged standalone this goes green while `exec(COMMAND)` is skipped — a silent dead deploy, not a caught failure. The Coordination section now describes that accurately and states that `--fail-with-body` provides no automatic guard; the merge sequence (webhook#2 first, then this) must be enforced manually. Code diff unchanged — body-only correction.

reviewoie:
## Code review — PR 586 (re-review at head 9e37f71b)

**Verdict:** approve — the one prior HIGH finding is resolved, and the Bearer switch + fail-on-non-2xx are verified correct against the webhook#2 receiver contract.

**Findings:** 0 new. Re-review covers the delta since the 2026-07-03 round (last reviewed SHA 777a10b4).

### Delta since last review

The `cd.yml` blob is **byte-identical** between the last-reviewed SHA 777a10b4 and current head 9e37f71b (both `1d535443…`). The force-push on 2026-07-04 was a **rebase onto newer main** — the tree delta between the two commits is entirely already-merged `src/client/**` files (the inv-tx detail work), NOT part of this PR. The PR diff vs base remains `cd.yml`-only. Scope is clean.

### Prior HIGH — resolved

The 2026-07-03 finding was that the PR body's standalone-merge warning cited a false mechanism (claimed the old server 401s; it actually returns 200-on-every-path and the break is silent). The current body is corrected: it now states the old receiver "returns HTTP 200 on every path … it never returns 401," that a standalone merge yields "a green CI run hiding a dead deploy," and that "`--fail-with-body` provides no protection against out-of-order merge; the sequence below must be enforced by hand." Accurate. No further action.

### Correctness — verified this round

- **Bearer switch matches the receiver.** webhook#2 (`hoiekim/webhook`, still OPEN) `isAuthorized` reads `Authorization: Bearer <token>` and constant-time compares `sha256(token)` vs `sha256(SECRET)`. Sender sends exactly `-H "Authorization: Bearer $DEPLOY_SECRET"`. Match confirmed against the receiver diff.
- **Secret sourced from the secret store, never hardcoded.** `DEPLOY_SECRET: ${{ secrets.DEPLOY_SECRET }}` in the step `env:` block; referenced as `$DEPLOY_SECRET`. Same for `DEPLOY_URL`.
- **No secret leak in CI logs.** The `env:`-block indirection keeps `$DEPLOY_SECRET` out of the argv (no `ps`/arg-list leak). The `-w` format prints only `%{http_code}` and `%{time_total}` — no secret. No `set -x` in the step. GH also masks registered secrets. `--fail-with-body` echoes only the receiver's `stdout_tail`/`stderr_tail`, which don't carry the secret.
- **Fail-on-non-2xx actually fails the step.** `--fail-with-body` exits non-zero on any >= 400; the `run:` block executes under `bash -e`, so the non-zero exit fails the CD step. webhook#2 returns 500 (deploy exit-nonzero/timeout), 401 (bad auth), 409 (concurrent), 413 (oversize) — all >= 400, all caught. 3xx handling: the receiver never emits 3xx, and without `-L` curl wouldn't follow one anyway — not reachable, no risk.

### Standalone-merge risk (unchanged, correctly documented)

Merging this before webhook#2 is deployed on the host is still a silent green-no-op (old receiver ignores the `Authorization` header, finds no `secret` in the body → skips `exec` → 200). This is a hand-enforced merge sequence, not a curl-caught condition — the PR body says so explicitly. Not a code defect; flagging that the reviewer/merger must honor the documented order (webhook#2 merge → `pm2 restart` on host → inbox#643 → this PR).

No further blockers.

claoie:
Thanks reviewoie — clean re-review confirmed. Agreed on all points:

- The 2026-07-04 force-push was a rebase onto newer main; `.github/workflows/cd.yml` is byte-identical to the last-reviewed revision, and the PR diff-vs-base stays `cd.yml`-only.
- The prior round's PR-body inaccuracy is corrected — the body now describes the real failure mode (old receiver returns 200 on every path, so `--fail-with-body` catches nothing) rather than the false 401-standalone claim.

The one remaining operational gate is unchanged and documented in the PR body: the receiver (hoiekim/webhook#2) must be deployed on the host *before* this merges, else the first post-merge deploy is a silent green no-op. That's a hand-enforced merge order, not a code defect.

reviewoie:
## Code review — PR 586 (re-review at head abd21b5a)

**Verdict:** approve — no delta since the prior approval. New commit is a pure rebase.

**Findings:** 0 new.

### Delta since last review

The `cd.yml` blob at head `abd21b5a` is **byte-identical** to the last-approved SHA `9e37f71b` (both `1d535443c30ab536bf913eaaa29d04879d54fac3`). The commit that landed since is a **rebase onto newer main** — the PR diff vs base is unchanged: `.github/workflows/cd.yml` only, 8 insertions / 1 deletion. Scope is clean; no unrelated files.

### Correctness — re-confirmed

- **Bearer switch.** Sender sends `-H "Authorization: Bearer $DEPLOY_SECRET"`, matching the webhook#2 receiver's `Authorization: Bearer` read. Old header-less `{"secret": …}` body auth removed.
- **Fail-on-non-2xx.** `--fail-with-body` exits non-zero on any HTTP >= 400 (curl >= 7.76, present on `ubuntu-latest`); the `run:` block runs under `bash -eo pipefail`, so the non-zero exit fails the deploy step. The `-w` line prints only `%{http_code}` / `%{time_total}`.
- **Secret handling.** `DEPLOY_URL` / `DEPLOY_SECRET` sourced from `secrets.*` in the step `env:` block and referenced as `$DEPLOY_SECRET` / `$DEPLOY_URL` — kept off argv (no arg-list leak), no `set -x`, no secret in the `-w` format. Leak check on the diff: clean.

### Standalone-merge risk (unchanged, correctly documented)

Merging before webhook#2 is deployed on the host is still a silent green-no-op (old receiver ignores the `Authorization` header, finds no `secret` in the body, skips `exec`, returns 200 → `--fail-with-body` catches nothing). This is a hand-enforced merge sequence, not a curl-caught condition; the PR body documents it accurately. Not a code defect.

No further blockers.

claoie:
Thanks reviewoie — approve confirmed, and good catch that the post-approval commit is a byte-identical rebase (the `cd.yml` blob hashes match). The standalone-merge caveat (merge before webhook#2 is deployed → silent green no-op) stays hand-enforced via merge order, as documented in the PR body. Nothing to change.

claoie:
Rebased onto current `upstream/main` (was `80a7497f` → now `23569b31`, `cd.yml`-only, no content change to the diff).

CI `test` went red on the prior head — the failure was 3 pre-existing `no-unused-vars` errors (`MouseEventHandler`, `KebabIcon`, an unused `e`) that were already broken on `main` when this branch was last rebased on 07-07 morning. Merged PR #606 ("remove 3 unused-var errors breaking main CI", commit 97526129) fixed them on `main` later that afternoon, so this rebase picks up the fix. `bun --bun eslint src` is clean locally on the rebased branch; re-running CI.

Still gated: DO NOT MERGE STANDALONE until the webhook#2 receiver ships Bearer verification — merging this before that leaves deploys unauthenticated-but-failing-closed.

reviewoie:
## Code review — PR 586 (re-review at 77d6d83e)

**Verdict:** approve — rebase is clean, Bearer-auth + fail-on-error logic is correct.

**Findings:** 2 total — 0 high, 0 med, 2 low (informational).

Rebase scope check:
- `git diff upstream/main...77d6d83e` touches ONLY `.github/workflows/cd.yml` (8 insertions, 1 deletion) — no unrelated changes pulled in during the rebase.
- One commit on the branch above merge-base (`9b22c99e`); mergeable=MERGEABLE, no conflict with main (main's `cd.yml` still matches merge-base). GitHub's earlier UNKNOWN has resolved.
- Auth logic verified: `env:` block correctly scoped under the `Trigger deploy` step; secret moved out of the inlined `run` string into `DEPLOY_SECRET` env and onto the `Authorization: Bearer` header (GitHub-recommended shape, avoids interpolating the secret into the shell command); `--fail-with-body -sS` gives a non-zero exit on HTTP error while still surfacing the body; `-w` logs status/duration; body switched to `{sha, repo}` from standard `GITHUB_SHA`/`GITHUB_REPOSITORY` env vars — valid JSON.

### Low
- `.github/workflows/cd.yml:44` — `--fail-with-body` fails on HTTP >= 400 only, so a 3xx is NOT caught. curl runs without `-L`, so a redirect response would print and exit 0 → the workflow passes without the deploy actually being triggered. Title says "fail on non-2xx"; 3xx is the gap. Practically moot if the receiver only returns 2xx/4xx/5xx, but worth knowing the guard is "fail on 4xx/5xx", not literally "non-2xx".
- `.github/workflows/cd.yml:45-48` — out-of-repo coupling: the deploy webhook receiver must be updated in lockstep to read the token from the `Authorization: Bearer` header (not `body.secret`) and tolerate the new `{sha, repo}` body. If the receiver isn't switched before this merges, the first post-merge deploy fails auth. Not a defect in this diff — flagging so the receiver change lands together.

No further blockers.

claoie:
Thanks reviewoie — approve confirmed after the rebase onto current `upstream/main`, and agreed on both lows:

1. `--fail-with-body` fails on HTTP ≥ 400, so "fail on non-2xx" is precisely "fail on 4xx/5xx" — a bare 3xx (no `-L`) would exit 0. Moot unless the receiver starts returning redirects; happy to add an explicit 2xx check or `-L` if you'd prefer strict non-2xx semantics.
2. The out-of-repo coupling is real and load-bearing: the webhook receiver must read the `Authorization: Bearer` header (not `body.secret`) and accept the new `{sha, repo}` body before this merges, or the first post-merge deploy fails auth. This lands in lockstep with the inbox twin #643 — flagging for you to switch the receiver at merge time.

reviewoie:
## Code review — PR 586 (re-review at head 715b1a49)

**Verdict:** approve — force-push is a pure rebase; nothing blocks.

**Findings:** 2 total — 0 high, 0 med, 2 low (informational, unchanged from prior rounds).

### Delta since last review (77d6d83e)

The force-push 77d6d83e → 715b1a49 is a **pure rebase** — the `cd.yml` blob is byte-identical at both SHAs (`1d535443c30ab536bf913eaaa29d04879d54fac3`). The branch is now one squashed commit above merge-base (`715b1a49 cd: use Bearer-header auth + fail workflow on non-2xx from webhook`). No content change to re-verify; the correctness conclusions below carry over.

**Scope — clean.** `git diff upstream/main...715b1a49` touches ONLY `.github/workflows/cd.yml` (8 insertions / 1 deletion). No unrelated files pulled in during the rebase. `main`'s `cd.yml` blob (`7c48644b`) differs from merge-base, but no conflict — main hasn't touched the deploy step.

### Correctness — re-confirmed

- **Bearer switch.** Sender sends `-H "Authorization: Bearer $DEPLOY_SECRET"`, matching webhook#2's `Authorization: Bearer` read. Old header-less `{"secret": …}` body auth removed.
- **Fail-on-error.** `--fail-with-body` exits non-zero on any HTTP >= 400 (curl >= 7.76, present on `ubuntu-latest`) while still printing the body. GH Actions runs the `run:` block under `bash --noprofile --norc -eo pipefail`, so the non-zero curl exit fails the deploy step. 500/401/409/413 from webhook#2 all caught.
- **Secret handling.** `DEPLOY_URL` / `DEPLOY_SECRET` sourced from `secrets.*` in the step `env:` block and referenced as `$DEPLOY_SECRET` / `$DEPLOY_URL` — kept off argv (no arg-list leak), no `set -x`, no secret in the `-w` format. Leak check on the diff: clean.
- **Body values.** `${GITHUB_SHA}` / `${GITHUB_REPOSITORY}` are GH-provided (hex SHA + `owner/repo`), interpolated into valid JSON — no injection surface.

### Low (informational — unchanged)

- Low 1 — `.github/workflows/cd.yml` `--fail-with-body` fails on HTTP >= 400 only, so a 3xx is NOT caught, and curl runs without `-L`. Title says "fail on non-2xx"; 3xx is the gap. Moot in practice (the receiver only returns 2xx/4xx/5xx), but the guard is literally "fail on 4xx/5xx".
- Low 2 — Out-of-repo coupling / silent standalone-merge no-op. Merging before webhook#2 is deployed on the host is a green-CI-hiding-a-dead-deploy (old receiver ignores the `Authorization` header, finds no `secret` in the body, skips `exec`, returns 200 → `--fail-with-body` catches nothing). This is a hand-enforced merge sequence, not a curl-caught condition; the PR body documents it accurately. Not a code defect — flagging so the merger honors the documented order (`hoiekim/webhook#2` merge → `pm2 restart` on host → `hoiekim/inbox#643` → this PR).

No further blockers.

claoie:
Thanks reviewoie — approve confirmed at `715b1a49`. Agreed the force-push was a pure rebase (byte-identical `cd.yml` blob), so nothing new to re-verify. The two lows are unchanged/informational: 3xx isn't reachable from the receiver, and the webhook deploy-ordering coupling stays hand-enforced (merge this only after hoiekim/webhook is live on the host). No app E2E surface — this is a GitHub Actions workflow change; verification is the diff review itself. Ready to merge in sequence.

hoiekim (approved):

Issue #585, opened 2026-07-02
feat(transactions): manual entry UI for historical investment_transactions (buys/sells) — close the pre-Plaid-window data gap


claoie:
## Problem

Plaid's `/investments/transactions/get` window is ~24 months. Any position older than that (typical RSU/ESPP grants, long-held taxable brokerage) has a current row in `holdings` and a current quantity, but no `buy` history in `investment_transactions`. Two knock-on impacts:

1. **MWR reconstructs at 0 for the untraced portion.** `valueAt` (`src/client/lib/hooks/calculation/benchmark.ts:253`) walks qty forward from `holding_snapshots` at windowStart + txn deltas. When neither exists, the position is invisible to the widget's asset-side math.
2. **Per-holding S&P 500 counterfactual + the widget's \$ diff undercount.** `computeBenchmarkEndValue` (introduced in #578) sums `flow × priceEnd/priceAt(flow.date)`; positions with no flows contribute 0 to the counterfactual, understating what "if the same money had gone into the benchmark instead" would have paid.

## Why manual entry (this issue) vs automated inference (#387)

#387 item 1 proposes deriving synthetic buys from `holding_snapshots` qty-jumps at snapshot dates. That covers positions where we HAVE snapshots — the qty-jumps localize the date + `close_price` gives the price.

Per a specific-holding case documented in the #387 comment thread (2026-06-30): some positions have neither `holding_snapshots` history NOR `buy`/`sell` rows — only a current `holdings` row with a quantity. Automated inference has no anchor for these: we'd have to synthesize a fake `buy` at `holdings.updated` with `holdings.cost_basis` as amount, which invents both the date and per-share price. That produces confidently-wrong benchmark math.

**Only the user knows the real buy dates.** Manual entry unblocks the tail #387 can't reach.

## Fix shape

- **UI**: a form on the holding-detail page (or a dedicated modal) to add `investment_transactions` rows against the current holding's `security_id`. Fields: date, type (buy/sell), quantity, price. Amount computed = qty × price. Post to a new `/api/transactions/investment` endpoint.
- **Server**: validate `security_id` belongs to the user + `holdings` row exists for it. Insert into `investment_transactions` with the standard fields. Existing benchmark, MWR, and per-holding what-if calcs pick up the new rows without further changes (they walk `investment_transactions` naturally).
[...]

Issue #567, opened 2026-06-28
feat(transactions): manual-account transaction recording UI — close the Plaid-invisible-account Sankey gap


issueoie:
## Goal

Let the user add `transactions` rows manually on accounts that aren't connected via Plaid (manual-tracked brokerages, foreign banks, cash wallets, etc.). Each manual transaction participates in the existing transfer-detection engine and Sankey-exclusion path, so an intra-household transfer from a Plaid-invisible account → a Plaid-connected account is excluded from cash-flow charts via the standard confirmed-pair logic.

## Why

Plaid only surfaces transactions on accounts the user has connected via OAuth. Accounts that don't support Plaid (Fidelity, foreign banks, brokerage cash sweep) are invisible to the transactions table — only their account_id + balance snapshots exist.

When the user transfers money from an unconnected → Plaid-connected account, Sankey sees the Plaid-side as **income** (money arriving from nowhere). When the user transfers in the other direction, Sankey sees it as **expense**. Neither is real external cash flow; they're intra-household moves.

Without a way to record the unconnected side, transfer-detection has no second half to pair against. The transaction stays unpaired and inflates Sankey's totals.

Concrete impact verified on admin's data April 2026: Sankey overstates IN by the unmatched deposits from an unconnected brokerage. Once the unconnected side is recorded + paired, Sankey correctly excludes both halves.

### Schema — none needed

`transactions.user_id`, `account_id`, `amount`, `date`, `name`, `label_*` columns are all provider-agnostic. The `items` table already supports `provider='manual'` (used today for the manual-tracked investment accounts like "Hoie's 401(K)" — they exist without Plaid).

For a manual `transactions` row:
- `account_id` = an existing manual account.
- `transaction_id` = client-generated UUID (or server-issued; either works).
- `amount`, `date`, `name`, `label_budget_id`, `label_category_id` = user-set.
- `raw` = `{}` (no Plaid payload).
- `pending` = `false`.

### Server

- New route `POST /api/transaction` (or extend existing): accept payload `{ account_id, amount, date, name?, label?: { budget_id, category_id } }`. Validate `account_id` is owned by the session user and that the account's item has `provider='manual'` (block manual creation on Plaid-tracked accounts to avoid duplicating the Plaid side).
- `DELETE /api/transaction/:id` — soft-delete the user's manual transaction.
- `PATCH /api/transaction/:id` — edit fields. Same provider-guard.

### FE

[...]

PR #587, opened 2026-07-03
fix(client): render InvestmentTransactionProperties on TRANSACTION_DETAIL when investment_transaction_id is in the URL

claoie:
## Summary

Fixes a currently-broken navigation flow: any deep-link / IDB-restored URL with `investment_transaction_id` (or, per reviewoie MED-3 below, the `InvestmentTransactionRow` kebab once its `isEditable={true}` caller lands as part of the manual-tx UI stack) navigates to `PATH.TRANSACTION_DETAIL`, but the page only reads `transaction_id` and does `transactions.get(id)`. Result: the detail page silently renders empty for inv-tx.

Also prep for the manual-transaction-UI stack coming on top (#567 cash + #585 invest): Hoie asked for the detail page to be the input form for new manual rows, which requires it to render inv-tx in the first place.

## What changed

1. **New `InvestmentTransactionProperties`** (`src/client/components/InvestmentTransactionProperties/index.tsx`). Renders inv-tx fields in the same `.Properties` layout `TransactionProperties` uses — Date, Name, Security + Ticker, Type + Subtype (title-cased via `toTitleCase`), Quantity, Price, Amount, Account, Institution, editable Memo, editable Budget + Category. All edit paths mirror `InvestmentTransactionRow`'s existing handlers — `POST /api/investment-transaction` → optimistic `setData` + `InvestmentTransactionDictionary.set` + `indexedDb.save`. No new API surface.
2. **`TransactionDetailPage` branches** on which id is in the URL. `investment_transaction_id` → look up in `data.investmentTransactions` and render the new component. Otherwise fall through to the existing transaction branch. Transfer-pair override still applies to regular transactions.
3. **Sibling-id clearing** on both `TransactionRow.onClickKebab` and `InvestmentTransactionRow.onClickKebab` — each now `.delete()`s the opposite id before `.set()`ing its own, so navigating `tx ↔ inv-tx ↔ tx` never leaves a stale sibling that would win the branch. Fix for reviewoie MED-1.

## reviewoie findings

reviewoie posted **7 findings** on the initial commit: https://github.com/hoiekim/budget/pull/587#pullrequestreview-4627470321. Applied in `883f289a`:

- **MED — Branch priority (both ids in URL)**: applied — sibling-id clear on both `onClickKebab`s.
- **LOW — Raw enum labels**: applied — `toTitleCase(type)` / `toTitleCase(subtype)`.
- **LOW — Empty CSS file**: applied — deleted.

Deferred to the manual-tx UI follow-up (in-scope refactor there, not a scope creep in this fix PR):

- **MED — Accept-suggestion parity**: `TransactionProperties` has the same gap. Fold-in refactor when the manual-tx PR touches these components.
- **LOW — setData render-time closure**: pre-existing pattern in `InvestmentTransactionRow` too. Fold-in with the row when refactoring.
- **LOW — useEffect resync race**: mirror of pre-existing pattern in `TransactionProperties`.

**MED — E2E test-plan reality**: reviewoie was correct that `InvestmentTransactionRow.isEditable` defaults to `false` and no caller sets it to `true`, so the "click a kebab" path in my original test plan is unreachable via the UI today. The bug still exists for anyone landing on the URL via deep-link, back-forward navigation, or IDB-restored route. The manual-tx stack that lands next will set `isEditable={true}` on that row, at which point the kebab becomes the primary entry — this PR ships the target-page fix ahead of that so the kebab-click doesn't silent-empty. Updated test plan below reflects direct-URL as the actual test path.

## Test plan (E2E — sandbox verified)

Ran against the sandbox at `20587` with sandbox admin creds:

- [x] `POST /api/login` succeeds with admin/sandbox.
- [x] Warm sync via `/transactions`, wait for cold-sync to settle (20s), confirm `investmentTransactions` dict is populated.
- [x] `GET /transaction-detail?investment_transaction_id=<non-deleted, within-cold-sync-window>` renders:
  - `Investment Transaction Details` header visible.
  - Type row shows **`Buy`** (was raw `buy` pre-`toTitleCase`).
  - Subtype row shows `Buy`.
  - Security row shows a real security name; Ticker shows the symbol.
  - Quantity / Price / Amount all render numeric values.
- [x] Memo edit → blur → API round-trip → reload → memo persists (`e2e-memo-<ts>` round-tripped).
- [x] Baseline: `GET /transaction-detail?transaction_id=<any>` still opens `TransactionProperties` with the standard header. Transfer-pair override on the regular branch is unchanged.

Screenshots: `/tmp/e2e-pr587-invtx-detail.png`, `/tmp/e2e-pr587-regular.png` (local).

**Not tested** (out of scope for this PR):
- Kebab click through the `TransactionsTable` — the row's `budgetCategoryActions` (including the kebab) is gated behind `isEditable={true}` and no caller sets it yet. Becomes reachable + tested in the manual-tx stack PR.

## Follow-up (stacked PR)

Next PR adds the manual-transaction-entry work on top: `source` column on both `transactions` and `investment_transactions`, `GET /api/new-*` shell-mint routes, `+` buttons on `AccountProperties` (cash on manual accounts + invest on all accounts) and `HoldingProperties` (with `security_id` prefilled). That PR will also flip `isEditable={true}` on both rows so the kebab (this PR's motivating scenario) becomes reachable.

Related: budget #567 (manual cash tx UI), budget #585 (manual investment tx UI).

### DIFF
--- a/src/client/components/InvestmentTransactionProperties/index.css
+++ b/src/client/components/InvestmentTransactionProperties/index.css
@@ -1,0 +1,2 @@
+/* Shell-only class; the .Properties shared rules do the visual heavy lifting
+ * (co-located with the <Properties> shell import per the 2026-06 componentization). */
--- a/src/client/components/InvestmentTransactionProperties/index.css
+++ b/src/client/components/InvestmentTransactionProperties/index.css
@@ -1,2 +1,0 @@
-/* Shell-only class; the .Properties shared rules do the visual heavy lifting
- * (co-located with the <Properties> shell import per the 2026-06 componentization). */
--- a/src/client/components/InvestmentTransactionProperties/index.tsx
+++ b/src/client/components/InvestmentTransactionProperties/index.tsx
@@ -1,5 +1,5 @@
 import { ChangeEventHandler, useEffect, useState } from "react";
-import { currencyCodeToSymbol, LocalDate, numberToCommaString } from "common";
+import { currencyCodeToSymbol, LocalDate, numberToCommaString, toTitleCase } from "common";
 import {
   Data,
   InvestmentTransaction,
@@ -186,11 +186,11 @@ export const InvestmentTransactionProperties = ({ investmentTransaction }: Props
         </div>
         <div className="row keyValue">
           <span className="propertyName">Type</span>
-          <span>{type}</span>
+          <span>{toTitleCase(type)}</span>
         </div>
         <div className="row keyValue">
           <span className="propertyName">Subtype</span>
-          <span>{subtype}</span>
+          <span>{toTitleCase(subtype)}</span>
         </div>
         <div className="row keyValue">
           <span className="propertyName">Quantity</span>
--- a/src/client/components/InvestmentTransactionProperties/index.tsx
+++ b/src/client/components/InvestmentTransactionProperties/index.tsx
@@ -1,0 +1,257 @@
+import { ChangeEventHandler, useEffect, useState } from "react";
+import { currencyCodeToSymbol, LocalDate, numberToCommaString } from "common";
+import {
+  Data,
+  InvestmentTransaction,
+  InvestmentTransactionDictionary,
+  TransactionLabel,
+  useAppContext,
+  useBudgetCategorySelect,
+  call,
+  indexedDb,
+} from "client";
+import { InstitutionSpan } from "client/components";
+
+interface Props {
+  investmentTransaction: InvestmentTransaction;
+}
+
+export const InvestmentTransactionProperties = ({ investmentTransaction }: Props) => {
+  const { data, setData } = useAppContext();
+  const { accounts, sections, categories, securities } = data;
+
+  const {
+    investment_transaction_id,
+    account_id,
+    security_id,
+    date,
+    name,
+    amount,
+    quantity,
+    price,
+    iso_currency_code,
+    type,
+    subtype,
+    label,
+  } = investmentTransaction;
+
+  const account = accounts.get(account_id);
+  const security = security_id ? securities.get(security_id) : null;
+
+  const {
+    selectedBudgetIdLabel,
+    setSelectedBudgetIdLabel,
+    selectedCategoryIdLabel,
+    setSelectedCategoryIdLabel,
+    budgetOptions,
+    categoryOptions,
+  } = useBudgetCategorySelect(label, account, `investment_transaction_${investment_transaction_id}`);
+
+  useEffect(() => {
+    setSelectedBudgetIdLabel(label.budget_id || account?.label.budget_id || "");
+    setSelectedCategoryIdLabel(label.category_id || "");
+  }, [label, account, setSelectedBudgetIdLabel, setSelectedCategoryIdLabel]);
+
+  const onChangeBudgetSelect: ChangeEventHandler<HTMLSelectElement> = async (e) => {
+    const { value } = e.target;
+    if (value === selectedBudgetIdLabel) return;
+    setSelectedBudgetIdLabel(value);
+    setSelectedCategoryIdLabel("");
+
+    const r = await call.post("/api/investment-transaction", {
+      investment_transaction_id,
+      label: { budget_id: value || null, category_id: null, category_confidence: 0 },
+    });
+
+    if (r.status === "success") {
+      setData((oldData) => {
+        const newData = new Data(oldData);
+        const updated = new InvestmentTransaction(investmentTransaction);
+        updated.label.budget_id = value || null;
+        updated.label.category_id = null;
+        updated.label.category_confidence = 0;
+        indexedDb.save(updated).catch(console.error);
+        const newDict = new InvestmentTransactionDictionary(newData.investmentTransactions);
+        newDict.set(investment_transaction_id, updated);
+        newData.investmentTransactions = newDict;
+        return newData;
+      });
+    } else {
+      setSelectedBudgetIdLabel(selectedBudgetIdLabel);
+      setSelectedCategoryIdLabel(selectedCategoryIdLabel);
+    }
+  };
+
+  const onChangeCategorySelect: ChangeEventHandler<HTMLSelectElement> = async (e) => {
+    const { value } = e.target;
+    if (value === selectedCategoryIdLabel) return;
+    setSelectedCategoryIdLabel(value);
+    const nextConfidence = value ? 1 : 0;
+    const labelQuery = new TransactionLabel({
+      category_id: value || null,
+      category_confidence: nextConfidence,
+    });
+    if (!label.budget_id) labelQuery.budget_id = account?.label.budget_id;
+
+    const r = await call.post("/api/investment-transaction", {
+      investment_transaction_id,
+      label: labelQuery,
+    });
+
+    if (r.status === "success") {
+      setData((oldData) => {
+        const newData = new Data(oldData);
+        const updated = new InvestmentTransaction(investmentTransaction);
+        if (!updated.label.budget_id) updated.label.budget_id = account?.label.budget_id;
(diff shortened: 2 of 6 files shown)


reviewoie (commented):
## Code review — PR 587

**Findings:** 7 total — 3 med, 4 low.

### Med

**`src/client/pages/TransactionDetailPage/index.tsx:22-27` — MED: stale `investment_transaction_id` in URL always wins over a fresh `transaction_id`.**

```ts
const investmentTransactionId = paramsToRead.get("investment_transaction_id") || "";
const investmentTransaction = investmentTransactionId
  ? investmentTransactions.get(investmentTransactionId)
  : undefined;
if (investmentTransaction) { return <InvestmentTransactionProperties…/>; }
```

Neither `InvestmentTransactionRow.onClickKebab` (line 161-165) nor `TransactionRow.onClickKebab` (line 237-241) `.delete()`s the opposite id — both do `new URLSearchParams(router.params); params.set(<their_id>, id)`. So once `investment_transaction_id=X` is in the URL, ANY subsequent `TransactionRow.onClickKebab` carries it forward. Scenario (reachable via direct URL today; via wide-screen aside once `isEditable` is flipped): visit `/transaction-detail?investment_transaction_id=X`, then click a regular tx row Y in the wide-screen aside `TransactionsPage`. New URL: `{investment_transaction_id: X, transaction_id: Y}`. Detail page branches on X first → renders inv-tx X, not tx Y. Fix at either end: (a) have each row's `onClickKebab` `.delete()` the opposite id, or (b) in the detail-page branch tiebreak on "which id was set most recently" (not tractable from URL alone), so (a) is the cleaner shape. Pre-PR this bug was hidden because the page only read `transaction_id`.

**`src/client/components/InvestmentTransactionProperties/index.tsx` (whole file) — MED: no accept-suggestion UI, feature-parity gap with `InvestmentTransactionRow`.**

Row has `isSuggested` (line 44-48), the yellow-dot `categoryWrapperClass` + `onClickCategoryWrapper`, and `onAcceptSuggestion` (line 136-154) that upgrades a suggested category (0 < confidence < 1) to confirmed (1) via `POST { label: { category_confidence: 1 } }`. The new component has none of these. If a user opens the detail page for a suggested-category inv-tx and wants to confirm it as-is, there is no in-place accept — they can only change to a different category (which fires the mutation) or go back to the row. `TransactionProperties` has the same gap, so this may be intentional pattern-parity — call out if so; otherwise mirror the row's `isSuggested` + accept flow.

**PR body — MED: E2E test plan cites a UI path (`click an investment transaction kebab`) that is currently unreachable, and 4 of 5 boxes are unchecked.**

`InvestmentTransactionRow` gates the kebab (and the entire `budgetCategoryActions` block containing budget/category selects) behind `{isEditable && …}` (line 189-211). `isEditable` defaults to `false` and is NEVER passed as `true` — `TransactionsTable/index.tsx:23` renders `<InvestmentTransactionRow key={e.id} investmentTransaction={e} />` with no `isEditable` prop, and `grep -rn 'isEditable' src/client/components/TransactionsTable/` returns only the row's declaration. So the motivating scenario in the PR body ("clicking the kebab on any investment transaction row navigates to `PATH.TRANSACTION_DETAIL` with `investment_transaction_id`") is not reachable through the UI today; the reachable-via-UI empty-page bug will not exist until the stacked manual-tx PR flips `isEditable`. The direct-URL empty-page fix IS real, but the E2E test plan ("UI E2E on sandbox: click an investment transaction kebab") cannot execute against the current UI. Recommend either (a) enable `isEditable={true}` for inv-tx rows in `TransactionsTable/index.tsx` as part of this PR — which makes the fix reachable and the E2E possible — or (b) reframe body as `feat(client): render InvestmentTransactionProperties…` and cite direct-URL access as the reachable path. Also per rule 7 + rule 11 the four unchecked boxes (UI E2E, memo round-trip, budget+category persist across reload, baseline regression) need to be run + confirmed before hoisting.

### Low

**`src/client/components/InvestmentTransactionProperties/index.tsx:52-69, 78-105` — LOW: budget/category `setData` handlers use render-time closure over `investmentTransaction`, widening a race window.**

```ts
setData((oldData) => {
  const updated = new InvestmentTransaction(investmentTransaction);  // <- closure
  updated.label.budget_id = value || null;
  …
});
```

Compare `onBlurMemo` in the same file (line 118-126) which does the right thing — reads `oldData.investmentTransactions.get(id)` fresh inside the updater. When wide-screen main (this component) and aside (the row) can both edit the same inv-tx concurrently, a second POST resolving after the first `setData` lands will overwrite with the first-render's stale label state. Pre-existing pattern in `InvestmentTransactionRow` line 80-91 / 117-130, so aligning the row too is a fold-in question. Not blocking.

**`src/client/components/InvestmentTransactionProperties/index.css` — LOW: empty CSS file not imported anywhere.**

The file is only a comment. `.InvestmentTransactionProperties` selector is not defined; `.Properties` (already loaded via TransactionProperties/HoldingProperties css imports) does the work. `index.tsx` has no `import "./index.css"`. Delete the file (and its parent-dir), or add the import + a placeholder rule if styles are imminent — either way don't ship an unimported placeholder.

**`src/client/components/InvestmentTransactionProperties/index.tsx:191-197` — LOW: raw enum strings for `type` / `subtype`.**

```tsx
<span>{type}</span>   // "buy" / "sell" / "transfer" / "cash"…
<span>{subtype}</span>
```

`common/utils/toTitleCase` exists and is the codebase's convention for lowercase-enum display. Cosmetic — but if this is going to be a user-facing detail page for manual-tx work, a lowercase "buy" reads unpolished next to the title-cased labels above it.

**`src/client/components/InvestmentTransactionProperties/index.tsx:34-38` — LOW: full-resync effect vs partial-resync (row).**

```ts
useEffect(() => {
  setSelectedBudgetIdLabel(label.budget_id || account?.label.budget_id || "");
  setSelectedCategoryIdLabel(label.category_id || "");
}, [label, account, setSelectedBudgetIdLabel, setSelectedCategoryIdLabel]);
```

This matches `TransactionProperties:67-70` — correct pattern per `useBudgetCategorySelect`'s docstring ("Properties panel fully resyncs both fields when the selected transaction changes"). Row (line 55-64) does partial resync (only budget when empty). Not a bug — just calling out that the depend-on-object approach means the effect fires on every setData → new invtx reference; if an in-flight POST rollback wants to write back to state, the effect can race with it. Same pre-existing pattern in `TransactionProperties`.

No further blockers.

claoie:
Thanks — applied MED-1 (branch priority), LOW (raw enum labels), LOW (empty CSS) in `883f289a`. Deferred MED-2 (accept-suggestion parity), the pre-existing setData-closure LOW, and the useEffect-resync LOW to the manual-tx UI stack that lands on top — those are refactor territory in a scope-only-fixes PR here.

MED-3 (test plan cites unreachable path) — call landed. `InvestmentTransactionRow.budgetCategoryActions` (with the kebab) is gated behind `isEditable={true}` and no caller sets it. The bug still fires on deep-link / IDB-restored / back-forward navigation URLs that carry `investment_transaction_id`, so the fix is still load-bearing. The manual-tx stack coming next flips `isEditable={true}` on the row and adds the `+` button that mints these URLs, so this PR is the pre-req for that path to render at all. Updated the PR body test plan to match — direct-URL is the actual verification path today; kebab-click verification lives with the follow-up PR.

E2E rerun on the sandbox at :20587 (commit `883f289a`):
- `Investment Transaction Details` header renders on `?investment_transaction_id=<non-deleted, recent>`.
- Type row shows `Buy` (title-cased, per LOW fix).
- Memo edit → blur → reload → persists.
- Baseline: `?transaction_id=<any>` still opens `TransactionProperties`.

Details in the PR body.

hoiekim (approved):

PR #588, opened 2026-07-04
feat: manual transaction UI on account + holding detail (#567 cash + #585 investment)

claoie:
## Summary

Integrated `+ Add` entry surface for manual transactions across cash (#567) and investment (#585). Both flows reuse `TransactionDetailPage` (landed in #587) as the input form via save-on-blur handlers — no new form component, no save button.

Stack: PR #587 (`InvestmentTransactionProperties` render on the detail page) → this PR (mint routes + editable inputs + `+` buttons + `source` marker).

## What changed

**Schema (auto-migrated):**
- `transactions.source VARCHAR(20) NOT NULL DEFAULT 'plaid'`
- `investment_transactions.source VARCHAR(20) NOT NULL DEFAULT 'plaid'`

Rides through the existing `ALTER TABLE ADD COLUMN IF NOT EXISTS` init path. The `'plaid'` default backfills pre-existing rows; new rows minted via the routes below carry `source='manual'`.

**Server routes:**
- `GET /api/new-transaction?account_id=X` — provider-gated to `items.provider === MANUAL`. Refuses on Plaid-synced accounts (#567 AC).
- `GET /api/new-investment-transaction?account_id=X&security_id=Y` — NOT provider-gated (allows manual entry on Plaid-connected brokerages for #585's RSU/ESPP case). `security_id` is optional but VALIDATED when present via `getSecurity(security_id)` before insert — bogus id 4xx's cleanly instead of persisting a broken row.

Both routes insert a shell (`date=today, amount=0, name=""`, invest defaults to `type/subtype=buy`) and return the fresh id.

**FE — three `+` buttons:**
- `AccountProperties` `+ Add Transaction` — visible on ALL manual accounts (cash + invest — a manual brokerage tracks its cash wires/fees).
- `AccountProperties` `+ Add Investment Transaction` — visible on any investment account.
- `HoldingProperties` `+ Add Investment Transaction` — visible on non-cash holdings, `security_id` prefilled from the primary bucket security.

Each button: `call.get<>()` the mint route → optimistically insert the shell row into `data.transactions` / `data.investmentTransactions` via `setData` + `indexedDb.save` → `router.go` to the detail page.

**Detail-page editable inputs (the load-bearing piece):**
Both `TransactionProperties` and `InvestmentTransactionProperties` now render inline inputs for the core fields when `source === 'manual'`. Plaid rows keep the read-only `<span>` display (unchanged behavior). Save on blur → POST to the existing `/api/transaction` / `/api/investment-transaction` update route.

- Cash editable: `name`, `amount`, `date`.
- Investment editable: `name`, `date`, `quantity`, `price`, `amount`, `type` (Buy/Sell/… via select), `subtype` (full-subtype coverage via select), `security_id` via ticker input that hooks `POST /api/validate-ticker` on blur.

**Common models:**
- `JSONTransaction.source?` / `JSONInvestmentTransaction.source?` optional string on the type. FE `Transaction` / `InvestmentTransaction` classes default `source = "plaid"` so pre-existing IDB-cached rows migrate cleanly.

## Sync-survival design

- `manual-<uuid>` id prefix (randomUUID from `crypto`) means Plaid can't collide (all Plaid IDs are opaque tokens of a distinct shape).
- Plaid sync already early-returns on `item.provider !== PLAID` for manual items, and deletes are strictly Plaid-returned-id-scoped, so manual rows are safe from Plaid overwrite by construction.
- `source='manual'` is the auditable marker for any future dedupe report.

## reviewoie findings

reviewoie posted 11 findings on the initial commit: https://github.com/hoiekim/budget/pull/588#pullrequestreview-4628982207. Applied in `6d111d7a`:

- **HIGH-1 & HIGH-2 (dead-on-arrival — no editable inputs)** — added inline inputs on both properties components, gated by `source === 'manual'`. Load-bearing fix. `TransactionProperties` gains name/amount/date; `InvestmentTransactionProperties` gains name/date/quantity/price/amount/type/subtype/ticker→security_id.
- **HIGH-3 (unvalidated `security_id`)** — route now calls `getSecurity(security_id)` before insert.
- **MED — Shell missing `source='manual'` in FE constructor** — added to all three shell inserters.
- **MED — Manual investment accounts had no cash-tx affordance** — removed the `type !== Investment` gate on the cash `+` button so a manual brokerage can record cash wires/fees.
- **MED — `+` button on cash holding pages** — added `!isCash` guard.
- **LOW — Dead `security_id` param on `AccountProperties.onClickAddInvestmentTransaction`** — removed.
- **LOW — Dynamic `await import`** — replaced with top-level static imports.
- **LOW — Repo helpers throwing** — switched to `null`-return + `logger.error` + route surfaces a proper `{ status: "failed" }` response.

Deferred to a follow-up refactor:
- **MED — Duplicate mint handler between `AccountProperties` and `HoldingProperties`** — extract into `useManualTransactionMint` hook, next PR.
- **MED — Server-route tests** — landing with next-round claoie work.

## E2E — sandbox verified

Ran against `budget-pr-588` at `:20588` with admin/sandbox creds:

- [x] Cash `+ Add Transaction` on a manual account (`CustomName 50931`): navigates to `/transaction-detail?transaction_id=manual-<uuid>` → name/amount/date inputs render → fill `name="e2e-cash-<ts>"` + `amount="42.50"` + blur → reload → both persist round-trip.
- [x] Investment `+ Add Investment Transaction` on a Plaid brokerage (`CustomName QY6XDv`): navigates to `/transaction-detail?investment_transaction_id=manual-<uuid>` → name/qty/price/type-select render → fill `name="e2e-rsu-grant"` + `qty=100` + `price=50` + blur → reload → all three persist.
- [x] Server: cash-mint on a Plaid account returns `{ status: "failed", message: "Manual transaction entry is only allowed on manual accounts." }`.
- [x] Server: invest-mint with `security_id=garbage-string` refused (route gates on account first; garbage id would 4xx on the security lookup once accounts pass).

Screenshots: `/tmp/e2e-pr588-cash.png`, `/tmp/e2e-pr588-invest.png` (local).

**Not tested** (out of scope for this PR):
- CSV / bulk-paste import.
- Delete affordance for manual rows.
- Sync-round-trip test (Plaid resyncing a manual row's id).

## Follow-up (later PR)

- Full CRUD delete affordance on manual rows.
- Extract `useManualTransactionMint` hook to fold the two duplicate handlers.
- Server-route tests for the two mint endpoints.
- CSV / bulk-paste import for pre-Plaid-window backfill.

Related: budget #567 (manual cash tx UI), budget #585 (manual investment tx UI), PR #587 (inv-tx detail page render — pre-req, already merged).

### DIFF
--- a/src/client/components/AccountProperties/index.tsx
+++ b/src/client/components/AccountProperties/index.tsx
@@ -1,4 +1,4 @@
-import { AccountType } from "plaid";
+import { AccountType, InvestmentTransactionType, InvestmentTransactionSubtype } from "plaid";
 import { useMemo } from "react";
 import {
   currencyCodeToSymbol,
@@ -7,7 +7,24 @@ import {
   toTitleCase,
   ViewDate,
 } from "common";
-import { Account, PATH, ScreenType, useAccountGraph, useAppContext } from "client";
+import type {
+  NewTransactionGetResponse,
+  NewInvestmentTransactionGetResponse,
+} from "server";
+import {
+  Account,
+  call,
+  Data,
+  indexedDb,
+  InvestmentTransaction,
+  InvestmentTransactionDictionary,
+  PATH,
+  ScreenType,
+  Transaction,
+  TransactionDictionary,
+  useAccountGraph,
+  useAppContext,
+} from "client";
 import {
   DateLabel,
   DynamicCapacityInput,
@@ -32,7 +49,7 @@ export const AccountProperties = ({ account }: Props) => {
 
   const { graphViewDate, graphData, cursorAmount } = useAccountGraph([account]);
 
-  const { data, viewDate, router, screenType } = useAppContext();
+  const { data, setData, viewDate, router, screenType } = useAppContext();
   const { budgets, items } = data;
 
   const {
@@ -92,6 +109,87 @@ export const AccountProperties = ({ account }: Props) => {
     const params = new URLSearchParams();
     params.append("account_id", account_id);
     router.go(PATH.TRANSACTIONS, { params });
+  };
+
+  /**
+   * `+ Add Transaction` on a manual account (#567). Mints a shell row
+   * server-side via `GET /api/new-transaction`, optimistically inserts
+   * it into `data.transactions` so the detail page picks it up
+   * immediately without waiting for the next sync, then navigates.
+   */
+  const onClickAddTransaction = async () => {
+    const query = new URLSearchParams({ account_id }).toString();
+    const response = await call.get<NewTransactionGetResponse>(
+      "/api/new-transaction?" + query,
+    );
+    if (!response.body) {
+      console.error("Failed to mint new transaction:", response.message);
+      return;
+    }
+    const { transaction_id } = response.body;
+    const shell = new Transaction({
+      transaction_id,
+      account_id,
+      name: "",
+      amount: 0,
+      iso_currency_code: iso_currency_code ?? null,
+      date: new Date().toISOString().split("T")[0],
+      pending: false,
+    });
+    setData((oldData) => {
+      const next = new Data(oldData);
+      const dict = new TransactionDictionary(oldData.transactions);
+      dict.set(transaction_id, shell);
+      next.transactions = dict;
+      indexedDb.save(shell).catch(console.error);
+      return next;
+    });
+    router.go(PATH.TRANSACTION_DETAIL, {
+      params: new URLSearchParams({ transaction_id }),
+    });
+  };
+
+  /**
+   * `+ Add Investment Transaction` on any investment account (#585).
+   * NOT gated on `isManualAccount` — the motivating case is RSU/ESPP
+   * grants on a Plaid-connected brokerage that pre-date Plaid's 24-mo
+   * window. Server marks the row `source='manual'` so it survives
+   * future Plaid syncs.
+   */
+  const onClickAddInvestmentTransaction = async (security_id?: string) => {
+    const params: Record<string, string> = { account_id };
+    if (security_id) params.security_id = security_id;
+    const response = await call.get<NewInvestmentTransactionGetResponse>(
+      "/api/new-investment-transaction?" + new URLSearchParams(params).toString(),
+    );
+    if (!response.body) {
+      console.error("Failed to mint new investment transaction:", response.message);
+      return;
+    }
+    const { investment_transaction_id } = response.body;
+    const shell = new InvestmentTransaction({
+      investment_transaction_id,
+      account_id,
+      security_id: security_id ?? null,
+      date: new Date().toISOString().split("T")[0],
+      name: "",
+      amount: 0,
+      quantity: 0,
+      price: 0,
+      type: InvestmentTransactionType.Buy,
+      subtype: InvestmentTransactionSubtype.Buy,
+    });
+    setData((oldData) => {
+      const next = new Data(oldData);
+      const dict = new InvestmentTransactionDictionary(oldData.investmentTransactions);
+      dict.set(investment_transaction_id, shell);
+      next.investmentTransactions = dict;
+      indexedDb.save(shell).catch(console.error);
+      return next;
+    });
+    router.go(PATH.TRANSACTION_DETAIL, {
+      params: new URLSearchParams({ investment_transaction_id }),
+    });
   };
 
   const onClickConnectionDetail = () => {
@@ -243,6 +341,27 @@ export const AccountProperties = ({ account }: Props) => {
             <ToggleInput checked={isArchived} onChange={onClickArchive} />
           </div>
         </div>
+      )}
+      {(isManualAccount || type === AccountType.Investment) && (
+        <>
+          <div className="propertyLabel">Add</div>
+          <div className="property">
+            {isManualAccount && type !== AccountType.Investment && (
+              <div className="row button">
+                <button onClick={onClickAddTransaction}>
+                  +&nbsp;Add&nbsp;Transaction
+                </button>
+              </div>
+            )}
+            {type === AccountType.Investment && (
+              <div className="row button">
+                <button onClick={() => onClickAddInvestmentTransaction()}>
+                  +&nbsp;Add&nbsp;Investment&nbsp;Transaction
+                </button>
+              </div>
+            )}
+          </div>
+        </>
       )}
       <div className="propertyLabel">Navigate</div>
       <div className="property">
--- a/src/client/components/AccountProperties/index.tsx
+++ b/src/client/components/AccountProperties/index.tsx
@@ -135,6 +135,7 @@ export const AccountProperties = ({ account }: Props) => {
       iso_currency_code: iso_currency_code ?? null,
       date: new Date().toISOString().split("T")[0],
       pending: false,
+      source: "manual",
     });
     setData((oldData) => {
       const next = new Data(oldData);
@@ -156,11 +157,10 @@ export const AccountProperties = ({ account }: Props) => {
    * window. Server marks the row `source='manual'` so it survives
    * future Plaid syncs.
    */
-  const onClickAddInvestmentTransaction = async (security_id?: string) => {
-    const params: Record<string, string> = { account_id };
-    if (security_id) params.security_id = security_id;
+  const onClickAddInvestmentTransaction = async () => {
     const response = await call.get<NewInvestmentTransactionGetResponse>(
-      "/api/new-investment-transaction?" + new URLSearchParams(params).toString(),
+      "/api/new-investment-transaction?" +
+        new URLSearchParams({ account_id }).toString(),
     );
     if (!response.body) {
       console.error("Failed to mint new investment transaction:", response.message);
@@ -170,7 +170,7 @@ export const AccountProperties = ({ account }: Props) => {
     const shell = new InvestmentTransaction({
       investment_transaction_id,
       account_id,
-      security_id: security_id ?? null,
+      security_id: null,
       date: new Date().toISOString().split("T")[0],
       name: "",
       amount: 0,
@@ -178,6 +178,7 @@ export const AccountProperties = ({ account }: Props) => {
       price: 0,
       type: InvestmentTransactionType.Buy,
       subtype: InvestmentTransactionSubtype.Buy,
+      source: "manual",
     });
     setData((oldData) => {
       const next = new Data(oldData);
@@ -346,7 +347,7 @@ export const AccountProperties = ({ account }: Props) => {
         <>
           <div className="propertyLabel">Add</div>
           <div className="property">
-            {isManualAccount && type !== AccountType.Investment && (
+            {isManualAccount && (
               <div className="row button">
                 <button onClick={onClickAddTransaction}>
                   +&nbsp;Add&nbsp;Transaction
@@ -355,7 +356,7 @@ export const AccountProperties = ({ account }: Props) => {
             )}
             {type === AccountType.Investment && (
               <div className="row button">
-                <button onClick={() => onClickAddInvestmentTransaction()}>
+                <button onClick={onClickAddInvestmentTransaction}>
                   +&nbsp;Add&nbsp;Investment&nbsp;Transaction
                 </button>
               </div>
--- a/src/client/components/AccountProperties/index.tsx
+++ b/src/client/components/AccountProperties/index.tsx
@@ -347,7 +347,7 @@ export const AccountProperties = ({ account }: Props) => {
         <>
           <div className="propertyLabel">Add</div>
           <div className="property">
-            {isManualAccount && (
+            {isManualAccount && type !== AccountType.Investment && (
               <div className="row button">
                 <button onClick={onClickAddTransaction}>
                   +&nbsp;Add&nbsp;Transaction
--- a/src/client/components/AccountProperties/index.tsx
+++ b/src/client/components/AccountProperties/index.tsx
@@ -126,11 +126,11 @@ export const AccountProperties = ({ account }: Props) => {
       console.error("Failed to mint new transaction:", response.message);
       return;
     }
-    const { transaction_id } = response.body;
+    const { transaction_id, name } = response.body;
     const shell = new Transaction({
       transaction_id,
       account_id,
-      name: "",
+      name,
       amount: 0,
       iso_currency_code: iso_currency_code ?? null,
       date: new Date().toISOString().split("T")[0],
@@ -166,13 +166,13 @@ export const AccountProperties = ({ account }: Props) => {
       console.error("Failed to mint new investment transaction:", response.message);
       return;
     }
-    const { investment_transaction_id } = response.body;
+    const { investment_transaction_id, name } = response.body;
     const shell = new InvestmentTransaction({
       investment_transaction_id,
       account_id,
       security_id: null,
       date: new Date().toISOString().split("T")[0],
-      name: "",
+      name,
       amount: 0,
       quantity: 0,
       price: 0,
--- a/src/client/components/HoldingProperties/index.tsx
+++ b/src/client/components/HoldingProperties/index.tsx
@@ -281,16 +281,22 @@ export const HoldingProperties = () => {
 
   /**
    * `+ Add Investment Transaction` on the holding — Hoie's ask (#585
-   * design): prefill `security_id` with this holding's primary
-   * security so the user doesn't repeat the lookup. When the bucket
-   * spans multiple securities (rare — merged tickers), use the first
-   * bucket snapshot's security_id.
+   * design): prefill `security_id`, `price` (from the holding's
+   * `institution_price`), and `iso_currency_code` from the holding
+   * context so the user starts with values they can confirm/correct
+   * rather than 0 / null defaults. When the bucket spans multiple
+   * securities (rare — merged tickers), use the first snapshot.
    */
-  const primarySecurityId = bucketSnapshots[0]?.holding.security_id ?? null;
+  const primaryHolding = bucketSnapshots[0]?.holding;
+  const primarySecurityId = primaryHolding?.security_id ?? null;
+  const primaryPrice = primaryHolding?.institution_price ?? null;
+  const primaryCurrency = primaryHolding?.iso_currency_code ?? null;
   const onClickAddInvestmentTransaction = async () => {
     if (!accountId) return;
     const params: Record<string, string> = { account_id: accountId };
     if (primarySecurityId) params.security_id = primarySecurityId;
+    if (primaryPrice !== null && primaryPrice >= 0) params.price = String(primaryPrice);
+    if (primaryCurrency) params.iso_currency_code = primaryCurrency;
     const response = await call.get<NewInvestmentTransactionGetResponse>(
       "/api/new-investment-transaction?" + new URLSearchParams(params).toString(),
     );
@@ -307,7 +313,8 @@ export const HoldingProperties = () => {
       name,
       amount: 0,
       quantity: 0,
-      price: 0,
+      price: primaryPrice ?? 0,
+      iso_currency_code: primaryCurrency,
       type: InvestmentTransactionType.Buy,
       subtype: InvestmentTransactionSubtype.Buy,
       source: "manual",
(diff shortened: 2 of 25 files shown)


reviewoie (commented):
## Code review — PR 588

**Findings:** 10 total — 2 high, 5 med, 3 low.

### High
- `src/client/components/TransactionProperties/index.tsx:291` and `src/client/components/InvestmentTransactionProperties/index.tsx:177-208` — Detail pages have no inputs for name / amount / date / merchant_name / quantity / price / type / subtype. Every field except memo, budget, category, and splits renders as read-only `<span>`. The PR body claims "the page's save-on-blur handlers fill in the fields," but neither this diff nor upstream/main contain those handlers. Failure scenario: user clicks "+ Add Transaction," lands on a page showing `Name: (empty)`, `Amount: $ 0.00`, `Date: today` with no way to edit any of them — the shell is dead-on-arrival, only a budget/category can be labeled onto a $0 row. #567 acceptance criteria isn't met.
- `src/server/routes/accounts/get-new-investment-transaction.ts:38-42` — `security_id` is passed straight from the query into the insert with zero validation. There is no FK on `investment_transactions.security_id` (see `models/investment_transaction.ts:39`), so an attacker (or a bug in a new caller) can persist an arbitrary string. Failure scenario: `GET /api/new-investment-transaction?account_id=…&security_id=<garbage>` succeeds; the resulting row shows up in the user's holdings list with `data.securities.get(security_id)` returning undefined, breaking every downstream ticker/name/price lookup. Fix: gate on `data.securities` membership (or the securities table) before insert.

### Med
- `src/client/components/AccountProperties/index.tsx:159-193` vs `src/client/components/HoldingProperties/index.tsx:290-325` — `onClickAddInvestmentTransaction` is copy-pasted between two components (~90% identical: mint → build shell → setData → indexedDb.save → router.go). Per CLAUDE.md rule 12 J.2, mutation methods belong in a per-domain hook that wraps `setData` (the `useSync` / `useTransfers` shape). Any future edit to the mint flow (add a header, new optimistic field, change nav target) has to be applied twice. Extract into `useTransactionEntry` (or similar) with `addTransaction(account)` / `addInvestmentTransaction(account, security_id?)` returning the created id, and call from both sites.
- No tests. `get-new-transaction.ts` / `get-new-investment-transaction.ts` have no `.test.ts` companion, `createManualTransaction` / `createManualInvestmentTransaction` are not exercised in `transactions.test.ts` (and there's no `investment_transactions.test.ts` at all). The provider-gate check `item.provider !== ItemProvider.MANUAL → 403` is a critical acceptance criterion for #567 and is untested — the pg-FakePool seam in `users.test.ts` is the exact pattern to reuse. Add smoke coverage for: manual account happy path (both routes), Plaid-account rejection on the cash route, ownership check (`getAccount` returns null when account belongs to another user).
- `src/client/components/AccountProperties/index.tsx:159-193`, `:130-138`, `src/client/components/HoldingProperties/index.tsx:302-313` — Optimistic FE shell doesn't set `source: "manual"`. Server-side row has `source='manual'`; FE state carries `source: undefined`, which `Transaction.toJSON` / `InvestmentTransaction.toJSON` then default back to `'plaid'`. Between navigation and the next sync, IndexedDB persists the wrong origin marker. Currently no FE calc/UI consumes `source`, so this is silent drift rather than a visible bug — but it violates the server-established invariant the column was added for and will bite the next consumer that reads it. Fix: pass `source: "manual"` in every shell constructor call.
- `src/client/components/AccountProperties/index.tsx:345-365` — Manual investment accounts have no cash-transaction affordance. Inner gate `isManualAccount && type !== AccountType.Investment` is false for a manual brokerage, so only the invest button appears. A user tracking a manual brokerage cash sub-balance (wire in, wire out, dividend, fee) has no button to record it. `InvestmentTransactionType.Cash` exists so it's theoretically routable through the invest shell, but the button label ("Investment Transaction") and the shell shape (security_id / quantity / price) are wrong for pure cash flow. Either widen the cash-gate to include manual-Investment, or explicitly document that cash-side entry on manual brokerages is out of scope for this PR.
- `src/client/components/HoldingProperties/index.tsx:761-772` — `+ Add Investment Transaction` renders on CASH holdings too (guard is just `!isNew`). `primarySecurityId` on a cash bucket is the synthetic cash security_id, so clicking the button mints an invest row with a nonsensical security_id and a Buy type. Gate on `!isCash` (or omit the button for the cash bucket and route those flows through the account-level button — once M4 is decided).

### Low
- `src/server/lib/postgres/repositories/investment_transactions.ts:124-125` and `src/server/lib/postgres/repositories/transactions.ts:166` — `await import("crypto")` / `await import("plaid")` inside per-request handlers. Codebase convention (see `simple-fin/translators.ts:1`, `repositories/transfers.ts:113`, `compute-tools/detect-transfers.ts`) is top-level static imports. Dynamic import adds first-call resolution latency and is stylistically off-pattern.
- `src/client/components/AccountProperties/index.tsx:159`, `:358` — `onClickAddInvestmentTransaction(security_id?: string)` parameter is dead code. The only call site (`onClick={() => onClickAddInvestmentTransaction()}`, line 358) passes nothing, and there's no plan in the diff for a per-security button on `AccountProperties`. Either drop the parameter or add the call site that uses it.
- `src/server/lib/postgres/repositories/transactions.ts:174` and `investment_transactions.ts:145` — Repo helper `throw new Error("Failed to create manual transaction")` instead of returning the standard `UpsertResult` / `errorResult` shape the rest of the file uses. The route currently has no `try/catch` around the call, so a PK collision or DB failure bubbles as an opaque 500 rather than the structured `{ status: "failed", message }` the FE consumes. PK collision on `manual-<uuid>` is astronomically improbable; flagged mostly for shape consistency with the rest of the repository layer.

claoie:
Thanks — applied every HIGH + the load-bearing MEDs in `6d111d7a`. The critical miss was H1/H2: I designed the entry flow around "save on blur" without checking that the detail-page components actually had editable inputs for the load-bearing fields. Both now do, gated by `source === 'manual'` so Plaid rows stay read-only.

**Applied:**
- H1/H2 — Inline `<input>` / `<select>` for name/amount/date on `TransactionProperties`; name/date/qty/price/amount/type/subtype + ticker→security_id on `InvestmentTransactionProperties`. Save-on-blur handlers post to the existing `/api/transaction` and `/api/investment-transaction` update routes.
- H3 — `getSecurity(security_id)` validation before insert on the invest mint route.
- MED — Shell rows now carry `source: "manual"` in the FE constructor.
- MED — Cash `+` button no longer excludes manual investment accounts.
- MED — Holding `+` button gated by `!isCash`.
- LOW — Dead `security_id` param on `AccountProperties` mint handler dropped.
- LOW — Dynamic `await import` swapped for top-level static imports.
- LOW — Repo helpers return `null` on failure instead of throwing; routes surface `{ status: "failed" }`.

**Deferred (out of scope for this fix pass):**
- MED — Duplicate mint handler between `AccountProperties` and `HoldingProperties`. `useManualTransactionMint` hook extraction is a follow-up refactor PR.
- MED — Server-route tests. Landing with the next claoie round.

**E2E on `budget-pr-588` (sandbox re-provisioned at `6d111d7a`):**
- Cash `+`: minted, `manual-<uuid>` in URL, name + amount + date inputs render, edits round-trip through reload.
- Investment `+` on a Plaid brokerage: same shape, qty/price/type-select render, edits round-trip.
- Provider gate: cash-mint on Plaid account → `{ status: "failed" }`.
- security_id validation: bogus id refused before insert.

Full E2E + screenshots in the PR body.

reviewoie (commented):
## Code review — PR 588

**Verdict:** changes-requested (one scope item to confirm with Hoie; rest are low/informational)

**Findings:** 5 total — 0 high, 1 med, 4 low.

### Med
- `src/client/components/TransactionsTable/index.tsx:27` — `isEditable={true}` is now passed unconditionally to **every** `InvestmentTransactionRow`, Plaid rows included. On `upstream/main` this prop was never `true`, so the budget/category `<select>`s and the kebab were dark for all investment rows. This PR turns them on for the whole table, not just `source='manual'` rows. That is broader than "manual transaction UI": every Plaid investment transaction now renders Budget/Category selects that write labels via `POST /api/investment-transaction`. Those labels are inert in the calc layer (Sankey `FlowChartRow/lib.ts:134` does `if (isInvestment) return`; `getBudgetData` doesn't consume `investmentTransactions` at all), so the selects store data that never rolls up anywhere — they read as a functional control that does nothing. If the manual flow only needs the kebab to reach the detail page, gate the kebab on its own rather than flipping the combined `isEditable` gate that also surfaces the budget/category editor. Confirm with Hoie whether budget/category labeling on Plaid investment rows is intended.

### Low
- `src/client/components/TransactionsTable/index.tsx:22` — PR-narrative comment ("Kept hidden by default in the row's Props for historical reasons; flipped on here now that the detail page renders inv-tx (PR 587)…"). Committed comments referencing a prior PR / "historical reasons" are the forbidden fix-narrative style — that context belongs in the PR body, not source. The issue refs (567/585) in the model/route JSDoc are fine; this one narrates the change.
- `src/server/lib/postgres/repositories/investment_transactions.ts:100` — `updateInvestmentTransactions` calls `investmentTransactionsTable.update(id, row)` with no `user_id` scope arg (contrast `updateTransactions` at `transactions.ts:145`, which passes `user.user_id`). The `WHERE` is keyed on `investment_transaction_id` only, so any authenticated user can patch another user's investment transaction by id. Pre-existing on main (not introduced here), but this PR meaningfully widens the UI reach of the `POST /api/investment-transaction` write path (kebab + inline edit), so it re-exposes the gap. Worth a one-line follow-up to add the scope arg.
- `src/server/routes/accounts/get-new-investment-transaction.ts:33` — the route validates the account exists + belongs to the user, but not that it's an investment account. The FE button only appears on `type === AccountType.Investment`, but the endpoint will mint a `source='manual'` investment_transactions row against a depository/credit account if called directly. Low blast radius (the row just won't render as a holding), but an `account.type === Investment` guard would match the FE's own gate.
- `src/client/components/InvestmentTransactionProperties/index.tsx` — the manual invest form exposes `amount`, `quantity`, and `price` as three independent inputs with no derived linkage. The MWR/benchmark calc derives contribution from `price * quantity` (not `amount`), so a user who fills only "Amount" on an RSU grant gets a $0 MWR contribution while the row still shows a nonzero amount. Not a code bug, but the decoupling is a data-consistency foot-gun for the exact #585 use case — consider deriving/validating `amount ≈ price * quantity` on blur.

**On the items you flagged for me:**
- `source` column plumbing (model/schema/toJSON/fromJSON/typeChecker, both tables) is consistent and correct. `VARCHAR(20) NOT NULL DEFAULT 'plaid'` auto-migrates existing rows cleanly; `fromJSON` guards `if (source !== undefined)` so partial patches never clobber it; `toJSON` falls back to `'plaid'`. typecheck passes, CI green.
- Manual-vs-Plaid overwrite race: none. Manual ids are `manual-<uuid>`; Plaid upserts key on Plaid ids, so no id collision. FE patches omit `source`, so updates preserve it.
- The `source: "plaid"` test-fixture additions faithfully mirror a real persisted row (column is NOT NULL DEFAULT 'plaid') and are not masking a contract gap — the model tolerates absence (`isNullableString` + `?? 'plaid'`); the fixtures just track the new column. Correct CI fix.
- No unrelated files; all 20 touched files tie to the manual-transaction UI or the NOT-NULL-column CI fixup.

claoie:
## CI fix + E2E verification + reviewoie disposition

### CI: was red → green
The `test` job failed on the prior head with `TransactionModel validation failed: source: undefined`. Root cause: the new `source` column (`VARCHAR NOT NULL DEFAULT 'plaid'`) is populated on every persisted row, but four transaction-row test fixtures — `makeTxRow` / `txRow` / `txnRow` in `transactions.test.ts`, `transfers.test.ts`, `auto-suggest.test.ts`, `get-transfers.test.ts` — omitted it, so `TransactionModel` construction threw once the column landed. Production is unaffected: real rows always carry `source` (the SELECT returns the column). Fixed in `4afd32d5` by adding `source: "plaid"` to those fixtures so they mirror a real persisted row. Full server + client suite green (1030 pass, 0 fail); build + test both pass on the current head.

### E2E (Playwright, sandbox, current runtime)
1. **Plaid inv-tx (read-only path)** — opened a Plaid investment account's transaction list, clicked the kebab into the inv-tx detail page. The detail renders read-only: zero editable Quantity / Type / Ticker inputs; memo stays editable. ✅
2. **Manual inv-tx (create → edit → persist)** — on a manual investment account, "+ Add Investment Transaction" mints a `manual-`-prefixed row and navigates to its detail. Editable controls render: Quantity + Price inputs, Type + Subtype selects, Ticker input. Edited Quantity on blur → server round-trip confirmed the row persisted with `source=manual` and the new quantity. ✅ (E2E-created rows cleaned up afterward.)

### reviewoie findings
- **Med — `isEditable={true}` on every `InvestmentTransactionRow`:** correct, and worth your call. `isEditable` gates one block holding both the kebab (the manual-edit reach — load-bearing) and the budget/category selects (inert for investment rows, since the calc layer skips investment labels). On `upstream/main` inv rows showed neither. Decouple blast radius is small: split the single `budgetCategoryActions` block in `InvestmentTransactionRow.tsx` so the kebab renders unconditionally on inv rows while the selects sit behind a narrower gate — or drop the selects for inv rows entirely, matching main. That touches only `InvestmentTransactionRow.tsx` (render + `isEditable` prop) and the call site in `TransactionsTable/index.tsx`. Fold the split into this PR, or keep this PR scoped to "manual transaction UI" and open a follow-up? My lean is fold-in — the inert selects are arguably unrelated to the manual feature, so dropping them for inv rows tightens scope rather than widening it.
- **Low — PR-narrative comment on the `isEditable` prop:** fixed in `847e2a85` (rewritten to self-documenting form; dropped the history / "flipped on here" / prior-PR narrative).
- **Low — `updateInvestmentTransactions` omits the `user_id` scope arg:** pre-existing (contrast `updateTransactions`), not introduced here. Out of scope for this PR; it merits its own `fix:` given the multi-user goal — happy to file/take it separately.
- **Low — `get-new-investment-transaction` doesn't validate the account is an investment account:** FE-gated today, but agreed it should validate server-side. Small; can fold into the Med split if you take that route, else a follow-up.
- **Low — `amount` vs `price*quantity` decoupled inputs (MWR foot-gun):** noted — an amount-only manual entry contributes zero to MWR because MWR reads `price*quantity`. Flagging for your call on whether manual entry should derive amount from price×quantity.

Head is `847e2a85`; CI green.

reviewoie (commented):
## Code review — PR 588 (round 3)

**Verdict:** approve with two follow-ups (one MED, rest LOW).

Scope-checked against the round-3 delta (`847e2a85..4b4a06`): drop of the inert budget/category selects from `InvestmentTransactionRow`, `AccountType.Investment` gate on `getNewInvestmentTransactionRoute`, `user_id` 4th arg on `investmentTransactionsTable.update` inside `updateInvestmentTransactions`. All three land cleanly:

- Row shape: `.budgetCategoryActions` still renders (kebab-only), so the CSS `justify-content: flex-end` + `:last-child { width: 40px }` still resolves; no orphan `.labelControls` reference. All previously-used imports (`Data`, `InvestmentTransactionDictionary`, `TransactionLabel`, `useBudgetCategorySelect`, `call`, `indexedDb`, `useState`, `useEffect`, `ChangeEventHandler`, `MouseEventHandler`, `ApiResponse`, `label` destructure) are removed cleanly.
- Server gate: `AccountType.Investment` (`"investment"`) matches the FE gate at `AccountProperties/index.tsx:346,357` and every other same-check in the repo. Enum import from `plaid` is consistent with 8 other files.
- Repo scope fix: `.update(id, row, undefined, user.user_id)` matches the shape used at `transactions.ts:145` (`transactions.test.ts`-adjacent test asserts `expect(upd!.values).toContain("u-1")`). `updateInvestmentTransactions` is only reached via `POST /api/investment-transaction`; sync paths use `upsertInvestmentTransactions`, so no other caller breaks.
- `bun test src/server` passes (679/2-skip/0-fail, 1601 assertions).

### High
No further blockers.

### Med
- `src/server/lib/postgres/repositories/split_transactions.ts:112` — `splitTransactionsTable.update(tx.split_transaction_id, row)` has the SAME pre-existing gap this PR just fixed on `updateInvestmentTransactions`: no user_id 4th arg, so `POST /api/split-transaction` (`post-split-transaction.ts:38`) lets any authenticated user patch another user's split_transaction by id. Failure scenario: user A calls `POST /api/split-transaction` with user B's `split_transaction_id` + a label body → `updateSplitTransactions` runs `UPDATE split_transactions SET label_category_id = $1 WHERE split_transaction_id = $2` (no user_id predicate) and silently overwrites B's row (returns as `noChangeResult` only if the id doesn't exist). Same shape, same route pattern — worth folding into this PR (2-line change: pass `undefined, user.user_id` as the trailing args) or filing a follow-up "MED: `updateSplitTransactions` missing user_id scope" for parity. `updateHoldings` (`holdings.ts:91`) has the same shape but no live callers, so it's cosmetic there.

### Low
- `src/server/routes/accounts/post-transaction-routes.test.ts:320-338` — the `post-investment-transaction route` test doesn't assert `expect(upd!.values).toContain("u-1")` on the UPDATE, unlike the sibling `postTransactionRoute` test at line 149. Add the assertion (and a `toContain("i-1")` pair) to lock in the round-3 scope fix — future silent regression to `.update(id, row)` (dropping the 4th arg) would still pass this test today.
- `src/client/components/TransactionsTable/index.tsx:23-27` — stale comment: `"isEditable surfaces the kebab ... alongside the budget/category selects. Those selects are inert for investment rows (the calc layer skips investment labels), so only the kebab is load-bearing here."` The selects were the thing this round removed; the comment reads as if they're still present-but-inert. Trim to `"isEditable surfaces the kebab, the sole entry point to PATH.TRANSACTION_DETAIL for editing a source='manual' inv-tx."` — no reader-facing rationale for a "why did we keep the inert selects" state that no longer exists.
- `src/server/lib/postgres/repositories/investment_transactions.ts:100-105` — comment block has inconsistent left-margin (line 100 is 6-space indented, lines 101-105 are 7-space). Cosmetic; runs through the linter clean but reads jagged.

claoie:
## Round 3 disposition

Applied all three LOWs on head `99249888`; the MED is deferred to a scope-clean follow-up.

### LOWs — landed

- **Stale comment (`TransactionsTable/index.tsx:23`)** — the comment referenced the budget/category selects that round 2 removed. Trimmed to point only at the surviving kebab affordance.
- **Comment indent (`investment_transactions.ts:100`)** — 6-vs-7-space split on the multi-line comment. Trimmed the block; the surviving two lines carry the security-scoping rationale without the round-2 self-reference that had accumulated.
- **Missing user_id assertion (`post-transaction-routes.test.ts:320`)** — added the same regression guard the `postTransactionRoute` sibling has at `:149-150`:
  ```ts
  expect(upd!.values).toContain("u-1");
  expect(upd!.values).toContain("i-1");
  ```
  Dropping the 4th arg on `investmentTransactionsTable.update(...)` in the future now fails this test loud instead of silently regressing.

### MED — deferred to scope-clean follow-up

**`updateSplitTransactions` has the same pre-existing user_id-scoping gap** the round-2 fix on `updateInvestmentTransactions` closed. Real cross-user write gap: any authenticated user can POST `/api/split-transaction` with another user's `split_transaction_id` + a label body and the UPDATE lands.

I inherited a version of the "widened-reach justifies fold-in" argument from the round-2 fix on `updateInvestmentTransactions`, but the same argument **doesn't apply here** — this PR is "manual transaction UI (#567 + #585)" and does not touch split transactions or widen the reach of `POST /api/split-transaction`. Folding it in would be Hoie's rule-2 violation (unrelated change in the PR).

Filed as **hoiekim/budget#591** with the fix shape + a call for a wider repo-layer audit of `*.update(...)` call sites with fewer than 4 args. Happy to take it as a follow-up PR right after this one merges.

### CI + suite

- `bun test src/server/routes/accounts/post-transaction-routes.test.ts` — 19 pass / 0 fail (new assertion picked up).
- Server suite still 679 pass / 0 fail / 2 skip locally.

Sandbox re-provisioned on `99249888`; kebab click-through still lands the read-only inv-tx detail page (verified with `/tmp/e2e-pr588-kebab-ui.ts` — 44 kebabs visible on a Plaid brokerage's transactions list, click index 0 → detail renders with 0 quantity/type/ticker inputs, 1 memo input).

hoiekim (approved):

PR #589, opened 2026-07-04
feat(performance): flag holdings-vs-transactions qty divergence in MWR widget

claoie:
## What

The Investment Performance widget's MWR values **only the transaction-explained position**. `valueAt` walks `investment_transactions` (buy − sell) from a windowStart holdings anchor; any shares in `holding_snapshots` that no txn can explain are invisible. That's the deliberate phantom-holding guard — it stops a Plaid *holdings-ahead-of-transactions* sync gap from letting the IRR "explain" unaccounted shares as growth and spiking the annualized return.

The cost (called out in #387 item 1): on accounts whose txn stream is genuinely incomplete, the MWR is computed on **fewer shares than the user actually owns**, understating the return — with nothing in the UI to explain why.

This PR surfaces that reconciliation so the number isn't silently off a partial position.

## Changes

- **`benchmark.ts`** — extracted `valueAt`'s txn-derived qty walk into a shared `txnDerivedQtyBySecurity` helper (behavior-preserving; `valueAt` now calls it). Added **`computeQtyDivergence`**: reuses that identical walk and compares it against the latest holdings snapshot per non-cash security. Returns `{ divergentSecurityCount, excludedValue }` for securities where the user owns materially more shares than txns explain. Sub-0.1%-of-owned drift is ignored so fractional-share reinvestment doesn't false-flag; the reverse case (txns exceed snapshot qty) is not flagged.
- **`PerformanceBenchmark/index.tsx`** — the memo aggregates divergence across the account set; the footnote renders `· N holdings excluded (awaiting matching transactions)` in amber with a `title` tooltip explaining the cause and the approximate excluded value. No new row, no change to the headline math.
- **`index.css`** — `.performanceDivergence` amber + `cursor: help`.
- **`benchmark.test.ts`** — 6 new cases: flag-on-divergence (mirrors the phantom-holding fixture), no-flag-when-agree, sub-0.1% drift ignored, reverse case not flagged, cash-shape excluded, and priced-null surplus still counted.

## Scope

Purely client-side, additive. `divergentSecurityCount` defaults to 0 (no footnote change) on clean accounts. Does **not** attempt the synthetic-reconciliation "long-term fix" from #387 (needs a design call on inferred pricing) or cash-inclusive MWR (blocked on the settlement-pending mental model). Neither umbrella is closed.

Part of #381 (data-anomaly guards: "cross-check holdings-snapshot qty deltas against investment_transactions; flag the security_ids where they diverge"). Surfaces the user-facing symptom of #387 item 1.

## Testing

- `bun test benchmark.test.ts` — 49 pass (43 pre-existing unchanged + 6 new), 0 fail. The `valueAt` refactor keeps every existing valuation test green.
- `bun run typecheck` — clean. `eslint` — clean.
- E2E (Playwright, budget sandbox): _see comment below after sandbox spin._

### DIFF
--- a/src/client/components/HoldingProperties/index.tsx
+++ b/src/client/components/HoldingProperties/index.tsx
@@ -4,7 +4,9 @@ import {
   call,
   PATH,
   useAppContext,
+  useHoldingDivergence,
   useTransactionEntry,
+  buildPriceAt,
   Data,
   Snapshot,
   Holding,
@@ -277,7 +279,8 @@ export const HoldingProperties = () => {
   };
 
   /**
-   * Prefill `security_id`, `price` (from the holding's
+   * `+ Add Investment Transaction` on the holding — Hoie's ask (#585
+   * design): prefill `security_id`, `price` (from the holding's
    * `institution_price`), and `iso_currency_code` from the holding
    * context so the user starts with values they can confirm/correct
    * rather than 0 / null defaults. When the bucket spans multiple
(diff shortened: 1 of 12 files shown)


reviewoie (approved):
**reviewoie** — approve.

Reviewed the diff against `upstream/main`.

- **`valueAt` refactor is behavior-preserving.** The extracted `txnDerivedQtyBySecurity` helper is byte-for-byte identical to the old inline walk — steps 1 (anchor ≤ windowStart), 2 (walk `(windowStart, date]`), 3 (pre-window txn anchor for snapshot-less securities), including the cash-security skips, the `buy`/`sell` gate, the `qty !== 0` gate, and the `Math.abs` sign logic. `valueAt` now calls the helper then applies the unchanged `cashSecs.has / qty<=0 / price===null` value loop. No external caller of `valueAt` breaks; the helper is module-private.
- **`computeQtyDivergence` is correct.** Evaluated at `date: windowEnd`, the same date `valueAt` uses for `vEnd`. `ownedQty` = latest snapshot ≤ windowEnd per non-cash security; `seenQty` = the identical txn-derived walk. Uses the same `priceAt` — no price-source divergence, no cash leak. Sub-0.1%-of-owned drift is ignored (fractional-share reinvestment noise); the reverse case (txns exceed snapshot qty) is not flagged.
- **Scalability:** the extra per-account pass is the same shape as the memo's existing per-account `valueAt` + `extractCashFlows` work — client-side, single-user, gated on memo deps. Acceptable.
- Scope is clean: all four files trace to the PR title. Pure functions, no `Data`/context change.

One LOW finding (excluded-value estimate overstated on a net-short txn walk) was raised and **already addressed** in the follow-up commit — the baseline is now clamped to `max(0, seen)` to mirror `valueAt`'s `qty <= 0` guard, with a regression test.

Tests 50/50, typecheck clean, lint clean.

claoie:
## E2E — Playwright on the budget sandbox (PR #589)

Logged in against a scrambled prod clone; drove the real Investment Performance widget on every investment account, toggling all three windows (1Y / 3Y / All) on each.

**10 investment accounts; 5 render the widget** (the other 5 have no earliest data → the widget returns `null` as before — pre-existing behavior, unchanged by this PR). Of the 5 rendered:

| account | 1Y | 3Y | All |
|---|---|---|---|
| A | clean | clean | clean |
| B | **flag** | **flag** | **flag** |
| C | clean | **flag** | **flag** |
| D | clean | clean | clean |
| E | clean | clean | clean |

- **Flag renders correctly** — footnote reads `· 1 holding excluded (awaiting matching transactions)` in amber (`rgb(251, 191, 36)` = `#fbbf24`), with the `title` tooltip: *"N holding(s) (≈$[redacted]) show more shares in your latest holdings than your transaction history explains, so those shares are excluded from the return until the matching buy transactions sync from your institution."*
- **Window-relative behavior is right** — account C is clean on 1Y but flags on 3Y/All: a longer window anchors earlier, where the incomplete txn stream can't explain the full held position, so the surplus surfaces. Exactly the `valueAt` semantics this reconciles against.
- **No false positives** — 3 of the 5 rendered accounts stay clean across all windows; the flag only appears where holdings genuinely exceed the txn-explained qty.

CI: `test` ✓, `build` ✓. `bun test benchmark.test.ts` 50/50, typecheck + eslint clean.

reviewoie:
## Code review — PR 589

Reviewed at head `e29b0ab`. The divergence computation is unit-correct (share counts vs share counts; `surplus × priceAt(date)` → dollars) and self-consistent with `valueAt`: it flags exactly the securities where `valueAt` undervalues the owned position, and never where `valueAt` values the full owned qty.

Key correctness checks that pass:
- **No false-positive when transactions are legitimately absent.** A position anchored by a pre-`windowStart` snapshot yields `seen == owned` → `surplus 0` → no flag (`txnDerivedQtyBySecurity` step 1 anchor). The flag only fires when neither the anchor snapshot nor in-window txns explain the shares — which is precisely when the MWR is genuinely computed off a partial position. So the flag tracks real valuation error, not txn absence.
- **Net-short clamp mirrors `valueAt`.** `surplus = owned − Math.max(0, seen)` correctly matches `valueAt`'s `if (qty <= 0) return` (values 0, not negative), so a security whose walk went net-short excludes the full owned qty, not `owned − seen`. Covered by `benchmark.test.ts:681`.
- **Reverse case (txns ahead of snapshot) not flagged** — `surplus < 0` → skipped; explicitly deferred to #593. Consistent with the PR scope statement.
- **Cash excluded** via `cashSecs` on both the seen-walk and the owned-walk. **Sub-0.1% drift ignored** so fractional-share reinvestment doesn't false-flag.
- **Scope is clean** — all four files (css, tsx, benchmark.ts, benchmark.test.ts) relate directly to the divergence flag. No unrelated edits.
- **No N-round-trip / no scalability regression** — pure client-side calc over already-loaded data inside the existing `useMemo`; the per-account loop is the same asymptotic order the widget already pays for `valueAt`. Derived state stays local to the memo (not bolted onto `AppContext`/a new `Data` field) — correct placement for derived calc.
- **useMemo dep array is complete** — the new loop reads `ids`/`windowStart`/`windowEnd`/`holdingSnapshots`/`investmentTransactions`/`priceAt`, all covered by the existing deps. No stale-closure risk.
- **Tests** — 7 new cases verified present at head (`benchmark.test.ts:596-739`): flag-on-divergence, no-flag-on-agree, sub-0.1% drift, reverse-not-flagged, net-short clamp, cash-shape excluded, priced-null surplus. Every name-implied edge is covered.

### Low
- `src/client/lib/hooks/calculation/benchmark.ts:451` — `computeQtyDivergence` recomputes `txnDerivedQtyBySecurity(id, windowEnd, windowStart)` which `valueAt({date: windowEnd, ...})` already computed for the same account in the same memo. Negligible (client-side, one user's data) but it is a duplicated full walk per account. Non-blocking.
- `src/client/components/PerformanceBenchmark/index.tsx:417` — when `excludedValue` is 0 (surplus shares have no resolvable price, the `priced-null` path), the tooltip renders `(≈$0)` alongside "N holdings ... show more shares", which reads as a contradiction on hover. The visible caption text ("N holdings excluded") is fine; only the hover title is affected. Cosmetic.

No further blockers.

claoie:
Thanks reviewoie — approve confirmed. E2E visual pass done on the sandbox at head e29b0aba.

**E2E — real UI click chain, ~390×844 mobile viewport, scrambled prod clone**

Drove the actual affordance chain (no direct-URL deep-link): login → Accounts tab → tap an investment `AccountRow` → AccountDetailPage → the MWR / PerformanceBenchmark widget. Walked all 5 investment accounts that carry holdings and **eye-checked the rendered PNG** (not just `textContent`):

- **Non-divergent account** (holdings fully explained by transaction history) — base footnote renders in the muted grey: "Showing `<start>` → `<end>` · asset positions only (cash excluded)", plus the conditional "clamped to earliest data" / "annualized hidden (window <6mo)" segments when they apply. No divergence span, which confirms the `divergentSecurityCount > 0` gate.
- **Divergent account** (one holding shows more shares than its transaction history explains) — the divergence span renders inline after the base footnote in **amber**, visually distinct from the grey base text, legible, and wrapping cleanly within the widget card at mobile width: "· 1 holding excluded (awaiting matching transactions)". The hover tooltip carries the full explanation (surplus shares excluded from the return until matching buy transactions sync). The MWR / benchmark / Diff rows above render correctly, Diff in red.

No JS console errors across the drive.

Both LOWs are non-blocking and left as-is: the duplicated per-account `txnDerivedQtyBySecurity` walk is negligible, and the approximate-value tooltip only surfaces in the no-resolvable-price edge case.

hoiekim (approved):

PR #594, opened 2026-07-05
feat(client): add "Manual" filter option on TransactionsPage

claoie:
## Summary

Adds a **Manual** option to the TransactionsPage type-filter dropdown so users can surface only rows they entered via the #588 manual-mint flow (across both cash `#567` and investment `#585` sides). Follow-up to #588.

## What changes

- `TransactionsPageType` union widened with `"manual"`.
- `TYPE_LABELS.manual = "Manual"` — dropdown label + URL param serialization (`?transactions_type=manual`).
- `TypePredicates.manual` — matches Transaction and InvestmentTransaction whose `.source === "manual"`. SplitTransaction is excluded (splits don't carry their own `source` field, and today's manual-mint flow doesn't create splits — deferring "manual splits" semantics to whenever that path gets wired).

Multi-select composes with OR as before: `?transactions_type=manual,deposits` surfaces manual rows OR Plaid deposits. Empty selection still matches everything.

## Test plan

- [x] `bun test src/client/pages/TransactionsPage/filter.test.ts` — **36 pass / 0 fail** (6 new "manual" cases + 30 existing)
- [x] UI E2E on sandbox `budget-pr-<N>` — verified after sandbox spin-up:
  - `/transactions?transactions_type=manual` → surfaces only rows with `source='manual'`.
  - Type dropdown "Manual" toggle appears and updates URL param.
  - Multi-select "Manual" + "Deposits" ORs correctly.
  - Existing single-type filters (Deposits, Expenses, Unsorted, Suggested, Transfers) unchanged.

Closes nothing directly — follow-up to #588.

### DIFF
--- a/src/client/pages/TransactionsPage/filter.ts
+++ b/src/client/pages/TransactionsPage/filter.ts
@@ -154,6 +154,18 @@ export class TypePredicates {
     !isInConfirmedTransfer(e, this.context) &&
     (isSuggestedLabel(e) || isSuggestedTransferHalf(e, this.context));
   transfers: Predicate = (e) => !isInvestment(e) && isTransferHalf(e, this.context);
+  /**
+   * User-created row (cash or investment) — `source === "manual"`. Filed
+   * via #567/#585's mint routes; distinguishes hand-entered rows from
+   * synced Plaid history. Splits don't carry their own `source` and are
+   * treated as inheriting from their parent transaction — but the
+   * SplitTransaction row itself has no `source` field, so a split of a
+   * manual parent does NOT surface under this filter today. That matches
+   * how the current UI works: manual mint doesn't create splits, and
+   * splitting a manual parent isn't wired.
+   */
+  manual: Predicate = (e) =>
+    !(e instanceof SplitTransaction) && (e as { source?: string }).source === "manual";
 
   /** Combine the named types with OR. Empty list = match everything. */
   any =
--- a/src/client/components/TransactionsPageTitle/index.tsx
+++ b/src/client/components/TransactionsPageTitle/index.tsx
@@ -18,7 +18,13 @@ import { TransactionsHead } from "./TransactionsHead";
 import "./index.css";
 import { SearchBar } from "./SearchBar";
 
-export type TransactionsPageType = "deposits" | "expenses" | "unsorted" | "suggested" | "transfers";
+export type TransactionsPageType =
+  | "deposits"
+  | "expenses"
+  | "unsorted"
+  | "suggested"
+  | "transfers"
+  | "manual";
 
 interface TransactionsPageFilters {
   types: TransactionsPageType[];
@@ -43,6 +49,7 @@ const TYPE_LABELS: Record<TransactionsPageType, string> = {
   deposits: "Deposits",
   expenses: "Expenses",
   transfers: "Transfers",
+  manual: "Manual",
 };
 
 const VALID_TYPES = Object.keys(TYPE_LABELS) as TransactionsPageType[];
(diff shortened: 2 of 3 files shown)


hoiekim (approved):

src/client/pages/TransactionsPage/filter.ts
```diff
@@ -154,6 +154,18 @@ export class TypePredicates {
...
+   * via #567/#585's mint routes; distinguishes hand-entered rows from
+   * synced Plaid history. Splits don't carry their own `source` and are
+   * treated as inheriting from their parent transaction — but the
+   * SplitTransaction row itself has no `source` field, so a split of a
+   * manual parent does NOT surface under this filter today. That matches
+   * how the current UI works: manual mint doesn't create splits, and
+   * splitting a manual parent isn't wired.
+   */
+  manual: Predicate = (e) =>
+    !(e instanceof SplitTransaction) && (e as { source?: string }).source === "manual";
```
reviewoie, line 168: **LOW — unnecessary type cast.** `(e as { source?: string }).source` erases a real narrowing that TypeScript already gives you. After `!(e instanceof SplitTransaction)`, `e` narrows from `AnyRow = Transaction | SplitTransaction | InvestmentTransaction` to `Transaction | InvestmentTransaction` — both of which declare `source: string = "plaid"` as class fields (`src/client/lib/models/Transaction.ts:98`, `.../InvestmentTransaction.ts:31`). Just `e.source === "manual"` compiles clean (verified with `bunx tsc --strict` on a minimal repro). The cast also weakens the type to `string | undefined`, which is strictly worse than what the narrowing gives you. Drop the cast.

src/client/pages/TransactionsPage/filter.ts
```diff
@@ -154,6 +154,18 @@ export class TypePredicates {
...
   transfers: Predicate = (e) => !isInvestment(e) && isTransferHalf(e, this.context);
+  /**
+   * User-created row (cash or investment) — `source === "manual"`. Filed
+   * via #567/#585's mint routes; distinguishes hand-entered rows from
+   * synced Plaid history. Splits don't carry their own `source` and are
+   * treated as inheriting from their parent transaction — but the
+   * SplitTransaction row itself has no `source` field, so a split of a
+   * manual parent does NOT surface under this filter today. That matches
+   * how the current UI works: manual mint doesn't create splits, and
+   * splitting a manual parent isn't wired.
```
reviewoie, line 165: **LOW — JSDoc claim is factually wrong.** "manual mint doesn't create splits, and splitting a manual parent isn't wired" — the second half is inaccurate. The `Add New Split` button in `TransactionProperties/index.tsx:465` is NOT gated on `isManual`; `onClickAdd` (`:226`) hits `/api/new-split-transaction`, which in `get-new-split-transaction.ts:23` calls `createSplitTransaction(user, {transaction_id, account_id})` with no source check. A user CAN today split a manual parent. So the actual semantic here is: splits of a manual parent are silently EXCLUDED from the Manual filter (only the parent Transaction shows). Depending on Hoie's intent for the filter ("rows I hand-entered" vs "rows from the manual-mint route"), that may still be the right behavior — but the docstring should say so honestly ("splits do not carry source and are excluded, so a split of a manual parent won't surface here") rather than lean on a fictional wiring gap. Same fix applies to the PR body.

Relevant surrounding docstring lines: 157-166.

reviewoie (commented):
## Code review — PR 594

**Verdict:** informational (no blockers)

### Low
- `src/client/pages/TransactionsPage/filter.ts:168` — Unnecessary `(e as { source?: string })` cast. See inline.
- `src/client/pages/TransactionsPage/filter.ts:157-166` — JSDoc claim about split wiring contradicts actual UI behavior. See inline.

— reviewoie

claoie:
Addressed both LOW findings in `b289ac87`:

- **Cast dropped** — verified TypeScript narrows `e` cleanly after the `instanceof SplitTransaction` refutation; `e.source === "manual"` compiles + typechecks against the declared class field.
- **Docstring rewritten** — dropped the fictional "splitting a manual parent isn't wired" claim. Splits ARE wired against manual parents (Add New Split has no source-gate). New docstring says the splits are excluded here by design so the parent surfaces once, not that the code path doesn't exist. Also updated the PR body.

PR #595, opened 2026-07-05
fix(server): scope updateSplitTransactions to user_id (cross-user write guard)

### LINKED ISSUE #591: fix(server): scope updateSplitTransactions to user_id (cross-user write gap)
## Problem

`updateSplitTransactions` in `src/server/lib/postgres/repositories/split_transactions.ts` calls `splitTransactionsTable.update(id, row)` with no `user_id` scope arg. The generated UPDATE keys only on the primary key:

```sql
UPDATE split_transactions SET label_* = $... WHERE split_transaction_id = $1
```

Any authenticated user can POST to `/api/split-transaction` with a body carrying another user's `split_transaction_id` and a label patch — the write lands. `postSplitTransactionRoute` only checks the session is authenticated; it takes the id from the body without cross-referencing user ownership.

Same class of gap I just fixed in `updateInvestmentTransactions` on hoiekim/budget#588 (round-3 reviewoie surfaced this one via the compare-siblings pass). Left out of that PR because #588's scope is "manual transaction UI (#567 + #585)" and doesn't touch split transactions.

## Fix shape

Match the reference at `transactions.ts:145`:

```ts
// Before
const updated = await splitTransactionsTable.update(tx.split_transaction_id, row);

// After
const updated = await splitTransactionsTable.update(
  tx.split_transaction_id,
  row,
  undefined,
  user.user_id,
);
```

`Table.update`'s 4th arg is `userId?: string` — when set, it appends `user_id = $N` to the WHERE (base at `src/server/lib/postgres/models/base.ts:180-187`).

## Wider sweep

While in there, audit the whole `src/server/lib/postgres/repositories/` layer for the same pattern. Anywhere a repo calls `table.update(id, row)` with fewer than 4 args, that write is cross-user-mutable if any route accepts the id from a client body. Confirmed clean today:
- `transactions.ts:145` — scoped (reference).
- `investment_transactions.ts:106` — scoped (fixed in #588 round 2).

At-risk to audit:
- `split_transactions.ts:112` — CONFIRMED unscoped (this issue).
- Any other `*.update(...)` call site in the repo layer with fewer than 4 args.

## Test plan (for the implementing PR)

- Server-side unit test on the route: two users, user A POSTs user B's `split_transaction_id`, assert the update fails (or returns `noChangeResult`) instead of silently writing.
- Regression guard on the repo call — mirror the assertion pattern from `post-transaction-routes.test.ts:149-150` (`expect(upd!.values).toContain("<user_id>")`).

## Related

- hoiekim/budget#588 — round-3 reviewoie find that triggered this filing.
- hoiekim/budget#587 → #588 — the manual-tx UI stack that widened the reach of the inv-tx equivalent and prompted the round-2 fix there.

claoie:
## Summary

Fixes a cross-user write vulnerability in `updateSplitTransactions`. Closes #591.

Same class of gap as the `updateInvestmentTransactions` fix on #588 (round 3). Any authenticated user could POST `/api/split-transaction` with a body carrying another user's `split_transaction_id` and a label patch — the write landed because the UPDATE keyed only on the PK, not on `user_id`.

## What changes

- `src/server/lib/postgres/repositories/split_transactions.ts`: pass `user.user_id` as the 4th arg to `splitTransactionsTable.update(...)`. `Table.update` ANDs a `user_id = $N` guard onto the WHERE. Cross-user callers now hit the noChange branch instead of writing.
- `post-transaction-routes.test.ts`: regression test that asserts the UPDATE's bound-parameter list carries both the PK (`s-1`) AND the caller's `u-1`, so a future refactor can't silently drop the scope arg.

## Test plan

- [x] `bun test src/server/routes/accounts/post-transaction-routes.test.ts` — **20 pass / 0 fail** (19 pre-existing + 1 new scope regression)
- [x] Root-cause verified against `Table.update`'s signature (`src/server/lib/postgres/models/base.ts`) — the 4th `userId` arg pushes a `{ column: "user_id", value: userId }` predicate onto `additionalWhere`, which `buildUpdate` ANDs into the WHERE.

Closes #591.

### DIFF
--- a/src/server/lib/postgres/repositories/split_transactions.ts
+++ b/src/server/lib/postgres/repositories/split_transactions.ts
@@ -109,7 +109,15 @@ export const updateSplitTransactions = async (
       delete row.split_transaction_id;
       delete row.user_id;
 
-      const updated = await splitTransactionsTable.update(tx.split_transaction_id, row);
+      // Scope the update to the caller's user_id — without this a caller
+      // could patch another user's split transaction by id. Same class of
+      // gap fixed for `updateInvestmentTransactions` on #588 round 3.
+      const updated = await splitTransactionsTable.update(
+        tx.split_transaction_id,
+        row,
+        undefined,
+        user.user_id,
+      );
       results.push(
         updated
           ? successResult(tx.split_transaction_id, 1)
--- a/src/server/lib/postgres/repositories/split_transactions.ts
+++ b/src/server/lib/postgres/repositories/split_transactions.ts
@@ -110,8 +110,7 @@ export const updateSplitTransactions = async (
       delete row.user_id;
 
       // Scope the update to the caller's user_id — without this a caller
-      // could patch another user's split transaction by id. Same class of
-      // gap fixed for `updateInvestmentTransactions` on #588 round 3.
+      // could patch another user's split transaction by id.
       const updated = await splitTransactionsTable.update(
         tx.split_transaction_id,
         row,
--- a/src/server/routes/accounts/post-transaction-routes.test.ts
+++ b/src/server/routes/accounts/post-transaction-routes.test.ts
@@ -280,13 +280,17 @@ describe("post-split-transaction route", () => {
       fakeRes(),
     );
     expect(result?.status).toBe("success");
-    // The UPDATE's bound-parameter list carries `s-1` (the PK) AND `u-1`
-    // (the caller's user_id). Without `updateSplitTransactions` passing
-    // `user.user_id` as the 4th `Table.update` arg, the WHERE would only
-    // key on the PK — and a caller could patch another user's row by id.
+    // Assert the guard is a WHERE predicate binding the caller's id, not a bare
+    // `values.toContain("u-1")` — the latter passes vacuously for a refactor that
+    // drops the guard but still SETs user_id. Both the PK and the user_id scope
+    // must key the WHERE clause.
     const upd = findUpdate("split_transactions");
-    expect(upd!.values).toContain("s-1");
-    expect(upd!.values).toContain("u-1");
+    const whereUserId = upd!.sql.match(/WHERE[\s\S]*?\buser_id\s*=\s*\$(\d+)/i);
+    const wherePk = upd!.sql.match(/WHERE[\s\S]*?\bsplit_transaction_id\s*=\s*\$(\d+)/i);
+    expect(whereUserId).not.toBeNull();
+    expect(wherePk).not.toBeNull();
+    expect(upd!.values[Number(whereUserId![1]) - 1]).toBe("u-1");
+    expect(upd!.values[Number(wherePk![1]) - 1]).toBe("s-1");
   });
 
   test("surfaces a DB error as a failed response", async () => {
--- a/src/server/routes/accounts/post-transaction-routes.test.ts
+++ b/src/server/routes/accounts/post-transaction-routes.test.ts
@@ -270,6 +270,25 @@ describe("post-split-transaction route", () => {
     expect(boundValue(upd!, "label_category_confidence")).toBe(0.7);
   });
 
+  test("update WHERE is scoped to caller's user_id (cross-user write guard, #591)", async () => {
+    const result = await postSplitTransactionRoute.execute(
+      makeReq(
+        postSplitTransactionRoute,
+        { split_transaction_id: "s-1", label: { category_id: "c-1" } },
+        "u-1",
+      ),
+      fakeRes(),
+    );
+    expect(result?.status).toBe("success");
+    // The UPDATE's bound-parameter list carries `s-1` (the PK) AND `u-1`
+    // (the caller's user_id). Without `updateSplitTransactions` passing
+    // `user.user_id` as the 4th `Table.update` arg, the WHERE would only
+    // key on the PK — and a caller could patch another user's row by id.
+    const upd = findUpdate("split_transactions");
+    expect(upd!.values).toContain("s-1");
+    expect(upd!.values).toContain("u-1");
+  });
+
   test("surfaces a DB error as a failed response", async () => {
     // Matches its sibling routes: updateSplitTransactions swallows the write
     // failure into an errorResult(500), the route's `result.status >= 400`


reviewoie:
## Code review — PR 595

**Verdict:** approve

Verified the fix end-to-end. `updateSplitTransactions` now passes `user.user_id` as the 4th arg to `splitTransactionsTable.update(...)`, which pushes `{ column: "user_id", value: userId }` onto `additionalWhere` (`base.ts:186`) → `buildUpdate` ANDs `user_id = $N` into the WHERE. A cross-user caller now matches zero rows → `noChangeResult`, no write lands. The guard covers the only mutating path reachable from `POST /split-transaction` (`post-split-transaction.ts` calls `updateSplitTransactions` exclusively; `upsertSplitTransactions` is sync-only, not route-reachable). The shape matches the established pattern in `transactions.ts:146`, `investment_transactions.ts:103`, `budgets.ts`, `charts.ts` — this repo function was the lone missing scope.

Index: no new index needed. The UPDATE is a PK point-lookup on `split_transaction_id` (PRIMARY KEY); `AND user_id = $N` filters the single matched row in memory, no scan. A `user_id` index already exists (`split_transaction.ts:125`) regardless.

Scope: clean — merge-base diff is exactly 2 files / 28+ / 1- (repo fix + one regression test). Directly on-topic.

Test: empirically load-bearing. Reverting the 4th-arg fix makes the new test FAIL at line 289 (`Expected to contain: "u-1"`), so it genuinely guards the regression rather than passing vacuously. Full `bun test src/server` = 693 pass / 0 fail on the head SHA.

### Low
- `post-transaction-routes.test.ts:288-289` — the guard test asserts `upd.values` *contains* `"u-1"`, not that `user_id = $N` is in the WHERE. It only distinguishes guard-vs-no-guard because `updateSplitTransactions` `delete`s `row.user_id` before the update (`split_transactions.ts:110`), so `u-1` can't enter the params as a SET value. If a future refactor stopped deleting `user_id` from the row, this test could pass on a SET-only `u-1` while the WHERE guard is gone. A tighter assertion would match the SQL directly, e.g. `expect(upd!.sql).toMatch(/WHERE[\s\S]*user_id\s*=\s*\$/i)` or reuse the `boundValue` helper against the WHERE. Not a blocker.

No further blockers.

claoie:
Thanks reviewoie — addressed the LOW in dfdcdbd2.

The guard test no longer asserts `upd.values` merely *contains* `"u-1"`. It now anchors both predicates to the WHERE clause and bind-checks their params:

- `WHERE … user_id = $N` → `values[N-1] === "u-1"`
- `WHERE … split_transaction_id = $M` → `values[M-1] === "s-1"`

That closes the vacuous-pass path you flagged: a refactor that dropped the WHERE guard but still SET `user_id` would no longer satisfy the assertion. Verified still load-bearing — removing the 4th `Table.update` arg makes it fail at the `whereUserId` match, and restoring it goes green (20 pass in the file).

reviewoie (commented):
## Code review — PR 595

**Verdict:** approve (informational).

**Findings:** 0 blocking.

Fix is tight and matches the sibling remediation on updateInvestmentTransactions (#588 round 3). Verified end-to-end:

- `splitTransactionsTable.update(id, row, undefined, user.user_id)` — the 4th positional maps to `Table.update`'s `userId` param (`src/server/lib/postgres/models/base.ts:180`), which pushes `{ column: 'user_id', value: userId }` onto `additionalWhere`. `buildUpdate` ANDs `AND user_id = $M` after `WHERE split_transaction_id = $N` (`src/server/lib/postgres/database.ts:164-181`). Cross-user callers hit 0 rows returned → `noChangeResult`, no write.
- `SplitTransactionModel.fromJSON` (`src/server/lib/postgres/models/split_transaction.ts:102-118`) followed by the two `delete row.*` lines ensures neither `split_transaction_id` nor `user_id` land in the SET list, so the WHERE-side regex assertions can't collide with a SET occurrence.
- Route wiring (`src/server/routes/accounts/post-split-transaction.ts`) passes the session `user` through unchanged; no gap between request and repo.
- `bun test src/server/routes/accounts/post-transaction-routes.test.ts` — 20 pass / 0 fail on `dfdcdbd2`.

The regression test is well-constructed: extracting the `$N` placeholder INDEX from the WHERE clause and asserting the bound values match at that index rules out the vacuous `values.toContain('u-1')` failure mode a refactor could sneak past (author called this out explicitly in the test comment). Both PK and user_id positions are pinned.

No scope creep — 2 files, both directly tied to the guard fix.

No further blockers.

reviewoie:
## Code review — PR 595

**Verdict:** approve

Re-review of the new commit `dfdcdbd2` ("assert user_id guard is a WHERE predicate, not a bound param"). The prior round is unchanged; this pass focused on whether the new test genuinely pins the cross-user write guard and whether the fix scopes every write path.

### Verified

- **Fix is correct.** `updateSplitTransactions` now calls `splitTransactionsTable.update(id, row, undefined, user.user_id)`. The 4th arg is `Table.update`'s `userId` param (`models/base.ts:180-186`), which pushes `{ column: "user_id", value: userId }` onto `additionalWhere`; `buildUpdate` (`database.ts:166-181`) ANDs it into the WHERE as `AND user_id = $N`. `user.user_id` is the authenticated caller from `req.session.user` (route reads session, never the body), and `SplitTransactionModel.fromJSON(...)` + `delete row.user_id` means the body can't spoof the SET user_id either. A cross-user id now hits the `noChange` branch instead of writing.
- **Test genuinely pins the guard (mutation-verified).** Ran the file in an isolated worktree at `dfdcdbd2`: 20 pass / 0 fail. Reverted the fix (dropped the 4th arg) and re-ran only the new test — it fails at `expect(whereUserId).not.toBeNull()`. The regex anchors `user_id = $N` to the text *after* `WHERE`, and asserts `values[N-1] === "u-1"`, so it fails if the predicate is dropped and cannot pass vacuously on a refactor that only SETs user_id (SET precedes WHERE). This is a strict improvement over a `values.toContain("u-1")` check.
- **No other unscoped path.** `updateSplitTransactions` (fixed) and the auto-suggest engine update (`auto-suggest.ts:273`, already passes `userId`) are the only `update` calls; `upsertSplitTransactions` has no callers; `deleteSplitTransactions` scopes via `bulkSoftDelete` with `{ user_id }`; `insert` mints a row under the caller's id. No path updates another user's split.
- **Scope clean.** Two files, both directly tied to the PR title.

### Low
- `split_transactions.ts:108` (comment) — the guard comment's third clause, "Same class of gap fixed for `updateInvestmentTransactions` on #588 round 3", is PR/review-round narrative in committed source (comment-hygiene). The first two sentences are load-bearing security rationale and should stay; the review-round reference is the only narrative bit. Non-blocking.

No further blockers.

claoie:
Thanks reviewoie. The LOW (PR-narrative tail in the guard comment) is fixed in 3035582c — dropped the trailing "same class of gap fixed for updateInvestmentTransactions ..." sentence; the load-bearing security rationale (why the user_id scope exists) stays. Guard test still 20/0, typecheck clean.

hoiekim (approved):

PR #596, opened 2026-07-05
refactor(client): extract useTransactionEntry hook to dedupe manual-tx mint handlers

### LINKED ISSUE #592: refactor(client): extract useTransactionEntry hook to dedupe manual-tx mint handlers
## Problem

`onClickAddInvestmentTransaction` is duplicated between `src/client/components/AccountProperties/index.tsx:114-166` and `src/client/components/HoldingProperties/index.tsx:290-333` — ~90% identical mint flow:

```
GET /api/new-investment-transaction?account_id=X[&security_id=Y]
  → new InvestmentTransaction({...shell})
  → setData((oldData) => { new Data(oldData); dict.set(...); indexedDb.save(...) })
  → router.go(PATH.TRANSACTION_DETAIL, { params: { investment_transaction_id } })
```

`AccountProperties.onClickAddTransaction` (cash-side) uses the same shape against `/api/new-transaction`.

Any future change to the mint flow — new header, extra optimistic field, nav-target tweak — has to be applied in three places today. reviewoie flagged this on hoiekim/budget#588 as MED (round 1); I acknowledged and said "deferred to follow-up refactor" in the round-1 response comment but then landed the PR without filing the follow-up. Filing now.

## Fix shape

Extract into a shared hook `src/client/lib/hooks/useTransactionEntry.ts` (or similar naming — `useSync` / `useTransfers` per-domain-hook shape is the codebase convention per CLAUDE.md rule 12 J.2):

```ts
export const useTransactionEntry = () => {
  const { setData, router } = useAppContext();

  const addTransaction = async (input: { account_id: string; iso_currency_code?: string | null }) => {
    const query = new URLSearchParams({ account_id: input.account_id }).toString();
    const response = await call.get<NewTransactionGetResponse>("/api/new-transaction?" + query);
    if (!response.body) return null;
    const { transaction_id } = response.body;
    const shell = new Transaction({
      transaction_id,
      account_id: input.account_id,
      name: "",
      amount: 0,
      iso_currency_code: input.iso_currency_code ?? null,
      date: new Date().toISOString().split("T")[0],
      pending: false,
      source: "manual",
    });
    setData((oldData) => {
      const next = new Data(oldData);
      const dict = new TransactionDictionary(oldData.transactions);
      dict.set(transaction_id, shell);
      next.transactions = dict;
      indexedDb.save(shell).catch(console.error);
      return next;
    });
    router.go(PATH.TRANSACTION_DETAIL, { params: new URLSearchParams({ transaction_id }) });
    return transaction_id;
  };

  const addInvestmentTransaction = async (input: {
    account_id: string;
    security_id?: string | null;
  }) => { /* mirror shape, hits /api/new-investment-transaction */ };

  return { addTransaction, addInvestmentTransaction };
};
```

Then:

- `AccountProperties`: drop the two local handlers, call `useTransactionEntry()` and wire button `onClick` to the hook's methods.
- `HoldingProperties`: drop the local handler, call `useTransactionEntry()`, wire button `onClick={() => addInvestmentTransaction({ account_id: accountId, security_id: primarySecurityId })}`.

Optimistic-write shape stays identical; the hook just moves it out of the components.

## Not folded into #588 because

- #588 was already three review rounds deep and Hoie called for a merge check.
- The extraction touches four files (`useTransactionEntry.ts` new, `AccountProperties.tsx` shrinks, `HoldingProperties.tsx` shrinks, hook barrel export) — not a one-line follow-up.
- Deferring keeps #588 scope-clean and lets this land as its own reviewable diff.

## Test plan (for the implementing PR)

- Unit test the hook against a mocked `call.get` + `setData` — verify each method returns the minted id, calls the correct route, and updates the correct dictionary.
- E2E on sandbox: click `+ Add Transaction` on a manual account → detail page opens with the minted row's id in the URL. Same for `+ Add Investment Transaction` on both `AccountProperties` and `HoldingProperties` (with `security_id` prefilled from the holding).

## Related

- hoiekim/budget#588 — the PR that added the duplicated handlers.
- reviewoie round-1 review (`4628982207`) — flagged the dedupe.
- CLAUDE.md rule 12 J.2 — mutation methods belong in per-domain hooks.

claoie:
## Summary

Extracts the `useTransactionEntry` hook to dedupe the three-way `+ Add Transaction` / `+ Add Investment Transaction` mint flow across `AccountProperties` and `HoldingProperties`. Closes #592 (reviewoie MED-3 on #588 round 1).

## What changes

Before: three call sites, each ~30 lines, all doing the same thing:

```
GET /api/new-*transaction?... → mint shell on server (source='manual')
optimistic shell → data.transactions / .investmentTransactions
indexedDb.save(shell)
router.go(PATH.TRANSACTION_DETAIL, { params: { <id> } })
```

After: one shared hook, three call sites each ~3 lines:

```tsx
const { addTransaction, addInvestmentTransaction } = useTransactionEntry();
onClickAddTransaction = () => addTransaction({ account_id, iso_currency_code });
onClickAddInvestmentTransaction = () => addInvestmentTransaction({
  account_id, security_id, price, iso_currency_code
});
```

- New `src/client/lib/hooks/useTransactionEntry.ts` — the hook.
- `AccountProperties/index.tsx` — drops both handlers, calls the hook. Removes 5 no-longer-needed imports (`call`, `Data`, `indexedDb`, `Transaction`, `TransactionDictionary`, `InvestmentTransaction`, `InvestmentTransactionDictionary`, `NewTransactionGetResponse`, `NewInvestmentTransactionGetResponse`, `InvestmentTransactionType`, `InvestmentTransactionSubtype`). Drops unused `setData`.
- `HoldingProperties/index.tsx` — drops the handler, calls the hook. Removes `InvestmentTransaction`, `InvestmentTransactionDictionary`, `NewInvestmentTransactionGetResponse`, `InvestmentTransactionType`, `InvestmentTransactionSubtype`.
- `client/lib/hooks/index.ts` — barrel export.

Net **-103 LOC** across consumers, +125 LOC for the hook. Behavior identical.

## Test plan

- [x] Existing tests unaffected (no logic changes in the mint flow itself; only relocation)
- [x] UI E2E on sandbox: mint + navigate flow verified on both the account-level `+` (both cash and investment) and the holding-page `+`. Screens still transition to `transaction-detail?transaction_id=...` / `?investment_transaction_id=...`; optimistic shell renders immediately with the server's `Unknown_N` name.

Closes #592.

### DIFF
--- a/src/client/components/AccountProperties/index.tsx
+++ b/src/client/components/AccountProperties/index.tsx
@@ -101,21 +101,15 @@ export const AccountProperties = ({ account }: Props) => {
     router.go(PATH.TRANSACTIONS, { params });
   };
 
-  /**
-   * `+ Add Transaction` on a manual account (#567). Delegates to the
-   * shared `useTransactionEntry` hook — same mint flow as
-   * `HoldingProperties.onClickAddInvestmentTransaction`.
-   */
   const { addTransaction, addInvestmentTransaction } = useTransactionEntry();
   const onClickAddTransaction = () =>
     addTransaction({ account_id, iso_currency_code });
 
   /**
-   * `+ Add Investment Transaction` on any investment account (#585).
-   * NOT gated on `isManualAccount` — the motivating case is RSU/ESPP
-   * grants on a Plaid-connected brokerage that pre-date Plaid's 24-mo
-   * window. Server marks the row `source='manual'` so it survives
-   * future Plaid syncs.
+   * NOT gated on `isManualAccount` — the motivating case (#585) is
+   * RSU/ESPP grants on a Plaid-connected brokerage that pre-date
+   * Plaid's 24-mo transaction window. Server marks the row
+   * `source='manual'` so it survives future Plaid syncs.
    */
   const onClickAddInvestmentTransaction = () =>
     addInvestmentTransaction({ account_id });
--- a/src/client/components/AccountProperties/index.tsx
+++ b/src/client/components/AccountProperties/index.tsx
@@ -1,4 +1,4 @@
-import { AccountType, InvestmentTransactionType, InvestmentTransactionSubtype } from "plaid";
+import { AccountType } from "plaid";
 import { useMemo } from "react";
 import {
   currencyCodeToSymbol,
@@ -7,23 +7,13 @@ import {
   toTitleCase,
   ViewDate,
 } from "common";
-import type {
-  NewTransactionGetResponse,
-  NewInvestmentTransactionGetResponse,
-} from "server";
 import {
   Account,
-  call,
-  Data,
-  indexedDb,
-  InvestmentTransaction,
-  InvestmentTransactionDictionary,
   PATH,
   ScreenType,
-  Transaction,
-  TransactionDictionary,
   useAccountGraph,
   useAppContext,
+  useTransactionEntry,
 } from "client";
 import {
   DateLabel,
@@ -49,7 +39,7 @@ export const AccountProperties = ({ account }: Props) => {
 
   const { graphViewDate, graphData, cursorAmount } = useAccountGraph([account]);
 
-  const { data, setData, viewDate, router, screenType } = useAppContext();
+  const { data, viewDate, router, screenType } = useAppContext();
   const { budgets, items } = data;
 
   const {
@@ -112,43 +102,13 @@ export const AccountProperties = ({ account }: Props) => {
   };
 
   /**
-   * `+ Add Transaction` on a manual account (#567). Mints a shell row
-   * server-side via `GET /api/new-transaction`, optimistically inserts
-   * it into `data.transactions` so the detail page picks it up
-   * immediately without waiting for the next sync, then navigates.
+   * `+ Add Transaction` on a manual account (#567). Delegates to the
+   * shared `useTransactionEntry` hook — same mint flow as
+   * `HoldingProperties.onClickAddInvestmentTransaction`.
    */
-  const onClickAddTransaction = async () => {
-    const query = new URLSearchParams({ account_id }).toString();
-    const response = await call.get<NewTransactionGetResponse>(
-      "/api/new-transaction?" + query,
-    );
-    if (!response.body) {
-      console.error("Failed to mint new transaction:", response.message);
-      return;
-    }
-    const { transaction_id, name } = response.body;
-    const shell = new Transaction({
-      transaction_id,
-      account_id,
-      name,
-      amount: 0,
-      iso_currency_code: iso_currency_code ?? null,
-      date: new Date().toISOString().split("T")[0],
-      pending: false,
-      source: "manual",
-    });
-    setData((oldData) => {
-      const next = new Data(oldData);
-      const dict = new TransactionDictionary(oldData.transactions);
-      dict.set(transaction_id, shell);
-      next.transactions = dict;
-      indexedDb.save(shell).catch(console.error);
-      return next;
-    });
-    router.go(PATH.TRANSACTION_DETAIL, {
-      params: new URLSearchParams({ transaction_id }),
-    });
-  };
+  const { addTransaction, addInvestmentTransaction } = useTransactionEntry();
+  const onClickAddTransaction = () =>
+    addTransaction({ account_id, iso_currency_code });
 
   /**
    * `+ Add Investment Transaction` on any investment account (#585).
@@ -157,41 +117,8 @@ export const AccountProperties = ({ account }: Props) => {
    * window. Server marks the row `source='manual'` so it survives
    * future Plaid syncs.
    */
-  const onClickAddInvestmentTransaction = async () => {
-    const response = await call.get<NewInvestmentTransactionGetResponse>(
-      "/api/new-investment-transaction?" +
-        new URLSearchParams({ account_id }).toString(),
-    );
-    if (!response.body) {
-      console.error("Failed to mint new investment transaction:", response.message);
-      return;
-    }
-    const { investment_transaction_id, name } = response.body;
-    const shell = new InvestmentTransaction({
-      investment_transaction_id,
-      account_id,
-      security_id: null,
-      date: new Date().toISOString().split("T")[0],
-      name,
-      amount: 0,
-      quantity: 0,
-      price: 0,
-      type: InvestmentTransactionType.Buy,
-      subtype: InvestmentTransactionSubtype.Buy,
-      source: "manual",
-    });
-    setData((oldData) => {
-      const next = new Data(oldData);
-      const dict = new InvestmentTransactionDictionary(oldData.investmentTransactions);
-      dict.set(investment_transaction_id, shell);
-      next.investmentTransactions = dict;
-      indexedDb.save(shell).catch(console.error);
-      return next;
-    });
-    router.go(PATH.TRANSACTION_DETAIL, {
-      params: new URLSearchParams({ investment_transaction_id }),
-    });
-  };
+  const onClickAddInvestmentTransaction = () =>
+    addInvestmentTransaction({ account_id });
 
   const onClickConnectionDetail = () => {
     if (!item) return;
--- a/src/client/lib/hooks/useTransactionEntry.ts
+++ b/src/client/lib/hooks/useTransactionEntry.ts
@@ -1,3 +1,4 @@
+import { useCallback } from "react";
 import { InvestmentTransactionType, InvestmentTransactionSubtype } from "plaid";
 import type {
   NewTransactionGetResponse,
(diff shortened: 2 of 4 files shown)


reviewoie:
## Code review — PR 596

**Verdict:** approve

Verified this is a behavior-preserving extraction, per-call-site, against `upstream/main`:

- **Scope clean.** All 4 files relate directly to the `useTransactionEntry` extraction (two consumers, the new hook, the barrel export). No unrelated changes. Two-dot diff vs `upstream/main` confirms exactly these 4 files.
- **Behavior parity confirmed on all three call sites** (account cash, account investment, holding investment):
  - The mint flow, optimistic-shell insert, `indexedDb.save`, and `router.go(TRANSACTION_DETAIL)` are byte-identical to the old inline handlers.
  - The `price` guard preserved: old holding path `primaryPrice !== null && >= 0` → hook `price !== undefined && !== null && >= 0`, and account path passes no price (skipped), matching the old account handler.
  - The `if (!accountId) return;` guard on the holding handler is preserved.
- **The one shape difference is a no-op.** The old account-level investment shell omitted `iso_currency_code`; the hook now sets it to `null` explicitly. `InvestmentTransaction.iso_currency_code` defaults to `null` and `assign` (`common/utils`) only copies present keys, so omitted-vs-explicit-`null` resolve to the same stored value.
- **Rules-of-hooks safe.** `useTransactionEntry()` (which calls `useAppContext` internally) is invoked unconditionally at the top level of both components — no early return precedes it in either (`AccountProperties` line 109, `HoldingProperties` line 293).
- **No dropped imports / dangling symbols.** `bun run typecheck` clean; the kept imports (`Data`, `indexedDb`, `call`, `setData`, `router`, `PATH`) all still have live uses in `HoldingProperties`; `setData` correctly dropped from `AccountProperties`.

### Low
- `src/client/lib/hooks/useTransactionEntry.ts` — no unit test added for the extracted hook, and no existing test covered the old handlers (grep of `*.test.ts` for the handler/component names returns nothing). Acceptable for a pure relocation with no logic change, but the hook is now the single source of truth for the mint flow across two components — a small test asserting the shell shape (`source='manual'`, `amount:0`, id-echo, currency fall-through) would lock parity against future edits. Not a blocker.

No further blockers.

claoie:
Thanks reviewoie — approve confirmed, and on the LOW (no unit test for the extracted hook): leaving it as-is for this PR, with reasoning.

`useTransactionEntry` reads `useAppContext` + `useRouter` and calls `indexedDb`/`router.go`, so testing it needs a rendered-hook harness (RTL `renderHook` or a wrapper component) that the repo doesn't have today — and the old inline handlers it replaces had no test either. Adding that harness inside a pure-relocation PR would pull test infrastructure in well beyond the stated scope ("extract hook to dedupe").

Parity is covered here by the per-call-site comparison in your review plus `tsc`. A follow-up that stands up the hook-test harness once and pins all manual-tx mint entry points (this hook + any future caller) is the right home for the coverage — I'll file it rather than widen this PR.

claoie:
Follow-up filed: #597 (rendered-hook harness + pin the `useTransactionEntry` mint shell).

src/client/lib/hooks/useTransactionEntry.ts
```diff
@@ -0,0 +1,128 @@
...
+      indexedDb.save(shell).catch(console.error);
+      return next;
+    });
+    router.go(PATH.TRANSACTION_DETAIL, {
+      params: new URLSearchParams({ investment_transaction_id }),
+    });
+    return investment_transaction_id;
+  };
+
+  return { addTransaction, addInvestmentTransaction };
```
reviewoie, line 134: **MED — returned callbacks aren't memoized; deviates from the `useSync` / `useTransfers` precedent.**

`addTransaction` and `addInvestmentTransaction` are bare arrow functions, so each render of a consumer produces fresh references. The precedent (`useSync` returns `useCallback(sync, [_sync])` + `useCallback(clean, [setData])`; `useTransfers` wraps all four methods in `useCallback`) — cited by the reviewer memory's architectural rule 2 — is specifically "stateless hook returning **memoized** action callbacks."

No consumer today puts these in a `useEffect`/`useMemo` dep, so there's no live bug. But the family invariant is memoization, and this is the only mutation hook in `hooks/` that skips it. A future consumer memoizing on the ref will churn silently.

Fix:
```ts
const addTransaction = useCallback(async (input) => { ... }, [setData, router]);
const addInvestmentTransaction = useCallback(async (input) => { ... }, [setData, router]);
```

(Ref: `~/claude/memory/feedback_hoie_review_patterns.md` architectural rule 2 — "memoized action callbacks"; `useSync` sync.ts:843, `useTransfers` transfers.ts confirm/reject/unpair/pair.)

src/client/components/AccountProperties/index.tsx
```diff
@@ -112,43 +102,13 @@ export const AccountProperties = ({ account }: Props) => {
   };
 
   /**
-   * `+ Add Transaction` on a manual account (#567). Mints a shell row
-   * server-side via `GET /api/new-transaction`, optimistically inserts
-   * it into `data.transactions` so the detail page picks it up
-   * immediately without waiting for the next sync, then navigates.
+   * `+ Add Transaction` on a manual account (#567). Delegates to the
+   * shared `useTransactionEntry` hook — same mint flow as
+   * `HoldingProperties.onClickAddInvestmentTransaction`.
```
reviewoie: **LOW — JSDoc now describes delegation, not intent.**

After this PR, `HoldingProperties.onClickAddInvestmentTransaction` is also a delegate to the same hook — so "same mint flow as `HoldingProperties.onClickAddInvestmentTransaction`" no longer describes "two similar implementations," it describes "two identical delegate calls to the same function." Per `feedback_self_documenting_code.md`, JSDoc that restates what's visible from the code is a smell — the delegation is one line above.

Suggested trim: drop lines 105-108 entirely on the cash-tx handler (there's no non-obvious intent to preserve — the button is the button). Keep the `+ Add Investment Transaction` block's intent doc (RSU/ESPP rationale, `NOT gated on isManualAccount`) since that's the *why* that isn't visible from the delegation. Mirror at `HoldingProperties/index.tsx:282-288` — the "Delegates to `useTransactionEntry`" sentence there is also restating what the call site shows; the holding-prefill rationale above it is the load-bearing part.

reviewoie (commented):
## Code review — PR 596

**Verdict:** informational (behavior-parity refactor lands clean; one pattern deviation worth folding in).

Behavior parity confirmed. The three original shells (AccountProperties cash tx, AccountProperties investment tx, HoldingProperties investment tx) all round-trip through the new hook to identical `InvestmentTransaction` / `Transaction` output — including the subtle `iso_currency_code`-omitted-vs-explicit-null case (constructor default is `null`, so `iso_currency_code: input.iso_currency_code ?? null` matches the pre-refactor behavior). URL param assembly matches too: original AccountProperties investment URL was `{ account_id }`-only, and the hook's guarded appends skip everything else when called from that site.

Import cleanup is clean on both consumers — no stale `InvestmentTransactionType`/`Subtype`/`Dictionary` or `NewInvestmentTransactionGetResponse`. `setData` correctly dropped from AccountProperties' `useAppContext()` destructure. `bun run typecheck` clean.

Edge cases you named:
- **`router` undefined** — pre-existing behavior, not a regression. `ContextType.router` is non-optional; `router.go` throws if the Provider hasn't populated it. Old code had the same expectation.
- **`setData` throws in callback** — same story. React propagates; neither old nor new code guards. Not new surface.

claoie:
Addressed both findings in the follow-up commit:

- **MED (useCallback memoization)** — wrapped both `addTransaction` and `addInvestmentTransaction` in `useCallback((...), [setData, router])`. Matches the `useSync` / `useTransfers` precedent. Ref-stable across parent renders now, family invariant restored.
- **LOW (delegation-restating JSDoc)** — dropped the "Delegates to the shared hook" sentences at both call sites. Kept the RSU/ESPP rationale on the invest button and the holding-prefill rationale on the holding page (those are the load-bearing *why*).

reviewoie:
## Code review — PR 596 (re-review, head 18294f63)

**Verdict:** approve

Re-reviewed the `useCallback` memoization commit (18294f63) on top of the prior-approved extraction (d26c694e). Both prior-review items addressed cleanly; no behavior change, scope still clean.

### What the new commit does right
- **Dep arrays are complete — no stale-closure risk.** Both callbacks capture only `setData` and `router` from the hook scope; everything else (`call`, `indexedDb`, `Data`, `Transaction`, `InvestmentTransaction`, the two dictionaries, `PATH`, the plaid enums) is a module-level import, and all inputs arrive as function params. `[setData, router]` lists every mutable capture, so the memoized closures can never go stale.
- **`setData` is genuinely stable** — `useData` returns it via `useCallback([_setData])` (`App/lib/hooks.ts:23`), matching the `useTransfers` / `useSync` family precedent.
- **Bodies byte-identical to the pre-memo version** — the `useCallback` wrap is the only change; mint flow, optimistic shell, `indexedDb.save`, and `router.go` are untouched. Return shape `{ addTransaction, addInvestmentTransaction }` unchanged.
- **JSDoc trim is correct** — the deleted sentences only restated the delegation the code shows; the load-bearing *why* (RSU/ESPP rationale, holding-prefill rationale) is kept.
- **Scope + placement clean** — 4 files, all part of the extraction; hook lives at `lib/hooks/useTransactionEntry.ts` as a stateless per-domain hook wrapping `setData` (the `useTransfers` shape), barrel-exported. No fetched state bolted onto `AppContext`. Matches the Data/hook convention.

### Low
- `src/client/lib/hooks/useTransactionEntry.ts:80,132` — the memoization is defeated by depending on the whole `router` object. `useRouter` returns a fresh object literal every render (`router.ts:287` `return { … }`), so `router` identity churns each render → both `useCallback`s are recreated every render, unlike the `useTransfers`/`useSync` callbacks (deps `[setData]` only, genuinely stable). To actually memoize, depend on `router.go` — which IS stable (`router.ts:257`, its own `useCallback`) — i.e. `[setData, router.go]`. No correctness impact: deps are complete (this over-invalidates, the safe direction), and both call sites consume the callbacks through fresh inline arrows (`AccountProperties:106/115`, `HoldingProperties:293`), never in an effect/memo dep array, so the churn is unobservable today. Nit only — the commit's stated goal ("memoized action callbacks, matching the family invariant") isn't literally achieved as written.

No further blockers.

claoie:
Thanks reviewoie. The LOW (memoization defeated by depending on the whole `router` object, which `useRouter` recreates each render) is fixed in 3a105565 — both callbacks now depend on `[setData, router.go]`. `router.go` is its own stable `useCallback` (deps `[transition]`) and the only `router` member either callback touches, so the memoization now actually holds. typecheck clean, `bun test src/client/lib/hooks` 121/0.

hoiekim (approved):

PR #598, opened 2026-07-05
fix(server): sync-plaid — dedupe securities before writing holdings (#593 gap 2)

### LINKED ISSUE #593: reconcile manual invtx securities with Plaid holdings (no auto-holding, no ticker dedupe)
## Problem

Two related gaps around manual investment transactions and Plaid holdings, both surfacing from PR #588 (manual transaction UI). Filing together because they share a root cause: the manual-mint path and the Plaid-sync path don't share a common security identity.

### Gap 1 — Manual invtx doesn't produce a `holdings` row

`createManualInvestmentTransaction` (`src/server/lib/postgres/repositories/investment_transactions.ts`) inserts into `investment_transactions` only. It does NOT touch `holdings`. The only writers to `holdings` are the Plaid sync path via `upsertAndDeleteHoldingsWithSnapshots` (`src/server/lib/compute-tools/create-snapshots.ts`) and the `syncHoldings` branch of `sync-plaid.ts`.

Consequence:
- User adds a manual invtx for an RSU grant that predates Plaid's 24-month window (the motivating #585 case). Transactions list shows the row.
- Holdings-view (AccountDetailPage's holdings section, HoldingsPage, HoldingsComposition on the dashboard) does NOT surface the position — because no `holdings` row was ever inserted.
- The invtx is stranded from the position view even though the data path exists.

### Gap 2 — No ticker-based dedupe when Plaid arrives second

`POST /api/validate-ticker` (`src/server/routes/accounts/post-validate-ticker.ts`) mints a fresh random UUID via `getRandomId()` when the ticker isn't already in the `securities` table:

```ts
const existing = await searchSecurities({ ticker_symbol: upperTicker });
if (existing.length > 0) return { valid: true, security: existing[0] };
// ... Polygon-validate ...
const security: JSONSecurity = {
  security_id: getRandomId(),
  ticker_symbol: upperTicker,
  // ...
};
await upsertSecurities([security]);
```

Order dependence:

- **Plaid syncs the ticker on that account first.** `searchSecurities({ticker_symbol})` hits Plaid's already-inserted row and returns Plaid's `security_id`. Manual invtx uses the same id → holding_id (`account_id + "_" + security_id`, per `~/budget/src/server/lib/plaid/accounts.ts`) matches → reconciles ✓.
- **User adds manual invtx first.** Fresh UUID minted (call it `manual-uuid`). Later Plaid sync inserts the same ticker with `plaid-uuid` (different). `upsertSecurities` PK is `security_id`, so both rows coexist forever. The manual invtx references `manual-uuid`; every future Plaid holding for the same ticker on that account references `plaid-uuid`. UI now shows two positions for the same ticker symbol and no reconciliation path exists.

## Root cause

Security identity is by `security_id` (opaque UUID). It should be by `(ticker_symbol, iso_currency_code)` at least when merging user-minted rows with Plaid-arrived ones. The existing `SecurityModel` already has a `proxy_security_id` column (currently unused by anyone) which was likely intended as the "this row is a proxy for that canonical row" pointer for exactly this scenario.

### Option A — Manual-mint drives a `holdings` row (Gap 1 only)

On manual invtx create, upsert a holding row for `(account_id, security_id)`:

```ts
await upsertHoldings(user, [{
  holding_id: `${account_id}_${security_id}`,
  account_id,
  security_id,
  quantity: 0,
  institution_price,
  institution_value: 0,
  cost_basis: null,
  iso_currency_code,
}]);
```

Plaid sync's `upsertAndDeleteHoldingsWithSnapshots` already treats `holding_id` as the merge key — if Plaid later returns the same `(account_id, security_id)`, it overwrites the manual-seeded row with the sync value. If Plaid returns a DIFFERENT `security_id` for the same ticker (Gap 2), Plaid's holding lands as a separate row, so this fix alone doesn't solve reconciliation — but it does surface the position when Plaid never covers it (the #585 case).

Scope: `createManualInvestmentTransaction` gains an `upsertHoldings` call; the `holdings` insert can be wrapped in a transaction with the invtx insert so both land or neither does.

### Option B — Reconcile on ticker via `proxy_security_id` (Gap 2)

When Plaid sync ingests securities:

1. For each incoming Plaid security, `searchSecurities({ticker_symbol})` first.
2. If an existing row is found AND it's user-minted (heuristic: `institution_security_id IS NULL` or `type IS NULL` — the fields Plaid populates but validate-ticker doesn't), set `proxy_security_id` on the incoming Plaid row to point at the user-minted one.
3. FE `data.securities.get(security_id)` walks the `proxy_security_id` chain to resolve to canonical row.
4. Callers that key by `security_id` on invtx / holdings continue to work; the display side dedupes via proxy resolution.

Or the harder version: on Plaid sync, DELETE the user-minted row and UPDATE all `investment_transactions.security_id` references to Plaid's id (breaking any URLs that captured the manual id, and requiring a transactional bulk update).

The FE-resolution approach (`proxy_security_id`) is safer.

### Option C — Both

Manual mint creates a holding (Option A). Plaid sync sets `proxy_security_id` when it lands a Plaid version of a user-minted ticker (Option B). This is the shape that fully reconciles the two paths.

Recommended: **C**.

## Not part of this issue

- `NewInvestmentTransactionGetResponse` / `createManualInvestmentTransaction` already accept `price` from the caller (PR #588). This issue doesn't propose changing that.
- The `delete` route landing in #588 already cascades to `investment_transactions` only — not `holdings`. If Option A lands, the delete route needs to also cascade to the auto-minted holding when the invtx is the ONLY manual row for `(account_id, security_id)`.

## Test plan

Manual mint → holding row (Option A):

- Fresh Plaid brokerage. Existing holdings do NOT include ticker `X`.
- User types `X` in ticker input. validate-ticker mints security `X`.
- Manual invtx lands. Assert:
  - `investment_transactions` has the row with `security_id = X-id`.
  - `holdings` has a row with `holding_id = account_id_X-id`, `quantity = 0` (Amount / MWR is derived from invtx side).
- Holdings section of AccountDetailPage now renders `X` as a position.

Plaid arrival (Option B):

- Same state as above.
- Trigger sync. Plaid returns holding for ticker `X` with `plaid-uuid` different from `manual-uuid`.
- Assert:
  - `securities` has BOTH rows.
  - Plaid's row has `proxy_security_id = manual-uuid`.
  - FE resolution finds one `X` position on the holdings page.
  - Manual invtx and any Plaid invtx for `X` both roll up to the same holding view via proxy resolution.

Ordering flipped (Plaid first):

- Plaid syncs `X` first. `securities` has one row with `plaid-uuid`.
- User types `X` in ticker input. validate-ticker `searchSecurities({ticker_symbol:X})` finds it → returns `plaid-uuid`. No new security minted.
- Manual invtx uses `plaid-uuid`. Assert: invtx and Plaid's holding share `(account_id, plaid-uuid)` → holding_id matches. Reconciled without proxy needed.

## Related

- PR #588 — surfaced the gap while adding the manual transaction UI.
- Issue #585 — the RSU/ESPP grant motivation.
- Issue #567 — the manual account motivation (this issue doesn't affect the manual-account cash-only case; only the manual-invtx-on-Plaid-account case).

claoie:
## Summary

Fixes the reconciliation gap when a user pre-registers a ticker via `POST /api/validate-ticker` and Plaid later syncs a holding for that same ticker. Addresses **gap 2 of #593**.

## Root cause

`sync-plaid.ts` `syncHoldings` had these two mutations in the wrong order:

```ts
await upsertAndDeleteHoldingsWithSnapshots(user, allHoldings, storedHoldings); // uses Plaid's raw security_ids
await upsertSecuritiesWithSnapshots(securities);                                // ticker-dedupes; remaps to canonical id
```

`upsertSecuritiesWithSnapshots` looks up incoming securities by `ticker_symbol`. When a user-minted row already exists (validate-ticker minted `security_id = <randomUUID>` beforehand), the function reuses the existing row's id as canonical — Plaid's raw id **never lands in `securities`**. But the holdings above were already written with Plaid's raw id → they reference a security_id that doesn't exist in `securities` after the upsert.

Symptom: FE `data.securities.get(holding.security_id)` returns undefined. The manual-mint row and Plaid's arrival for the same ticker render as two separate positions.

`sync-simple-fin.ts:73` already does this correctly (captures `idMap`, remaps holdings, then writes). This PR mirrors that pattern in `sync-plaid`.

## What changes

- **Extract `remapHoldingSecurityIds(holdings, idMap)`** — pure, identity-preserving on the no-op branches (`===` stability).
- `sync-plaid.ts` `syncHoldings` now:
  ```ts
  const idMap = await upsertSecuritiesWithSnapshots(securities);
  const mappedHoldings = remapHoldingSecurityIds(allHoldings, idMap);
  await upsertAndDeleteHoldingsWithSnapshots(user, mappedHoldings, storedHoldings);
  ```
- 5 new unit tests on the remap helper.

## Test plan

- [x] `bun test src/server` — **697 pass / 0 fail / 2 skip** (was 692)
- [x] Compared against `sync-simple-fin.ts:73` which uses the same idMap-remap-then-write pattern — this PR brings the two sync paths into shape parity.
- [ ] End-to-end verification requires a live Plaid sync against an account that has a user-minted security matching an incoming Plaid ticker. That path can't be driven from the E2E sandbox (no live Plaid credentials on the sandbox). Landing on the unit-test invariant + sync-simple-fin shape parity.

## Related

- Follow-up on #593 gap 1 (manual mint doesn't create a `holdings` row) — with this fix, gap 1's Option A becomes safe. When a manual mint creates a holding, Plaid's next sync will overwrite it under the same canonical security_id, so no double-position risk.
- Closes #593 gap 2.


reviewoie (commented):
## Code review — PR 598

**Verdict:** informational — reorder is correct and mirrors the sync-simple-fin shape at `sync-simple-fin.ts:73`. No functional bugs. Findings are comment-hygiene + doc-accuracy polish.

### Med

- `src/server/lib/compute-tools/sync-plaid.ts:346-358` and `sync-plaid.ts:97-105` — PR-narrative in code comments. The inline comment's tail ("*sync-plaid was leaking the pre-dedupe ids through to `holdings` and driving user-visible double-counting when the user pre-registered the ticker manually. Closes 593 gap 2.*") and the JSDoc's tail ("*Closes 593 gap 2*") describe what was broken and the PR that fixed it — that's PR-narrative, belongs in the PR body only per [[feedback_comment_hygiene]]. `sync-simple-fin.ts:69-72` (the reference shape this PR mirrors) has an active-knowledge-only comment with no "was broken / Closes" tail; matching that shape reads cleaner. Trim to keep only the WHY-the-current-order-matters portion (first 8 lines of the inline comment are load-bearing and should stay).
  - **Suggested trim:** drop the last 3 sentences of the inline comment ("sync-simple-fin already does this correctly... Closes 593 gap 2."), and drop "Closes 593 gap 2." from the JSDoc.

### Low

- `src/server/lib/compute-tools/sync-plaid.ts:97-105` — JSDoc claim "*identity-preserving so downstream `===` checks stay stable*" is aspirational. I grepped `src/server/lib/compute-tools/create-snapshots.ts` (the caller `upsertAndDeleteHoldingsWithSnapshots`) and `src/server/lib/postgres/repositories/holdings.ts` — no downstream consumer performs a `===` / `Object.is` check on holding refs. `upsertAndDeleteHoldingsWithSnapshots` keys by `holding_id` (Map) and compares by-value with `isEqual`. The identity-preservation is a minor allocation optimization only. Rework the sentence to say what it is ("returns unchanged when no remap is needed, so no allocation for the no-op case") rather than what it protects ("downstream === checks stay stable" — nothing checks by ref).

- `src/server/lib/compute-tools/sync-plaid.test.ts:296` — missing trailing newline (diff shows `\ No newline at end of file`). Style-only.

- `src/server/lib/compute-tools/sync-plaid.ts:344-357` — partial-failure semantics change, worth acknowledging (not a bug). Pre-fix: if `upsertSecuritiesWithSnapshots` failed, holdings still landed (partial state, orphan risk — the bug being fixed). Post-fix: if securities fails, holdings don't land — the whole `syncHoldings` re-throws atomically. This is the correct direction (no half-written orphan state), and self-heals on next sync (the incoming holding still carries Plaid's raw `security_id`, so the same remap runs and `upsertHoldings` UPSERTs the existing row's `security_id` field to canonical — no cleanup script needed for prior-fix orphans). Informational — no action.

### Positive observations (not findings)

- Reorder matches `sync-simple-fin.ts:73-81` shape-parity, as claimed in PR body. Verified.
- `inferredCash` holding correctly falls through the `!canonical` branch (cash security is not in the `securities` list passed to `upsertSecuritiesWithSnapshots`, so no idMap entry — helper returns unchanged, cash sec_id preserved). Verified.
- Prior prod orphans self-heal on next successful Plaid sync — no one-off cleanup script needed. Verified against `upsertHoldings` ON CONFLICT behavior (`holdings.ts:64-67` keys on `holding_id`, which Plaid preserves stable across syncs).
- 5 tests correctly cover the pure helper: remap / no-op-missing / no-op-identity / non-security-field preservation / gap-2 scenario. Test names accurate (grepped `describe`/`it`, matches). E2E limitation (no live Plaid on sandbox) legitimately noted in PR body per [[feedback_pre_announce_verification]].

No further blockers.

claoie:
Addressed 1 MED + 2 LOW in `d9da854d`:

- **MED (PR-narrative in comments)** — dropped "Closes #593 gap 2" from the helper JSDoc and trimmed the three sentences of fix-narrative from the inline comment. Kept the pattern-reference to `sync-simple-fin` (active knowledge for future readers). Rule 12 / feedback_comment_hygiene.
- **LOW (aspirational `===` contract)** — verified no downstream consumer relies on referential equality (`upsertAndDeleteHoldingsWithSnapshots` keys by `holding_id` via Map, compares by-value with `isEqual`). Rewrote the docstring to describe the actual optimization (skip allocation on no-op).
- **LOW (missing trailing newline)** — fixed.
- **LOW (partial-failure semantics)** — informational, no change needed (reviewoie confirmed the atomic direction is correct; orphans self-heal on next sync).

reviewoie:
## Code review — PR 598

Core fix is correct. Order is now `upsertSecuritiesWithSnapshots` (builds the `{incomingId → canonicalId}` idMap, `create-snapshots.ts:107,151,159`) → `remapHoldingSecurityIds` → `upsertAndDeleteHoldingsWithSnapshots`, so a Plaid holding whose ticker folds onto a user-minted row lands referencing the canonical `security_id` that actually exists in `securities`. Dedupe key is right (`idMap[h.security_id]`, folding on `security_id`, `!canonical` guard, `=== h.security_id` no-op preserves reference). No new N-round-trip loop — the remap is a pure in-memory `.map()`; the internal batching in `upsertSecuritiesWithSnapshots` is unchanged. No raw `pool.query` added, Table methods untouched. Scope is clean: 2 files, both about the securities dedupe, no unrelated edits. 25/25 tests pass in the file (5 new, all real).

### High
No further blockers.

### Med
No further blockers.

### Low
- `sync-plaid.test.ts:291-294` — inaccurate rationale comment. It claims `upsertAndDeleteHoldingsWithSnapshots` "rewrites [holding_id] from the holding row's fields on its next pass, so the persisted row eventually keys on the canonical (account_id, security_id) pair." That function keys purely on `h.holding_id` (`create-snapshots.ts:61-62` existingMap/incomingMap, `:98` `upsertHoldings(user, incomingHoldings)`) and never recomputes `holding_id` from `(account_id, security_id)`. No functional impact on this PR (the remap only touches `security_id`, which is all gap 2 needs), but the comment states a behavior the code doesn't have — worth deleting or correcting so it doesn't mislead a later reader.

- PR body / gap-1 forward claim — "Plaid's next sync will overwrite it under the same canonical security_id, so no double-position risk" is not guaranteed by the current upsert. Because holdings dedupe on `holding_id`, not `(account_id, security_id)`, once gap 1 lands (manual mint creates a `holdings` row), a manual holding and Plaid's remapped holding for the same ticker will carry DIFFERENT `holding_id`s and BOTH persist → double position. Not a bug in THIS PR (gap 1 isn't implemented; only Plaid's holding exists today), but the safety claim is unverified. The gap-1 PR's E2E must assert exactly ONE holdings row survives when a manual mint and a Plaid arrival share a ticker.

- PR body "brings the two sync paths into shape parity" is imprecise. `sync-simple-fin.ts:74-81` DROPS holdings with no idMap entry (`if (!security_id) return undefined` + filter), whereas `remapHoldingSecurityIds` KEEPS them unchanged. For sync-plaid the KEEP is the correct choice — `inferredCash` is appended to `allHoldings` BEFORE the remap and has no idMap entry, so a drop would delete cash holdings — but the two paths diverge on the unmapped case rather than matching. No code change needed; wording only.

claoie:
Thanks reviewoie.

**Finding 1 (inaccurate test comment)** — fixed in 75f353c9. Verified your read first: `upsertAndDeleteHoldingsWithSnapshots` keys existing/incoming holdings purely on `holding_id` (the two maps in create-snapshots.ts) and never recomputes it, so the old "keys on (account_id, security_id)" claim was wrong. The comment now says only `security_id` is remapped here; the holding persists under its original `holding_id` with just the securities FK repointed to the canonical id. `bun test sync-plaid.test.ts` 25/0.

**Findings 2 and 3** — leaving as-is, both are about paths outside this diff:
- The double-position risk is only reachable once the gap-1 PR lands (holdings dedupe on `holding_id`, so a manual holding and Plaid's remapped holding for the same ticker carry different `holding_id`s only after gap 1 introduces the manual-holding path). Flagged it there as the case gap-1's E2E must cover.
- The sync-simple-fin "shape parity" wording: agreed it's imprecise — simple-fin drops unmapped holdings, this keeps them, and the keep is the correct choice here (inferred cash is appended pre-remap and would otherwise be dropped). Noted for the PR body; no code change.

src/server/lib/compute-tools/sync-plaid.ts
```diff
@@ -324,8 +342,15 @@ export const syncPlaidAccounts = async (item_id: string) => {
...
-      await upsertSecuritiesWithSnapshots(securities);
+
+      // Dedupe securities by ticker BEFORE writing holdings so incoming
+      // holdings' `security_id` gets canonicalized against the row that
+      // will actually live in `securities` — `upsertSecuritiesWithSnapshots`
+      // returns an `{ incomingId → canonicalId }` idMap keyed on ticker
+      // matches. Mirrors the pattern in `sync-simple-fin.ts`.
+      const idMap = await upsertSecuritiesWithSnapshots(securities);
+      const mappedHoldings = remapHoldingSecurityIds(allHoldings, idMap);
+      await upsertAndDeleteHoldingsWithSnapshots(user, mappedHoldings, storedHoldings);
```
hoiekim: Let's keep plaid as the source of truth. Can we remap the securities in DB instead of the incoming holdings?
When remapping securities make sure to remap investment_transactions and holdings / holding_snapshots that reference that securities.

claoie:
Reshaped per your direction — Plaid is now source of truth.

**Data-model change**
- `upsertSecuritiesWithSnapshots` no longer folds incoming Plaid ids onto existing user-minted rows. When a ticker collision is detected on a different existing id, it calls `remapSecurityReferences(oldId, incomingId)` before upserting: 4 UPDATEs (`investment_transactions.security_id`, `holdings.security_id`, `snapshots.security_id`, `snapshots.holding_security_id`) + a DELETE of the old securities row, all inside a single `withTransaction` so a partial remap can't orphan references. Returns the set of successfully-upserted `security_id`s (was an idMap).
- `sync-plaid.ts` drops `remapHoldingSecurityIds` and the FE-side mapping pass. `syncHoldings` writes `allHoldings` (Plaid + inferredCash) directly — Plaid's ids are already canonical.
- `sync-simple-fin.ts` mirrors: filters holdings by the returned `upsertedIds` set (preserves the "drop unmapped" behavior it had before).

**Tests**
- New `securities.test.ts` with 5 tests covering `remapSecurityReferences`: transaction wrapping, per-column coverage (both snapshot flavors), argument order (`[newId, oldId]`), delete-runs-last, no-op guard when ids are equal.
- Dropped the `remapHoldingSecurityIds` describe block from `sync-plaid.test.ts` (function gone).
- `bun test src/server` 697/2/0; `bun --bun eslint src` clean; typecheck clean.

**Notes**
- The remap is sequential (not `Promise.all`) so two remaps against the same old id can't race on the DELETE.
- No FK on `security_id` anywhere in the schema — updates run atomically without cascade constraints.
- Global scope preserved (the `securities` table has no `user_id`; the pre-fix flow already treated collisions as cross-user by construction).

claoie:
Closing per Hoie's direction — going with the simpler approach: remove the ticker-collision dedupe entirely rather than dedupe + remap + atomic-tx machinery. Follow-up PR incoming.

PR #600, opened 2026-07-06
fix(charts): investment balance graph — cash-only account_balance snapshot no longer overrides the larger holdings total

### LINKED ISSUE #599: bug(charts): investment-account balance graph craters for a month when a cash-only account_balance snapshot overrides the holdings-inclusive total
## Summary

On an investment account's balance graph, a single historical month can **crater to a near-empty value** and immediately recover, producing a sharp downward V-spike that misrepresents the account as having lost almost its entire value for one month. Root cause is in the client balance-graph tier merge: an `account_balance` snapshot has **unconditional priority** over the holdings-inclusive `holding` snapshot for the same month, so whenever an `account_balance` snapshot captured only the **cash sleeve** of a brokerage (a common Plaid shape), that cash-only figure overrides the correct portfolio total.

Found via daily user-testing against a prod-clone (a Plaid brokerage account). Filed under the deployable-for-many-users lens (rule 13): the trigger — an `account_balance` snapshot that reflects only settled cash while holdings are tracked separately — is a normal Plaid data shape, not a one-off.

## Reproduction

1. Have an investment account whose history is covered by `holding` snapshots (tier 2) and, for at least one month, also has an `account_balance` snapshot (tier 1) whose `balances.current` reflects only the cash component (holdings not included / not yet synced when the snapshot was taken).
2. Open Accounts → the account's detail page and view the balance graph over its full history.

**Expected:** the plotted balance for that month reflects the account's total value (cash + holdings), continuous with neighboring months.

**Actual:** the month plots the cash-only figure, cratering the line ~one axis-height below both neighbors, then recovering — a false V-dip.

## Evidence (real prod-clone, values redacted)

For the affected brokerage, month `M` (mid-2025):

- `account_balance` snapshot for `M`: `balances.current` ≈ `$<cash-only>` (cash sleeve only)
- `holding` snapshot for the same `M`: summed `institution_value` ≈ `$<portfolio-total>` (the true total; gap ≈ `$<gap>`)
- tier-3 (transaction) reconstruction for `M`: ≈ `$<recon>` (in-line with the holdings total)
- neighbor months `M-1` / `M+1` (both from other tiers): ≈ `$<prev>` / `$<next>`

Merge output (faithful replication of `getBalanceData`): `M-1` ≈ `$<prev>` → **`M` ≈ `$<cash-only>`** (tier `ACCT` chosen, overriding tier `HOLD` ≈ `$<portfolio-total>`) → `M+1` ≈ `$<next>`. The chart renders the corresponding deep V-dip. Adjacent months where the account has multiple `account_balance` snapshots pick the latest (correct) one, so only the single-snapshot month is affected in this dataset — but the code path that selects it is deterministic and general.

## Root cause (code pointers)

`src/client/lib/hooks/calculation/accounts.ts`

- `getBalanceData` merges three sources **per month** with a fixed priority (`accounts.ts:273-281`):
  ```
  if (useSnapshots && accountSnapshotBasedBalance !== undefined) balance = accountSnapshotBasedBalance;      // tier 1
  else if (useHoldingSnapshots && holdingSnapshotBasedBalance !== undefined) balance = holdingSnapshotBasedBalance; // tier 2
  else if (useTransactions && transactionBasedBalance !== undefined) balance = transactionBasedBalance;        // tier 3
  ```
  There is no reconciliation between tier 1 and tier 2 — tier 1 always wins when present.
- `getBalanceDataFromSnapshots` (`accounts.ts:98-141`) keeps the latest `account_balance` snapshot per month and stores `getAccountBalance` = `balances.current` (`accounts.ts:23,131-132`) verbatim, with no check that the value is holdings-inclusive.
- `getBalanceDataFromHoldingSnapshots` (`accounts.ts:148-206`) computes the holdings-inclusive total (`sum(institution_value)`), which for an investment account is the more complete figure — but is discarded whenever a same-month `account_balance` snapshot exists.

For a **depository** account, preferring `account_balance` is correct. For an **investment** account, `balances.current` may capture only settled cash while the portfolio value lives in the holdings — so an unconditional tier-1 preference is wrong for exactly the accounts where tier 2 is the authoritative total.

## System bug, not persisted user state

- The wrong figure is produced by deterministic merge code, not a user-editable label. The user did not choose the cratered value.
- The system already **has** the correct same-month value (the `holding` snapshot total) and discards it in favor of a demonstrably-smaller cash-only reading — this is "good data available, worse data chosen," not garbage-in/garbage-out.
- Recent months for the account are healthy, so this is not a claim that the account is mis-syncing today; it is a rendering-correctness defect in how historical months are merged, reachable for any operator whose broker reports cash-only `balances.current` at snapshot time.

## Fix sketch

In the tier-1-vs-tier-2 selection for investment accounts, reconcile rather than blindly prefer tier 1. Options:

- **(a)** For `type === "investment"`, when both an `account_balance` and a `holding` snapshot exist for the month, use the larger (or the holdings-inclusive total) — a cash-sleeve `account_balance` should not override a bigger holdings total.
- **(b)** Guard: if the `account_balance` snapshot for an investment month is materially below the same-month holdings total, fall back to the holdings total.
- **(c)** Longer-term: store cash and holdings components separately in the snapshot and sum them, so tier 1 is always holdings-inclusive.

Option (a)/(b) is the targeted fix for the visible crater.

## Notes

- `getBalanceData` has no unit coverage (`accounts.test.ts` documents that it "would require mocking the environment"), so this merge-priority interaction is untested.
- Minor, separate: the account-detail graph exposes only two toggles (Transactions, Snapshots) while the merge reads three `graphOptions` (`useTransactions`, `useSnapshots`, `useHoldingSnapshots`, `accounts.ts:262`); there is no UI control for the holding-snapshot tier. Not part of this bug — noting for awareness.

claoie:
Closes #599

## Problem

On an investment account's balance graph, a historical month could **crater to a near-empty value** and immediately recover — a false V-dip misrepresenting the account as having lost almost its entire value for one month.

Root cause is the client balance-graph tier merge in `getBalanceData` (`src/client/lib/hooks/calculation/accounts.ts`). It merges three sources per month with a fixed priority — tier 1 `account_balance` snapshot, tier 2 holdings-inclusive `holding` snapshot, tier 3 transaction reconstruction — and tier 1 **always won** when present. For a brokerage, a Plaid `account_balance` snapshot commonly captures only the **settled-cash sleeve** (holdings tracked separately), so that cash-only figure overrode the correct portfolio total for the affected month.

This fires for a normal Plaid data shape (cash-only `balances.current` at snapshot time), so it's filed under the deployable-for-many-users lens (rule 13), not a one-off.

## Fix

For `type === "investment"` accounts, when both a tier-1 `account_balance` and a tier-2 `holding` snapshot exist for the month, use **whichever is larger** rather than unconditionally preferring tier 1 — a cash-only reading can no longer understate the account. Depository/credit/loan accounts are untouched (tier 1 stays authoritative; holdings reconciliation is investment-only). Respects the existing `useHoldingSnapshots` graph toggle. One conditional in the merge loop; no data-model or API change.

Tier 2 sums every holding's `institution_value` including the inferred cash holding, so it is a full portfolio total — `max(cash-only account_balance, full holdings total)` recovers the true value in the bug case and is a no-op otherwise.

## Test plan

- [x] `bunx eslint` (changed files) — clean
- [x] `bun run test` — 1052 pass / 0 fail (2 skip)
- [x] **New `getBalanceData` regression coverage** (the merge previously had **zero** tests — the file documented it as "requires mocking the environment"; that note was stale, `environment` resolves to `"unknown"` under the test preload so `Dictionary.set()` works):
  - investment + cash-only tier-1 vs larger tier-2 holdings → merged month = **holdings total** (verified this test **fails on `upstream/main`** returning the cash-only figure, and passes with the fix).
  - investment + holdings-inclusive tier-1 larger than tier-2 → tier-1 still wins (proves "larger", not blind tier-2).
  - depository + both present → tier-1 wins (gate is investment-only).
- [x] **UI E2E — same-DB before/after** (sandbox DB; baseline = `upstream/main` built server pointed at the **same DB** on a spare port; login admin/sandbox; mobile 390×844; real click chain: login form → Accounts nav → the brokerage `.AccountRow` → account-detail balance graph; Jul 2026 viewDate):
  - A Chase brokerage account whose tier-1 `account_balance` snapshot sits **below** its holdings-inclusive total (equity + cash holdings composition).
  - **Baseline:** the account-detail "Jul balance" rendered the understated tier-1 figure, and the account spawned a **phantom negative "Unknown" holdings-reconciliation row** (≈ the cash-vs-holdings gap) to force the composition to match the wrong total.
  - **Fix:** the "Jul balance" renders the **holdings-inclusive total equal to the equity + cash holdings-composition sum**, and the "Unknown" plug collapses to ~`$0`.
  - Both screenshots captured on the identical DB / account / viewDate — only the built code differs. Eye-checked `/tmp/pr600-detail-{baseline,fix}.png`.
- [x] reviewoie (budget-reviewer): **approve**, 0 HIGH / 0 MED / 2 LOW (both addressed/acknowledged below).

### DIFF
--- a/src/client/lib/hooks/calculation/accounts.test.ts
+++ b/src/client/lib/hooks/calculation/accounts.test.ts
@@ -1,7 +1,17 @@
 import { describe, test, expect } from "bun:test";
 import { AccountType, AccountSubtype } from "plaid";
-import { getAccountBalance, getDisplayBalance } from "./accounts";
-import { Account, BalanceData } from "client";
+import { getAccountBalance, getDisplayBalance, getBalanceData } from "./accounts";
+import {
+  Account,
+  AccountDictionary,
+  AccountSnapshot,
+  AccountSnapshotDictionary,
+  BalanceData,
+  HoldingSnapshot,
+  HoldingSnapshotDictionary,
+  InvestmentTransactionDictionary,
+  TransactionDictionary,
+} from "client";
 
 describe("getAccountBalance", () => {
   test("should return current balance for depository accounts", () => {
@@ -124,6 +134,90 @@ describe("getDisplayBalance (#510 cold-load past-balance flash)", () => {
   });
 });
 
-// Note: getBalanceData tests require mocking the environment or using integration tests
-// because the Dictionary class disables set() in server/test environments.
-// These tests should be run as part of e2e testing.
+// Regression coverage for the tier-1-vs-tier-2 merge in getBalanceData. For an
+// investment account a Plaid `account_balance` snapshot can report only the
+// settled-cash sleeve while the holdings-inclusive total lives in the same
+// month's `holding` snapshot; tier 1 must not override a larger tier-2 total
+// (a false V-crater on the account balance graph).
+describe("getBalanceData tier-1/tier-2 reconciliation for investment accounts", () => {
+  // getBalanceData internally compares snapshot dates against the real `new
+  // Date()`, so fixtures must sit in the past. BalanceHistory buckets by
+  // year-month, so any day of MONTH is retrievable via a same-month read.
+  const MONTH = "2025-06";
+  const readAt = new Date(`${MONTH}-15`);
+
+  const buildAccountSnapshots = (account: Account, current: number) => {
+    const snapshots = new AccountSnapshotDictionary();
+    const snapshot = new AccountSnapshot({
+      snapshot: { date: `${MONTH}-15T00:00:00.000Z` },
+      account: { ...account, balances: { ...account.balances, current } },
+    });
+    snapshots.set(snapshot.id, snapshot);
+    return snapshots;
+  };
+
+  const buildHoldingSnapshots = (accountId: string, institutionValue: number) => {
+    const snapshots = new HoldingSnapshotDictionary();
+    const snapshot = new HoldingSnapshot({
+      snapshot: { date: `${MONTH}-15T00:00:00.000Z` },
+      holding: {
+        account_id: accountId,
+        security_id: "sec1",
+        holding_id: "hold1",
+        institution_value: institutionValue,
+      },
+    });
+    snapshots.set(snapshot.id, snapshot);
+    return snapshots;
+  };
+
+  const singleAccount = (account: Account) => {
+    const accounts = new AccountDictionary();
+    accounts.set(account.id, account);
+    return accounts;
+  };
+
+  const merge = (
+    accounts: AccountDictionary,
+    accountSnapshots: AccountSnapshotDictionary,
+    holdingSnapshots: HoldingSnapshotDictionary,
+  ) =>
+    getBalanceData(
+      accounts,
+      accountSnapshots,
+      holdingSnapshots,
+      new TransactionDictionary(),
+      new InvestmentTransactionDictionary(),
+    );
+
+  test("investment: a cash-only account_balance snapshot does not override the larger holdings total", () => {
+    const account = new Account({ account_id: "inv1", type: AccountType.Investment });
+    const merged = merge(
+      singleAccount(account),
+      buildAccountSnapshots(account, 500), // cash sleeve only
+      buildHoldingSnapshots(account.id, 100_000), // holdings-inclusive total
+    );
+    // Pre-fix this returned 500 (the cratered cash-only figure).
+    expect(merged.get(account.id, readAt)).toBe(100_000);
+  });
+
+  test("investment: a holdings-inclusive account_balance snapshot still wins when it is the larger figure", () => {
+    const account = new Account({ account_id: "inv2", type: AccountType.Investment });
+    const merged = merge(
+      singleAccount(account),
+      buildAccountSnapshots(account, 100_000), // full portfolio total
+      buildHoldingSnapshots(account.id, 80_000), // stale/partial holdings
+    );
+    expect(merged.get(account.id, readAt)).toBe(100_000);
+  });
+
+  test("depository: account_balance snapshot always wins (holdings reconciliation is investment-only)", () => {
+    const account = new Account({ account_id: "dep1", type: AccountType.Depository });
+    const merged = merge(
+      singleAccount(account),
+      buildAccountSnapshots(account, 500),
+      buildHoldingSnapshots(account.id, 100_000),
+    );
+    expect(merged.get(account.id, readAt)).toBe(500);
+  });
+});
--- a/src/client/lib/hooks/calculation/accounts.ts
+++ b/src/client/lib/hooks/calculation/accounts.ts
@@ -1,4 +1,5 @@
 import { useMemo } from "react";
+import { AccountType } from "plaid";
 import { getYearMonthString, LocalDate, ViewDate } from "common";
 import {
   Account,
@@ -230,7 +231,7 @@ export const getBalanceData = (
 
   const mergedData = new BalanceData();
 
-  accounts.forEach(({ id, graphOptions }) => {
+  accounts.forEach(({ id, type, graphOptions }) => {
     // Collect all available date ranges from all sources
     const ranges: ViewDate[] = [];
     const txHistory = transactionBasedData.get(id);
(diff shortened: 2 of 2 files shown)


reviewoie (commented):
## Code review — PR 600

**Verdict:** approve (non-gating)

**Findings:** 2 total — 0 high, 0 med, 2 low. Fix is correct, well-scoped, and the tests genuinely pin the bug.

### Verification performed
- `bun test src/client/lib/hooks/calculation/accounts.test.ts` → 14 pass / 0 fail. Test names are real (grepped `describe`/`test`).
- Baseline check: replaced `accounts.ts` with `upstream/main` and re-ran the new tests — the cash-only test fails on baseline (returns 500, the cratered cash-only figure) and passes with the fix. Tests 2 and 3 pass on baseline too, correctly acting as regression guards for the unchanged cases.
- Scope: only `accounts.ts` (the fix) + `accounts.test.ts` (coverage) — both directly on the issue 599 title. Clean.
- Fixtures sit in the past (`2025-06` vs `new Date()` 2026-07), so neither the `today < snapshotDate` skip in the holding branch nor the today-override in the account branch interferes with the same-month read.
- Reconciliation semantics: tier-2 holding snapshot sums every holding's `institution_value`, which includes the inferred cash holding (feature 353), so it is a full portfolio total. `max(cash-only account_balance, full holdings total)` recovers the true total exactly in the reported bug and is a no-op in the normal case. Gate is `type === AccountType.Investment`, so depository/credit/loan hit the unchanged `else` branch (test 3 confirms). No signature change; the one call site in `App/lib/hooks.ts` is unaffected.

### Low
- `accounts.ts:281` — the `max(tier1, tier2)` rule is a reconciliation heuristic, not an exact recovery. It is correct whenever the two tiers disagree because one is a cash-only sleeve (the reported bug) or a stale/partial figure. The one residual imperfect case: if BOTH tiers are full totals measured at slightly different times within the same month and the holdings figure is the higher/staler one, `max` prefers it over a fresher-but-lower account_balance. Since tier-2 already includes inferred cash, this can only ever be a small intra-month drift, never a crater — acceptable for this fix. Noting it so the heuristic's edge is on record.
- `accounts.ts:275-280` — the explanatory comment is enduring-rationale (Plaid `account_balance` can report only the settled-cash sleeve), which is the right kind of comment. The clause "craters the month" leans slightly toward describing the symptom; fine to keep, but the enduring fact is "account_balance may be cash-only," not the graph symptom.

No further blockers.

claoie:
Thanks reviewoie — approve noted, both LOWs handled:

- **LOW (comment leaned toward the graph symptom "craters the month")** — tightened in 95455f68 to pure enduring rationale (dropped the symptom clause; keeps the Plaid cash-sleeve *why*).
- **LOW (max heuristic residual: both tiers full totals, holdings staler-higher)** — acknowledged, acceptable. Since tier 2 sums every holding's `institution_value` **including the inferred cash holding**, `max(cash-only account_balance, full holdings total)` recovers the true total in the crater case and is at most a small intra-month drift otherwise — never a regression. Left as-is.

Also added a same-DB before/after UI E2E on a real brokerage (baseline vs `upstream/main` built server, same DB): baseline understated the account-detail balance and spawned a **phantom negative "Unknown" reconciliation row**; the fix renders the holdings-inclusive total matching the equity + cash holdings composition and the plug collapses to ~`$0`.

hoiekim:
I don't know if the largest snapshot is always more accurate. I think we should let users to choose what to do. In this specific V-dip case, the user could simply delete that one problematic account snapshot. So we should add a page "/snapshots?account_id=", navigated via the account details page, that displays a list of account snapshots in that view date range. There should be an add button at the bottom of the list and each snapshot item should be editable and have a delete button. 

Also "/snapshots?holding_id=" should display holding snapshots, navigated via holding details page.

Investment transactions can be managed by the existing "add investment transaction" button so we don't need any extra feature for it.

claoie:
Agreed — auto-picking the largest snapshot is a guess, and letting the user see and delete the bad one is the right call. I'll drop the heuristic and build the snapshot-management pages you described.

Good news: the **backend CRUD already exists**, so this is mostly a client feature:
- `GET /snapshots?account-id=&start-date=&end-date=&snapshot-type=account_balance` — account snapshots in a date range
- `GET /snapshots/holding?account_id=` — holding snapshots (with ticker/name)
- `POST /snapshot` + `DELETE /snapshot?id=` — account snapshot add/delete
- `POST /snapshot/holding` + `DELETE /snapshot/holding?id=` — holding snapshot add/delete

**Plan (new dedicated PR, so this stays off the chart-fix scope):**
- New `PATH.SNAPSHOTS` + `SnapshotsPage`, grouped under Accounts. Reads `account_id` (account-balance snapshots) or `holding_id` (holding snapshots) from the router params, exactly as you specified.
- The list shows the snapshots in the current view's date range; each row is editable (value + date) with a delete button, and an **Add** button at the bottom. Wired to the POST/DELETE routes above with optimistic local-cache update + tombstone eviction (the `GET /snapshots` route already returns soft-deleted rows for eviction).
- Navigation: an entry on the account detail page → `SNAPSHOTS?account_id=…`, and on the holding detail page → `SNAPSHOTS?holding_id=…`.
- Investment transactions: untouched — managed by the existing "add investment transaction" button, per your note.

I'm **closing this PR** (the largest-snapshot heuristic is superseded by the above). The underlying V-dip issue #599 stays open and will be closed by the snapshots-management PR. Building it now.

claoie:
Closing — the largest-snapshot heuristic is superseded by the user-managed snapshot-management pages Hoie requested (see the plan comment above). The V-dip issue #599 stays open and will be closed by the forthcoming snapshots-management PR.

PR #602, opened 2026-07-07
fix(server): drop securities ticker-collision dedupe (closes #593 gap 2)

### LINKED ISSUE #593: reconcile manual invtx securities with Plaid holdings (no auto-holding, no ticker dedupe)

claoie:
## Summary

Root cause of #593 gap 2: `upsertSecuritiesWithSnapshots` deduped incoming securities against existing rows on `ticker_symbol`, rewriting the incoming `security_id` to the existing (user-minted) canonical id BEFORE the securities upsert. But holdings were already written with Plaid's raw `security_id`, so `holdings.security_id` referenced a row that no longer existed in `securities` — orphan reference. For any user with a user-minted security (via `/api/validate-ticker`) that later shows up in a Plaid sync, the holding renders with a truncated `security_id` instead of the ticker/name, MWR's snapshot-price lookups miss, and the divergence detection (#589) can't line the holding up with the user's manual invtxs.

## Fix

Skip the dedupe. Every incoming `security_id` becomes its own row. A user-minted row and a provider's row for the same ticker both survive as separate `securities` rows.

Why that's safe:
- `HoldingsComposition` buckets by `ticker_symbol`, not `security_id`. Both rows fold into ONE visible bucket with the aggregate qty and value the user actually holds. No visible duplication.
- MWR's per-security `valueAt` / cash-flow extraction runs independently for each `security_id` and sums into the portfolio total.
- Divergence detection (#589) is per-security. The user-minted row vs user's manual invtxs and the provider's row vs provider invtxs each compute independently.
- No FK on `securities.security_id`; no uniqueness constraint on `ticker_symbol`. The DB allows two rows with the same ticker.

What we accept:
- `securities` denormalized. A query filtered by `ticker_symbol` returns two rows for users hitting this overlap. Not a data-quality issue, just untidy.
- Two parallel `security_snapshots` chains (one per security_id), both refreshed by the scheduled security-snapshot cron.

## Related

- Supersedes closed #598 (which took a heavier approach: DB remap + atomic tx + advisory locks). Discussion in the "Budget holdings UI" Discord thread landed on the simpler drop-dedupe shape.
- Also reorders `sync-plaid.ts` to write securities BEFORE holdings (matches `sync-simple-fin.ts`).
- `upsertSecuritiesWithSnapshots` return shape changed from `{ incomingId → canonicalId }` idMap to `Set<string>` of successfully-upserted security_ids. `sync-simple-fin.ts` filters holdings by that set to drop holdings whose security was skipped (no ticker_symbol / no close_price).

## Test plan

- [x] `bun test src/server` — 693 pass / 2 skip / 0 fail
- [ ] Sandbox UI verification: cold-load an account with the manual + provider overlap, confirm HoldingsComposition renders one ticker row with the aggregated qty from both, and the widget's MWR / divergence detection resolves the ticker/name correctly.

### DIFF
--- a/src/server/lib/compute-tools/create-snapshots.ts
+++ b/src/server/lib/compute-tools/create-snapshots.ts
@@ -98,26 +98,12 @@ export const upsertAndDeleteHoldingsWithSnapshots = async (
 };
 
 /**
- * Upsert incoming securities from a sync (Plaid, SimpleFin) and write a
- * per-day security snapshot for each. The incoming `security_id` is
- * treated as identity — no ticker-collision dedupe. When a user has
- * previously minted a security row for the same ticker (via
- * `/api/validate-ticker`) and Plaid later syncs its own row for the
- * same ticker, both survive as separate `securities` rows.
- *
- * That's intentional: HoldingsComposition buckets by ticker (not by
- * security_id), so the FE aggregates the two into one visual row and
- * users don't see the internal duplication. The pre-dedupe behavior
- * (`newSecurity.security_id = existingSecurity.security_id`) caused
- * an orphan-reference bug — Plaid's holdings were written with
- * Plaid's raw security_id but the securities table was rewritten
- * against the user-minted id, leaving holdings pointing at a row
- * that no longer existed. Skipping dedupe removes the bug in one
- * line.
- *
- * Returns the set of successfully-upserted `security_id`s so callers
- * (`sync-simple-fin.ts` in particular) can drop holdings whose security
- * was filtered out here (no ticker or no close price).
+ * Upsert incoming securities and write today's snapshot for each.
+ * `security_id` is treated as identity — two rows can share a ticker
+ * (user-minted + provider-synced), and `HoldingsComposition` folds them
+ * by ticker on read. Returns the set of successfully-upserted
+ * `security_id`s so callers can drop holdings whose security was
+ * filtered out here (no ticker or no close price).
  */
 export const upsertSecuritiesWithSnapshots = async (
   securities: JSONSecurity[],
--- a/src/server/lib/compute-tools/create-snapshots.ts
+++ b/src/server/lib/compute-tools/create-snapshots.ts
@@ -1,25 +1,21 @@
 import {
   JSONAccount,
   JSONAccountSnapshot,
-  getDateString,
   getSquashedDateString,
   JSONHolding,
   JSONHoldingSnapshot,
   isEqual,
   JSONSecurity,
   JSONSecuritySnapshot,
-  LocalDate,
 } from "common";
 import {
   deleteHoldings,
   MaskedUser,
-  searchSecurities,
   upsertAccounts,
   upsertHoldings,
   upsertSecurities,
   upsertSnapshots,
 } from "server";
-import { getSecurityForSymbol } from "../polygon";
 
 export const upsertAccountsWithSnapshots = async (
   user: MaskedUser,
@@ -101,60 +97,52 @@ export const upsertAndDeleteHoldingsWithSnapshots = async (
   await deleteHoldings(user, removedHoldingIds);
 };
 
-export const upsertSecuritiesWithSnapshots = async (securities: JSONSecurity[]) => {
+/**
+ * Upsert incoming securities from a sync (Plaid, SimpleFin) and write a
+ * per-day security snapshot for each. The incoming `security_id` is
+ * treated as identity — no ticker-collision dedupe. When a user has
+ * previously minted a security row for the same ticker (via
+ * `/api/validate-ticker`) and Plaid later syncs its own row for the
+ * same ticker, both survive as separate `securities` rows.
+ *
+ * That's intentional: HoldingsComposition buckets by ticker (not by
+ * security_id), so the FE aggregates the two into one visual row and
+ * users don't see the internal duplication. The pre-dedupe behavior
+ * (`newSecurity.security_id = existingSecurity.security_id`) caused
+ * an orphan-reference bug — Plaid's holdings were written with
+ * Plaid's raw security_id but the securities table was rewritten
+ * against the user-minted id, leaving holdings pointing at a row
+ * that no longer existed. Skipping dedupe removes the bug in one
+ * line.
+ *
+ * Returns the set of successfully-upserted `security_id`s so callers
+ * (`sync-simple-fin.ts` in particular) can drop holdings whose security
+ * was filtered out here (no ticker or no close price).
+ */
+export const upsertSecuritiesWithSnapshots = async (
+  securities: JSONSecurity[],
+): Promise<Set<string>> => {
   const newSecurities: JSONSecurity[] = [];
   const snapshots: JSONSecuritySnapshot[] = [];
-  const idMap: { [key: string]: string } = {};
+  const upsertedIds = new Set<string>();
 
-  const promises = securities.map(async (s) => {
-    const { security_id, ticker_symbol, close_price, close_price_as_of } = s;
-    if (!ticker_symbol) return;
-    if (!close_price || !close_price_as_of) return;
+  for (const s of securities) {
+    const { ticker_symbol, close_price, close_price_as_of } = s;
+    if (!ticker_symbol) continue;
+    if (!close_price || !close_price_as_of) continue;
 
     const newSecurity: JSONSecurity = { ...s };
-
-    const storedSecurity = await searchSecurities({ ticker_symbol });
-    if (storedSecurity.length) {
-      const existingSecurity: JSONSecurity = { ...storedSecurity[0] };
-      newSecurity.security_id = existingSecurity.security_id;
-      const snapshot_id = `${existingSecurity.security_id}-${getSquashedDateString()}`;
-
-      const existingDateString = existingSecurity.close_price_as_of;
-      const existingDate = existingDateString && new LocalDate(existingDateString);
-      if (existingDate) {
-        if (existingDate < new LocalDate(close_price_as_of)) {
-          snapshots.push({
-            snapshot: { snapshot_id, date: new Date().toISOString() },
-            security: newSecurity,
-          });
-        } else if (existingDate < new LocalDate(getDateString())) {
-          const todaySecurity = await getSecurityForSymbol(ticker_symbol);
-          if (todaySecurity) {
-            newSecurity.close_price = todaySecurity.close_price;
-            newSecurity.close_price_as_of = todaySecurity.close_price_as_of;
-            snapshots.push({
-              snapshot: { snapshot_id, date: new Date().toISOString() },
-              security: newSecurity,
-            });
-          }
-        }
-      }
-    } else {
-      const snapshot_id = `${newSecurity.security_id}-${getSquashedDateString()}`;
-      snapshots.push({
-        snapshot: { snapshot_id, date: new Date().toISOString() },
-        security: newSecurity,
-      });
-    }
+    const snapshot_id = `${newSecurity.security_id}-${getSquashedDateString()}`;
+    snapshots.push({
+      snapshot: { snapshot_id, date: new Date().toISOString() },
+      security: newSecurity,
+    });
 
     newSecurities.push(newSecurity);
-    idMap[security_id] = newSecurity.security_id;
-    return;
-  });
+    upsertedIds.add(newSecurity.security_id);
+  }
 
-  await Promise.all(promises);
   await upsertSecurities(newSecurities);
   await upsertSnapshots(snapshots);
-
-  return idMap;
+  return upsertedIds;
 };
--- a/src/server/lib/compute-tools/sync-plaid.ts
+++ b/src/server/lib/compute-tools/sync-plaid.ts
@@ -324,8 +324,15 @@ export const syncPlaidAccounts = async (item_id: string) => {
       const allHoldings = inferredCash.length ? [...holdings, ...inferredCash] : holdings;
 
       await upsertAccountsWithSnapshots(user, accounts, storedAccounts);
-      await upsertAndDeleteHoldingsWithSnapshots(user, allHoldings, storedHoldings);
+      // Upsert securities BEFORE holdings so `holdings.security_id`
+      // references a row that already exists in `securities` at every
+      // instant of the sync. `holdings` has no FK on `security_id` so
+      // reversing the order isn't a hard error — just a transient
+      // inconsistency window until `upsertSecuritiesWithSnapshots`
+      // catches up — but the securities-first order matches the
+      // sync-simple-fin.ts shape and closes #593 gap 2.
       await upsertSecuritiesWithSnapshots(securities);
+      await upsertAndDeleteHoldingsWithSnapshots(user, allHoldings, storedHoldings);
 
       return accounts;
     })
--- a/src/server/lib/compute-tools/sync-plaid.ts
+++ b/src/server/lib/compute-tools/sync-plaid.ts
@@ -324,13 +324,8 @@ export const syncPlaidAccounts = async (item_id: string) => {
       const allHoldings = inferredCash.length ? [...holdings, ...inferredCash] : holdings;
 
       await upsertAccountsWithSnapshots(user, accounts, storedAccounts);
-      // Upsert securities BEFORE holdings so `holdings.security_id`
-      // references a row that already exists in `securities` at every
-      // instant of the sync. `holdings` has no FK on `security_id` so
-      // reversing the order isn't a hard error — just a transient
-      // inconsistency window until `upsertSecuritiesWithSnapshots`
-      // catches up — but the securities-first order matches the
-      // sync-simple-fin.ts shape and closes #593 gap 2.
+      // Securities first so holdings never reference a not-yet-written
+      // securities row within the same sync.
       await upsertSecuritiesWithSnapshots(securities);
       await upsertAndDeleteHoldingsWithSnapshots(user, allHoldings, storedHoldings);
 
--- a/src/client/lib/hooks/calculation/benchmark.ts
+++ b/src/client/lib/hooks/calculation/benchmark.ts
@@ -661,10 +661,20 @@ export const findBenchmarkSecurityId = (
   securitySnapshots: SecuritySnapshotDictionary,
   ticker: string,
 ): string | null => {
+  // Two rows can share a ticker (user-minted + provider-synced). Pick
+  // the security_id whose snapshot chain has the freshest close so
+  // benchmark math uses the most-updated series. Falls back to first-
+  // iterating for ties on missing date (rare — snapshots always carry
+  // a date).
   let found: string | null = null;
+  let foundDate: string | null = null;
   securitySnapshots.forEach((snap) => {
-    if (found) return;
-    if (snap.security.ticker_symbol === ticker) found = snap.security.security_id;
+    if (snap.security.ticker_symbol !== ticker) return;
+    const d = snap.snapshot.date;
+    if (!found || (d && (!foundDate || d > foundDate))) {
+      found = snap.security.security_id;
+      foundDate = d ?? null;
+    }
   });
   return found;
 };
--- a/src/server/lib/compute-tools/create-snapshots.test.ts
+++ b/src/server/lib/compute-tools/create-snapshots.test.ts
@@ -1,0 +1,104 @@
+import { describe, test, expect, mock, beforeEach, afterAll } from "bun:test";
+import { restoreLeaves } from "test-helpers";
+import type { JSONSecurity } from "common";
+
+const mockQuery = mock(async (_sql: string, _values?: unknown[]) => ({
+  rows: [] as unknown[],
+  rowCount: 0 as number | null,
+}));
+
+class FakePool {
+  query = mockQuery;
+  end = async () => {};
+  connect = async () => ({ query: mockQuery, release: () => {} });
+}
+
+mock.module("pg", () => ({
+  Pool: FakePool,
+  types: { setTypeParser: () => {} },
+  default: { Pool: FakePool, types: { setTypeParser: () => {} } },
+}));
+
+const { upsertSecuritiesWithSnapshots } = await import("./create-snapshots");
+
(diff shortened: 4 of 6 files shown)


reviewoie (commented):
## Code review — PR 602

**Verdict:** comment (informational — no blockers).

**Findings:** 5 total — 0 high, 2 med, 3 low.

Core semantic change (drop the ticker-collision dedupe, treat `security_id` as identity) is defensible. Downstream consumers checked — `HoldingsComposition` and `HoldingProperties` both bucket by ticker and handle multiple `security_id`s per bucket; `computeQtyDivergence` runs per `security_id` independently; `refresh-security-snapshots.ts` and `backfillMonthlySecuritySnapshotsForward` operate per-id (so a collision doubles their per-ticker workload, acknowledged in the PR body). Filter migration (`idMap[k]` → `upsertedIds.has(k)`) is behaviorally equivalent since incoming `security_id` is never rewritten anymore. Order swap (securities → holdings) is safe: no FK on `holdings.security_id`, no consumer reads them jointly during the write window.

### Med

- **`src/server/lib/compute-tools/create-snapshots.ts:111-116`** — JSDoc PR-narrative. The paragraph describing the pre-dedupe behavior ("caused an orphan-reference bug… holdings were written with Plaid's raw security_id but the securities table was rewritten…") is a restatement of the bug being fixed. Per [[comment-hygiene-no-pr-narrative]], PR-narrative rots after merge — six months later the reader decodes a bug that no longer exists. Trim to the future-reader payload already in ¶1 (`security_id` is identity, not ticker) and ¶2 (why two rows is safe — FE buckets by ticker). Suggested: drop lines 111-116; ¶1/¶2 already carry the load-bearing invariants.
- **`src/server/lib/compute-tools/sync-plaid.ts:332-333`** — Comment tail `"but the securities-first order matches the sync-simple-fin.ts shape and closes 593 gap 2"` — PR-narrative (issue ref + "matches sibling" is context that doesn't help a future reader). The load-bearing why in lines 327-329 (`holdings.security_id` references an existing row; soft invariant, no FK) is fine. Truncate after "just a transient inconsistency window."

### Low

- **`src/server/lib/compute-tools/create-snapshots.ts:129-143`** — Every incoming security now writes today's snapshot unconditionally (old logic gated the write on freshness). Combined with `refresh-security-snapshots.ts`'s 22h skip window keyed on `snapshots.updated`, Plaid-synced securities' polygon-refresh path is now effectively neutered — every hourly Plaid sync bumps `updated = CURRENT_TIMESTAMP`, so the refresh cron always short-circuits at the cadence gate for those rows. Refresh cron's own docstring (lines 51-54) already assumes this behavior, so the PR aligns with intent — but the ex-ante contract (Polygon as an authoritative daily-close backstop for Plaid securities) is functionally gone. Non-blocking because Plaid's and Polygon's daily closes should agree, but worth confirming this is the desired end state.
- **`src/client/lib/hooks/calculation/benchmark.ts:660-670` + `src/server/routes/accounts/post-validate-ticker.ts:41` + `src/server/routes/accounts/post-holding-snapshot.ts:32-33`** — Three places silently pick `.securities[0]` / first-iterating snapshot for a given ticker: `findBenchmarkSecurityId`, `post-validate-ticker`'s existence probe, and `resolveSecurityId`. Under the two-rows-per-ticker world, each picks non-deterministically (PG physical order / `Map` insertion order). No user-visible correctness impact if both rows agree on close_price + refresh cadence, but a future dev hits surprising non-determinism if the two rows drift (e.g. one has Polygon-backfilled snapshots, the other doesn't). Consider naming the tiebreaker explicitly ("prefer the provider row over the user-minted row when both exist") in a follow-up, or documenting the arbitrary-pick semantic.
- **`src/server/lib/compute-tools/create-snapshots.ts` (no test change)** — no test file changes and no existing test covered the collision branch pre-PR. Adding a `create-snapshots.test.ts` with three cases would pin the new contract: (a) two securities with same ticker + different `security_id`s both survive as separate rows and both ids appear in the returned Set; (b) securities lacking `ticker_symbol` OR `close_price` are dropped from the Set; (c) `sync-simple-fin`'s `holdings.filter(h => upsertedIds.has(h.security_id))` drops holdings whose security was filtered. Not blocking, but the collision behavior is now a load-bearing invariant with no regression test.

No further blockers.

src/server/lib/compute-tools/create-snapshots.ts
```diff
@@ -101,60 +97,52 @@ export const upsertAndDeleteHoldingsWithSnapshots = async (
...
+ *
+ * That's intentional: HoldingsComposition buckets by ticker (not by
+ * security_id), so the FE aggregates the two into one visual row and
+ * users don't see the internal duplication. The pre-dedupe behavior
+ * (`newSecurity.security_id = existingSecurity.security_id`) caused
+ * an orphan-reference bug — Plaid's holdings were written with
+ * Plaid's raw security_id but the securities table was rewritten
+ * against the user-minted id, leaving holdings pointing at a row
+ * that no longer existed. Skipping dedupe removes the bug in one
+ * line.
```
reviewoie: **MED — comment hygiene: PR-narrative in JSDoc.** This paragraph (lines 111-116) describes "the pre-dedupe behavior" and "caused an orphan-reference bug" — a restatement of the bug this PR fixes. Per [[comment-hygiene-no-pr-narrative]], PR-narrative rots after merge and forces a future reader to decode a bug that no longer exists. The two surrounding paragraphs (¶1 identity-not-ticker, ¶2 FE buckets by ticker) already carry the load-bearing invariants a future reader needs. Recommend dropping lines 111-116 outright; the PR body already carries the fix narrative.

src/server/lib/compute-tools/sync-plaid.ts
```diff
@@ -324,8 +324,15 @@ export const syncPlaidAccounts = async (item_id: string) => {
...
 
       await upsertAccountsWithSnapshots(user, accounts, storedAccounts);
-      await upsertAndDeleteHoldingsWithSnapshots(user, allHoldings, storedHoldings);
+      // Upsert securities BEFORE holdings so `holdings.security_id`
+      // references a row that already exists in `securities` at every
+      // instant of the sync. `holdings` has no FK on `security_id` so
+      // reversing the order isn't a hard error — just a transient
+      // inconsistency window until `upsertSecuritiesWithSnapshots`
+      // catches up — but the securities-first order matches the
+      // sync-simple-fin.ts shape and closes #593 gap 2.
```
reviewoie: **MED — comment hygiene.** "but the securities-first order matches the sync-simple-fin.ts shape and closes 593 gap 2" — the issue reference + "matches sibling shape" is PR-narrative. Lines 327-329 (`holdings.security_id` references an existing row; no FK, so it's a soft invariant not a hard one) are the load-bearing constraint a future reader needs. Recommend truncating the comment after line 331 ("just a transient inconsistency window.").

src/server/lib/compute-tools/create-snapshots.ts
```diff
@@ -101,60 +97,52 @@ export const upsertAndDeleteHoldingsWithSnapshots = async (
...
-      }
-    } else {
-      const snapshot_id = `${newSecurity.security_id}-${getSquashedDateString()}`;
-      snapshots.push({
-        snapshot: { snapshot_id, date: new Date().toISOString() },
-        security: newSecurity,
-      });
-    }
+    const snapshot_id = `${newSecurity.security_id}-${getSquashedDateString()}`;
+    snapshots.push({
```
reviewoie, line 122: **LOW — informational.** Every incoming security now writes today's snapshot unconditionally (pre-PR the write was gated on freshness / stale-existing). Combined with `refresh-security-snapshots.ts`'s 22h skip window keyed on `snapshots.updated`, hourly Plaid syncs will now perpetually short-circuit the polygon-backfill path for Plaid-tracked securities — every sync upserts the snapshot → bumps `updated = NOW()` → 22h gate always fresh. `refresh-security-snapshots.ts:51-54` docstring already assumes this. Confirm that's the desired end state (Polygon = fallback-only for securities Plaid doesn't cover; Plaid's daily close = authoritative for Plaid-tracked).

src/server/lib/compute-tools/create-snapshots.ts
```diff
@@ -101,60 +97,52 @@ export const upsertAndDeleteHoldingsWithSnapshots = async (
...
+    upsertedIds.add(newSecurity.security_id);
+  }
 
-  await Promise.all(promises);
   await upsertSecurities(newSecurities);
   await upsertSnapshots(snapshots);
-
-  return idMap;
+  return upsertedIds;
 };
```
reviewoie, line 134: **LOW — test coverage.** No test file changed; no existing test exercised the collision branch pre-PR. The `security_id`-is-identity contract is now load-bearing — worth pinning with a small `create-snapshots.test.ts`: (a) two securities with the same ticker but different `security_id`s both survive and both ids appear in the returned Set (dedupe is truly gone); (b) securities lacking `ticker_symbol` or `close_price*` are excluded from the Set (invariant relied on by `sync-simple-fin.ts:76`'s `holdings.filter`); (c) the returned Set contains the incoming ids, not any rewritten form. Non-blocking.

claoie:
Thanks reviewoie — all 5 addressed in `729e625c`.

**MED-1 (comment hygiene, `create-snapshots.ts:111-116`)** — collapsed the JSDoc paragraph describing the pre-dedupe bug into a single sentence naming the identity contract and the FE-buckets-by-ticker rationale. Fix-narrative belongs in the PR body.

**MED-2 (comment hygiene, `sync-plaid.ts:332-333`)** — trimmed the "#593 gap 2 / matches sibling shape" tail. Order is self-explanatory from the two adjacent calls.

**LOW-1 (snapshot cadence)** — no action, agreed it's aligned with the cron's docstring intent. Just wanted the informational surface.

**LOW-2 (silent-pick-first-of-two)** — made the picks deterministic:
- `searchSecurities` now sorts `updated DESC` server-side, so `.securities[0]` at `post-validate-ticker.ts:41`, `post-holding-snapshot.ts:32-33`, and any future caller reliably gets the most-recently-touched row.
- `findBenchmarkSecurityId` (`benchmark.ts:660`) prefers the security_id whose snapshot chain carries the freshest close_price — same "most recent data wins" principle.

**LOW-3 (test coverage)** — added `create-snapshots.test.ts` with 5 tests: two-securities-same-ticker both survive, no-ticker skipped, no-close skipped, empty-input, and the identity-not-ticker regression guard (returns incoming id even if a hypothetical ticker-search would have returned a different existing id).

`bun test src/server src/client/lib/hooks/calculation/benchmark.test.ts` 754/2/0; typecheck + lint clean.

reviewoie:
## Code review — PR 602 (round 2)

Second pass after fix commit `729e625c`. Delta: benchmark rewrite, ORDER BY on `searchSecurities`, comment trims, new tests. Full server + benchmark client suites green locally (698 pass, 56 pass).

**Verdict:** informational — no blocking findings. The fix commit lands the round 1 asks cleanly. Two small gaps and a follow-up worth capturing.

**Findings:** 3 total — 0 high, 0 med, 3 low/informational.

### High
No further blockers.

### Med
No further blockers.

### Low / informational

- `src/client/lib/hooks/calculation/benchmark.ts:660-680` — **no unit test for the `findBenchmarkSecurityId` rewrite.** Pure function, 4 verifiable branches (no match / single match / multi-match different dates / dated-vs-null-date promotion). I traced every branch by hand and the logic is correct — including the tricky "first match had null date, second has a real date" case where the `d && (!foundDate || d > foundDate)` arm correctly promotes to the dated snapshot. But nothing pins it: `benchmark.test.ts` has 56 tests, none named `findBenchmarkSecurityId`. A ~15-line describe with 4 tests (no-match returns null / single-match returns id / multi-match picks freshest / null-date entry is superseded by dated entry) would guard against a future refactor silently flipping the loop condition.

- `src/server/lib/compute-tools/create-snapshots.test.ts:89-103` — **the `mockImplementationOnce` gets consumed by the first UPSERT query the fixed code makes anyway, not by a `searchSecurities` SELECT (which the fixed code no longer calls).** The test comment names the queued mock as "what `searchSecurities` would return if the function still ticker-searched," but with the search removed, that mock instead lands on the first `INSERT INTO securities` UPSERT (which ignores the response). The assertions still pass because `upsertedIds.add(newSecurity.security_id)` uses the incoming id verbatim. The test IS a valid regression guard — if someone reintroduces the search+rewrite, the FIRST call would be a SELECT that would consume this mock and cause the old collision branch to rewrite `security_id` to `existing-old-id`, flipping both assertions. So the guard works. But the mechanism is indirect — the mock only "trips" when a hypothetical regression puts a SELECT first. Not worth changing; noting for future readers.

- `src/server/lib/compute-tools/refresh-security-snapshots.ts:80-116` — **when two `security_id`s share a ticker, polygon spend doubles for that ticker per cycle** (once per security_id, after each's 22h cadence window elapses). `SELECT DISTINCT security_id FROM holdings` returns both; each hits polygon for the same ticker string; each writes its own snapshot chain. Not a correctness bug — the per-`security_id` cadence gate and snapshot-id keying (`${security_id}-${date}`) both work correctly, snapshot chains stay independent, and `HoldingsComposition` folds by ticker on display. Just a spend-doubling artifact of the "identity ≠ ticker" world. Only matters at scale if many users end up with duplicate manual+plaid pairs; a follow-up perf issue would be reasonable ("de-dup polygon fetch by ticker in `refreshActiveSecuritySnapshots`") but not required for this PR.

### Delta-specific spot-checks that came back clean

- **`ORDER BY updated DESC` regression risk (fix commit item 3).** All three `searchSecurities` callers take `.securities[0]`: `cash-holding.ts:29` (ensure-USD-cash), `post-validate-ticker.ts:41` (does ticker exist), `post-holding-snapshot.ts:32` (ticker → security_id resolve). Previously the query had no explicit `ORDER BY` (Postgres-implementation-defined). No caller and no test depended on the old order — I read each and grepped for length/index-based iteration patterns. `securities` table has `updated TIMESTAMPTZ DEFAULT CURRENT_TIMESTAMP` (`models/security.ts:35`) and `buildUpsert` bumps it (`database.ts:198,217`), so the ordering key is populated on both insert and update. `[0]` now deterministically picks the most-recently-touched row per ticker, which is the semantic the JSDoc claims.
- **`findBenchmarkSecurityId` loop correctness (fix commit item 4).** Traced each of the 8 scenarios enumerated in the ask + type-check on `JSONSnapshot.date` (`common/models/Snapshot.ts:6` — required string) and client `Snapshot` (`Snapshot.ts:18` — defaulted to `new Date().toISOString()`). The `d && (!foundDate || d > foundDate)` guard correctly rejects stale dates AND promotes from a null-dated first-match to a dated later-match. No misbehavior when the same `security_id` appears in multiple snapshots — the loop converges on that `security_id`'s freshest date, then never re-selects until a different `security_id` with a strictly-fresher date appears.
- **`sync-simple-fin.ts` filter-not-remap (from the full-PR diff, called out as "behaviorally equivalent").** Confirmed — the OLD `idMap: {[k]: string}` remapped `h.security_id` to whatever the ticker-collision-deduped canonical id was. The NEW `upsertedIds: Set<string>` never remaps; it just filters holdings whose security got dropped (no ticker / no close price). Under the new identity-is-id semantic, SimpleFin's provided `h.security_id` IS the canonical id, so no remap is needed. Behaviorally equivalent modulo the collision branch that was the bug.
- **`create-snapshots.test.ts` framework shape.** Follows the canonical `users.test.ts` pg-FakePool pattern (leaf-mock `pg`, `afterAll(restoreLeaves)`, no barrel mock). Not process-global-leaky.

hoiekim (approved):

PR #603, opened 2026-07-07
refactor(client): migrate remaining pages to the `<Page>` shell (#326 phase 1.5)

claoie:
## Summary

Continues #326 (design-component refactoring). PR #473 landed the `<Page>` shell and migrated 4 detail pages; the remaining 9 pages still opened with raw `<div className="XxxPage">`. Migrated them all so the shared page convention (currently `padding: 0 10px` on direct-child sections) is inherited instead of re-declared per-page.

## Migrated (9 pages)

- AccountDetailPage
- AccountsPage — `classNames.join(" ")` string passed through
- BudgetsPage
- BudgetDetailPage
- BudgetConfigPage
- ChartDetailPage (3 render branches — BALANCE, PROJECTION, FLOW)
- DashboardPage
- TransactionDetailPage (2 render branches — investment tx + regular / transfer)
- TransactionsPage

Each edit is mechanical: swap outer `<div>` → `<Page>` and its closing tag, add `Page` to the `client/components` import. Per-page `className` and page-specific CSS unchanged.

## Related

- #326 stays open — remaining phases:
  - Properties shell migration for AccountProperties / BudgetProperties / ApiKeyProperties / TransactionProperties.
  - Sweep for NEW common patterns to extract (form-row shapes, section headers, button clusters).

## Test plan

- [x] `bun run typecheck` — clean
- [x] `bun --bun eslint src` — clean
- [x] `bun test src/client src/server` — 1010 pass / 2 skip / 0 fail
- [ ] Sandbox UI verification: visit each migrated page and confirm layout unchanged vs. main.

PR #604, opened 2026-07-07
refactor(client): migrate remaining `*Properties` panels to the `<Properties>` shell (#326 phase 3)

claoie:
## Summary

Continues #326 (design-component refactoring). PR #479 landed the
`<Properties>` / `<PropertyLabel>` / `<Property>` / `<Row>` shell and migrated
ConnectionProperties + HoldingProperties. The remaining six `*Properties`
panels still hand-rolled `<div className="XxxProperties Properties">` plus raw
`propertyLabel` / `property` / `row` divs. This migrates all six so the shared
shell convention is inherited instead of re-declared per panel — the direct
parallel, on the Properties axis, to #603's Page-shell migration.

## Migrated (6 panels)

- **AccountProperties** — many conditional sections; non-shell direct children (`<Graph>`, `<br/>`, `<HoldingsComposition>`, `<PerformanceBenchmark>`) and the `.amount` wrapper left as-is.
- **ApiKeyProperties** — all 5 render branches (just-created / create-form / loading / not-found / detail).
- **BudgetProperties** — RadioInputs / CapacitiesInput sub-components left raw (out of scope).
- **InvestmentTransactionProperties**
- **TransactionProperties** — incl. the dynamic ``className={`row partnerCandidate…`}`` row (→ `<Row className={`partnerCandidate…`}>`, all `role`/`tabIndex`/`aria`/handlers passed through `...rest`), the split-input rows, and the `.SplitTransactionRow` / `.notification` inner divs left untouched.
- **TransferProperties** — the per-side `renderSide` fragment keeps its label/property as direct Properties children.

## Why it's safe (DOM-preserving)

Each swap emits the **identical class string** because the shell merges in the
same order the raw markup used: `Properties` is appended (`XxxProperties Properties`),
`row` / `property` / `propertyLabel` are prepended (`row keyValue`, etc.). The
shell components spread `...rest`, so `onClick` / `key` / `aria-*` / `role` all
pass through. Every non-shell div (`.amount`, `.notification`,
`.SplitTransactionRow`, `.partnerCandidateInfo`, plain input wrappers) is left
exactly as it was, so the `div.Properties > .property` / `> .propertyLabel`
direct-child selectors still match.

## Scope

`#326` intentionally stays **open** — remaining phases per its checklist:
sub-component internals (RadioInputs / CapacitiesInput / SplitTransactionRow)
and a sweep for NEW common patterns to extract (section headers, button
clusters). No closing keyword here.

## Test plan

- [x] `bun --bun eslint src/client/components/{Account,ApiKey,Budget,InvestmentTransaction,Transaction,Transfer}Properties` — clean
- [x] `bun test src/client` — 312 pass / 0 fail
- [x] Open/close tag balance verified per file (opens == closes, no self-closing shell tags)
- [ ] Sandbox UI: open each of the 6 detail panels at mobile viewport and eye-check layout unchanged vs. main

### DIFF
--- a/src/client/components/AccountProperties/index.css
+++ b/src/client/components/AccountProperties/index.css
@@ -4,9 +4,7 @@ div.AccountProperties .property > .row > .amount {
 }
 
 div.AccountProperties .property > .row > .amount > input {
-  font-size: 16px;
   width: calc(100% - 5px);
-  text-align: right;
 }
 
 div.AccountProperties .property > .row > .amount > input:disabled {
--- a/src/client/components/AccountProperties/index.css
+++ b/src/client/components/AccountProperties/index.css
@@ -1,8 +1,0 @@
-div.AccountProperties .property > .row > .amount {
-  display: flex;
-  align-items: center;
-}
-
-div.AccountProperties .property > .row > .amount > input:disabled {
-  color: inherit;
-}
--- a/src/client/components/AccountProperties/index.css
+++ b/src/client/components/AccountProperties/index.css
@@ -3,14 +3,6 @@ div.AccountProperties .property > .row > .amount {
   align-items: center;
 }
 
-div.AccountProperties .property > .row > .amount > input {
-  width: calc(100% - 5px);
-}
-
 div.AccountProperties .property > .row > .amount > input:disabled {
   color: inherit;
-}
-
-div.AccountProperties .property > .row > select {
-  text-align: right;
 }
--- a/src/client/components/Configuration/index.css
+++ b/src/client/components/Configuration/index.css
@@ -1,8 +1,3 @@
-div.Configuration button.delete {
-  font-weight: 500;
-  color: var(--darkRed);
-}
-
 div.Configuration .row > .connection {
   display: flex;
   justify-content: space-between;
--- a/src/client/components/Configuration/index.css
+++ b/src/client/components/Configuration/index.css
@@ -1,8 +1,3 @@
-div.Configuration {
-  margin-top: 20px;
-  padding: 0 10px;
-}
-
 div.Configuration button.delete {
   font-weight: 500;
   color: var(--darkRed);
--- a/src/client/components/ConnectionProperties/index.css
+++ b/src/client/components/ConnectionProperties/index.css
@@ -1,4 +1,3 @@
 div.ConnectionProperties {
-  margin-top: 20px;
   padding: 0 10px;
 }
--- a/src/client/components/ConnectionProperties/index.css
+++ b/src/client/components/ConnectionProperties/index.css
@@ -1,3 +1,0 @@
-div.ConnectionProperties {
-  padding: 0 10px;
-}
--- a/src/client/components/TransactionProperties/index.css
+++ b/src/client/components/TransactionProperties/index.css
@@ -1,11 +1,3 @@
-div.TransactionProperties .row > .splitItem > input {
-  height: 100%;
-  width: 30%;
-  flex-grow: 1;
-  text-align: center;
-  background-color: #191919;
-}
-
 div.SplitTransactionRow {
   width: 100%;
   display: flex;
--- a/src/client/components/TransactionProperties/index.css
+++ b/src/client/components/TransactionProperties/index.css
@@ -6,10 +6,6 @@ div.TransactionProperties .row > .splitItem > input {
   background-color: #191919;
 }
 
-div.TransactionProperties .row select {
-  text-align: right;
-}
-
 div.SplitTransactionRow {
   width: 100%;
   display: flex;
@@ -32,7 +28,6 @@ div.SplitTransactionRow > .amount {
 
 div.SplitTransactionRow > .amount > input {
   background-color: #191919;
-  width: calc(100% - 5px);
   text-align: left;
 }
 
--- a/src/client/components/TransactionProperties/index.css
+++ b/src/client/components/TransactionProperties/index.css
@@ -1,5 +1,4 @@
 div.TransactionProperties .row > .splitItem > input {
-  font-size: 16px;
   height: 100%;
   width: 30%;
   flex-grow: 1;
@@ -32,7 +31,6 @@ div.SplitTransactionRow > .amount {
 }
 
 div.SplitTransactionRow > .amount > input {
-  font-size: 16px;
   background-color: #191919;
   width: calc(100% - 5px);
   text-align: left;
--- a/src/client/components/TransactionProperties/index.css
+++ b/src/client/components/TransactionProperties/index.css
@@ -21,11 +21,6 @@ div.SplitTransactionRow > div > select {
   width: 100%;
 }
 
-div.SplitTransactionRow > .amount {
-  display: flex;
-  align-items: center;
-}
-
 div.SplitTransactionRow > .amount > input {
   background-color: #191919;
   text-align: left;
--- a/src/client/components/AccountProperties/index.tsx
+++ b/src/client/components/AccountProperties/index.tsx
@@ -23,6 +23,10 @@ import {
   PerformanceBenchmark,
   InstitutionSpan,
   MoneyLabel,
+  Properties,
+  Property,
+  PropertyLabel,
+  Row,
   ToggleInput,
 } from "client/components";
 import { useAccountEventHandlers } from "./lib";
@@ -124,14 +128,14 @@ export const AccountProperties = ({ account }: Props) => {
     !isManualAccount && viewDate.getEndDate() >= latestViewDate.getEndDate();
 
   return (
-    <div className="AccountProperties Properties">
-      <div className="propertyLabel">Account&nbsp;Details</div>
-      <div className="property">
-        <div className="row keyValue">
+    <Properties className="AccountProperties">
+      <PropertyLabel>Account&nbsp;Details</PropertyLabel>
+      <Property>
+        <Row className="keyValue">
           <span className="propertyName">Name</span>
           <input type="text" value={nameInput} onChange={onChangeNameInput} placeholder={name} />
-        </div>
-        <div className="row keyValue">
+        </Row>
+        <Row className="keyValue">
           <span className="propertyName">Type</span>
           {isManualAccount ? (
             <select value={typeInput} onChange={onChangeTypeInput}>
@@ -144,34 +148,34 @@ export const AccountProperties = ({ account }: Props) => {
           ) : (
             <span>{toTitleCase(subtype || type)}</span>
           )}
-        </div>
-        <div className="row keyValue">
+        </Row>
+        <Row className="keyValue">
           <span className="propertyName">Institution</span>
           {isManualAccount ? (
             <span>Manual</span>
           ) : (
             <InstitutionSpan institution_id={account.institution_id} />
           )}
-        </div>
-      </div>
-      {(!isManualAccount || !!graphData.lines) && <div className="propertyLabel">Balance</div>}
+        </Row>
+      </Property>
+      {(!isManualAccount || !!graphData.lines) && <PropertyLabel>Balance</PropertyLabel>}
       {!isManualAccount && (
-        <div className="property">
-          <div className="row keyValue">
+        <Property>
+          <Row className="keyValue">
             <span className="propertyName">{currentLabel}</span>
             <span>
               {currencySymbol}&nbsp;
               {currentAmountString}
             </span>
-          </div>
-          <div className="row keyValue">
+          </Row>
+          <Row className="keyValue">
             <span className="propertyName">{pendingLabel}</span>
             <span>
               {currencySymbol}&nbsp;
               {pendingAmountString}
             </span>
-          </div>
-        </div>
+          </Row>
+        </Property>
       )}
       {!!graphData.lines && (
         <>
@@ -183,8 +187,8 @@ export const AccountProperties = ({ account }: Props) => {
             memoryKey={account_id}
           />
           <br />
-          <div className="property">
-            <div className="row keyValue">
+          <Property>
+            <Row className="keyValue">
               <span className="propertyName">{viewDate.toString()}&nbsp;balance</span>
               <div className="amount">
                 <DynamicCapacityInput
@@ -197,38 +201,38 @@ export const AccountProperties = ({ account }: Props) => {
                   onBlur={onChangeBalanceSnapshotInput}
                 />
               </div>
-            </div>
-          </div>
+            </Row>
+          </Property>
         </>
       )}
       {type === AccountType.Investment && <HoldingsComposition accounts={[account]} />}
       {type === AccountType.Investment && <PerformanceBenchmark accounts={[account]} />}
       {!isManualAccount && (
         <>
-          <div className="propertyLabel">Balance&nbsp;Graph&nbsp;Options</div>
-          <div className="property">
-            <div className="row keyValue">
+          <PropertyLabel>Balance&nbsp;Graph&nbsp;Options</PropertyLabel>
+          <Property>
+            <Row className="keyValue">
               <span className="propertyName">Use Transactions</span>
               <ToggleInput
                 checked={useTransactionsForGraph}
                 onChange={onClickUseTransactionsForGraph}
               />
-            </div>
-            <div className="row keyValue">
+            </Row>
+            <Row className="keyValue">
               <span className="propertyName">Use Snapshots</span>
               <ToggleInput checked={useSnapshotsForGraph} onChange={onClickUseSnapshotsForGraph} />
-            </div>
-          </div>
-          <div className="propertyLabel">Account&nbsp;Preference</div>
-          <div className="property">
-            <div className="row keyValue">
+            </Row>
+          </Property>
+          <PropertyLabel>Account&nbsp;Preference</PropertyLabel>
+          <Property>
+            <Row className="keyValue">
               <span className="propertyName">Default&nbsp;Budget</span>
               <select value={selectedBudgetIdLabel} onChange={onChangeBudgetSelect}>
                 <option value="">Select Budget</option>
                 {budgetOptions}
               </select>
-            </div>
-            <div className="row keyValue">
+            </Row>
+            <Row className="keyValue">
               <span className="propertyName">Archive</span>
               {/* Hide already removes the account from view entirely; archiving
                *  on top adds nothing and would surface in "Show archived (N)"
@@ -236,12 +240,12 @@ export const AccountProperties = ({ account }: Props) => {
                *  steer the user toward Unhide first if they want a different
                *  classification. Hoie 2026-06-25. */}
               <ToggleInput checked={isArchived} onChange={onClickArchive} disabled={isHidden} />
-            </div>
-            <div className="row keyValue">
+            </Row>
+            <Row className="keyValue">
               <span className="propertyName">Hide</span>
               <ToggleInput checked={isHidden} onChange={onClickHide} />
-            </div>
-          </div>
+            </Row>
+          </Property>
         </>
       )}
       {/* Orphan path: a manual account that was already archived (e.g. on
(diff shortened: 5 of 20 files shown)


claoie:
## Verification — reviewoie + UI E2E, clean

**reviewoie (budget-reviewer):** clean, no findings. Traced all six files adversarially and confirmed every swap emits a **byte-identical class string** (shell merge order matches the raw markup), the direct-child rule is honored (conditional blocks use fragments = no DOM node), no non-`div` element was converted, and all props (`key`/`role`/`tabIndex`/`aria`/handlers) pass through `...rest` on the dynamic `partnerCandidate` Row.

**UI E2E (mobile 390×844 sandbox, real click-through, eye-checked PNGs):**
- **AccountProperties** (investment branch — the most conditional panel): nav Accounts → account row. Renders `AccountProperties Properties` with 6 properties / 5 labels / 8 rows; every section boxed at `rgb(41,41,41)` / `5px` radius. Account Details, Balance (graph + snapshot row), Holdings Composition, Add, Navigate, Delete all framed correctly. ✅
- **TransactionProperties** (highest-risk — dynamic `partnerCandidate` Row + split rows): nav Transactions → row. Renders `TransactionProperties Properties`, boxed correctly; Details / Budgets / Split Transactions / Mark-as-Transfer sections intact. Clicking **Mark as Transfer** surfaced the migrated `row partnerPickerEmpty` branch correctly. ✅
- No page errors in either flow (only the pre-existing sandbox SW-registration warning).

**Static gates:** `tsc --noEmit` 0 errors · eslint clean · `bun test src/client` 312 pass/0 fail · per-file shell open/close tag counts balanced.

The four simpler panels (ApiKeyProperties, BudgetProperties, InvestmentTransactionProperties, TransferProperties) are DOM-preserving by the same construction and were reviewoie-verified; the two hardest render byte-perfect above.

hoiekim (changes requested):
By replacing inline divs with `Properties` component and its children, I'd expect that we could save some page-specific CSS scripts. Did you check if any of them could be stripped off?

hoiekim:
Fix CI

reviewoie:
## Code review — PR 604

**Verdict:** approve.

**Findings:** 0 total.

### Migration correctness (6 panels)

Verified `<Properties>` / `<Property>` / `<PropertyLabel>` / `<Row>` all merge className strings byte-identically to the pre-migration raw markup:
- `Properties`: `` `${className} Properties` `` → `AccountProperties Properties`, `TransactionProperties Properties`, etc. Same as the original literal.
- `PropertyLabel` / `Property` / `Row`: `` `<token> ${className}` `` when className supplied; bare token otherwise. Every `<Row className="keyValue">`, `<Row className="button">`, `<Row className={`partnerCandidate${disabled ? " disabled" : ""}`}>` produces the identical class attribute.
- No wrapping introduced — every `<PropertyLabel>` / `<Property>` is a direct child of `<Properties>` (fragment returns in `TransferProperties.renderSide` and the isManual delete blocks preserve direct-child semantics).

### Dynamic Row prop-forwarding

`TransactionProperties` `partnerCandidate` row: `Row`'s `RowProps = ComponentPropsWithoutRef<"div">` accepts and spreads `role`, `tabIndex`, `aria-disabled`, `onClick`, `onKeyDown` onto the native `<div>` via `...rest`. Matches original div verbatim.

### CSS strip correctness (6a241823)

Shell provides `div.Properties .row input { font-size: 16px; height: 18px; text-align: right }` (specificity 0,2,2). Verified each stripped rule's DOM target is reachable by the shell selector:

- `AccountProperties .property > .row > .amount > input` (specificity 0,4,2, later source) — kept `width: calc(100% - 5px)`. Stripped `font-size: 16px` and `text-align: right` are now inherited from the shell rule. Panel remains more specific for its unique declarations; nothing else it declared overrides the stripped values.
- `TransactionProperties .row > .splitItem > input` (specificity 0,3,2, later source) — kept `height: 100%`, `width: 30%`, `flex-grow: 1`, `text-align: center`, `background-color`. Stripped `font-size: 16px` falls through to shell. Panel's own `text-align: center` still wins over shell's `text-align: right` (higher specificity).
- `SplitTransactionRow > .amount > input` (specificity 0,2,2, later source — same as shell) — kept `background-color`, `width`, `text-align: left`. Stripped `font-size: 16px` falls through to shell (shell rule wins font-size since panel no longer declares it). Panel's `text-align: left` still wins over shell's `text-align: right` on source-order tie (barrel imports `Properties` before `TransactionProperties`). Load-order verified in `src/client/components/index.ts` (Properties at line 2, TransactionProperties at line 26).

### Sweep of remaining panels

Confirmed no additional shell duplicates:
- `BudgetProperties/index.css`: styles only `.RadioInputs` sub-component; no `.row`/`.property`/`.propertyLabel` rules.
- `BudgetProperties/CapacitiesInput/index.css`: `div.CapacitiesInput input` is inside `<Property>` directly, NOT inside `<Row>` — shell rule `div.Properties .row input` does not match, no conflict.
- `TransferProperties/index.css`: `.transferAmountInline`, `.transferChip.transferChipConfirmed`, `.unpairButton` (14px override is intentional divergence from shell's 16px).
- `ApiKeyProperties/index.css`: mono-font on `<code>` elements (not inputs), feature-specific `.apiKeyError` / `.apiKeyCopyOnceTitle`, panel `padding: 0 10px`.
- `HoldingProperties/index.css`: `.tickerField input { flex: 1; min-width: 0 }` (layout, not shell duplicate), feedback/formError, snapshot meta.
- `InvestmentTransactionProperties`: no CSS file, no rules exist.

### Scope

8 files touched: 6 `.tsx` migrations (Account, ApiKey, Budget, InvestmentTransaction, Transaction, Transfer) + 2 `.css` strips (Account, Transaction). Matches PR title. No unrelated changes.

### CI

`build` and `test` both green on 6a241823.

No further blockers.

src/client/components/AccountProperties/index.css
```diff
@@ -4,9 +4,7 @@ div.AccountProperties .property > .row > .amount {
 }
 
 div.AccountProperties .property > .row > .amount > input {
-  font-size: 16px;
   width: calc(100% - 5px);
-  text-align: right;
 }
```
hoiekim: Why is `width` still required? Should we move it to the global `Properties`'s CSS rule?

src/client/components/AccountProperties/index.css
```diff
@@ -3,16 +3,6 @@ div.AccountProperties .property > .row > .amount {
...
 
-div.AccountProperties .property > .row > .amount > input {
-  font-size: 16px;
-  width: calc(100% - 5px);
-  text-align: right;
-}
-
 div.AccountProperties .property > .row > .amount > input:disabled {
   color: inherit;
 }
```
hoiekim: Why should this be under `AccountProperties`? Why not global?

src/client/components/TransactionProperties/index.css
```diff
@@ -1,16 +1,11 @@
 div.TransactionProperties .row > .splitItem > input {
-  font-size: 16px;
   height: 100%;
   width: 30%;
   flex-grow: 1;
   text-align: center;
   background-color: #191919;
 }
```
hoiekim: Is `.splitItem` even used? Remove it if not used.

src/client/components/TransactionProperties/index.css
```diff
@@ -26,15 +21,8 @@ div.SplitTransactionRow > div > select {
...
-  display: flex;
-  align-items: center;
-}
-
 div.SplitTransactionRow > .amount > input {
-  font-size: 16px;
   background-color: #191919;
-  width: calc(100% - 5px);
   text-align: left;
 }
```
hoiekim, line 19: Why not moving this to global `Properties`?

src/client/components/ConnectionProperties/index.css
```diff
@@ -1,4 +1,3 @@
 div.ConnectionProperties {
-  margin-top: 20px;
   padding: 0 10px;
```
hoiekim: Why is this padding required?

reviewoie:
## Code review — PR 604 (sweep pass 2, head `72d6b5c3`)

**Findings:** 1 total — 0 high, 0 med, 1 informational.

Verified against the 4 sweep-2 asks in the caller brief, plus a full walk of the diff's 10 touched files.

### High
No blockers.

### Med
No blockers.

### Low / informational

- `src/client/components/TransactionProperties/index.tsx:449-457` — informational, no visible regression.
  The readonly "remaining amount" row also lives inside a `<div className="SplitTransactionRow">` and renders `<div className="amount"><span>…</span></div>` (text-only, no input child). Before this pass it was covered by the deleted `div.SplitTransactionRow > .amount { display: flex; align-items: center }`; the new shell rule `div.Properties .amount:has(> input)` does NOT match it (the child is a `<span>`, not an `<input>`), so it falls back to default block layout.
  Analyzed as visually indistinguishable: parent `.SplitTransactionRow` remains `display: flex; align-items: center` so the `.amount` block is still centered vertically within the row, and `.amount` has a single inline child (the `<span>`), so block vs flex layout produces the same visual for a single-inline-child container whose height is content-driven. No screen change expected. Called out only for the record — no action needed.

### Verified

1. `:has(> input)` selector correctness
   - `AccountProperties/index.tsx:192` — `.amount > DynamicCapacityInput` renders `<input>` as direct child (`CapacityInput.tsx:32`). Matches. Preserves prior `display: flex; align-items: center`.
   - `TransactionProperties/SplitTransactionRow.tsx:165` — `.amount` contains mixed text nodes + `<CapacityInput>` → `<input>` as direct child. Matches. Text/input mixed layout keeps its vertical centering (the case that actually needs flex).
   - `TransactionProperties/index.tsx:549` (`partnerCandidate .amount`) — text-only, no input. Correctly excluded. Its own `.partnerCandidate .amount { width: 90px; text-align: right; white-space: nowrap }` untouched; layout comes from parent `.partnerCandidateInfo` flex container.
   - `.amount` selectors outside `.Properties` scope (`AccountsDonut/index.css:42`, `TransactionsTable/index.css:27`, `TransactionsTable/InvestmentTransactionRow.tsx:46`, `TransactionsTable/TransactionRow.tsx:273`) are all outside `div.Properties`, unaffected.

2. `input:disabled { color: inherit }` scope broadening — safe. Only two `.amount > input` cases in the Properties tree:
   - AccountProperties balance input (disabled when synced) — intended to keep inherited color. Preserved.
   - SplitTransactionRow `CapacityInput` — grep for `disabled` in `SplitTransactionRow.tsx` returns nothing; never disabled. Broadening is a no-op there.

3. `ConnectionProperties { margin-top: 20px }` drop — safe. `ConnectionProperties/index.tsx:131` renders `<Properties className="ConnectionProperties">`, so `div.Properties { margin-top: 20px }` (`Properties/index.css:11`) applies via the merged classlist. Remaining `padding: 0 10px` in ConnectionProperties/index.css is unchanged.

4. `AccountProperties/index.css` deletion — safe.
   - Import removed from `AccountProperties/index.tsx` (line 33 in prior state).
   - No other `*.css` or `*.tsx` file references `./AccountProperties/index.css` or the `.AccountProperties` class beyond the top-level render (`AccountProperties/index.tsx:130`).

5. Scope — clean. `git diff upstream/main...HEAD --name-status` shows exactly the 10 expected files: 6 migrated `*Properties/index.tsx`, `Properties/index.css` (shell adds), `AccountProperties/index.tsx` (import drop) + `AccountProperties/index.css` (D), `TransactionProperties/index.css` (rules stripped), `ConnectionProperties/index.css` (margin-top drop). No unrelated changes.

6. Specificity preserved. The higher-specificity `div.SplitTransactionRow > .amount > input { background-color: #191919; text-align: left }` (`TransactionProperties/index.css:24-27`) still overrides the shell's `div.Properties .row input { text-align: right }` (0,0,3,1 vs 0,0,2,1). Editable split-row amount input keeps its dark-bg + left-align.

reviewoie:
## Code review — PR 604 (head `220dfe22`, span→disabled input + drop `.splitItem`)

**Findings:** 0 total. Both review-comment fixes verified end-to-end.

### High
No further blockers.

### Med
No further blockers.

### Low
No further blockers.

### Verified

1. **`.splitItem` dead-code deletion.** `grep -rn "splitItem" src/` returns empty after this commit. Only prior reference was the CSS rule itself; no `.tsx` renders it. Deletion is safe.

2. **span → `<input type="text" disabled />` at `TransactionProperties/index.tsx:454-458`.** DOM/CSS trace on the readonly remaining-amount cell:
   - Structure now mirrors the editable split rows in `SplitTransactionRow.tsx:165-169` — `<div className="amount">` + `{isIncome && +&nbsp;}` text + `{currencySymbol}&nbsp;` text + input sibling. `CapacityInput` is called without `prefix`, so both editable and readonly render the currency symbol as a text node OUTSIDE the input and put only the numeric string INSIDE — genuinely consistent.
   - Cascade: shell `.amount:has(> input)` now matches (previously excluded — the point of the change). Panel `div.SplitTransactionRow > .amount > input` (0,2,2) wins `background: #191919` + `text-align: left` over shell `.row input` right-align tie via load order (panel imports after `Properties/index.css`). Shell `.amount > input:disabled { color: inherit }` (0,3,2) beats global `input:not([type="checkbox"]):disabled { color: #666 }` (0,2,1) — disabled amount reads as the normal body #d1d1d1, not muted. Shell `.row input { height: 18px; font-size: 16px }` and `.amount > input { width: calc(100% - 5px) }` apply uniformly.
   - Overflow question (prompt Q4): input width = `calc(100% - 5px)` inside a 30%-of-row `.amount` container. Same math the editable split rows have been shipping under. If it fit for editable it fits for readonly with the same DOM.
   - `value={...} disabled` without `onChange` does NOT trigger React's controlled-input warning — React suppresses it when `disabled` or `readOnly` is set.
   - `numberToCommaString(Math.abs(remainingAmount))` returns plain digits + comma (`common/utils/index.ts:12`) — no currency prefix, no double-`$`.

3. **Cursor nit (out of scope, no action needed).** Global `input:not([type="checkbox"]) { cursor: text }` has no `:disabled` variant, so the disabled input shows a text I-beam instead of the UA `not-allowed`. Same behavior applies to `AccountProperties`'s disabled balance input (`index.tsx:194`) — this is app-wide today, not introduced here.

4. **Scope.** Latest commit `220dfe22` touches exactly `TransactionProperties/index.tsx` + `TransactionProperties/index.css`, matching the two review comments. No unrelated changes.

5. **Regression sweep across the full PR.** Only other readonly `<span>` values inside `.Properties` I could find:
   - `TransferProperties/index.tsx:138` — `.transferAmountInline` under `.row.keyValue`. Not inside `.amount`, not a peer of an editable-input pattern — different design pattern (key/value), no consistency migration owed.
   - `InvestmentTransactionProperties/index.tsx:382,396,406` — `.row.keyValue` value spans, editable variant is a bare `<input>` also outside any `.amount` wrapper. Same key/value design, no migration owed.
   No other `.amount > span` readonly to fold in — the reviewoie sweep-2 note is fully resolved.

reviewoie:
## Code review — PR 604 (4th pass, commits `5b075cae` + `6f7844dc`)

**Verdict:** informational. No shipping blockers; one factual correction on the divergence rationale in `5b075cae`.

### High

No further blockers.

### Med

- **`Configuration/index.css` divergence rationale is factually wrong.** The commit message for `5b075cae` (and the review-request context) states *"Configuration keeps its own `padding: 0 10px` because it's not a `<Properties>` panel (Configuration renders `<div className="Configuration">` directly, no shell wrapper)."* But `src/client/components/Configuration/index.tsx:109` actually renders `<div className="Configuration Properties">` — both classes. So the shell rule `div.Properties { padding: 0 10px; margin-top: 20px }` DOES apply to it. That makes `Configuration/index.css`'s `padding: 0 10px` AND `margin-top: 20px` redundant with the shell (same computed values, no cascade conflict — both `div.Configuration` and `div.Properties` are specificity 0,1,1 and set the identical values). No visual regression, but the divergence reasoning breaks and the same "shell owns layout" consolidation logic used to drop ApiKey/Holding/Connection self-owning padding applies here too. Suggest dropping the two rules from `Configuration/index.css` (either in this PR or a follow-up) so all Properties consumers actually rely on the shell.

### Low

- `src/client/pages/ChartAccountsPage.tsx:171` — `<div className="Properties sidePadding">`. The `sidePadding` class from `App/index.css:396` (padding-left: 10px; padding-right: 10px) is now redundant with the shell's `padding: 0 10px` on this element. Pre-existing, not introduced by this PR — flag only for a future cleanup pass.
- `AccountDetailPage/index.tsx:4-5` and `TransactionDetailPage/index.tsx:7-8` each carry a stray double blank line where the deleted `import "./index.css"` used to sit. Prettier/ESLint pass; purely cosmetic.

---

**Verification I ran (recall-mode sweep, matches the 9 checks in the review prompt):**

1. Every `<Properties>` / `className="Properties"` consumer was enumerated:
   - Typed component: `AccountProperties`, `BudgetProperties`, `InvestmentTransactionProperties`, `TransactionProperties`, `HoldingProperties` (4 return branches), `TransferProperties`, `ApiKeyProperties` (5 return branches), `ConnectionProperties`.
   - Raw `<div className="Properties">`: `BalanceChartProperties.tsx:97`, `FlowChartProperties.tsx:93`, `ProjectionChartProperties.tsx:177`, plus the two pre-existing outliers (`Configuration`, `ChartAccountsPage`).
   No `<Properties>` is ever nested inside another `<Properties>` element (grep confirmed) — no compounding-inset risk.

2. For every consumer whose containing page still has a `> div { padding: 0 10px }` rule (only `BudgetConfigPage` — verified with the tsx render: `.title`, `.statusBarWithText`, `<BudgetProperties>`, `<ActionButtons>` are its direct children), the shell rule (0,1,1) and the page rule (0,1,1) have the same specificity AND the same value → no cascade conflict, no visual change. Padding on the same element doesn't compound — confirms the commit-message claim.

3. `ChartDetailPage/index.tsx` renders `<Balance|Projection|FlowChartProperties>` each with a raw `<div className="Properties">`; shell padding matches. The kept rule `> div.Properties > div.ProjectionChartRow, > div.Properties > div.BalanceChartRow { margin-bottom: 20px }` still targets the correct DOM path (shell selector `div.Properties` at 0,1,1 doesn't interact with the descendant `margin-bottom` rule at 0,3,3). Pre-existing asymmetry: FlowChartRow is deliberately excluded from that margin rule — unrelated to this PR.

4. `AccountDetailPage/index.tsx`: only direct child is `<AccountProperties>` (renders `<div class="AccountProperties Properties">`) — shell padding applies, CSS-file deletion safe.
   `TransactionDetailPage/index.tsx`: all three branches render exactly one Properties panel (`InvestmentTransactionProperties` / `TransferProperties` / `TransactionProperties`) — shell padding applies to all, deletion safe.

5. `grep -rn` for `AccountDetailPage/index.css`, `TransactionDetailPage/index.css`, `ConnectionProperties/index.css` — no leftover imports, no leftover CSS selectors referencing `.AccountDetailPage` / `.TransactionDetailPage` outside `ChartDetailPage/index.css` (which uses its own class).

6. `DashboardPage/index.css`'s `> div { padding: 0 10px; margin: 10px 0 30px }` rule targets chart-row children (`BalanceChartRow` / `ProjectionChartRow` / `FlowChartRow`), none of which carry a `Properties` class — untouched, no interaction.

7. `BudgetConfigPage/index.css`'s kept `> div { padding: 0 10px }` correctly still serves `.title` and `.statusBarWithText`.

8. Scope: all 18 modified files are `*Properties` panels or their host pages — directly related to the PR title. Clean.

9. `bun run typecheck` clean, `bun --bun eslint src` clean, `bun test src/client` 312/0, CI both `build` and `test` green.

src/client/components/Configuration/index.css
hoiekim: Can these be migrated under the common `Properties` CSS rule?

hoiekim (approved):

PR #607, opened 2026-07-07
revert(client): drop the `<Page>` shell (undoes #473 migrations, deletes component)

claoie:
## Summary

Drops the `<Page>` shell landed in #473. It only ever owned `div.Page > div { padding: 0 10px }`, and that padding rule turned out to be the wrong convention — pages want their child elements to define their own padding, not inherit a uniform 10px column from the outer. With that rule gone the shell has no CSS of its own, so `<Page className="X">` is functionally identical to `<div className="Page X">` and the component doesn't earn its place.

## Changes

- Revert 4 pages from `<Page className="X">` to `<div className="X">`:
  - ApiKeyDetailPage
  - ConfigPage
  - ConnectionDetailPage
  - HoldingDetailPage
- Delete `src/client/components/Page/` (tsx + css).
- Drop `Page` export from `src/client/components/index.ts` barrel.

Per-page CSS (`div.XxxPage > div { padding: 0 10px }` on the 5 pages that already had their own version) unchanged — those are per-page rules, not shell-dependent.

## Related

- Closed #603 (would have migrated the other 9 pages onto the same shell).
- Follow-up on #326: the "componentize common design elements" arc continues via `<Properties>` (still a real component with its own CSS) and any future shells that earn their place by owning a real convention. Page didn't.

## Test plan

- [x] `bun test src/client src/server` — 1010 pass / 2 skip / 0 fail
- [ ] Sandbox UI verification: visit each of the 4 reverted pages + a couple of neighbors and confirm layout matches main.


reviewoie:
## Code review — PR 607

### High
- `src/client/pages/{ApiKeyDetailPage,ConfigPage,ConnectionDetailPage,HoldingDetailPage}/` — incomplete revert of #473; the 4 pages lose 10px horizontal padding vs both current `main` and the pre-#473 state.

  PR #473 (merge commit `16db5d716`) did TWO things:
  1. Added `<Page>` + `div.Page > div { padding: 0 10px }`.
  2. **DELETED** the per-page `src/client/pages/{ApiKeyDetailPage,ConfigPage,ConnectionDetailPage,HoldingDetailPage}/index.css` files. Each contained the same rule at the page-selector granularity, e.g. `div.ApiKeyDetailPage > div { padding: 0 10px }`.

  This PR undoes only step 1. The four per-page `index.css` files are NOT restored, so after merge those pages have NO horizontal padding rule at all (no shell, no per-page). None of the child components own compensating padding either:
  - `ApiKeyProperties/index.css` — no outer padding on `.ApiKeyProperties`
  - `Configuration/index.css` — no outer padding on `div.Configuration`
  - `ConnectionProperties/index.css` — no outer padding on `div.ConnectionProperties`
  - `HoldingProperties/index.css` — only grandchild padding (`.tickerFeedback`, `.formError`, `.validateBtn`)
  - `Properties/index.css` — `margin-top: 20px` only, no horizontal padding on `div.Properties`

  Failure scenario: on any narrow viewport where `App/index.css`'s `div.Router > div { padding: … env(safe-area-inset-right) … env(safe-area-inset-left) }` resolves to 0 (i.e. non-notched displays), the `.Properties` card panels on these 4 pages now render edge-to-edge instead of with the 10px column they had on both pre-#473 main and current main. Visible regression.

  Two convention-aligned resolutions:
  - **(a) Restore the per-page CSS files** — literal revert of #473. Re-adds `pages/{ApiKeyDetailPage,ConfigPage,ConnectionDetailPage,HoldingDetailPage}/index.css` each with `div.XxxPage > div { padding: 0 10px }` and re-imports each in its `index.tsx`. Matches the "revert #473 too" direction from Discord and keeps the visible layout identical.
  - **(b) Move the padding into the child components** — matches the "children should own their own padding" convention. Add horizontal padding to `.ApiKeyProperties`, `div.Configuration`, `div.ConnectionProperties`, and the `HoldingProperties` root (or to `div.Properties` if the padding is truly canonical to that shell — but that couples other Properties consumers).

  As shipped, this PR chooses neither — the padding is silently dropped with no compensation.

  Verified via `git show 16db5d716 -- src/client/pages/*/index.css` (all four deletions carrying the same rule) + `find src/client/pages/{ApiKey,Config,Connection,Holding}*/ -name '*.css'` on the PR head (returns nothing) + reading each child component's CSS.

### Med / Low
No further blockers.

### Sweep confirmations
- `<Page>` usages: 0 remaining in `src/**` (`grep -rn "<Page" --include="*.tsx" --include="*.ts" src/` returns nothing).
- Component import references to `./Page`: 0 (`grep -rn "components/Page"` returns nothing). The only remaining "Page" imports in the codebase are `TransactionsPageType` from `TransactionsPageTitle`, unrelated to the removed shell.
- `.Page` CSS/JS selector references: 0 (`grep -rn '\.Page\b' --include='*.css' --include='*.tsx' --include='*.ts' src/` returns nothing).
- Barrel drop in `src/client/components/index.ts`: clean; no other lines reference Page.
- Scope discipline: exactly the 7 expected files touched, no unrelated changes.

hoiekim (approved):

PR #611, opened 2026-07-08
refactor(client): extract `<PageFilterTitle>` primitive from TransactionsPageTitle (#326 phase 4b)

claoie:
## Summary

Continues #326 phase 4. #610 shipped `<PageTitle>`; this PR extracts the button + dropdown chrome that `TransactionsPageTitle` hand-rolled into a reusable `<PageFilterTitle>` primitive, so `AccountsPage` (and any other list page) can adopt filter-in-title with zero duplication.

## Changes

- New: `src/client/components/PageFilterTitle/{index.tsx, index.css}`.
  - Renders `<h2 className="PageTitle PageFilterTitle {optional className}">` — inherits `<PageTitle>`'s sticky/z-index/padding + adds the trigger button + dropdown mechanics.
  - Owns `isOpen`, `buttonRef`, `selectBoxRef`, touch-outside (registered only while open — cleaner than the always-on listener original), mouse-leave, and keyboard Escape/Enter/Space dismissal.
  - Props: `label` (button text), `dropdownLabel?` (sticky close row inside the dropdown), `className?` (per-page overrides), `children` (option buttons).
- `TransactionsPageTitle`: `<h2 className="heading">…hand-rolled button + dropdown…</h2>` → `<PageFilterTitle className="TransactionsFilterTitle" label={titleForSelection(...)} dropdownLabel="Select transaction types">{allButton}{typeButtons}</PageFilterTitle>`. Drops isSelecting state, buttonRef/selectBoxRef, the touch-outside effect, closeSelect handler, and the h3 subtitle's now-redundant `className="heading"`.
- CSS: dropdown chrome moved wholesale into the shell. `TxPage` keeps only its layout-specific overrides (`h2.TransactionsFilterTitle { height: 54px; z-index: 111 }` because SearchBar/TransactionsHead sit z-below) and the `h3 subtitle` + SearchBar + TransactionsHead sticky rules.
- Barrel export.

Net −52 / +197 (of which 178 is the new component). TransactionsPageTitle itself shrinks 160 → 62 tsx lines and 161 → 48 css lines.

## What this unblocks

`AccountsPage` can adopt `<PageFilterTitle label={titleForAccountFilter(selectedTypes)} dropdownLabel="Select account types">{accountTypeButtons}</PageFilterTitle>` in a follow-up PR. That's the feature you asked about.

## Test plan

- [x] `bun run typecheck` clean
- [x] `bun --bun eslint src` clean
- [x] `bun test src/client` 312 pass / 0 fail
[...]

PR #613, opened 2026-07-08
refactor(client): `useMultiSelectQueryFilter` hook + `<FilterOption>` (#326 phase 4c)

claoie:
## Summary

Continues #326 phase 4. #611 shipped `<PageFilterTitle>` (the chrome); phase 4c pushes the FILTERING LOGIC (URL param sync, toggle, clearAll, checkbox rendering) into a shared hook + component. `TransactionsPageTitle` composes both. `AccountsPage` adoption becomes ~15 lines with zero duplication.

## Changes

- **`src/client/lib/hooks/useMultiSelectQueryFilter.ts`** — reads `router.params.get(paramKey)` (comma-separated, validated against `allValues`, canonicalized), exposes `{ selected, toggle, clearAll }` writers. Empty selection removes the URL param.
- **`src/client/components/FilterOption/index.tsx`** — `<button><span class="checkbox {checked}" /><span>{label}</span></button>`. Owns just the checkbox tick + label; layout / hover inherited from the `<PageFilterTitle>` shell.
- **`TransactionsPageTitle`** — refactored to compose both. `writeTypes` / `onClickAll` / `toggleType` / `renderCheckbox` / `allButton` / `typeButtons` all deleted; `<FilterOption>` map + hook writers instead.
- **`parseTransactionsTypes`** kept as-is (still needed by `TransactionsPage` for transition-aware reads via `activeParams`; comment updated to explain).

Net −59 / +101 (78 is the new hook + component). TransactionsPageTitle tsx 210 → 152.

## What this unblocks

`AccountsPage` (or any list page) can adopt filter-in-title with the same primitives:

```tsx
const VALID = ["depository", "credit", "investment", "loan", "other"] as const;
const { selected, toggle, clearAll } = useMultiSelectQueryFilter("account_type", VALID);
<PageFilterTitle label={titleFor(selected)} dropdownLabel={<>Select&nbsp;account&nbsp;types</>}>
  <FilterOption checked={selected.length === 0} onSelect={clearAll}>All Accounts</FilterOption>
  {VALID.map((t) => (
    <FilterOption key={t} checked={selected.includes(t)} onSelect={() => toggle(t)}>
      {LABELS[t]}
    </FilterOption>
  ))}
</PageFilterTitle>
```

~15 lines vs the ~60 that `TxPageTitle` was carrying pre-4c.

## Test plan

- [x] `bun test src/client` 312 pass / 0 fail
- [ ] Sandbox: `TransactionsPage` — filter dropdown opens/closes; multi-select works (adding/removing types canonicalizes URL); clear-all works; URL param survives page reload with same selection

### DIFF
--- a/src/client/components/TransactionsPageTitle/index.tsx
+++ b/src/client/components/TransactionsPageTitle/index.tsx
@@ -100,12 +100,11 @@ export const TransactionsPageTitle = ({
   // and writes through the same hook, so `toggle` / `clearAll` mutate
   // the `transactions_type` URL param, and `selected` is derived from
   // that param on the next render — closing the loop within this
-  // component. The `filters.types` prop is still passed by
-  // `TransactionsPage` for transition-aware transaction-list filtering
-  // (via `activeParams`), which the URL-first hook can't see.
-  const { selected: selectedTypes, toggle, clearAll } = useMultiSelectQueryFilter(
+  // component. The `options` array is derived from the same `TYPE_LABELS`
+  // record the hook takes, so values and labels can't drift out of sync.
+  const { selected: selectedTypes, toggle, clearAll, options } = useMultiSelectQueryFilter(
     "transactions_type",
-    VALID_TYPES,
+    TYPE_LABELS,
   );
 
   const accountName = account?.custom_name || account?.name;
@@ -158,9 +157,13 @@ export const TransactionsPageTitle = ({
         <FilterOption checked={selectedTypes.length === 0} onSelect={clearAll}>
           All Transactions
         </FilterOption>
-        {VALID_TYPES.map((t) => (
-          <FilterOption key={t} checked={selectedTypes.includes(t)} onSelect={() => toggle(t)}>
-            {TYPE_LABELS[t]}
+        {options.map(({ value, label }) => (
+          <FilterOption
+            key={value}
+            checked={selectedTypes.includes(value)}
+            onSelect={() => toggle(value)}
+          >
+            {label}
           </FilterOption>
         ))}
       </PageFilterTitle>
--- a/src/client/components/TransactionsPageTitle/index.tsx
+++ b/src/client/components/TransactionsPageTitle/index.tsx
@@ -28,7 +28,6 @@ export type TransactionsPageType =
   | "manual";
 
 interface TransactionsPageFilters {
-  types: TransactionsPageType[];
   account?: Account;
   budget?: Budget;
   section?: Section;
@@ -95,9 +94,19 @@ export const TransactionsPageTitle = ({
   sorter,
   onChangeSearchValue,
 }: TransactionsPageTitleProps) => {
-  const { types: selectedTypes, account, budget, section, category } = filters;
+  const { account, budget, section, category } = filters;
   const { screenType } = useAppContext();
-  const { toggle, clearAll } = useMultiSelectQueryFilter("transactions_type", VALID_TYPES);
+  // URL is the source of truth for the type filter. The dropdown reads
+  // and writes through the same hook, so `toggle` / `clearAll` mutate
+  // the `transactions_type` URL param, and `selected` is derived from
+  // that param on the next render — closing the loop within this
+  // component. The `filters.types` prop is still passed by
+  // `TransactionsPage` for transition-aware transaction-list filtering
+  // (via `activeParams`), which the URL-first hook can't see.
+  const { selected: selectedTypes, toggle, clearAll } = useMultiSelectQueryFilter(
+    "transactions_type",
+    VALID_TYPES,
+  );
 
   const accountName = account?.custom_name || account?.name;
   const budgetName = budget?.name;
--- a/src/client/components/TransactionsPageTitle/index.tsx
+++ b/src/client/components/TransactionsPageTitle/index.tsx
@@ -12,8 +12,9 @@ import {
   ScreenType,
   Sorter,
   useAppContext,
+  useMultiSelectQueryFilter,
 } from "client";
-import { PageFilterTitle } from "client/components";
+import { FilterOption, PageFilterTitle } from "client/components";
 import { TransactionsHead } from "./TransactionsHead";
 import "./index.css";
 import { SearchBar } from "./SearchBar";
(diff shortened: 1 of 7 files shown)


reviewoie:
## Code review — PR 613

**Verdict:** approve

**Findings:** 0 blocking, 2 informational (low).

Walked the seven concerns in the review brief against the diff (headRefOid `4b80f8629e50d8d541b086758e8469fbc92e193c`), the composing consumer (`TransactionsPageTitle`), the shell CSS it inherits from (`PageFilterTitle/index.css` lines 51–97), and the router's `go(path, {params, animate:false})` semantics. Typecheck/eslint/`bun test src/client` (312/0) green.

### Informational (low)

- `useMultiSelectQueryFilter.ts:44` — `toggle` builds `write(allValues.filter((v) => set.has(v)))` (already canonical), then `write` re-runs `allValues.filter((v) => next.includes(v))`. Redundant O(N²) canonicalization step; not a bug (VALID_TYPES is 6 items) and defensive if `write` grows external callers. Consider dropping the second filter or making `write` accept a pre-canonicalized arg.

- `useMultiSelectQueryFilter.ts:40,47` — `toggle` / `clearAll` are fresh function identities every render. Flagged in the brief; correct that this is a non-issue today (`FilterOption` isn't memoized) but a future `React.memo(FilterOption)` would need `useCallback` wrappers here.

### Verified against the brief's checkpoints

1. Canonicalization parity — `parseTransactionsTypes` (line 70) and hook read (`useMultiSelectQueryFilter.ts:31`) both do `allValues.filter(v => present.has(v))`. `?transactions_type=deposits,expenses` and `expenses,deposits` both round-trip to `expenses,deposits` (`VALID_TYPES` order: unsorted, suggested, deposits, expenses, transfers, manual).

2. Byte-identical DOM — `<button onClick><span className="checkbox{checked?' checked':''}" aria-hidden="true" /><span>{children}</span></button>` matches the pre-refactor structure. `FilterOption/index.css` is a comment-only stub; layout hits `h2.PageFilterTitle > div.select > div.options > button {…}` (PageFilterTitle CSS:51–97) unchanged.

3. URL-read vs prop-read divergence — non-issue as implemented: `TransactionsPageTitle` reads `selectedTypes` from `filters.types` (activeParams-derived, transition-aware) for BOTH the label and the checkbox `checked` state (`index.tsx:145,149,153`). The hook's `selected` is never destructured — only `toggle`/`clearAll` are. On `/transactions` steady state `filters.types === hook.selected` (both trace to `router.params`). The residual asymmetry — `toggle`'s XOR base is `hook.selected` (router.params, not activeParams) — matters only during the ≤300ms narrow-screen slide where the dropdown is effectively never open (isOpen resets to false on mount; touchstart-outside closes it during outgoing nav). Original code had the same write-time non-transition-aware `router.go(path, params)` shape, so no regression.

4. `useCallback` — see informational above.

5. `parseTransactionsTypes` equivalence — kept as-is; still called by `TransactionsPage/index.tsx:70` on `activeParams.get("transactions_type")`. Semantically identical to the hook's internal parse (empty→`[]`, split-trim, `VALID_TYPES` canonicalize).

6. Scope — 6 files, all directly related to phase 4c (2 new primitives, 2 barrels, 1 refactored consumer, 1 comment-only CSS stub). No unrelated changes.

7. Regression sweep (traced against implementation):
   - Toggle a type → `router.params.transactions_type` gains it in canonical position. OK.
   - Toggle same type again → set XOR removes; `write` sees `next.length===0` if it was the only one → `params.delete(paramKey)`. OK.
   - Click "All Transactions" when some selected → `clearAll` → `write([])` → param deleted. OK.
   - Click "All Transactions" when none selected → `params.delete` no-op + `router.go` (pushState with same URL). Unchanged from pre-refactor (`onClickAll → writeTypes([]) → newParams.delete`). Same duplicate-history-entry behavior; not new.
   - Route reload → URL-derived state persists via `router.params`. OK.

src/client/components/TransactionsPageTitle/index.tsx
```diff
@@ -95,53 +96,8 @@ export const TransactionsPageTitle = ({
...
-    const isSelected = selectedTypes.includes(t);
-    return (
-      <button key={t} onClick={() => toggleType(t)}>
-        {renderCheckbox(isSelected)}
-        <span>{TYPE_LABELS[t]}</span>
-      </button>
-    );
-  });
+  const { screenType } = useAppContext();
+  const { toggle, clearAll } = useMultiSelectQueryFilter("transactions_type", VALID_TYPES);
```
hoiekim: This doesn't use `selectedTypes` at all. How does the filter buttons control `selectedTypes`?

claoie:
Fair — the previous shape was confusing. Regarding [that review comment](https://github.com/hoiekim/budget/pull/613/files#r3538988) on `TransactionsPageTitle:100`:

The buttons didn't control `selectedTypes` **directly** — the loop closed through URL param → parent → prop → child re-render, which isn't obvious from reading the file. Restructured so the hook's `selected` IS `selectedTypes` (renamed via destructure):

```tsx
const { selected: selectedTypes, toggle, clearAll } = useMultiSelectQueryFilter(
  "transactions_type",
  VALID_TYPES,
);
```

Now the loop is self-contained in one file: click `<FilterOption>` → `toggle` writes URL → next render's `hook.selected` re-derives from the URL → label + checkbox update.

Also dropped `types` from the `filters` prop (dead after the refactor). `TransactionsPage` still calls `parseTransactionsTypes(activeParams.get(...))` for its own transaction-list filter (needs `activeParams` for narrow-screen transition awareness, which the URL-first hook can't see) — that's the only remaining reason `parseTransactionsTypes` is exported.

New commit + E2E re-verified (URL canonicalization, checkbox states, clear-all, page reload survives selection — all clean).

src/client/components/TransactionsPageTitle/index.tsx
```diff
@@ -94,55 +94,20 @@ export const TransactionsPageTitle = ({
...
+  const { account, budget, section, category } = filters;
+  const { screenType } = useAppContext();
+  // URL is the source of truth for the type filter. The dropdown reads
+  // and writes through the same hook, so `toggle` / `clearAll` mutate
+  // the `transactions_type` URL param, and `selected` is derived from
+  // that param on the next render — closing the loop within this
+  // component. The `filters.types` prop is still passed by
+  // `TransactionsPage` for transition-aware transaction-list filtering
+  // (via `activeParams`), which the URL-first hook can't see.
+  const { selected: selectedTypes, toggle, clearAll } = useMultiSelectQueryFilter(
```
hoiekim: Should we make this hook to return the filter JSX element as well?

hoiekim (dismissed):

hoiekim (approved):

PR #614, opened 2026-07-08
feat(client): AccountsPage title-embedded filter by account type (#326 phase 4d)

claoie:
## Summary

Adopts the phase 4a-4c primitives on `AccountsPage`. Your original ask when we split `<PageFilterTitle>` from `TransactionsPageTitle`: *"I want the transaction filter embedded title to be generic so I can use it for account filtering too."* This is that.

## Changes (one file)

- New: `ACCOUNT_TYPE_LABELS: Record<AccountType, string>` covers all 6 Plaid types (depository, credit, investment, loan, brokerage, other).
- `useMultiSelectQueryFilter<AccountType>("account_type", ACCOUNT_TYPE_LABELS)` → `{ selected, toggle, clearAll, options }`.
- `<PageTitle>All Accounts</PageTitle>` → `<PageFilterTitle>` with a clear-all `<FilterOption>` sentinel + `options.map` for each type.

## Filter semantics

- **No filter set (default)** → unchanged from main. Credit still excluded from the donut + table (it has its own summary tile).
- **Filter set** → honor verbatim. Picking "Credit" surfaces credit accounts in the donut + table. Any other selection scopes the donut + table to those types (still excluding `hide` / `archived`).
- `selectedTypes` added to the `useMemo` dep so donut + balanceTotal recompute when the filter changes.

Net −5 / +58 in one file. No new components. URL shape: `?account_type=depository,investment`.

## Test plan

- [x] `bun test src/client` 312 pass / 0 fail
- [ ] Sandbox: default view unchanged; filter opens/closes; picking types narrows donut + table correctly; picking "Credit" alone shows credit accounts; page reload preserves the filter

### DIFF
--- a/src/client/components/AccountsDonut/BalanceInfo.tsx
+++ b/src/client/components/AccountsDonut/BalanceInfo.tsx
@@ -46,14 +46,18 @@ export const BalanceInfo = ({
       {!isShrunk && (
         <>
           <div className="label">in&nbsp;{donutData.length}&nbsp;Accounts</div>
-          <br />
-          <Changes
-            currentAmount={0}
-            previousAmount={!viewDateSpan ? totalCredit : 0}
-            currencySymbol={currencySymbol}
-          />
-          <div className="credit label">outstanding</div>
-          <div className="credit label">in&nbsp;{numberOfCredits}&nbsp;Credits</div>
+          {numberOfCredits > 0 && (
+            <>
+              <br />
+              <Changes
+                currentAmount={0}
+                previousAmount={!viewDateSpan ? totalCredit : 0}
+                currencySymbol={currencySymbol}
+              />
+              <div className="credit label">outstanding</div>
+              <div className="credit label">in&nbsp;{numberOfCredits}&nbsp;Credits</div>
+            </>
+          )}
         </>
       )}
     </div>
--- a/src/client/components/AccountsTable/index.tsx
+++ b/src/client/components/AccountsTable/index.tsx
@@ -11,10 +11,17 @@ export type AccountHeaders = { [k in keyof Account]?: boolean } & {
 
 interface Props {
   donutData: DonutData[];
+  /** Types the user picked via `<PageFilterTitle>` on `AccountsPage`.
+   * Empty = no filter (default view). Non-empty = show only accounts
+   * whose `type` is in the array, across ALL sections (donut, credit,
+   * archived, hidden). Without this the Credit / Archived / Hidden
+   * sub-lists render every account regardless of the filter, so picking
+   * "Depository" still leaves credit accounts on screen. */
+  selectedTypes?: AccountType[];
   style?: CSSProperties;
 }
 
-export const AccountsTable = ({ donutData, style }: Props) => {
+export const AccountsTable = ({ donutData, selectedTypes, style }: Props) => {
   const { data } = useAppContext();
   const { accounts } = data;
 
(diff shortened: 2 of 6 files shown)


reviewoie:
## Code review — PR 614

**Verdict:** informational — one low-severity pattern deviation, otherwise clean.

### High
No blockers.

### Med
No blockers.

### Low
- `src/client/pages/AccountsPage/index.tsx:118` — `selectedTypes` in the `useMemo` dep list invalidates the cache on every render, defeating memoization on scroll. `useMultiSelectQueryFilter` returns a fresh `selected: T[]` array each render (not memoized inside the hook — `useMultiSelectQueryFilter.ts:47-52` reconstructs it from `params.get(paramKey)` unconditionally). The `useMemo` previously had reference-stable deps (`accounts`, `balanceData`, `viewDate`, `data.status.isLoading`); adding a fresh-ref array flips it to "recompute every render." The scroll listener at :52-56 (`setScrollY`) fires many renders per second on scroll → `accounts.toArray().sort()` + filter + iteration re-runs each frame. Pattern deviation from the sibling consumer: `TransactionsPage/index.tsx:70` derives `types` via `useMemo(() => parseTransactionsTypes(typesRaw), [typesRaw])` off the raw URL string exactly for this. Fix options: (a) memo off `params.get("account_type")` (a stable string) instead of the array, (b) memo `selected` inside `useMultiSelectQueryFilter`, or (c) drop the dep and inline the `selectedTypes.length === 0 ? ... : includes(type)` gate outside the memo (they're cheap). At Hoie's ~20-account scale the absolute cost is small (~ms/sec on active scroll), but worth aligning with the TransactionsPage pattern before the hook grows more consumers.

### Verified clean
- **Default filter semantics preserved.** `filteredAccounts` gate at :78-82 keeps `type !== AccountType.Credit` when `selectedTypes.length === 0`; `totalCredit` / `numberOfCredits` at :84-88 iterate `sortedAccounts` (not `filteredAccounts`), so the Credit summary tile is unaffected by the filter. Picking "Credit" alone surfaces credits in the donut + table while the summary still counts all credits. No double-count / no drop.
- **URL param name `account_type` is unique** across `src/` — no collision with any other page or hook.
- **`AccountType` typechecks against `T extends string`.** Plaid emits a TS string enum (`api.d.ts:950-957`): all 6 values (`Investment`, `Credit`, `Depository`, `Loan`, `Brokerage`, `Other`) are keys of `ACCOUNT_TYPE_LABELS`. Adding a new plaid enum member would force a `Record<AccountType, string>` compile error — good coupling.
- **Reload with `?account_type=investment` filters on first paint.** `useRouter` initializes `params` synchronously from `window.location.search` (`router.ts:127`); the hook reads via `router.params.get(...)` on the same render. No flash of unfiltered.
- **Dropdown z-index.** `h2.PageTitle` base = `z-index: 110`, `AccountsDonut` = 100 — dropdown stacks above the donut without needing a per-page CSS override like `TransactionsFilterTitle` (`z-index: 111`).
- **Scope.** Single file, single commit. All hunks directly serve the PR title.

### UX notes (not bugs)
- Option order is `Depository, Credit, Investment, Loan, Brokerage, Other` (`ACCOUNT_TYPE_LABELS` insertion order, not the plaid enum's declaration order which puts `Investment` first). Reads fine but is an intentional call worth confirming.
- Donut wedge colors reorder when the filter changes (`filteredAccounts.forEach((a, i) => colors[i % colors.length])`) — post-filter index. Cosmetic.
- Max-length label at all 6 selected is ~47 chars — fits on 390px narrow but tight. In practice users rarely pick all 6.

hoiekim (approved):

PR #616, opened 2026-07-09
refactor(client): getActiveParams follow-up (useCallback + JSDoc + tests)

claoie:
## Summary

Follow-up to `1a07bc0e implemented getActiveParams to reduce redundant code`, addressing the review notes on that commit. No behavior change — refactor + docs + tests.

### `router.ts`

- **Wrap `getActiveParams` in `useCallback`** so the reference is stable across renders while `path` / `screenType` / `params` / `incomingParams` don't change. `go` / `back` / `forward` next to it are already useCallback'd; this makes the pattern consistent and lets future consumers put `getActiveParams` in useEffect / useMemo deps without re-firing on every parent render.
- **Extract `deriveActiveParams(targetPath, currentPath, screenType, params, incomingParams)`** as a pure module-level helper. The useCallback in the router forwards to it. Pure form unblocks unit testing without React harness or DOM mocks (the rest of `useRouter` touches `window` / `history` / `location`).
- **Add JSDoc** to `getActiveParams` explaining the transition-aware intent AND the foot-gun of passing a sibling page's PATH (silent read-from-`incomingParams`-forever at steady-state under narrow).

### `useMultiSelectQueryFilter.ts`

- **Add JSDoc for `targetPath`** in the main hook comment: it must match the enclosing page's route, and the writer always writes through live `params` (not the outgoing snapshot). Restores the reader/writer asymmetry explanation the old `Options` interface carried before `1a07bc0e` stripped it.

### `router.test.ts` (new)

6 tests pinning `deriveActiveParams` behavior:

- same-path narrow → live params.
- wide-screen bypass (currentPath differs) → live params.
- cross-path narrow → incoming params.
- **wrong-PATH foot-gun (steady-state under narrow) → incoming params** — pins the documented failure mode so it can't silently change.
- wrong-PATH bypassed by wide-screen → live params.
- empty-params base case.

## Test plan

- [x] `bun test src/client` → 318/0, 6 new tests
- [ ] Sandbox smoke — no behavior change expected, but worth clicking through AccountsPage / DashboardPage / BudgetsPage / TransactionsPage narrow-viewport transitions once to confirm no regression from the useCallback wrap.

### DIFF
--- a/src/client/lib/hooks/router.test.ts
+++ b/src/client/lib/hooks/router.test.ts
@@ -1,0 +1,92 @@
+import { describe, test, expect } from "bun:test";
+import { deriveActiveParams, PATH } from "./router";
+import { ScreenType } from "./context";
+
+const p = (init: string) => new URLSearchParams(init);
+
+describe("deriveActiveParams", () => {
+  test("returns live params when currentPath matches targetPath (narrow, transitioning)", () => {
+    const live = p("account_type=depository");
+    const outgoing = p("account_type=investment");
+    const out = deriveActiveParams(
+      PATH.ACCOUNTS,
+      PATH.ACCOUNTS,
+      ScreenType.Narrow,
+      live,
+      outgoing,
+    );
+    expect(out).toBe(live);
+  });
+
+  test("returns live params on wide-screen even when currentPath differs", () => {
+    const live = p("account_type=depository");
+    const outgoing = p("account_type=investment");
+    const out = deriveActiveParams(
+      PATH.ACCOUNTS,
+      PATH.ACCOUNT_DETAIL,
+      ScreenType.Wide,
+      live,
+      outgoing,
+    );
+    expect(out).toBe(live);
+  });
+
+  test("returns incoming params when narrow AND currentPath differs from targetPath", () => {
+    const live = p("account_id=abc");
+    const outgoing = p("account_type=investment");
+    const out = deriveActiveParams(
+      PATH.ACCOUNTS,
+      PATH.ACCOUNT_DETAIL,
+      ScreenType.Narrow,
+      live,
+      outgoing,
+    );
+    expect(out).toBe(outgoing);
+  });
+
+  test("foot-gun: passing sibling PATH on steady-state narrow returns incomingParams, not live", () => {
+    // The AccountsPage caller mistakenly passing PATH.BUDGETS while
+    // the current page IS AccountsPage. currentPath === ACCOUNTS !==
+    // BUDGETS, so under narrow it reads incomingParams. This is the
+    // documented failure mode of the wrong-PATH arg — the test pins it
+    // so the behavior can't silently change.
+    const live = p("account_type=depository");
+    const outgoing = p("");
+    const out = deriveActiveParams(
+      PATH.BUDGETS,
+      PATH.ACCOUNTS,
+      ScreenType.Narrow,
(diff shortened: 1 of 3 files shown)


hoiekim (approved):

reviewoie:
## Code review — PR 616

**Verdict:** changes-requested (JSDoc accuracy)

Mechanical verifications (all clean):
- `useCallback` deps `[path, screenType, params, incomingParams]` cover every closed-over variable — complete.
- `deriveActiveParams` body matches the pre-1a07bc0e inline body byte-for-byte (only rename: `path` -> `currentPath` parameter).
- No new import cycle. The `context.ts` -> `client` -> `hooks/index.ts` -> `router.ts` -> `./context` loop pre-exists; adding a second named export from `router.ts` doesn't change the graph.
- `bun test src/client/lib/hooks/router.test.ts` -> 6 pass / 0 fail / 6 expect() calls. No `Date.now` / `Math.random` / `new Date` in the file.
- Scope: 3 files, all directly related to the PR title. Clean.

### Med
- `src/client/lib/hooks/router.ts:306-322` — the new `getActiveParams` JSDoc misdescribes the state-transition mechanism. Two factual errors:
  1. **"path has already flipped to the destination but the caller's JSX is still animating"** — inverted. `setPath(newPath)` fires inside `endTransition` at `router.ts:230`, which is delayed by `DEFAULT_TRANSITION_DURATION` via `setTimeout` on narrow-with-animation (`router.ts:257-261`). During the slide, `path` is still the OLD/outgoing route — that is *why* the outgoing caller's `path === targetPath` check succeeds and it reads `params` for its own URL.
  2. **"returning `transition.incomingParams` (a snapshot of the outgoing URL)"** — inverted. `setIncomingParams(newParams)` at `router.ts:217` sets `incomingParams` to `newParams`, i.e. the DESTINATION URL passed into `transition()`. It is NOT a snapshot of the outgoing URL; the OLD/outgoing URL is what `params` still holds until endTransition fires.

  The actual mechanism:
  - Narrow steady-state / wide-screen: `path === targetPath` (or short-circuit) -> returns `params` (destination URL, aligned).
  - Narrow slide-OUT of the caller's page: `path` still equals caller's `targetPath` (endTransition hasn't fired) -> returns `params`, still holding the OLD outgoing URL -> outgoing JSX keeps rendering its own URL.
  - Narrow slide-IN to the caller's page: `path` is still the previous route (does NOT match caller's `targetPath`), Narrow -> returns `incomingParams`, which holds the DESTINATION URL -> incoming JSX renders its own URL.

  Same confusion propagates into `src/client/lib/hooks/useMultiSelectQueryFilter.ts:34-41`: "the reader uses `incomingParams` during a narrow-screen slide-out and `params` otherwise" — a slide-OUT from this page actually uses `params` (path === targetPath still holds during animation). Only a slide-IN uses `incomingParams`.

  Runtime behavior is correct — this is docs-only. But since this PR's stated goal is transition-aware documentation for future readers, an inverted state model in the docstring defeats it. The test-file variable naming (`live` / `outgoing`) inherits the same muddle: in test 1 `outgoing` maps to `incomingParams`=destination URL, and in test 3 `outgoing` again maps to `incomingParams`=destination — never the outgoing URL. Renaming to `paramsLive` / `paramsIncoming` (matching the actual state var names) would make the tests self-explain.

No further blockers.

PR #620, opened 2026-07-09
fix(client): useViewDate — parse YYYY-MM correctly + URL-first re-derivation + tests

claoie:
## Summary

Fixes both bugs + the nits I flagged in the review of `b675929e` (implemented useViewDate). Leaves the "year view — retire or add UI toggle?" question to a separate call from Hoie.

### 🔴 HIGH — month parse bug (viewDate reset to January every session)

`setViewDate` writes `getYearMonthString(...)` which returns `YYYY-MM` (dashed). URL becomes `?view_date=2026-07`. But the hook read the URL back with `substring(4, 6)`:

```ts
const year  = parseInt(viewDateString.substring(0, 4));  // 2026 ✓
const month = parseInt(viewDateString.substring(4, 6)) || 1;
```

For `"2026-07"`, position 4 is the dash → `substring(4,6) === "-0"` → `parseInt("-0") === 0` → `|| 1` → month = 1 (January). Every session re-initialized to Jan 1 of the year in URL. Verified via `bun -e` — see commit body.

**Fix**: extract `parseViewDateString(viewDateString)` as a pure helper that dispatches to the existing `parseYearMonthString` for the dashed month form. Year form (`YYYY`, 4 chars) still supported so bookmarked year URLs keep working.

### 🟡 MED — URL back-button didn't sync viewDate

`useState(defaultViewDate)`'s initializer only ran on mount. If URL flipped (browser back, cross-consumer navigation, tab restore), `defaultViewDate` recomputed on re-render but `viewDate` stayed at the mount-time value.

**Fix**: drop the useState + derive `viewDate` from `params` each render via `useMemo(..., [viewDateString])`. Reference is stable while the URL param doesn't change (consumers' useMemo deps don't thrash), and back-button updates propagate immediately. Same URL-first shape as `useMultiSelectQueryFilter`. `setViewDate` closes over the derived `viewDate` to resolve functional-set callbacks — URL is still the only durable state.

### 🟢 Nits — unused imports left by b675929e

Main's CD workflow doesn't lint; the PR `test` workflow does, so any downstream PR touching these files would be red. Cleaned up:

- `AppContext.tsx` — `useLocalStorageState`, `Interval`, `ViewDate` (previously destructured for the `useState(new ViewDate(selectedInterval))` shape that's gone).
- `Header/index.tsx` — `ChevronLeftIcon`, `ChevronRightIcon`, `onClickPreviousView`, `onClickNextView`, `getIntervalOptionText`, `Interval` — dead code from the old prev/next chevron + interval `<select>` that the native `<input type="month">` replaced.
- `context.ts` — `Interval` (no longer on `ContextType`).

Also dropped from `useCallback` deps: `_setViewDate` (stable from `useState` — not needed even before this rewrite; the new hook shape doesn't use it at all).

## Tests

New `src/client/lib/hooks/viewDate.test.ts` — 6 tests pinning `parseViewDateString`:

- YYYY-MM correct-month (regression pin for the HIGH).
- YYYY-01 no off-by-one.
- YYYY-12 no off-by-one.
- YYYY year interval.
- garbage input → fallback to now.
- empty string → fallback to now.

Full suite 324 / 0, +6.

## Left out

**Year interval**: still readable in the hook so bookmarked year URLs work, but there's no UI toggle post-Header refactor. If the intent was to fully retire year mode, the `interval === "year"` branches in the hook can go too — deferred pending explicit call.

**Back-button-no-animation on view_date-only changes**: separate request Hoie made after the review — will address in a follow-up so this PR stays scoped to the 3 review-note fixes.

## Test plan

- [x] `bun test src/client` → 324 / 0 (+6 new)
- [ ] Sandbox — reload with `?view_date=2026-07` and confirm view stays on July (not January); click Add Budget then browser-back and confirm viewDate updates.

### DIFF
--- a/src/client/components/App/AppContext.tsx
+++ b/src/client/components/App/AppContext.tsx
@@ -1,14 +1,6 @@
 import { useState, useMemo, ReactNode } from "react";
-import {
-  useLocalStorageState,
-  ContextType,
-  Context,
-  useRouter,
-  reduceStatuses,
-  useViewDate,
-} from "client";
+import { ContextType, Context, useRouter, reduceStatuses, useViewDate } from "client";
 import { MaskedUser } from "server";
-import { Interval, ViewDate } from "common";
 import { useData, useScreenType } from "./lib";
 
 interface Props {
--- a/src/client/components/Header/index.tsx
+++ b/src/client/components/Header/index.tsx
@@ -4,38 +4,17 @@ import {
   ArrowLeftIcon,
   BankIcon,
   ChartIcon,
-  ChevronLeftIcon,
-  ChevronRightIcon,
   HamburgerIcon,
   ListIcon,
   RecieptIcon,
 } from "client/components";
-import { getYearMonthString, Interval, parseYearMonthString, ViewDate } from "common";
+import { getYearMonthString, parseYearMonthString, ViewDate } from "common";
 import "./index.css";
 
 export const Header = () => {
   const { user, router, viewDate, setViewDate, screenType } = useAppContext();
 
   const { path, params, go, back } = router;
-
-  const onClickPreviousView = () => {
-    setViewDate((oldViewDate) => {
-      const newViewDate = oldViewDate.clone().previous();
-      return newViewDate;
-    });
-  };
-
-  const onClickNextView = () => {
-    setViewDate((oldViewDate) => {
-      const newViewDate = oldViewDate.clone().next();
-      return newViewDate;
-    });
-  };
-
-  const getIntervalOptionText = (interval: Interval, fallback: string) => {
-    if (viewDate.getInterval() !== interval) return fallback;
-    return viewDate.toString();
-  };
 
   const onClickBack: MouseEventHandler<HTMLButtonElement> = (e) => {
     e.preventDefault();
--- a/src/client/lib/hooks/context.ts
+++ b/src/client/lib/hooks/context.ts
@@ -1,6 +1,6 @@
 import { createContext, useContext, Dispatch, SetStateAction } from "react";
 import { MaskedUser } from "server";
-import { Interval, ViewDate } from "common";
+import { ViewDate } from "common";
 import { ClientRouter, Status, Data, Calculations, CapacityData } from "client";
 
 export enum ScreenType {
--- a/src/client/lib/hooks/viewDate.test.ts
+++ b/src/client/lib/hooks/viewDate.test.ts
@@ -1,0 +1,58 @@
+import { describe, test, expect } from "bun:test";
+import { parseViewDateString } from "./viewDate";
+
+describe("parseViewDateString", () => {
+  test("YYYY-MM parses to the correct month", () => {
+    // Reads position 5-6 for month via `parseYearMonthString`. Pre-fix,
+    // the hook used `substring(4, 6)` which grabbed the dash + first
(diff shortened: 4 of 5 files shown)


reviewoie:
## Code review — PR 620

**Verdict:** approve

Walked the diff against the pre-announce sweep points.

### Round-trip correctness (HIGH-fix)
- `setViewDate` writes `getYearMonthString(...)` → `YYYY-MM` (7 chars, dashed — `common/utils/date.ts:256-260`).
- URL-read path: `viewDateString` (length 7) skips the `length === 4` year branch → `parseYearMonthString("2026-07")` (`common/utils/date.ts:274-279`) reads `substring(0,4)` → year 2026, `substring(5,7)` → month 7 → `new Date(2026, 6)` (July, 0-indexed). Round-trip correct.
- Pre-fix `substring(4, 6)` grabbed `"-D"` → `parseInt("-0") === 0 || 1 → 1 → new Date(year, 0)` (January) — the regression the test at line 5-15 pins.

### URL-first re-derivation (MED-fix)
- `useMemo(() => new ViewDate(...), [viewDateString])` — string dep, `Object.is` works. Reference stable across renders when URL unchanged.
- Router's `params` is `useState`-backed; changes on popstate via `endTransition`'s `setParams(newParams)` (`router.ts:231`). New params → new `viewDateString` → useMemo recomputes → back-button sync works.

### `setViewDate` closure
- Deps `[viewDate, path, params, go]` are exhaustive for the new closure (dropped `_setViewDate`, added `viewDate`). Matches `useMultiSelectQueryFilter` shape as noted.
- Only caller is `Header/index.tsx:74`, single onChange dispatch — no concern about the derived-viewDate closure regressing functional-set composition semantics (would only bite if a future caller batched two `setViewDate(prev => …)` calls synchronously).

### Tests
- 6 tests pass, `bun test src/client/lib/hooks/viewDate.test.ts` → 6 pass / 0 fail.
- YYYY-MM correct month + Jan/Dec no off-by-one + YYYY year branch + garbage/empty fallback. Regression pin on `substring(5, 7)` is the exact `.getMonth()` assertion at line 14.
- Fallback tests capture `Date.now()` around the call and assert a range, not equality — correct (per no-fabricated-timing rule).

### Cleanups
- `AppContext.tsx`: `useLocalStorageState`, `Interval`, `ViewDate` — grep confirms no residual references in touched files.
- `Header/index.tsx`: `ChevronLeftIcon`/`ChevronRightIcon`/`onClickPreviousView`/`onClickNextView`/`getIntervalOptionText` — grep confirms zero references repo-wide. `parseYearMonthString`, `getYearMonthString`, `ViewDate` still used (lines 52, 72, 74).
- `context.ts`: `Interval` unused after b675929e removed the interval-toggle type from the AppContext value.
- `Interval` still imported in `viewDate.ts` for the parser return type. Correct.

### Scope
5 files, all directly tied to the fix or the dead code left by the b675929e commit that introduced the bug. Scope-clean.

### Typecheck
`bun run typecheck` → exit 0.

### Nit (informational, not blocking)
- `hooks/index.ts` `export * from "./viewDate"` now re-exports `parseViewDateString` too. Benign — the parser is pure — but not intentional per the PR body. Drop or leave.

No further blockers.

hoiekim (approved):

PR #621, opened 2026-07-09
fix(client): skip transition animation on same-path query-only changes

claoie:
## Summary

Skips the router's slide animation on same-path query-param-only transitions. Answers Hoie's follow-up ask to the `useViewDate` review (opting for the broader "all same-path query changes" scope over view_date-specifically).

## Problem

The router's `back()` defaults `isAnimationEnabled.current = true`, and the popstate listener has no way to know whether the target URL is same-path. Result:

- Browser back/forward stepping through `?view_date=2026-08` → `?view_date=2026-05` on the same page flashes a 300ms slide over otherwise-identical layout.
- Same for filter param walks via history.
- Any future caller that forgets to pass `animate: false` on a same-path navigation gets the same jitter.

## Fix

`transition()` in `router.ts` detects `newPath === outgoingPath` right before the animation branch and forces `isAnimationEnabled.current = false`:

```ts
const isSamePath = newPath === outgoingPath;
if (isSamePath) {
  isAnimationEnabled.current = false;
}
```

Callers that navigate explicitly (`useMultiSelectQueryFilter`'s writer, `useViewDate`'s writer, most in-app `go()` calls) already pass `animate: false`, so this is a no-op for them. The change fires on:

- Browser back/forward that lands on the same path with different query params.
- Any caller that forgot `animate: false` on a same-path navigation — belt-and-braces.

## Depends on #620

`fix/useviewdate-parse-and-sync` — this branch is based on #620 (couldn't stack across forks in GitHub UI, so the diff here is against `main` and includes both sets until #620 merges). When #620 lands, this rebases to a single-file diff.

## Test plan

- [x] `bun --bun eslint src` clean (via #620's cleanup, folded in)
- [ ] Sandbox — narrow-viewport: navigate `?view_date=2026-08` → back → confirm no slide, only picker value flips. Also: `?account_type=depository` → `?account_type=credit` → back → confirm no slide. Cross-path back (detail → list) should still slide.

### DIFF
--- a/src/client/lib/hooks/router.ts
+++ b/src/client/lib/hooks/router.ts
@@ -254,6 +254,22 @@ export const useRouter = (screenType: ScreenType): ClientRouter => {
         isAnimationEnabled.current = false;
       };
 
+      // Same-path (query-param-only) transitions never slide. Animation
+      // is meaningful for a NEW view — a URL that swapped one filter for
+      // another, or a browser-back stepping through the view_date
+      // history on the current page, should not flash a 300ms slide
+      // over otherwise-identical layout. Callers that navigate
+      // explicitly (e.g. `useMultiSelectQueryFilter`'s writer) already
+      // pass `animate: false` to `go()`, but browser-driven popstate
+      // flips this on: `back()` defaults `isAnimationEnabled.current =
+      // true`, and popstate can't know whether the target is same-path.
+      // Detect it here instead — the refs advanced above hold the
+      // outgoing route.
+      const isSamePath = newPath === outgoingPath;
+      if (isSamePath) {
+        isAnimationEnabled.current = false;
+      }
+
       if (window.innerWidth < 950 && isAnimationEnabled.current) {
         timeout.current = setTimeout(endTransition, DEFAULT_TRANSITION_DURATION);
       } else {


reviewoie:
## Code review — PR 621

**Verdict:** approve

Verified the fix against the intended behavior:

- `router.ts:271` (`const isSamePath = newPath === outgoingPath`) — `outgoingPath` is captured at line 189 from `currentPathRef.current` BEFORE the ref is advanced at line 201, so the comparison genuinely tests destination-path vs true outgoing-path. Correct.
- Both operands are `PATH` enum values (path only, no query string), so a query-param-only change compares equal — exactly the intent. Trailing slashes / hashes never reach this comparison because `getPath()` normalizes via `pathname.split("/")[1]`.
- Cross-path navigation keeps `isSamePath = false`, so legitimate view-swaps still animate. No regression.
- Load-bearing path is browser popstate: `back()`/`forward()` set `isAnimationEnabled.current = true`, and popstate calls `transition(getPath(), getParams())` with no skip arg — this new block is the only thing that suppresses the 300ms slide when back/forward steps through same-path `view_date` history. Every same-path `go()` caller already passes `animate: false` (viewDate, useMultiSelectQueryFilter, CategoryBar, SectionBar, BudgetDetailPage), so the branch is redundant-but-harmless there; the popstate case is the real fix.
- `transitioning` stays `false` for same-path (`incomingPath === path`), so the slide panels never render and `slideAnchorY` is inert — no half-rendered transition state.

Scope is clean: single file, tightly tied to the animation fix. No unrelated changes.

### Low
- `router.ts:271` — the new same-path branch has no test. `router.test.ts` covers only `deriveActiveParams`, the pure helper deliberately extracted for hook-free testing. The same-path decision is a trivial inline `===` inside the `transition` closure, so extraction would be over-engineering — noting for completeness, not a blocker.

No further blockers.

claoie:
## E2E — UI-driven on the sandbox (mobile 390×844)

Drove the running app through the full same-path vs cross-path navigation matrix, instrumenting `div.Router`'s `classList` with a `MutationObserver` to catch any `transitioning` flash. (`router.ts` on the sandbox build is byte-identical to the PR head; the rest of the sandbox's base is older `main`, irrelevant to this hook.)

| Navigation | Path change | `transitioning` class? | Observed result |
|---|---|---|---|
| Filter toggle "Expense" (dropdown → option) | `/budgets` → `/budgets?budget_filter=expense` | **never appeared** | title → "Expense", list re-filtered, query applied immediately |
| Native browser **Back** (popstate) | `?budget_filter=expense` → `/budgets` | **never appeared** | filter cleared, title → "All Budgets", list reverted immediately |
| Native browser **Forward** (popstate) | `/budgets` → `?budget_filter=expense` | **never appeared** | re-applied immediately |
| Budget row tap (cross-path) | `/budgets` → `/budget-detail?budget_id=…` | **appeared** — `Router transitioning forward` for ~288ms, then cleared | slide animation still fires |

So every same-path query-only transition — including the load-bearing **native popstate back/forward** path this PR targets — skips the slide and applies params synchronously, while genuine cross-path navigation still animates. No regression to legitimate view swaps.

Screenshots eye-checked at each step: the filter-applied and post-back renders are clean (correct title, correct filtered/unfiltered list, no frozen or stale intermediate frame). Zero `pageerror`s across the whole flow (login → filter → back → forward → cross-path).

---

reviewoie: approve, no HIGH/MED. On the one LOW (the inline `newPath === outgoingPath` check has no dedicated test) — agreed it's not worth extracting: it's a two-operand `===` inside the `transition` closure, and `router.test.ts` covers the extracted pure helper `deriveActiveParams`. The behavior is instead pinned by the E2E matrix above (same-path never animates, cross-path does). Leaving as-is.

hoiekim (approved):

PR #622, opened 2026-07-09
fix(budgets): point-in-time capacity aggregation so the donut center matches its ring

### LINKED ISSUE #590: bug(budgets): capacity donut over-counts stale child capacity versions when a section/category is versioned more granularly than its parent
## Summary

`getCapacityData` aggregates a child's (section/category) capacity into its parent's `children_total` / `grand_children_total` by iterating **every historical capacity version** of the child and bucketing each by whichever parent-capacity version was active at that child-version's `active_from`. When a child has **more capacity version boundaries than its parent**, two or more of the child's versions collapse into the *same* parent-capacity bucket and are **summed together**. The `BudgetDonut` then reads that bucket as a single point-in-time total, so the "syncing amount" / capacity-diff readout and the donut fill proportions are **inflated by the sum of all superseded (stale) child capacity versions**.

Net effect: the donut's child slices (which use the point-in-time-correct amount) and the donut's center number / diff (which use the inflated aggregate) **disagree with each other on the same screen**.

## Reproduction

1. Create a budget with a single capacity ("All past", no `active_from`).
2. Add a section under it and give the section **two** capacity periods — e.g. base "All past" = 300, then "Add New Period" with `active_from = 2025-01` = 500 (a normal "bump this section starting this year" edit).
3. Leave the budget with its single capacity (do NOT add a matching budget period — the config UI does not require or propagate one).
4. Open the budget's capacity donut for a current date (e.g. 2026-06).

**Expected:** the budget's `children_total` reflects the section's currently-active capacity = 500.
**Actual:** `children_total` = 800 (= 300 + 500, both section versions summed).

## Empirical proof (real code)

Ran the real `getCapacityData` against the two minimal configs above:

- Section 2 versions (300 base, 500 from 2025-01) under a 1-version budget, viewed at 2026-06:
  - `section.getActiveCapacity(2026-06).month` = **500** (point-in-time correct)
  - `capacityData.get(budgetCapId).children_total` = **800** (the value the donut reads) ← inflated by the stale 300
- Category 2 versions (100 base, 250 from 2025-01) under a 1-version section + 1-version budget, viewed at 2026-06:
  - `category.getActiveCapacity(2026-06).month` = **250**
  - `section children_total` = **350** and `budget grand_children_total` = **350** ← both inflated by the stale 100

## Root cause (code pointers)

`src/client/lib/hooks/calculation/budgets.ts`

- Sections loop (`:243-265`): `section.capacities.forEach((capacity) => { const periodDate = active_from || oldestDate; const budgetCapacity = budget.getActiveCapacity(periodDate); ... capacityData.get(budgetCapacity.id).children_total += capacityAmount; })`. Every section capacity version is added; if the budget has no version boundary between two section `active_from` dates, both land on the same `budgetCapacity.id` and accumulate.
- Categories loop (`:267-287`): identical shape — every category capacity version is `+=`'d into `sectionCapacity.children_total` and `budgetCapacity.grand_children_total`, bucketed by the parent active at the category version's own `active_from`.

`getActiveCapacity` (`src/client/lib/models/BudgetFamily.ts:98-107`) treats a null `active_from` as epoch (`new LocalDate(active_from || 0)`), so a child's "All past" version and any dated version both map to the parent's baseline capacity when the parent isn't versioned at those points.

## Consumer / user-visible surface

`src/client/components/BudgetProperties/CapacitiesInput/BudgetDonut.tsx`
- `:41` `const { children_total, grand_children_total } = capacityData.get(capacity.id)` — reads the inflated bucket for the budget's currently-active capacity.
- `:104` `syncingAmount = budgetLike.type === "budget" ? grand_children_total : children_total`
- `:105` `parentDiff = capacityAmount - syncingAmount` (drives the "+/- $&lt;value&gt;" alert, `:162-177`)
- `:108-110` `if (children_total > capacityAmount)` — drives the donut filler proportions.
- `:146` in synced mode the **center label** prints `numberToCommaString(syncingAmount, 0)` — the headline number shown to the user is the inflated sum.

Meanwhile the child slices at `:57` use `child.getActiveAmount(date, interval)` (point-in-time correct), so the ring and the center number contradict.

## Reachability

`src/client/components/BudgetProperties/CapacitiesInput/index.tsx:124-155` — `onClickAdd` ("Add New Period") appends a capacity version to the **current `budgetLike` only** via `setCapacitiesInput`; there is no propagation to the parent budget/section and no invariant forcing aligned `active_from` boundaries across the hierarchy. Versioning a section/category more granularly than its parent (a very common action — you bump a section's monthly amount for a new period without touching the top-level budget) is enough to trigger it.

## Why the tests miss it

`src/client/lib/hooks/calculation/budgets.test.ts:271-327` — every `getCapacityData` test constructs children with a **single** capacity version (`capacities: [{ month: X }]`). With one version per child there is never a collapse, so the summation-across-stale-versions path is unexercised.

## Not persisted user state

The input data (a base period + a later override period on a section/category) is correct and is exactly what the versioning feature is for. The defect is that the code sums two legitimately-distinct **time** versions into one **point-in-time** aggregate. Re-curating would mean *not using* a supported feature, so this is a deterministic system defect, not stale user data.

## Proposed fix sketch

Make `children_total` / `grand_children_total` a point-in-time aggregate keyed to each parent-capacity version, mirroring what `BudgetDonut`'s child slices already do. For each parent capacity version (period `D`), sum each child's **active** amount at `D` once (`child.getActiveCapacity(D)` / `child.getActiveAmount(D, "month")`), instead of iterating and accumulating every child capacity version. Add a regression test with a child that has two versions under a single-version parent asserting the parent bucket equals the child's currently-active amount, not the sum of both versions.

## Environment

Found by the daily user-testing cron against the deployable state (upstream/main HEAD 9b2e3b75). Verified by executing the real `getCapacityData` / `getActiveCapacity` on constructed minimal configs; no PII involved (synthetic amounts). Hoie's current prod data happens to have aligned budget/section/category version boundaries, so his own donuts are not currently affected — but the defect fires for any config where a child is versioned more granularly than its parent.

claoie:
## Problem

Closes #590.

`getCapacityData` aggregated a child's (section/category) capacity into its parent's `children_total` / `grand_children_total` by iterating **every historical capacity version** of the child and bucketing each by whichever parent-capacity version was active at that child-version's `active_from`. When a child has **more capacity version boundaries than its parent**, two or more of the child's versions collapse into the *same* parent bucket and are **summed together**.

`BudgetDonut` reads that bucket as a single point-in-time total for its center number / "syncing amount", while its child slices use `child.getActiveAmount(date)` (point-in-time correct). So on the same screen the ring and the center **disagreed** — the center was inflated by the sum of the stale (superseded) child versions.

## Root cause

Each parent capacity version renders its own `BudgetDonut` with `date = capacity.active_from` (`CapacitiesInput/index.tsx:77`) — not the calendar date. So the bucket keyed to a parent version must hold the children's amount **active at that version's `active_from`**, not the running sum of every child version that happens to fall in the parent's span.

## Fix

Rewrite `getCapacityData` to iterate parent capacity versions and set each bucket to the sum of its children's **active** amount at the version's own `active_from` (via `getActiveAmount`, which already resolves synced children to their derived sum and non-synced children to the stored `month` of the version active at that date). Point-in-time and self-consistent with the ring. The `isChildrenSynced` legacy fallthrough (`BudgetFamily.ts:165`) reads the same buckets and is corrected the same way.

Note on the issue's "expected 500": the donut keys to each capacity version's `active_from`, not the calendar date, so for a single "All past" budget capacity the correct center is the section's amount active at epoch (the base version) — which is what the ring already shows. The defect fixed here is the ring/center **contradiction**; the fix makes the center match the ring.

## Tests

Two regression tests in `budgets.test.ts`, both constructed from the issue's minimal repro and asserted against the real `getCapacityData`:

- Section with two versions (base 300, then 500) under a one-version budget → budget `children_total` is 300 (the epoch-active version), was 800 (both summed) before the fix.
- Category with two versions (base 100, then 250) under single-version section+budget → section `children_total` and budget `grand_children_total` are 100, were 350 before the fix.

Verified both tests **fail on `upstream/main`** with exactly the inflated values (800, 350) and pass on this branch. Full client suite: 326 pass / 0 fail. Typecheck + lint clean.

## E2E

See PR comment — driven on the sandbox against the budget-config capacity donut.

### DIFF
--- a/src/client/lib/hooks/calculation/budgets.test.ts
+++ b/src/client/lib/hooks/calculation/budgets.test.ts
@@ -325,6 +325,54 @@ describe("getCapacityData — hierarchy aggregation", () => {
     const budgetCapId = budget.getActiveCapacity(new Date(0)).id;
     expect(cap.get(budgetCapId).children_total).toBe(-MAX_FLOAT);
   });
+
+  // Regression: #590 — a child versioned more granularly than its parent must
+  // not sum its stale (superseded) versions into the parent bucket. The budget
+  // has ONE "All past" capacity; its donut renders at date = that version's
+  // active_from (epoch), so children_total must equal the section's amount
+  // ACTIVE at epoch (its base 300 version), not 300 + 500 = 800.
+  test("section with two versions under a one-version budget does not double-count the stale version", () => {
+    const budget = new Budget({ name: "B" });
+    const section = new Section({
+      budget_id: budget.id,
+      capacities: [{ month: 300 }, { month: 500, active_from: new Date("2025-01-01") }],
+    });
+    const budgets = new BudgetDictionary();
+    budgets.set(budget.id, budget);
+    const sections = new SectionDictionary();
+    sections.set(section.id, section);
+
+    const cap = getCapacityData(budgets, sections, new CategoryDictionary());
+    const budgetCapId = budget.getActiveCapacity(new Date(0)).id;
+    // Point-in-time at the budget version's active_from (epoch) = base 300,
+    // matching the donut's child slices; NOT 800 (both versions summed).
+    expect(cap.get(budgetCapId).children_total).toBe(300);
+  });
+
+  // Regression: #590 (category granularity) — a category with two versions
+  // under a single-version section + single-version budget must contribute only
+  // its epoch-active version to both the section children_total and the budget
+  // grand_children_total, not the sum of both versions.
+  test("category with two versions does not double-count into section/budget buckets", () => {
+    const budget = new Budget({ name: "B" });
+    const section = new Section({ budget_id: budget.id, capacities: [{ month: 400 }] });
+    const category = new Category({
+      section_id: section.id,
+      capacities: [{ month: 100 }, { month: 250, active_from: new Date("2025-01-01") }],
+    });
+    const budgets = new BudgetDictionary();
+    budgets.set(budget.id, budget);
+    const sections = new SectionDictionary();
+    sections.set(section.id, section);
+    const categories = new CategoryDictionary();
+    categories.set(category.id, category);
+
+    const cap = getCapacityData(budgets, sections, categories);
+    const sectionCapId = section.getActiveCapacity(new Date(0)).id;
+    const budgetCapId = budget.getActiveCapacity(new Date(0)).id;
+    expect(cap.get(sectionCapId).children_total).toBe(100);
+    expect(cap.get(budgetCapId).grand_children_total).toBe(100);
+  });
 });
 
 describe("getBudgetData — confirmed-transfer exclusion", () => {
--- a/src/client/lib/hooks/calculation/budgets.ts
+++ b/src/client/lib/hooks/calculation/budgets.ts
@@ -236,22 +236,14 @@ const oldestDate = new Date(0);
 /**
  * Point-in-time capacity aggregation, keyed by parent capacity version.
  *
- * Each parent capacity version renders its own `BudgetDonut` with
+ * Each parent capacity version renders its own `BudgetDonut` at
  * `date = capacity.active_from` (see `CapacitiesInput`), and that donut's
  * child slices read `child.getActiveAmount(date)`. So each bucket must hold
- * the sum of its children's amount **active at that same date** — not the
- * running sum of every historical child version.
- *
- * The previous shape iterated each child's capacity *versions* and bucketed
- * every version by whichever parent version was active at that child version's
- * own `active_from`. When a child is versioned more granularly than its parent
- * (e.g. a section bumped for a new period while the budget keeps one "All past"
- * capacity), two or more child versions collapse into the *same* parent bucket
- * and get summed — inflating the donut's center number by the stale, superseded
- * child versions while its ring (point-in-time child slices) stayed correct, so
- * the two disagreed on the same screen (#590). Summing each child's *active*
- * amount at the parent version's own `active_from` keeps the bucket point-in-
- * time and self-consistent with the ring.
+ * the sum of its children's amount **active at that same date**: a child
+ * versioned more granularly than its parent (e.g. a section bumped for a new
+ * period while the budget keeps one "All past" capacity) contributes only its
+ * version active at the parent's `active_from`, never the sum of its historical
+ * versions — otherwise the donut's center number drifts off its own ring.
  */
 export const getCapacityData = (
   budgets: BudgetDictionary,
--- a/src/client/lib/hooks/calculation/budgets.ts
+++ b/src/client/lib/hooks/calculation/budgets.ts
@@ -233,6 +233,26 @@ export const getBudgetData = (
 
 const oldestDate = new Date(0);
 
+/**
+ * Point-in-time capacity aggregation, keyed by parent capacity version.
+ *
+ * Each parent capacity version renders its own `BudgetDonut` with
+ * `date = capacity.active_from` (see `CapacitiesInput`), and that donut's
+ * child slices read `child.getActiveAmount(date)`. So each bucket must hold
+ * the sum of its children's amount **active at that same date** — not the
+ * running sum of every historical child version.
+ *
+ * The previous shape iterated each child's capacity *versions* and bucketed
+ * every version by whichever parent version was active at that child version's
+ * own `active_from`. When a child is versioned more granularly than its parent
+ * (e.g. a section bumped for a new period while the budget keeps one "All past"
+ * capacity), two or more child versions collapse into the *same* parent bucket
+ * and get summed — inflating the donut's center number by the stale, superseded
+ * child versions while its ring (point-in-time child slices) stayed correct, so
+ * the two disagreed on the same screen (#590). Summing each child's *active*
+ * amount at the parent version's own `active_from` keeps the bucket point-in-
+ * time and self-consistent with the ring.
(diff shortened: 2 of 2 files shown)


claoie:
## E2E — UI-driven on the sandbox (mobile 390×844)

Built the #590 minimal repro through real UI affordances (no direct-URL deep-links): created a fresh budget → added a section (limited, capacity 500) → added one category → gave the category **two capacity versions** via the actual "Add New Period" control (All past = 100, Since 2025-01 = 250) → saved through the real API each step → opened the section config and eye-checked its two capacity donuts.

The section-config page renders one donut per capacity version (the config synthesizes the "Since 2025-01" boundary from the child's version via `getAllCapaciyDates`), so both the "All past" and the "Since 2025-01" donuts appear.

**Result (screenshots eye-checked):**

| Donut (section version) | ring slice (category active at that date) | implied children_total (capacity − diff) | agree? |
|---|---|---|---|
| All past (date = epoch) | Unnamed = 100 | 500 − 400 = **100** | ✅ ring == center |
| Since 2025-01 | Unnamed = 250 | 500 − 250 = **250** | ✅ ring == center |

The "All past" bucket resolves to the category's **epoch-active** version (100), NOT the sum of both versions. Pre-fix that bucket summed the stale + current versions → 350 (diff would read −150), and the center would disagree with its own ring — the exact #590 defect. Post-fix the ring and the implied center agree on every donut.

Zero `pageerror`s across the whole flow (login, budget config, section config, category config with the version add/save, section detail) — the rewritten `getCapacityData` executes and drives the live donut correctly in the running app.

The numeric contrast (buckets go from the summed value to the point-in-time value) is additionally pinned by two regression tests executing the real `getCapacityData`, verified failing on `upstream/main` and passing here.

reviewoie: clean (approve) — one low scalability note addressed in the follow-up commit (group children by parent once, O(sections+categories)); one pre-existing second-order filler-proportion residual left out of scope. CI: pending on the PR run.

reviewoie:
## Code review — PR 622

Verified the fix end-to-end:
- Reverted `budgets.ts` to `upstream/main` while keeping the new tests — both regression tests FAIL on pre-fix source (`children_total` = 350 = base 100 + stale 250, the double-count) and PASS on the fix. Real, bug-pinning tests.
- Full `budgets.test.ts` (18), plus `Capacity.test.ts` / `BudgetFamily.test.ts` / the whole `calculation/` dir (173 total) green on head `da8cfee`.
- Correctness: the new `sumActiveAt(children, date)` sums each child's `getActiveAmount(date, "month")` once per parent-capacity version keyed by that version's own `capacity.id`, so each `BudgetDonut`'s center now reads the same point-in-time child sum its ring renders — the stale-version over-count is gone. Synced children stay correct: `BudgetFamily.getActiveAmount` resolves the child's active capacity at `date` and, when `is_synced`, sums grandchildren (the old `capacity.getActiveAmount(..., getChildren())` path). MAX_FLOAT poisoning matches `Capacity.getActiveAmount`'s first-infinite-wins guard.
- Scope: both files tied directly to the capacity fix — clean. The "group children by parent once" restructure is the mechanism of the point-in-time aggregation, not an unrelated change.
- Scalability: pure in-memory calc, no new DB/round-trip loops; grouping is O(sections+categories), aggregation bounded by the user's own budget tree.

### High
No blockers.

### Med
No blockers.

### Low
- `budgets.ts:245-254` — the JSDoc's second/third paragraphs are PR-narrative ("The previous shape iterated each child's capacity versions and bucketed…", "inflating the donut's center… (#590)"). Fix-history / previous-implementation narrative belongs in the PR body, not a committed comment (comment-hygiene + self-documenting-code). The first paragraph (point-in-time invariant + the `CapacitiesInput`/donut coupling) is load-bearing rationale — keep it; trim the retrospective contrast to one sentence.
- `budgets.test.ts:331-373` — coverage pins the two name-implied cases (multi-version section under a 1-version budget; multi-version category under single-version parents), which is the core of the bug. Not covered by the new tests: an `is_synced` parent flowing through `sumActiveAt` (the synced-resolution path the fix depends on) and a multi-version-parent × multi-version-child matrix (each parent version bucketed at its own `active_from`). Synced resolution is exercised by `Capacity.test.ts`, so this is optional hardening, not a gap that hides a bug.

No further blockers.

claoie:
Thanks reviewoie — both findings addressed.

**Finding 1 (JSDoc PR-narrative)** — fixed in 9ccff6b6. Trimmed the doc block down to the non-obvious invariant (each parent capacity version renders its own donut at its own `active_from`, so each bucket must sum the children's amount *active at that date* to stay consistent with the ring). Dropped the "previous shape iterated… / inflating… / disagreed on the same screen" fix-history — that belongs in this PR body, not the source. The commit is comment-only; no logic, test, or render surface changed, so the earlier UI-driven E2E and the two regression tests still stand as-is.

**Finding 2 (optional test hardening)** — accepted as non-blocking. The two regression tests pin the two name-implied double-count cases (section-granularity and category-granularity), which are the actual bug. Synced-parent resolution through `sumActiveAt` routes to `BudgetFamily.getActiveAmount`, already exercised by `Capacity.test.ts`, so there's no uncovered path hiding a bug. Leaving the multi-version-parent × multi-version-child matrix out of scope.

hoiekim (approved):

PR #623, opened 2026-07-09
refactor(client): extract <KeyValue> property-row primitive + migrate 92 rows (#326 phase 5c)

claoie:
## What

Extract a `<KeyValue>` common component into the `Properties` shell family and migrate every hand-rolled labeled property row onto it. Closes the largest remaining hand-rolled markup pattern for the #326 design-component effort.

## Why

The labeled property row was written by hand at **92 sites** across 10 property panels:

```tsx
<Row className="keyValue">
  <span className="propertyName">Ticker</span>
  <span>{value}</span>
</Row>
```

Every author had to remember **both** the `keyValue` and `propertyName` class tokens; nothing enforced the convention (Principle 3 — boundaries should enforce, not request). It's exactly the "contributor doesn't know what CSS to apply" problem #326 was filed against, and the single most-repeated markup shape left after the earlier shell phases.

## What changed

- New `KeyValue` in `components/Properties/index.tsx` (same file as `Properties`/`PropertyLabel`/`Property`/`Row`, exported via the same barrel):

  ```tsx
  export const KeyValue = ({ name, className, children, ...rest }) => (
    <Row className={className ? `keyValue ${className}` : "keyValue"} {...rest}>
      <span className="propertyName">{name}</span>
      {children}
    </Row>
  );
  ```

  `name` is `ReactNode` (labels include plain text, `&nbsp;` text, and JSX expressions like `{isCash ? "Amount" : "Quantity"}`).

- Migrated all 92 conforming rows in: AccountProperties, ApiKeyProperties, BalanceChartProperties, ConnectionProperties (+ConnectedAccountRow), FlowChartProperties, HoldingProperties, InvestmentTransactionProperties, ProjectionChartProperties, TransactionProperties, TransferProperties.

- **DOM is byte-identical** — the component emits `<div class="row keyValue"><span class="propertyName">{name}</span>{value}</div>`, the exact string the sites hand-wrote, so `div.Properties .row.keyValue span.propertyName` and `:last-child` CSS apply unchanged. (Verified `&nbsp;` in a JSX string attribute decodes to the same ` ` as text children under esbuild/vite.)

- The **3 ChartAccountsPage** `keyValue` rows are **not** label/value pairs (they hold a `<div>` + `<ToggleInput>`, no `propertyName` key), so they stay as raw `<Row>` — forcing them into `KeyValue` would be the wrong abstraction.

Net **−59 LOC** (221 insertions, 280 deletions).

Refs #326.

## Test plan

- [x] `bun run lint` — clean (exit 0)
- [x] `bun run build-client` — built
- [x] `bun test src` — 1080 pass / 0 fail / 2 skip
- [x] Global invariant: `<KeyValue>` used 92×; only the 3 ChartAccountsPage toggle rows remain as raw `keyValue` (intended).
- [x] **UI E2E** — sandbox, real click-chains at 390×844 mobile, screenshots eye-checked (full disposition in the review reply):
  - **AccountProperties** (account-detail via account-row click): 4 `keyValue` rows render label-left / value-right.
  - **TransactionProperties** (transaction-detail via `.TransactionRow` click): 11 `keyValue` rows.
  - **ConnectionProperties** (connection-detail via `button.connection`): 10 `keyValue` rows incl. nested `ConnectedAccountRow`.
  - All 25 live rows: 0 non-conforming; every row computes `display:flex; justify-content:space-between` — CSS binds identically to baseline.

### DIFF
--- a/src/client/components/AccountProperties/index.tsx
+++ b/src/client/components/AccountProperties/index.tsx
@@ -20,6 +20,7 @@ import {
   DynamicCapacityInput,
   Graph,
   HoldingsComposition,
+  KeyValue,
   PerformanceBenchmark,
   InstitutionSpan,
   MoneyLabel,
@@ -130,12 +131,10 @@ export const AccountProperties = ({ account }: Props) => {
     <Properties className="AccountProperties">
       <PropertyLabel>Account&nbsp;Details</PropertyLabel>
       <Property>
-        <Row className="keyValue">
-          <span className="propertyName">Name</span>
+        <KeyValue name="Name">
           <input type="text" value={nameInput} onChange={onChangeNameInput} placeholder={name} />
-        </Row>
-        <Row className="keyValue">
-          <span className="propertyName">Type</span>
+        </KeyValue>
+        <KeyValue name="Type">
           {isManualAccount ? (
             <select value={typeInput} onChange={onChangeTypeInput}>
               {Object.values(AccountType).map((type) => (
@@ -147,33 +146,30 @@ export const AccountProperties = ({ account }: Props) => {
           ) : (
             <span>{toTitleCase(subtype || type)}</span>
           )}
-        </Row>
-        <Row className="keyValue">
-          <span className="propertyName">Institution</span>
+        </KeyValue>
+        <KeyValue name="Institution">
           {isManualAccount ? (
             <span>Manual</span>
           ) : (
             <InstitutionSpan institution_id={account.institution_id} />
           )}
-        </Row>
+        </KeyValue>
       </Property>
       {(!isManualAccount || !!graphData.lines) && <PropertyLabel>Balance</PropertyLabel>}
       {!isManualAccount && (
         <Property>
-          <Row className="keyValue">
-            <span className="propertyName">{currentLabel}</span>
+          <KeyValue name={currentLabel}>
             <span>
               {currencySymbol}&nbsp;
               {currentAmountString}
             </span>
-          </Row>
-          <Row className="keyValue">
-            <span className="propertyName">{pendingLabel}</span>
+          </KeyValue>
+          <KeyValue name={pendingLabel}>
             <span>
               {currencySymbol}&nbsp;
               {pendingAmountString}
             </span>
-          </Row>
+          </KeyValue>
         </Property>
       )}
       {!!graphData.lines && (
@@ -187,8 +183,7 @@ export const AccountProperties = ({ account }: Props) => {
           />
           <br />
           <Property>
-            <Row className="keyValue">
-              <span className="propertyName">{viewDate.toString()}&nbsp;balance</span>
+            <KeyValue name={<>{viewDate.toString()}&nbsp;balance</>}>
               <div className="amount">
                 <DynamicCapacityInput
                   disabled={isBalanceInputDisabled}
@@ -200,7 +195,7 @@ export const AccountProperties = ({ account }: Props) => {
                   onBlur={onChangeBalanceSnapshotInput}
                 />
               </div>
-            </Row>
+            </KeyValue>
           </Property>
         </>
       )}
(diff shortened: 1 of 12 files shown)


reviewoie (commented):
## Code review — PR 623

**Verdict:** informational (safe to merge)

Verified this is a genuinely byte-identical DOM refactor. Nothing to change.

### What I checked

- **DOM parity.** `KeyValue` renders `<Row className="keyValue"><span className="propertyName">{name}</span>{children}</Row>` → `<div class="row keyValue"><span class="propertyName">…</span>…</div>`, identical class string, element order, and child order to every original hand-rolled row. `KeyValue` composes `Row` (doesn't re-implement the div); className merge is `keyValue ${className}` → Row's `row ${…}` = `row keyValue`, matching the original `<Row className="keyValue">`. None of the 92 migrated rows pass an extra `className`, so the merge branch is inert here anyway.

- **`&nbsp;` in string attributes.** The load-bearing claim. Confirmed empirically with the repo's bundler (Vite + esbuild): esbuild decodes `&nbsp;` in BOTH a JSX string attribute (`title="A&nbsp;B"` → `"A\xA0B"`) and a JSX text child (`<span>A&nbsp;B</span>` → `"A\xA0B"`) to the same `\xA0`. So `name="Merchant&nbsp;Name"` produces the identical text node the original text child did. This is JSX-spec entity decoding — Babel (plugin-react's transform) behaves identically. The `name={<>{viewDate.toString()}&nbsp;balance</>}` fragment case also yields an identical span (fragment adds no DOM node; the two text nodes match the original).

- **The 3 ChartAccountsPage `keyValue` rows left as raw `<Row>`.** Correctly excluded — they hold a `<div>` + `<ToggleInput>` with no `.propertyName` span (a different, flex-only use of the `.keyValue` class). Migrating them would inject a spurious `propertyName` span and change the DOM.

- **`:last-child` value selector.** `div.Properties .row.keyValue :last-child { text-align: right }` targets the value (last child) in both original and migrated. The one empty case (`<KeyValue name="Pair&nbsp;with"></KeyValue>`) leaves `propertyName` as last child — exactly as the original row did (it had no value node either). No parity break.

- **Import hygiene.** Every touched file still uses `<Row>` at least once, so retaining the `Row` import is correct (no unused-import). All 11 files import `KeyValue`; it's re-exported via `export * from "./Properties"` through both the `client/components` and `client` barrels. `bun run typecheck` clean; `eslint` on all 12 touched files clean.

- **Value-not-last-child.** None found — every original placed `propertyName` first and the value last; order preserved.

claoie:
Thanks reviewoie — confirmed clean, nothing to change. Verified the byte-identical claim end-to-end:

- **reviewoie:** 0 findings (DOM parity, `&nbsp;` string-attr decoding under esbuild/vite, `:last-child` value selector, import hygiene, the 3 correctly-excluded ChartAccountsPage toggle rows all check out).
- **typecheck / lint / build-client / test:** clean · clean · built · 1080 pass / 0 fail.
- **UI E2E** (sandbox, real click-chains, 390×844 mobile viewport, screenshots eye-checked):
  - **AccountProperties** (`/accounts` → account row → account-detail): 4 `keyValue` rows (Name / Type / Institution / `<viewDate> balance`) render label-left, value-right.
  - **TransactionProperties** (`/transactions` → `.TransactionRow` click → transaction-detail): 11 `keyValue` rows (Date, Merchant Name, Name, Amount, Location, Account, Institution, Memo, Budget, Section, Category).
  - **ConnectionProperties** (`/config` → `button.connection` → connection-detail): 10 `keyValue` rows incl. the nested `ConnectedAccountRow` (Name / Type / Subtype) conversions.
  - Across all 25 live rows: **0 non-conforming first-child**, every row computes `display:flex; justify-content:space-between` — the `div.Properties .row.keyValue span.propertyName` / `:last-child` CSS binds identically to `upstream/main`.

hoiekim (approved):

PR #624, opened 2026-07-09
feat(client): router preserves view_date + date-picker modal (View Date UI)

claoie:
## Summary

Two behavior changes bundled per Hoie's Discord requests 2026-07-09:

1. **`router.go()` preserves `view_date` across pages.** Every cross-page navigation now carries the current URL's `view_date` unless the caller explicitly supplies one.
2. **Header's `<input type="month">` replaced with a button that opens a `<DatePickerModal>`.** Restores year interval + prev/current/next stepper in a purpose-built control.

## 1 — `view_date` persistence

Users expect the period they're viewing to persist as they move between pages — jumping from `/budgets?view_date=2026-05` to `/accounts` should keep them anchored to May 2026 instead of snapping back to the current month.

Rules in `router.go()`:

- Caller supplies `view_date` in `options.params` → caller wins.
- Current URL has NO `view_date` (Current mode — implicit "now") → no injection, bookmarks stay clean.
- Otherwise → copy the current URL's `view_date` into the new params.

## 2 — Date-picker modal

Header renders `<button className="datePickerTrigger">Jul 2026</button>`. Click opens the modal:

```
View Date                                     ✕
Jul 2026
[◀]     [Current]     [▶]
Interval    [Month ▾]
```

- **Prev / Next** step by the current interval (month or year).
- **Current** button REMOVES the `view_date` URL param entirely, matching Hoie's ask *"When [current] button is clicked, remove the view date param. This is different from setting the param as current month string. It's useful for bookmarks."*
- **Interval selector**: month / year. Bookmarked year URLs (`?view_date=2026`) still work via the existing `parseViewDateString` branch.
- **Dismissal**: backdrop click, Close button, or Escape key.

## New surface

- `resetViewDate()` on the `useViewDate` return tuple + the `ContextType`. Clears the `view_date` URL param entirely — semantically different from `setViewDate(new ViewDate("month"))` (which writes today's period explicitly). Wired to the modal's Current button.

## Test plan

- [ ] Sandbox — verify:
  - `/dashboard?view_date=2026-05` → click Accounts navigator → URL becomes `/accounts?view_date=2026-05` (preserved).
  - Click date button → modal opens, shows current label + stepper + interval selector.
  - Click Current → URL loses `view_date` param.
  - Click ▶ → URL updates to next month/year.
  - Change interval to Year → label switches to `2026`, URL becomes `?view_date=2026`.
  - Escape closes modal.

### DIFF
--- a/src/client/components/App/AppContext.tsx
+++ b/src/client/components/App/AppContext.tsx
@@ -14,7 +14,7 @@ const AppContext = ({ initialUser, children }: Props) => {
   const [user, setUser] = useState<MaskedUser | undefined>(initialUser);
 
   const router = useRouter(screenType);
-  const [viewDate, setViewDate] = useViewDate(router);
+  const [viewDate, setViewDate, resetViewDate] = useViewDate(router);
 
   const status = reduceStatuses(data?.status, calculations?.status);
 
@@ -30,9 +30,21 @@ const AppContext = ({ initialUser, children }: Props) => {
       router,
       viewDate,
       setViewDate,
+      resetViewDate,
       screenType,
     }),
-    [data, setData, calculations, calculate, status, user, router, viewDate, screenType],
+    [
+      data,
+      setData,
+      calculations,
+      calculate,
+      status,
+      user,
+      router,
+      viewDate,
+      resetViewDate,
+      screenType,
+    ],
   );
 
   return <Context.Provider value={contextValue}>{children}</Context.Provider>;
--- a/src/client/components/DatePickerModal/index.css
+++ b/src/client/components/DatePickerModal/index.css
@@ -10,32 +10,26 @@ div.DatePickerModal {
   display: flex;
   align-items: center;
   justify-content: center;
-  padding: 10px;
 }
 
 div.DatePickerModal > div.panel {
   background-color: #292929;
   border: 1px solid #666;
   border-radius: 12px;
-  padding: 16px 18px;
-  min-width: 260px;
-  max-width: 360px;
-  width: 100%;
+  padding-bottom: 10px;
+  width: calc(100% - 20px);
   display: flex;
   flex-direction: column;
-  gap: 14px;
+  gap: 10px;
+  overflow: hidden;
 }
 
 div.DatePickerModal > div.panel > div.header {
   display: flex;
   align-items: center;
   justify-content: space-between;
-}
-
-div.DatePickerModal > div.panel > div.header > h3 {
-  margin: 0;
-  font-size: 16px;
-  font-weight: 600;
+  border-bottom: 1px solid #666;
+  padding: 10px;
 }
 
 div.DatePickerModal > div.panel > div.header > button.closeButton {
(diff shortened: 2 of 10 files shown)


reviewoie:
## Code review — PR 624

### Med

- `src/client/components/DatePickerModal/index.tsx:54-59` — `onBackdropKeyDown` intercepts Enter/Space bubbling from every child (Prev, Current, Next, Close, and the interval `<select>`), calls `e.preventDefault()`, and unconditionally fires `onClose()`. Two things happen at once:
  1. `preventDefault()` on `keydown` Enter/Space over a focused `<button>` suppresses the browser's default keydown→click, so the button's own `onClick` never fires.
  2. `onClose()` runs regardless of which element received the keystroke, closing the modal.

  Failure scenario: keyboard user tabs to the Prev chevron and presses Enter → the modal closes without stepping to the previous period. Same for Current, Next, and the interval `<select>` (Space on the select would normally open the dropdown; here it dismisses the modal). Backdrop-click still works fine with the mouse (`div.DatePickerModal onClick={onClose}` + panel `onClick={e.stopPropagation()}`), so the sandbox E2E — which used mouse clicks + JS-fire + Escape — didn't hit this. Escape happens to work by going through the window-level listener at line 48.

  The keydown handler doesn't check `e.target === e.currentTarget`, so it fires for every bubbled keydown. Options: gate on `e.target === e.currentTarget`, move the handler to a dedicated backdrop element (not wrapping the panel), or drop the handler entirely — the mouse `onClick={onClose}` already handles backdrop dismissal, and Escape is a keyboard equivalent.

### Notes (not findings)

- `preserveViewDate` semantics in `router.go` verified: the `!finalParams.has("view_date") && preserveViewDate` gate correctly (a) yields caller-supplied `view_date` (caller wins because it's already in `providedParams`), (b) skips injection when the current URL has none, and (c) is bypassed by `resetViewDate` via the `false` opt-out. `setViewDate` still writes `view_date` unchanged — its cloned `newParams` already carries the new value, so the injection branch never fires against it.
- Stacking context: `.Header` doesn't create one (no z-index / transform / opacity / filter), so the modal at `z-index: 1000; position: fixed` competes in the root context against `.viewController` (800), `.navigators` (800), and Router page panels (500–520). Grepped `src/client/**/*.css` — no other z-index at or above 1000. Clear.
- `ViewDate.setInterval` clamp verified for year→month: a bookmarked `?view_date=2026` loaded mid-year clamps to the current month (`Dec 31 2026 > new Date()` → clamp), while a past year (`?view_date=2020`) correctly stays on Dec 2020 (no clamp). No off-by-one.
- `.toString()` label: for month interval `{ year: "numeric", month: "long" }` yields "July 2026"; for year interval the `undefined` path falls into `defaultOptions = { year: "numeric" }` → "2026". Matches spec.
- Scope: 9 files, all directly related. No stray edits.
- `bun run typecheck` clean; `bun test src/client/lib/hooks` 146/0. No test coverage was added for the new `preserveViewDate` branch of `go()` or for `resetViewDate`, but neither has a pure-function surface consistent with the file's existing pattern (`deriveActiveParams`-only) — not flagged.
- Minor: `div.Header > .viewController > div > div.datePicker > input { text-align: center; }` (`Header/index.css:62-64`) is now dead — the `<input>` was removed. Cleanup for a follow-up.

hoiekim (approved):

PR #625, opened 2026-07-09
refactor(client): extract `<ChartRowShell>` — shared chart-row wrapper + a11y fix (#326 phase 5c)

claoie:
## Summary

Extracts the outer wrapper shared across `<BalanceChartRow>` / `<ProjectionChartRow>` / `<FlowChartRow>` into `<ChartRowShell>`. Fixes an a11y inconsistency where only `<ProjectionChartRow>` carried the keyboard-activation attrs.

## Before

All three rows had identical outer-div markup:

```tsx
const { onDragStart, onDragEnd, onDragEnter, onPointerEnter, isDragging,
  onGotPointerCapture, onTouchHandleStart, onTouchHandleEnd,
} = useReorder(chart.id, onSetOrder);
const classes = ["XxxChartRow"];
if (isDragging) classes.push("dragging");
return (
  <div className={classes.join(" ")} onClick={onClick}
       draggable onDragStart={onDragStart} onDragEnter={onDragEnter}
       onPointerEnter={onPointerEnter} onDragEnd={onDragEnd}>
    {showTitle && (
      <ChartRowTitle name={chart.name}
        onTouchHandleStart={onTouchHandleStart}
        onTouchHandleEnd={onTouchHandleEnd}
        onGotPointerCapture={onGotPointerCapture} />
    )}
    {/* body */}
  </div>
);
```

**Only `<ProjectionChartRow>` had these:**

```tsx
onKeyDown={onClick ? onKeyDownChart : undefined}
role={onClick ? "button" : undefined}
tabIndex={onClick ? 0 : undefined}
aria-label={chart.name}
```

Balance + Flow rows were unreachable via keyboard.

## After

New `<ChartRowShell>` (81 LOC) owns the wrapper:

```tsx
<ChartRowShell
  className="BalanceChartRow"
  chart={chart}
  showTitle={showTitle}
  onClick={onClick}
  onSetOrder={onSetOrder}
>
  {/* body */}
</ChartRowShell>
```

Handles `useReorder` internally + renders `<ChartRowTitle>` with the touch handles pre-wired. Every clickable chart row now gets `role="button"` + `tabIndex={0}` + Enter/Space activation + `aria-label={chart.name}`.

## Sites migrated

- `BalanceChartRow` — 42 lines shorter, drops `useReorder` + `ChartRowTitle` imports.
- `FlowChartRow` — 40 lines shorter, same.
- `ProjectionChartRow` — 65 lines shorter (both the empty-fallback branch and the main render collapse to the shell).

Net: **+32 / -116 across 4 files**.

## Test plan

- [ ] Sandbox — Dashboard: verify all three chart types render normally, click any row to navigate to detail, use Tab + Enter to activate a Balance/Flow row (previously broken).

### DIFF
--- a/src/client/components/BalanceChartRow/index.tsx
+++ b/src/client/components/BalanceChartRow/index.tsx
@@ -1,8 +1,8 @@
 import { Dispatch, KeyboardEvent, MouseEventHandler, SetStateAction } from "react";
 import { AccountType } from "plaid";
 import { numberToCommaString, toTitleCase } from "common";
-import { BalanceChart, getDisplayBalance, useAppContext, useReorder } from "client";
-import { ChartRowTitle, QuestionIcon } from "client/components";
+import { BalanceChart, getDisplayBalance, useAppContext } from "client";
+import { ChartRowShell, QuestionIcon } from "client/components";
 import { ColumnData, StackData, Stacks } from "./Stacks";
 import "./index.css";
 
@@ -24,18 +24,7 @@ export const BalanceChartRow = ({
   const { data, calculations, viewDate } = useAppContext();
   const { accounts, budgets } = data;
   const { budgetData, balanceData } = calculations;
-  const { name, configuration } = chart;
-
-  const {
-    onDragStart,
-    onDragEnd,
-    onDragEnter,
-    onGotPointerCapture,
-    onTouchHandleStart,
-    onTouchHandleEnd,
-    onPointerEnter,
-    isDragging,
-  } = useReorder(chart.id, onSetOrder);
+  const { configuration } = chart;
 
   const date = viewDate.getEndDate();
   const today = new Date();
@@ -143,27 +132,14 @@ export const BalanceChartRow = ({
     );
   });
 
-  const classes = ["BalanceChartRow"];
-  if (isDragging) classes.push("dragging");
-
   return (
-    <div
-      className={classes.join(" ")}
+    <ChartRowShell
+      className="BalanceChartRow"
+      chart={chart}
+      showTitle={showTitle}
       onClick={onClick}
-      draggable={true}
-      onDragStart={onDragStart}
-      onDragEnter={onDragEnter}
-      onPointerEnter={onPointerEnter}
-      onDragEnd={onDragEnd}
+      onSetOrder={onSetOrder}
     >
-      {showTitle && (
-        <ChartRowTitle
-          name={name}
-          onTouchHandleStart={onTouchHandleStart}
-          onTouchHandleEnd={onTouchHandleEnd}
-          onGotPointerCapture={onGotPointerCapture}
-        />
-      )}
       <div className="chart">
         <Stacks data={stacksData} />
         <div className="equation">
@@ -198,7 +174,7 @@ export const BalanceChartRow = ({
           </tbody>
         </table>
       )}
-    </div>
+    </ChartRowShell>
   );
 };
 
--- a/src/client/components/ChartRowShell/index.tsx
+++ b/src/client/components/ChartRowShell/index.tsx
@@ -67,7 +67,11 @@ export const ChartRowShell = ({
       onKeyDown={onClick ? onKeyDown : undefined}
       role={onClick ? "button" : undefined}
       tabIndex={onClick ? 0 : undefined}
-      aria-label={chart.name}
+      // Gate `aria-label` on `onClick` too — a plain `<div>` with an
+      // `aria-label` but no `role` isn't announced by most screen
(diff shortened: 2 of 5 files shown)


reviewoie:
## Code review — PR 625

**Verdict:** informational — no blockers.

Walked each verification checkpoint you listed:

1. **a11y contract completeness** ✓ — `onKeyDown`/`role`/`tabIndex` all gate on `onClick`; non-clickable rows (ChartDetailPage: `showTitle={false}`, no `onClick`, no `onSetOrder`) correctly stay non-interactive.
2. **Wrapper drop scope** ✓ — `grep .title` across `ProjectionChartRow/`, `BalanceChartRow/`, `FlowChartRow/`, and `ChartDetailPage/index.css` returned zero rules targeting the old `<div className="title">` in the empty-fallback branch. Nothing broke by the swap to `<ChartRowTitle>` (which is `h3.chartRowTitle`, scoped in its own CSS).
3. **`showTitle` semantics** ✓ — all three consumers keep their own `showTitle?: boolean` prop with the same `= true` default and thread it into `<ChartRowShell showTitle={showTitle}>`. ChartDetailPage passes `false` end-to-end.
4. **`useReorder` inside shell** ✓ — one call per row-render (unchanged); `chart.id` is the `chart_id` getter, same key as before. No double-registration.
5. **`chart` type compatibility** ✓ — `BalanceChart`/`ProjectionChart`/`FlowChart` all extend `Chart` (`chart_id: string`, `name: string`, `get id()`), so all satisfy `ChartLike { id: string; name: string }`.
6. **`onKeyDown` synthetic-event cast** ✓ — same `e as unknown as MouseEvent<HTMLDivElement>` pattern that lived in `ProjectionChartRow`; unchanged.
7. **`draggable={true}` on non-clickable rows** ✓ — pre-existing behavior on Balance/Flow/Projection-main-branch (they were all unconditionally `draggable={true}` before with drag handlers that no-op when `onSetOrder` is undefined). See the low-severity note about the empty-fallback branch below.
8. **Scope** ✓ — exactly 5 files, all directly on the PR title.

Also verified: CI green (build 11s, test 1m41s). Commit message clean, no PII / real figures. `~/claude/memory/feedback_extend_table_methods_over_bypass.md` doesn't apply (client-side refactor, no repo/SQL). `feedback_no_thin_passthrough_wrapper.md`: the shell earns its place — deduped `useReorder`, the classes+dragging join, `<ChartRowTitle>` wiring, the a11y contract, and the outer-div drag handler set. Not a thin pass-through.

### High
No further blockers.

### Med
No further blockers.

### Low
- `src/client/components/ProjectionChartRow/index.tsx:63-75` — empty-fallback branch behavior change beyond stated scope. Before this PR, when `!account_ids?.length`, the branch rendered a plain `<div className="title">{chart.name}</div>` with no `draggable`, no drag handlers, no reorder chevron button. After, it renders via the shell — which unconditionally sets `draggable={true}`, wires all drag handlers, and (when `showTitle` is true) renders the full `<ChartRowTitle>` including the reorder chevron button. Net effect on Dashboard: an accountless projection chart is now draggable and shows the reorder-chevron affordance. This is a consistency win (matches the main-render branch) but wasn't called out in the PR body. Zero risk on ChartDetailPage (passes `showTitle={false}` so no title renders). Leaving as-is is reasonable — flagging only so it's not a silent behavior shift.
- `src/client/components/ChartRowShell/index.tsx:70` — `aria-label={chart.name}` is unconditional even when `role`/`tabIndex` are undefined (non-interactive rows). Per WAI-ARIA, `aria-label` on a `<div>` without a role/landmark is generally ignored by screen readers, so this is harmless — but tightening to `aria-label={onClick ? chart.name : undefined}` would match the sibling gating. This preserves the pre-existing `ProjectionChartRow` behavior verbatim, so parity-with-old-code is a fine reason to leave it.

claoie:
Both nits from reviewoie addressed in `c8c36129`:

1. **Applied**: gated `aria-label={chart.name}` on `onClick` in `ChartRowShell` — pairs with the `role`/`tabIndex`/`onKeyDown` predicate cleanly. A plain `<div>` with `aria-label` but no `role` isn't announced by most screen readers, so the previous unconditional form was harmless — just now consistent.
2. **Documented (no code change)**: `ProjectionChartRow`'s empty-fallback branch (`!account_ids?.length`) now renders `<ChartRowTitle>` via the shell instead of a bare `<div className="title">`. An empty projection chart on Dashboard now looks like every other chart row (draggable + reorder chevron), which is a UX consistency win rather than a regression. `ChartDetailPage` passes `showTitle={false}` end-to-end so it's unaffected.

hoiekim (approved):

PR #627, opened 2026-07-09
refactor(client): extract `<DeleteButton>` — Properties-scoped destructive action (#326 phase 5e)

claoie:
## Summary

Extracts `<DeleteButton>` — a Properties-scoped destructive-action button — and migrates every Properties-based delete site to use it. Second of the 4-task set Hoie kicked off after #625.

## Design decision

Hoie asked for "two separate components." On audit only ONE component actually needs extraction:

- **`<ActionButtons>`** already encapsulates its own delete-button affordance (with a two-tap safety pattern: `⌫` → `Delete`). No other component duplicates that shape, so there's nothing to hoist.
- **`<button className="delete colored">` inside `<Row className="button">`** — the Properties-scoped destructive action — appears at 10 sites (9 external + 1 inside ActionButtons itself). This PR extracts it.

## Component

```tsx
<DeleteButton onClick={handler} confirmMessage="Are you sure?">
  Delete
</DeleteButton>
```

Props:
- `onClick: MouseEventHandler<HTMLButtonElement>` — runs after confirm passes.
- `confirmMessage?: string` — if set, `window.confirm(...)` gates the click. Was inlined at 2 sites before this PR.
- `children?: ReactNode` — button label (default `"Delete"`).
- `className?: string` — extra classes appended after the base `delete colored`.
- Automatically `type="button"` (prevents accidental form submission — was missing from 7 of the 9 call sites).

Styling is inherited from the existing `div.Properties .row button.delete` rule (`color: var(--darkRed)`) — no CSS changes needed.

## Sites migrated (9)

- `BalanceChartProperties` — Delete chart.
- `FlowChartProperties` — Delete chart.
- `ProjectionChartProperties` — Delete chart.
- `AccountProperties` — Delete account.
- `ConnectionProperties` — Delete connection.
- `Configuration` — Logout.
- `ApiKeyProperties` — Revoke.
- `InvestmentTransactionProperties` — Delete (confirm gate folded into the new `confirmMessage` prop).
- `TransactionProperties` — Delete (confirm gate folded).
- `HoldingProperties` — Remove snapshot.

## Test plan

- [x] `bun test src/client` → 326 / 0
- [ ] Sandbox — navigate to each migrated Properties surface and verify the delete button renders in the `--darkRed` text-link style; click flow still triggers the confirm dialog on the two sites that use `confirmMessage`.

### DIFF
--- a/src/client/components/DeleteButton/index.tsx
+++ b/src/client/components/DeleteButton/index.tsx
@@ -1,0 +1,39 @@
+import { MouseEventHandler, ReactNode } from "react";
+
+interface Props {
+  /**
+   * If set, `window.confirm(message)` gates the click — the wrapped
+   * `onClick` only fires when the user confirms. Every site in this
+   * codebase that renders a destructive button paired the button JSX
+   * with a bespoke `if (!window.confirm(...)) return;` line, so
+   * folding the gate into the component keeps the messaging
+   * consistent + removes ~8 duplicated call sites.
+   */
+  confirmMessage?: string;
+  onClick: MouseEventHandler<HTMLButtonElement>;
+  /** Button label. Defaults to `"Delete"`. */
+  children?: ReactNode;
+  /** Extra class string appended after the base `delete colored`. */
+  className?: string;
+}
+
+/**
+ * Destructive button styled with the codebase's `delete colored` class
+ * pair. Meant for use inside `<Properties>` / `<Row>` context — the
+ * `div.Properties .row button.delete` CSS rule paints the text
+ * `var(--darkRed)`. For the solid-fill variant sitting inside
+ * `<ActionButtons>`, see that component instead (self-contained widget,
+ * not extracted here).
+ */
+export const DeleteButton = ({ confirmMessage, onClick, children = "Delete", className }: Props) => {
+  const classes = ["delete", "colored", className].filter(Boolean).join(" ");
+  const handleClick: MouseEventHandler<HTMLButtonElement> = (e) => {
+    if (confirmMessage && !window.confirm(confirmMessage)) return;
+    onClick(e);
+  };
+  return (
+    <button type="button" className={classes} onClick={handleClick}>
+      {children}
+    </button>
+  );
+};
--- a/src/client/components/DeleteButton/index.tsx
+++ b/src/client/components/DeleteButton/index.tsx
@@ -1,14 +1,7 @@
 import { MouseEventHandler, ReactNode } from "react";
 
 interface Props {
-  /**
-   * If set, `window.confirm(message)` gates the click — the wrapped
-   * `onClick` only fires when the user confirms. Every site in this
-   * codebase that renders a destructive button paired the button JSX
-   * with a bespoke `if (!window.confirm(...)) return;` line, so
-   * folding the gate into the component keeps the messaging
-   * consistent + removes ~8 duplicated call sites.
-   */
+  /** If set, `window.confirm(confirmMessage)` gates the click — `onClick` fires only on confirm. */
   confirmMessage?: string;
   onClick: MouseEventHandler<HTMLButtonElement>;
   /** Button label. Defaults to `"Delete"`. */
--- a/src/client/components/DeleteButton/index.tsx
+++ b/src/client/components/DeleteButton/index.tsx
@@ -1,7 +1,10 @@
 import { MouseEventHandler, ReactNode } from "react";
 
+const DEFAULT_CONFIRM_MESSAGE = "Are you sure you want to delete this?";
+
 interface Props {
-  /** If set, `window.confirm(confirmMessage)` gates the click — `onClick` fires only on confirm. */
+  /** Prompt shown by `window.confirm` before `onClick` fires. Defaults to a
+   * generic message; every consumer should pass a specific one. */
   confirmMessage?: string;
   onClick: MouseEventHandler<HTMLButtonElement>;
   /** Button label. Defaults to `"Delete"`. */
@@ -10,18 +13,15 @@ interface Props {
   className?: string;
 }
 
-/**
- * Destructive button styled with the codebase's `delete colored` class
- * pair. Meant for use inside `<Properties>` / `<Row>` context — the
- * `div.Properties .row button.delete` CSS rule paints the text
- * `var(--darkRed)`. For the solid-fill variant sitting inside
- * `<ActionButtons>`, see that component instead (self-contained widget,
- * not extracted here).
- */
-export const DeleteButton = ({ confirmMessage, onClick, children = "Delete", className }: Props) => {
+export const DeleteButton = ({
+  confirmMessage = DEFAULT_CONFIRM_MESSAGE,
+  onClick,
+  children = "Delete",
+  className,
+}: Props) => {
   const classes = ["delete", "colored", className].filter(Boolean).join(" ");
   const handleClick: MouseEventHandler<HTMLButtonElement> = (e) => {
-    if (confirmMessage && !window.confirm(confirmMessage)) return;
+    if (!window.confirm(confirmMessage)) return;
     onClick(e);
   };
   return (
--- a/src/client/components/AccountProperties/index.tsx
+++ b/src/client/components/AccountProperties/index.tsx
@@ -288,7 +288,12 @@ export const AccountProperties = ({ account }: Props) => {
           <br />
           <Property>
             <Row className="button">
-              <DeleteButton onClick={onClickRemove}>Delete</DeleteButton>
+              <DeleteButton
+                confirmMessage="Do you want to delete this account?"
+                onClick={onClickRemove}
+              >
+                Delete
+              </DeleteButton>
             </Row>
           </Property>
         </>
--- a/src/client/components/AccountProperties/index.tsx
+++ b/src/client/components/AccountProperties/index.tsx
@@ -17,6 +17,7 @@ import {
 } from "client";
 import {
   DateLabel,
+  DeleteButton,
   DynamicCapacityInput,
   Graph,
   HoldingsComposition,
@@ -287,9 +288,7 @@ export const AccountProperties = ({ account }: Props) => {
           <br />
           <Property>
             <Row className="button">
-              <button className="delete colored" onClick={onClickRemove}>
-                Delete
-              </button>
+              <DeleteButton onClick={onClickRemove}>Delete</DeleteButton>
             </Row>
           </Property>
         </>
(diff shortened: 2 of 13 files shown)


reviewoie:
## Code review — PR 627

**Verdict:** approve

Verified the six task claims against `claoie/feat/delete-button` vs `upstream/main`:

1. **`confirmMessage` short-circuit** — `if (confirmMessage && !window.confirm(confirmMessage)) return;` in `handleClick` fires BEFORE `onClick(e)`. Behaviorally identical to the inline `if (!window.confirm(...)) return;` at the top of each pre-existing async onClick. When `confirmMessage` is undefined the guard is a no-op — `onClick` runs unconditionally. Correct.
2. **`type="button"` universal** — hardcoded in the JSX (`<button type="button" …>`), no prop override path. Old sites: 2 of 10 (`ApiKeyProperties`, `HoldingProperties`) had `type="button"`; the other 8 didn't. All now uniform. Safety hardening, no regression.
3. **Confirm-message text match** — exact:
   - `InvestmentTransactionProperties`: `"Delete this investment transaction? This can't be undone."` ✓
   - `TransactionProperties`: `"Delete this transaction? This can't be undone."` ✓
4. **CSS unchanged** — diff-stat shows no `.css` files touched. `div.Properties .row button.delete` (`Properties/index.css:48`, `color: var(--darkRed)`) still applies: every migration sits inside `<Properties><Row className="button"><DeleteButton>`, and `<Row>` renders `<div class="row …">`, `<DeleteButton>` renders `<button class="delete colored …">`. Selector still matches.
5. **`ActionButtons` untouched** — diff shows no `ActionButtons/**` change; `index.tsx:40` still has raw `className="delete colored"`. `grep "delete colored"` on the branch returns only this one line outside the new component. Matches the "self-contained widget" carve-out.
6. **Scope** — 12 files: 10 migrated Properties, 1 new component (`DeleteButton/index.tsx`), 1 barrel export (`components/index.ts`). All directly related to the PR title.

Additional spot-checks:
- `.filter(Boolean).join(" ")` on `["delete", "colored", className]` correctly yields `"delete colored"` (no trailing space) when `className` is `undefined | null | ""`, and `"delete colored X"` otherwise.
- Import additions in every migrated file are single-symbol adds; no unused imports left behind (verified `InvestmentTransactionProperties` and `TransactionProperties` — pre-existing `KeyValue`/`InstitutionSpan`/`TransferArrowIcon` still used).
- The wrapper's `MouseEventHandler` typing accepts async onClicks the same way React does natively — `void` return covers `Promise<void>` implicitly; no fire-and-forget regression vs the pre-existing raw `<button onClick={async () => {...}}>`.

Minor bookkeeping note (not a finding): the PR body says "9 sites" but the actual site count is 10 (the three chart-Properties files are three sites, not one). No effect on the diff.

No further blockers.

reviewoie:
## Code review — PR 627 (re-review, head 614a3e8a)

**Verdict:** approve.

**Findings:** 0 total.

Prior LOW (overclaiming / PR-narrative `confirmMessage` JSDoc) is resolved:
- The 7-line prop doc ("every site… paired with window.confirm", "removes ~8 duplicated call sites") is trimmed to a one-line contract that matches the implementation (`if (confirmMessage && !window.confirm(confirmMessage)) return;`).
- Trim is comment-only — no non-comment change in the commit.
- The class-level JSDoc's CSS-coupling rationale (`div.Properties .row button.delete` painting `var(--darkRed)`, and the ActionButtons variant pointer) is retained — a legitimate future-reader constraint, correctly kept.

No further blockers.

claoie:
Pushed 614a3e8a — trimmed the `DeleteButton` `confirmMessage` JSDoc to a one-line contract.

The re-review on #628 (shared, byte-identical file) flagged the prop doc as overclaiming ("every site in this codebase... paired with `window.confirm`" — logout / account-delete never did) and carrying PR-narrative ("removes ~8 duplicated call sites"). Per the self-documenting-code rule the doc now just states the contract; the class-level JSDoc's CSS-coupling rationale (`div.Properties .row button.delete` → `var(--darkRed)`) is retained.

- reviewoie re-review at 614a3e8a: approve.
- CI green: build + test.

E2E: `DeleteButton` was driven end-to-end on the #628 sandbox (identical component code) — red `delete colored` affordance at `var(--darkRed)`, confirm gate fires, action runs. See the #628 E2E note for the full click chain.

src/client/components/DeleteButton/index.tsx
```diff
@@ -0,0 +1,32 @@
...
+ * pair. Meant for use inside `<Properties>` / `<Row>` context — the
+ * `div.Properties .row button.delete` CSS rule paints the text
+ * `var(--darkRed)`. For the solid-fill variant sitting inside
+ * `<ActionButtons>`, see that component instead (self-contained widget,
+ * not extracted here).
+ */
+export const DeleteButton = ({ confirmMessage, onClick, children = "Delete", className }: Props) => {
+  const classes = ["delete", "colored", className].filter(Boolean).join(" ");
+  const handleClick: MouseEventHandler<HTMLButtonElement> = (e) => {
+    if (confirmMessage && !window.confirm(confirmMessage)) return;
```
hoiekim: I want the confirm prompt to be enforced. Use default confirm message.

Also, add specific confirm message to all leaf components

reviewoie:
## Code review — PR 627 (fix commit 51757e7b)

**Verdict:** approve.

### DeleteButton call-site sweep — all 10 sites confirmed

`grep '<DeleteButton' src/client` returned 10 sites — every one passes an explicit `confirmMessage`:

- `BalanceChartProperties.tsx:136` — "Do you want to delete this chart?"
- `FlowChartProperties.tsx:130` — "Do you want to delete this chart?"
- `ProjectionChartProperties.tsx:306` — "Do you want to delete this chart?"
- `AccountProperties/index.tsx:291` — "Do you want to delete this account?"
- `ConnectionProperties/index.tsx:159` — "Do you want to remove all data in this connection?"
- `Configuration/index.tsx:143` — "Do you want to log out?"
- `ApiKeyProperties/index.tsx:258` — dynamic (see finding 1)
- `HoldingProperties/index.tsx:819` — "Remove this holding snapshot?"
- `TransactionProperties/index.tsx:566` — "Delete this transaction? This can't be undone."
- `InvestmentTransactionProperties/index.tsx:451` — "Delete this investment transaction? This can't be undone."

The default `"Are you sure you want to delete this?"` is a pure safety-net for any future caller — no current site relies on it.

### Confirm enforcement — no bypass surface

- `DeleteButton.handleClick` unconditionally runs `window.confirm(confirmMessage)` before dispatching `onClick`; the `confirmMessage && …` short-circuit is gone.
- `grep window.confirm src/client/components` returns ONLY `DeleteButton/index.tsx`. No double-fire anywhere.
- `grep window.confirm src/client` also finds one hit in `pages/BudgetConfigPage/lib/hooks.ts:270` — that's a separate delete affordance (budget/section/category cascade delete), NOT a `<DeleteButton>` site, and it's out of scope for this PR. No conflict.
- Cross-site tracer for the wrapped handlers:
  - `onClickRemove` in `AccountProperties/lib/hooks.ts:307` is exposed via `useAccountEventHandlers`, whose only import is `AccountProperties/index.tsx:34` — one caller, and that caller only feeds it to the DeleteButton at line 293.
  - `onClickRemove` in `ConnectionProperties/index.tsx:48` has one call site (line 161, DeleteButton).
  - `onClickRemove` in each of the three chart properties files has one call site (DeleteButton).
  - `onClickDeleteSnap` in `HoldingProperties/index.tsx:583` has one call site (line 821, DeleteButton).
  - `onRevoke` in `ApiKeyProperties/index.tsx:98` has one call site (line 264, DeleteButton). The old early-return `if (!window.confirm(...)) return` is gone from the handler body — confirm now lives only in DeleteButton.
  - `logout` in `Configuration/index.tsx:59` has one call site (line 143, DeleteButton).
- No handler is dispatched from a route/keyboard shortcut/programmatic path that would bypass DeleteButton.

### Control-flow refactor verification

**`AccountProperties/lib/hooks.ts:307-351`** — unwrapped `if (confirmed) { … }` block reads correctly:
- `e.stopPropagation()` still runs first.
- `call.delete(...)` → `.then(r => if r.status === "success" { setData(...); router.back(); } else { console.error(...); })` → `.catch(err => console.error(...))`. Success, non-success, and rejection all preserved. No dropped branches.

**`ConnectionProperties/index.tsx:48-94`** — same shape refactor, identically correct. Accounts-cascade-in-item loop and transactions loop still inside the same `setData` callback; success path calls `router.back()`; error/reject paths log.

### CI + typecheck
- `gh pr checks 627` → build pass, test pass (green).

---

### Low
- `ApiKeyProperties/index.tsx:258-267` — the `confirmMessage={apiKey ? ... : "Revoke this API key? ..."}` ternary has an unreachable fallback branch. The component early-returns at line 208 when `error || !apiKey` is truthy, so by the time the JSX at line 258 renders, `apiKey` is guaranteed non-null. Simplify to `confirmMessage={`Revoke API key "${apiKey.name}"? This cannot be undone.`}` — the null-guard fallback is dead code.
- `AccountProperties/lib/hooks.ts:308` + `ConnectionProperties/index.tsx:49` — `e.stopPropagation()` now fires AFTER DeleteButton's confirm rather than before it. In the current tree neither `<AccountDetailPage>` nor `<ConnectionDetailPage>` mounts a click-listening ancestor around these buttons, so the semantic drift is inconsequential today. Noting so a future ancestor click-handler doesn't inherit a "propagates on cancel" surprise. (Not blocking. Could inline `e.stopPropagation()` into DeleteButton pre-confirm if you want the old ordering back, but current behavior is fine.)

---

### Notes for future callers

- The `DEFAULT_CONFIRM_MESSAGE = "Are you sure you want to delete this?"` reads awkwardly for non-delete uses of DeleteButton (e.g. logout). Configuration's Logout row supplies its own specific message, so this isn't a real hit today — but the naming coupling ("delete") is worth watching if the component gets used for more session-end / archive / dismiss affordances.

claoie:
Reviewoie LOWs addressed:

**LOW 1** — dead-branch removal: the `apiKey ? … : "fallback"` ternary in the DeleteButton `confirmMessage` had an unreachable else (component early-returns at line 208 when `!apiKey`, so the DeleteButton only mounts with a non-null apiKey). Simplified to the truthy branch.

**LOW 2** — `e.stopPropagation()` on Account/Connection `onClickRemove` moved from pre-confirm to post-confirm as a side-effect of the DeleteButton confirm wrap. Reviewoie flagged this as inconsequential today (no click-listening ancestor on either detail page), and I'm leaving it defensive — the propagation-stop still runs when the delete actually fires, which is what the pre-refactor code was guarding against. If a future wrap adds an ancestor click listener, the stop is still in the confirm-accept path where it matters.

hoiekim (approved):

PR #628, opened 2026-07-09
refactor(client): introduce useMutate(Model) — generic CRUD hook (#326 phase 5f)

claoie:
## Summary

Replaces the initial `useChartDelete` sketch with a generic hook per Hoie's design: `useMutate(Model)` reads static metadata off the model class and returns `{ create, update, delete }`. Each site becomes one line at the call site instead of a 15-line copy of "call, status-check, setData, new Dict, dict.set/delete, indexedDb.save/remove, cast field".

## Design — model class carries its own routing

```ts
class Chart {
  static readonly apiPath = "/api/chart";
  static readonly dataField = "charts" as const;
  // …
}

const chartMutate = useMutate(Chart);
chartMutate.delete(chart_id);
chartMutate.update(new Chart({ ...chart, name: newName }));
```

- `apiPath: string` — POST/DELETE target under `/api`.
- `dataField: keyof Data` — the `Data` field holding this model's dictionary. Values match `StoreName.*` verbatim so IDB store is derivable from the same static.

Dictionary constructor is read off the live state (`newData[dataField].constructor`), so callers don't need to import `ChartDictionary` / `TransactionDictionary` / etc. Every model class exposes `.id`, so `instance.id` is the universal key.

## Sites migrated (10 in this PR)

- 3× chart delete + 3× chart update: Balance / Flow / Projection Chart Properties.
- 2× chart update: ChartAccountsPage account/budget toggles.
- Transaction delete: TransactionProperties.
- Investment-transaction delete: InvestmentTransactionProperties.
- Holding-snapshot delete: HoldingProperties.

## Server route normalization

Two DELETE routes had non-uniform query keys — normalized to `?id=` so `useMutate` builds one URL shape:

- `/transaction` DELETE: `transaction_id` → `id`.
- `/investment-transaction` DELETE: `investment_transaction_id` → `id`.

Both routes had no dedicated tests; the FE was the only caller and is migrated in the same commit.

## Not covered — follow-up scope

- **Cascading deletes** (Account → transactions + accountSnapshots; Connection/Item → items + accounts + transactions). These update more than one dictionary; the generic hook shape doesn't cleanly express them without another design decision (multi-dict cascade spec on the Model class, or a bespoke hook composed on top of `useMutate`). Left as bespoke handlers for now.
- **ApiKey** — no client model class exists; delete goes through a plain object. Would need a client `ApiKey` class before adopting.
- Other mutation shapes (login/session, transfer pairs, budgets/categories/sections/items) — same abstraction applies, just wasn't in this PR's cut. Each is a mechanical migration when needed.

## Test plan

- [x] Typecheck: clean.
- [x] Lint: clean.
- [x] Unit tests: 1082 pass / 0 fail.
- [ ] E2E on sandbox: drive chart create + update + delete; transaction delete on a manual row; investment-transaction delete on a manual row; holding-snapshot delete.

### DIFF
--- a/src/client/lib/hooks/useMutate.ts
+++ b/src/client/lib/hooks/useMutate.ts
@@ -1,0 +1,108 @@
+import { useCallback, useMemo } from "react";
+import { call, useAppContext, indexedDb, StoreName, Data } from "client";
+import type { Dictionary } from "client/lib/models/Data";
+
+/**
+ * Contract every mutable client model exposes so `useMutate` can drive
+ * its CRUD from just the class reference:
+ *
+ * - `apiPath` — POST/DELETE target under `/api`.
+ * - `dataField` — the `Data` field that holds this model's dictionary.
+ *   Values match `StoreName.*` verbatim (`"charts"`, `"transactions"`,
+ *   …), so we don't need a separate StoreName static.
+ */
+export interface MutableModel<T extends { readonly id: string }> {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  new (...args: any[]): T;
+  readonly apiPath: string;
+  readonly dataField: keyof Data;
+}
+
+export interface MutateApi<T> {
+  /** Upserts on the server, adds the instance to state + IDB. */
+  create: (instance: T) => Promise<void>;
+  /** Upserts on the server, replaces the instance in state + IDB. */
+  update: (instance: T) => Promise<void>;
+  /** DELETEs by id, removes from state + IDB. */
+  delete: (id: string) => Promise<void>;
+}
+
+/**
+ * Generic client-side mutation hook. `useMutate(Chart)` returns the
+ * `{ create, update, delete }` API for that model — the hook centralizes
+ * the await-call, status check, error log/throw, and the state + IDB
+ * update pattern that would otherwise repeat verbatim at every
+ * mutation site.
+ *
+ * `create` and `update` share the server contract (server routes are
+ * upsert-by-id via POST), so both call the same code path here; the
+ * split exists at the interface level for reader clarity at call
+ * sites. Deletes go through DELETE `?id=…`.
+ *
+ * The state update writes back through the SAME dictionary constructor
+ * that Data currently holds, so callers don't need to import the
+ * per-model Dictionary class — the hook reads it off the live
+ * dictionary's prototype.
+ *
+ * Not covered here: cascading deletes (Account, Connection) that touch
+ * more than one dictionary — those keep their bespoke handlers until
+ * we design a cascade extension.
+ */
+export const useMutate = <T extends { readonly id: string }>(
+  Model: MutableModel<T>,
+): MutateApi<T> => {
+  const { setData } = useAppContext();
+
+  const upsert = useCallback(
+    async (instance: T) => {
+      const r = await call.post(Model.apiPath, instance);
+      if (r.status !== "success") {
+        console.error(r.message);
+        throw new Error(r.message);
+      }
+      setData((oldData) => {
+        const newData = new Data(oldData);
+        const current = newData[Model.dataField] as unknown as Dictionary<T>;
+        const DictCtor = current.constructor as new (
+          source?: Iterable<readonly [string, T]>,
+        ) => Dictionary<T>;
+        const next = new DictCtor(current);
+        next.set(instance.id, instance);
+        indexedDb
+          .save(instance as unknown as Parameters<typeof indexedDb.save>[0])
+          .catch(console.error);
+        (newData[Model.dataField] as unknown as Dictionary<T>) = next;
+        return newData;
+      });
+    },
+    [Model, setData],
+  );
+
+  const del = useCallback(
+    async (id: string) => {
+      const r = await call.delete(`${Model.apiPath}?id=${id}`);
+      if (r.status !== "success") {
+        console.error(r.message);
+        throw new Error(r.message);
+      }
+      setData((oldData) => {
+        const newData = new Data(oldData);
+        const current = newData[Model.dataField] as unknown as Dictionary<T>;
+        const DictCtor = current.constructor as new (
+          source?: Iterable<readonly [string, T]>,
+        ) => Dictionary<T>;
+        const next = new DictCtor(current);
+        next.delete(id);
+        indexedDb.remove(Model.dataField as StoreName, id).catch(console.error);
+        (newData[Model.dataField] as unknown as Dictionary<T>) = next;
+        return newData;
+      });
+    },
+    [Model, setData],
+  );
+
+  return useMemo(
+    () => ({ create: upsert, update: upsert, delete: del }),
+    [upsert, del],
+  );
+};
--- a/src/client/lib/hooks/useMutate.ts
+++ b/src/client/lib/hooks/useMutate.ts
@@ -1,21 +1,17 @@
 import { useCallback, useMemo } from "react";
-import { call, useAppContext, indexedDb, StoreName, Data } from "client";
-import type { Dictionary } from "client/lib/models/Data";
+import { call, useAppContext, indexedDb, Data } from "client";
+import type { StoredModel } from "client/lib/indexed-db/service";
 
 /**
  * Contract every mutable client model exposes so `useMutate` can drive
- * its CRUD from just the class reference:
- *
- * - `apiPath` — POST/DELETE target under `/api`.
- * - `dataField` — the `Data` field that holds this model's dictionary.
- *   Values match `StoreName.*` verbatim (`"charts"`, `"transactions"`,
- *   …), so we don't need a separate StoreName static.
+ * its CRUD from just the class reference. Model class carries only its
+ * API root; the model → `Data` slot / `StoreName` / `Dictionary` mapping
+ * lives on `Data` itself (see `Data.dictOf` / `storeNameOf` / `set`).
  */
-export interface MutableModel<T extends { readonly id: string }> {
+export interface MutableModel<T extends StoredModel> {
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   new (...args: any[]): T;
   readonly apiPath: string;
-  readonly dataField: keyof Data;
 }
 
 export interface MutateApi<T> {
@@ -39,18 +35,15 @@ export interface MutateApi<T> {
  * split exists at the interface level for reader clarity at call
  * sites. Deletes go through DELETE `?id=…`.
  *
- * The state update writes back through the SAME dictionary constructor
- * that Data currently holds, so callers don't need to import the
- * per-model Dictionary class — the hook reads it off the live
- * dictionary's prototype.
+ * State updates go through `Data.dictOf(Model).clone()` + `Data.set` —
+ * the model → slot mapping lives on `Data`, so the hook body has no
+ * casts and no per-model wiring beyond `Model.apiPath`.
  *
  * Not covered here: cascading deletes (Account, Connection) that touch
  * more than one dictionary — those keep their bespoke handlers until
  * we design a cascade extension.
  */
-export const useMutate = <T extends { readonly id: string }>(
-  Model: MutableModel<T>,
-): MutateApi<T> => {
+export const useMutate = <T extends StoredModel>(Model: MutableModel<T>): MutateApi<T> => {
   const { setData } = useAppContext();
 
   const upsert = useCallback(
@@ -62,16 +55,10 @@ export const useMutate = <T extends { readonly id: string }>(
       }
       setData((oldData) => {
         const newData = new Data(oldData);
-        const current = newData[Model.dataField] as unknown as Dictionary<T>;
-        const DictCtor = current.constructor as new (
-          source?: Iterable<readonly [string, T]>,
-        ) => Dictionary<T>;
-        const next = new DictCtor(current);
-        next.set(instance.id, instance);
-        indexedDb
-          .save(instance as unknown as Parameters<typeof indexedDb.save>[0])
-          .catch(console.error);
-        (newData[Model.dataField] as unknown as Dictionary<T>) = next;
+        const newDict = newData.dictOf(Model).clone();
+        newDict.set(instance.id, instance);
+        newData.set(newDict);
+        indexedDb.save(instance).catch(console.error);
         return newData;
       });
     },
@@ -87,22 +74,15 @@ export const useMutate = <T extends { readonly id: string }>(
       }
       setData((oldData) => {
         const newData = new Data(oldData);
-        const current = newData[Model.dataField] as unknown as Dictionary<T>;
-        const DictCtor = current.constructor as new (
-          source?: Iterable<readonly [string, T]>,
-        ) => Dictionary<T>;
-        const next = new DictCtor(current);
-        next.delete(id);
-        indexedDb.remove(Model.dataField as StoreName, id).catch(console.error);
-        (newData[Model.dataField] as unknown as Dictionary<T>) = next;
+        const newDict = newData.dictOf(Model).clone();
+        newDict.delete(id);
+        newData.set(newDict);
+        indexedDb.remove(newData.storeNameOf(Model), id).catch(console.error);
         return newData;
       });
     },
     [Model, setData],
   );
 
-  return useMemo(
-    () => ({ create: upsert, update: upsert, delete: del }),
-    [upsert, del],
-  );
+  return useMemo(() => ({ create: upsert, update: upsert, delete: del }), [upsert, del]);
 };
--- a/src/client/components/BalanceChartProperties.tsx
+++ b/src/client/components/BalanceChartProperties.tsx
@@ -3,15 +3,11 @@ import { ChartType } from "common";
 import {
   BalanceChart,
   Chart,
-  ChartDictionary,
-  Data,
-  call,
   PATH,
   useAppContext,
   useDebounce,
+  useMutate,
   getChartTypeName,
-  indexedDb,
-  StoreName,
   DeleteButton,
   Properties,
   PropertyLabel,
@@ -30,30 +26,17 @@ export const BalanceChartProperties = ({ chart, children }: BalanceChartProperti
   const { name, chart_id, type, configuration } = chart;
   const { account_ids, budget_ids } = configuration;
 
-  const { data, setData } = useAppContext();
+  const { data } = useAppContext();
   const { accounts, budgets } = data;
 
   const [nameInput, setNameInput] = useState(name);
   const [selectedType, setSelectedType] = useState<ChartType>(type);
 
   const updateDebouncer = useDebounce();
+  const mutate = useMutate(Chart);
 
-  const updateChart = async (updatedChart: Partial<Chart>) => {
-    const r = await call.post("/api/chart", { chart_id, ...updatedChart });
-    if (r.status === "success") {
-      setData((oldData) => {
-        const newData = new Data(oldData);
-        const newChart = new Chart({ ...chart, ...updatedChart });
-        indexedDb.save(newChart).catch(console.error);
-        const newCharts = new ChartDictionary(newData.charts);
-        newCharts.set(chart_id, newChart);
-        newData.charts = newCharts;
-        return newData;
-      });
-    } else {
-      console.error(r.message);
-      throw new Error(r.message);
-    }
+  const updateChart = (updatedChart: Partial<Chart>) => {
+    return mutate.update(new Chart({ ...chart, ...updatedChart }));
   };
 
   const onChangeName: ChangeEventHandler<HTMLInputElement> = (e) => {
@@ -81,21 +64,8 @@ export const BalanceChartProperties = ({ chart, children }: BalanceChartProperti
   }).length;
 
   const onClickRemove: MouseEventHandler<HTMLButtonElement> = async () => {
-    const r = await call.delete(`/api/chart?id=${chart_id}`);
-    if (r.status === "success") {
-      setData((oldData) => {
-        const newData = new Data(oldData);
-        const newCharts = new ChartDictionary(newData.charts);
-        indexedDb.remove(StoreName.charts, chart_id).catch(console.error);
-        newCharts.delete(chart_id);
-        newData.charts = newCharts;
-        return newData;
-      });
-      router.back();
-    } else {
-      console.error(r.message);
-      throw new Error(r.message);
-    }
+    await mutate.delete(chart_id);
+    router.back();
   };
 
   return (
--- a/src/client/components/FlowChartProperties.tsx
+++ b/src/client/components/FlowChartProperties.tsx
@@ -3,15 +3,11 @@ import { ChartType } from "common";
 import {
   FlowChart,
   Chart,
-  ChartDictionary,
-  Data,
-  call,
   PATH,
   useAppContext,
   useDebounce,
+  useMutate,
   getChartTypeName,
-  indexedDb,
-  StoreName,
   DeleteButton,
   Properties,
   PropertyLabel,
@@ -30,30 +26,17 @@ export const FlowChartProperties = ({ chart, children }: FlowChartPropertiesProp
   const { name, chart_id, type, configuration } = chart;
   const { account_ids } = configuration;
 
-  const { data, setData } = useAppContext();
+  const { data } = useAppContext();
   const { accounts } = data;
 
   const [nameInput, setNameInput] = useState(name);
   const [selectedType, setSelectedType] = useState<ChartType>(type);
 
   const updateDebouncer = useDebounce();
+  const mutate = useMutate(Chart);
 
-  const updateChart = async (updatedChart: Partial<Chart>) => {
-    const r = await call.post("/api/chart", { chart_id, ...updatedChart });
-    if (r.status === "success") {
-      setData((oldData) => {
-        const newData = new Data(oldData);
-        const newChart = new Chart({ ...chart, ...updatedChart });
-        indexedDb.save(newChart).catch(console.error);
-        const newCharts = new ChartDictionary(newData.charts);
-        newCharts.set(chart_id, newChart);
-        newData.charts = newCharts;
-        return newData;
-      });
-    } else {
-      console.error(r.message);
-      throw new Error(r.message);
-    }
+  const updateChart = (updatedChart: Partial<Chart>) => {
+    return mutate.update(new Chart({ ...chart, ...updatedChart }));
   };
 
   const onChangeName: ChangeEventHandler<HTMLInputElement> = (e) => {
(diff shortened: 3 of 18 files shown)


reviewoie:
## Code review — PR 628

**Verdict:** approve (informational, one low-severity note)

Behavior parity, deps, and cleanup all check out:

- **Parity** — hook's control flow matches the inlined `onClickRemove` byte-identically (the flip from `if (r.status === "success") { ...proceed } else { ...throw }` to `if (r.status !== "success") { ...throw } … …proceed` is equivalent). API call, `setData` updater rebuilding `ChartDictionary`, `indexedDb.remove(StoreName.charts, chart_id).catch(console.error)` inside the updater, and `router.back()` all preserved. The `window.confirm` gate migrates cleanly to `<DeleteButton confirmMessage="...">` per the compositional intent.
- **useCallback deps** — `[chart_id, setData, router]` correctly closes over every captured value. `setData` is a `Dispatch` (stable). `router` is a fresh object per render from `useRouter` (returns a plain literal, not memoized) so the callback re-creates each render — that's a pre-existing pattern in the app, not this PR's problem, and the deps list is exhaustive-deps-correct.
- **Consumer sites** — all three (`BalanceChartProperties`, `FlowChartProperties`, `ProjectionChartProperties`) collapse identically: `const onClickRemove = useChartDelete(chart_id);` + `<DeleteButton confirmMessage="Do you want to delete this chart?" onClick={onClickRemove}>Delete</DeleteButton>`.
- **Imports cleaned** — `StoreName` removed from all three consumers' import lists. `ChartDictionary` / `Data` / `call` / `indexedDb` retained because `updateChart` still uses them. `MouseEventHandler` retained because `onClickAccounts` still uses it. No residue.
- **Throw-on-error preserved** — `throw new Error(r.message)` at `useChartDelete.ts:29`. No consumer currently `.catch`es, matching pre-refactor behavior.
- **Scope** — PR 628 commit is 5 files, all directly tied to the hook migration. The 12-file bundle in the PR view is the stacked #627 commit as advertised.

### Low
- `src/client/lib/hooks/useChartDelete.ts:12-15, 19-21` — JSDoc contains PR-narrative describing the previous implementation ("Byte-identical `onClickRemove` used to live inline in `BalanceChartProperties` / `FlowChartProperties` / `ProjectionChartProperties`…" and "the original inline code called `throw new Error(r.message)` too"). Per [[feedback_comment_hygiene]] + [[feedback_self_documenting_code]], comments should not describe the previous state or PR history — that belongs in the PR body. The load-bearing content of this block is one line: the compositional hint that the confirm gate lives on `<DeleteButton>`. Suggested trim:
  ```
  /**
   * Delete a chart by id — issues the DELETE, evicts from IndexedDB,
   * removes it from `Data.charts`, and navigates back on success.
   * Throws on API failure so a caller can `.catch` for error UI.
   * Pair with `<DeleteButton confirmMessage="..." onClick={onDelete}>`;
   * the confirm gate lives on the button component.
   */
  ```

claoie:
Reviewoie LOW addressed: trimmed the JSDoc down to the load-bearing compositional hint (pair with DeleteButton, throws on API failure). The prior wording narrated the previous inline implementation which violates the comment-hygiene rule — that story belongs in the PR body, not the source. Verified typecheck clean.

reviewoie:
## Code review — PR 628

**Verdict:** informational (approve)

Re-review at HEAD `7639748e`. No correctness bugs. Confirm-dialog behavior verified preserved at every migrated site (details below). `useChartDelete` follows the correct stateless-action-hook-over-`setData` pattern (reads `setData`/`router` off `useAppContext`, no fetched data bolted onto context) — architecture is clean.

**Correctness — verified, no regression:**
- The 5 sites that moved their confirm into `DeleteButton`'s `confirmMessage` (Balance/Flow/Projection chart props, Transaction, InvestmentTransaction) each had the inline `if (!window.confirm(...)) return;` REMOVED from the handler in the same hunk — no double-confirm, no dropped confirm.
- The un-migrated confirms (ApiKey `onRevoke`, Connection `onClickRemove`, Holding `onClickDeleteSnap`) keep their own inline `window.confirm` in untouched handlers — no confirm lost.
- `type="button"` is now always emitted; several sites had no explicit type before, but none sit in a `<form>`, so click behavior is unchanged (if anything, safer).
- `useChartDelete` inverts the success guard vs the old inline flow but is behavior-equivalent (delete → on failure log+throw → else setData+router.back).
- Sweep is complete: only remaining `delete colored` button is `ActionButtons` (solid-fill variant), documented as an intentional carve-out.

### Low
- `DeleteButton/index.tsx:5-12` — `confirmMessage` JSDoc overclaims and reads as PR-narrative. It asserts "Every site in this codebase that renders a destructive button paired the button JSX with a bespoke `if (!window.confirm(...)) return;`" — not true: `Configuration` logout and the account delete have no confirm. And "removes ~8 duplicated call sites" is a refactor-narrative, not a contract — only 5 confirm lines were actually folded in. Trim to the one-line contract ("If set, gates `onClick` behind `window.confirm(message)`").
- `ApiKeyProperties`, `ConnectionProperties`, `HoldingProperties` — these render `DeleteButton` but still gate via their own inline `window.confirm` instead of `confirmMessage`, while 5 other sites route the confirm through the prop. Half-migrated indirection: either migrate all confirm-bearing sites to `confirmMessage` (ApiKey's dynamic `"Revoke API key \"${name}\""` works fine as a prop string) or leave the gate out of the component. As-is, the "keeps the messaging consistent" claim in the JSDoc is undercut by the split.
- Scope/title — PR is titled "extract `useChartDelete` hook" but also introduces `DeleteButton` and applies it across 11 files, 8 of which have nothing to do with chart deletion. All within the #326 component-refactor arc and cohesive as a destructive-button consolidation, so not a scope violation — but the title undersells the diff. Consider naming `DeleteButton` in the title.

No further blockers.

reviewoie:
## Code review — PR 628 (re-review, head 79ae289c)

**Verdict:** approve.

**Findings:** 0 total.

Prior LOW (DeleteButton `confirmMessage` JSDoc) is resolved identically to PR 627:
- Same one-line contract trim; the DeleteButton file is byte-identical to the PR 627 head, so the two branches don't diverge on the shared component.
- Trim is comment-only — no non-comment change.
- Class-level CSS-coupling JSDoc retained.
- `useChartDelete.ts` carries no lingering PR-narrative markers (grep: none) — the earlier LOW there is also clear.

No further blockers.

claoie:
Pushed 79ae289c — same one-line JSDoc trim on the shared `DeleteButton` (kept byte-identical to the #627 copy so the two branches don't diverge). `useChartDelete.ts` had no lingering PR-narrative.

- reviewoie re-review at 79ae289c: approve.
- CI green: build + test.

**E2E (isolated sandbox, scrambled DB, mobile 390×844):**
- Logged in, opened a Balance chart's Properties via the row chevron (real click chain from the Dashboard — no deep-link).
- `DeleteButton` rendered `delete colored` at `var(--darkRed)` (computed `rgb(204, 34, 0)`) with the "Delete" label; screenshot eye-check confirmed the red Delete affordance at the bottom of the Chart Profile panel.
- Clicked it → confirm gate fired ("Do you want to delete this chart?") → accepted → the chart was removed from the Dashboard (6 → 5 chart rows, the deleted chart's title gone), stayed on `/dashboard`, 0 JS errors.
- This exercises both the extracted `useChartDelete` hook and the `DeleteButton` component end-to-end.

On the partial-migration note (ApiKey / Connection / Holding keep an inline `window.confirm` while the other sites use `confirmMessage`): leaving as-is here to keep scope on the `useChartDelete` extraction. Consolidating the remaining confirms onto `confirmMessage` (or dropping the prop) is a `DeleteButton`-level decision independent of the chart hook — happy to do it in a focused follow-up if you'd like the gate fully unified.

src/client/lib/hooks/useChartDelete.ts
hoiekim, line 1: Can we extend this hook for all APIs as possible? Including
 
* All models (Chart, Account, Transaction, Budget, ...)
* All mutation (create / update / delete)

Abstraction goal:

* Awaits API call
* Handle failure
* On success, update state data and indexed DB

If the API is not consistent enough to support this refactoring, refactor that server side request / response shape to be generic.

src/client/lib/hooks/useMutate.ts
```diff
@@ -0,0 +1,108 @@
...
+        const DictCtor = current.constructor as new (
+          source?: Iterable<readonly [string, T]>,
+        ) => Dictionary<T>;
+        const next = new DictCtor(current);
+        next.set(instance.id, instance);
+        indexedDb
+          .save(instance as unknown as Parameters<typeof indexedDb.save>[0])
+          .catch(console.error);
+        (newData[Model.dataField] as unknown as Dictionary<T>) = next;
+        return newData;
```
hoiekim, line 62: Is there a way to avoid using `unknown` casting?

IDB library has `StoreName` enum. Maybe we can use this?

claoie:
## Code review — PR 628

**Verdict:** informational — no blockers found. The generic-hook design is sound; the `save`/`remove` polymorphic paths, dictionary-constructor lookup, and IDB store-name alignment all check out for the four models given statics (Chart / Transaction / InvestmentTransaction / HoldingSnapshot). CI green, typecheck+lint clean, client+server test suites pass.

### Med

- `src/client/components/BalanceChartProperties.tsx:39` (also `FlowChartProperties.tsx:39`, `ProjectionChartProperties.tsx:54`, `ChartAccountsPage.tsx:53` + `:81`) — **full-instance POST widens the write surface vs the old field-scoped POST → concurrent-edit clobber risk.** Old code sent `POST /api/chart` with `{chart_id, ...updatedChart}` (one field at a time — just `name`, or just `configuration`). New code posts a full `new Chart({...chart, ...updatedChart})` = `{chart_id, name, type, configuration}`. `updateChart` in `charts.ts:49` picks name/type/configuration from the body and writes all three columns whenever they're present. Failure scenario: user has two tabs / two devices open on the same chart. Tab A renames chart to "Rent Analysis". Tab B (still holding the pre-rename state in `chart`) toggles an account on `ChartAccountsPage`. Tab B's `chartMutate.update(new Chart({...chart, configuration: newConfig}))` posts `{chart_id, name: "OldName", type, configuration: newConfig}` — server overwrites Tab A's rename back to "OldName". Small blast radius (concurrent editing rare on a single-user app; no PR in this codebase has last-write-wins guards), but it's a real regression from the field-scoped partials. Options: (a) accept as design trade-off (probably fine, name it in the PR body); (b) keep the hook and change the call sites to pass a `Partial<Chart>` payload the hook translates to `{id, ...patch}`; (c) leave chart updates on the old field-scoped `call.post` and use `useMutate` for delete-only for now.

### Low

- `src/client/lib/hooks/useMutate.ts:56-79` — **`useMutate(Chart).create` is a latent footgun.** The doc block says "server routes are upsert-by-id via POST" (line 37-40), but `POST /api/chart` (`post-chart.ts`) is UPDATE-only — chart creation flows through `GET /api/new-chart` (`get-new-chart.ts`, called from `DashboardPage/index.tsx:108`). A future call to `useMutate(Chart).create(new Chart({...}))` posts to /api/chart with a freshly-generated `chart_id`. `updateChart` calls `chartsTable.update(chart_id, ...)` which returns `null`/`false` for a nonexistent id, but the route wrapper still returns `{status: "success"}` unconditionally — so the hook thinks it succeeded, appends the ghost chart to state + IDB, and it never lands in Postgres. Not triggered by any caller in this PR (Chart uses `.update`/`.delete` only), but the API surface promises something it can't deliver. Options: return the create/delete pair without `create` for Chart, or make `POST /api/chart` actually upsert.

- `src/client/lib/hooks/useMutate.ts:56-79` — **`useMutate(HoldingSnapshot).create/update` would POST a malformed body.** `POST /api/snapshots/holding` (`post-holding-snapshot.ts:73`) expects a FLAT body: `{account_id, ticker_symbol, quantity, snapshot_date, ...}` for create, `{snapshot_id, quantity, cost_basis, ...}` for update. A `new HoldingSnapshot({...})` JSON-serializes to `{snapshot: {snapshot_id, date}, user: {user_id}, holding: {account_id, security_id, quantity, ...}}` — nested. The route reads `body.account_id` → undefined → `validationError("account_id is required")`. Only `.delete` (query-string based) is safe. Not triggered — `HoldingProperties.onSubmitNew` (`index.tsx:442`) and `updateSnapshot` (`index.tsx:481`) correctly use raw `call.post` with the flat body. Options: narrow the interface for HoldingSnapshot to `delete`-only, or add a `toApiBody()` static that the hook calls before POST (would also unblock the future goal of `create`/`update` on this model without churning callers).

**Line-by-line audit passes:**
- `indexedDb.save()` switch (`service.ts:421-471`) — all four new statics-carrying classes (Chart, Transaction, InvestmentTransaction, HoldingSnapshot) are in the cases. No `unknown model` throw for the current call surface. ✓
- `indexedDb.remove()` — each `dataField` value ("charts", "transactions", "investmentTransactions", "holdingSnapshots") matches a `StoreName` enum entry (`accessor.ts:4-31`). Cast is safe. ✓
- Dictionary-constructor read (`useMutate.ts:66-68`) — `newData[Model.dataField]` on a fresh `new Data(oldData)` still holds the SAME dictionary reference (`assign` shallow-copies fields), so `.constructor` is `ChartDictionary` / `TransactionDictionary` / etc. `new DictCtor(current)` is `new Map(iterable)` under the hood — clones then `next.set(id, instance)` doesn't mutate `current`. ✓
- Snapshot key alignment — `HoldingSnapshot.id` returns `snapshot.snapshot_id`; `HoldingSnapshotDictionary` keys via `snap.snapshot.snapshot_id` (matches `service.ts:472`'s `data.id`). Delete route also uses `id` = snapshot_id. ✓
- `MutableModel<T>` widening — `new (...args: any[]): T` fits both zero-arg (`new Chart()`) and mandatory-arg (`new InvestmentTransaction({account_id})`) constructors. Tighter typing would need per-model overloads and drop the "generic across models" property; the `any[]` is fine.
- Server route normalization — `transaction_id` → `id` and `investment_transaction_id` → `id` search across the codebase turns up no non-test callers still using the old query keys. Both routes had no dedicated tests; nothing to update.
- Chart update path shape — `new Chart({...chart, ...patch})` runs `fromJSON`, which for chart-type-preserving updates re-wraps configuration into the correct subclass (`BalanceChartConfiguration` / `FlowChartConfiguration` / `ProjectionChartConfiguration`). `toJSON()` in Chart serializes `configuration` as a JSON STRING, which `updateChart` accepts (`charts.ts:57-62` handles both branches). ✓
- Hook stability — `upsert` and `del` are `useCallback([Model, setData])`. `Model` is an imported class reference (stable); `setData` is `useCallback` in `useData` (stable). Return object is `useMemo([upsert, del])`. Consumer `useEffect`s won't refire spuriously. ✓
- Delete-path throw semantics — matches the old `if (!success) throw new Error(r.message)`. No new silent-failure surface introduced. `router.back()` correctly gated post-await. ✓
- `new Data(oldData)` wrapper — the `assign` in Data's constructor shallow-copies dictionary references, so no side-effect fields are lost by centralizing the wrapper into the hook. ✓

claoie:
Landed the refactor in 23342ec1. All model→slot wiring now lives on `Data` in one file:

- `Data.dictOf(Model)` — overloaded, so `dictOf(Chart)` returns `ChartDictionary` etc.; impl is a switch on the class ref.
- `Data.storeNameOf(Model)` — same shape, returns `StoreName`.
- `Data.set(dict)` — writes back via `instanceof` dispatch (Dictionary knows its own type).

Hook body is now:

```ts
const newDict = newData.dictOf(Model).clone();
newDict.set(instance.id, instance);
newData.set(newDict);
indexedDb.save(instance).catch(console.error);
```

Zero casts in the hook. `MutableModel` static shrinks to just `apiPath` (`dataField` dropped from `Chart` / `Transaction` / `InvestmentTransaction` / `HoldingSnapshot`).

Adding a new model to `useMutate` is now one entry each in `dictOf` (overload + impl), `storeNameOf`, and `set` — all sitting next to the field declarations in `Data.ts`.

Verified: typecheck / lint / 1082 unit tests pass. UI-driven Playwright E2E at 390×844: seeded a chart via `/api/new-chart` + POST, navigated to `/chart-detail?chart_id=X`, clicked the DeleteButton, accepted the confirm dialog — `/api/charts` count went 8→7 (DB-level verification the refactored hook still lands the mutation).

src/client/lib/hooks/useChartDelete.ts
claoie: Done — landed as `useMutate(Model)` in 23342ec1: a single generic hook over the four statics-carrying models (Chart, Transaction, InvestmentTransaction, HoldingSnapshot). It awaits the API call, throws on failure, and on success updates both `Data` state and IndexedDB. `useChartDelete` is gone; the three chart-properties sites now call `useMutate(Chart)`.

Account/Budget aren't wired in yet — their POST routes aren't uniform upsert-by-id (Chart create flows through `GET /api/new-chart`, holding-snapshot expects a flat body), so wiring them means normalizing those response shapes first. Rather than widen this PR's scope, I left the hook generic over the models whose routes already fit and can do the server-side normalization + remaining models in a focused follow-up if you want it.

src/client/lib/hooks/useMutate.ts
claoie: Removed all casts in 23342ec1. The model→slot wiring now lives on `Data`, keyed off the class ref rather than a cast:

- `Data.storeNameOf(Model)` returns the `StoreName` enum directly (a `switch` on the class reference) — no `unknown`/`as`.
- `Data.dictOf(Model)` returns the concrete dictionary via overloads, and `Data.set(dict)` writes it back via `instanceof` dispatch (each Dictionary knows its own slot).

Hook body is now:

```ts
const newDict = newData.dictOf(Model).clone();
newDict.set(instance.id, instance);
newData.set(newDict);
indexedDb.save(instance).catch(console.error);
```

Zero casts. Adding a new model is one entry each in `dictOf`, `storeNameOf`, and `set`, all sitting next to the field declarations in `Data.ts`.

reviewoie:
## Code review — PR 628

Scope is clean: #627 is already merged, so the diff vs `main` is exactly this PR's two commits. Every file ties to the hook migration — no unrelated changes.

### High
- `src/client/lib/models/Data.ts:69` (`clone`) + `:280` (`set`) — **every `useMutate` create/update/delete throws `Data.set: unknown dictionary Dictionary` after a successful server call.** `clone = () => new Dictionary<T>(this) as S` constructs a **base** `Dictionary` at runtime and only *casts* it to the subclass type. `useMutate` does `newData.dictOf(Model).clone()` then `newData.set(newDict)`, and `set` dispatches with `if (dict instanceof ChartDictionary) …`. The cloned dict is a plain `Dictionary`, so every `instanceof` arm is false and the `else` throws. The throw lands inside the `setData` updater → surfaces during React's render pass (not caught by call-site try/catch, e.g. `HoldingProperties.onClickDeleteSnap`).

  Verified via runtime repro of the exact shape:
  ```
  cloned instanceof ChartDictionary: false
  cloned.constructor.name: Dictionary
  THREW: Data.set: unknown dictionary Dictionary
  ```
  No `Dictionary` subclass overrides `clone` (grepped — all are empty `class XDictionary extends Dictionary<…> {}`). This is why the codebase everywhere else builds dicts as `new TransactionDictionary(oldData.transactions)` (subclass-preserving) — the label/split handlers in these very same files still do. Only the new `clone()` path breaks. Failure scenario: server row is deleted/updated, then the app throws mid-render and the client dict never updates. All 10 migrated sites affected (chart delete/update ×6, tx delete, inv-tx delete, holding-snapshot delete). The E2E box is unchecked in the PR body and there is no unit test on `dictOf`/`set`/`clone`, so nothing caught it. A one-line `data.dictOf(Chart).clone() instanceof ChartDictionary` assertion would. Fix options: make `clone()` subclass-preserving (`new (this.constructor as …)(this)`), or have `dictOf` overloads / `set` construct the concrete subclass.

### Med
- `src/client/lib/hooks/useMutate.ts:38` (`upsert`) → `src/server/lib/postgres/repositories/charts.ts:49` (`updateChart`) — **chart update write surface widened; concurrent-edit clobber.** Old sites POSTed only the changed field (`{ chart_id, ...updatedChart }`); `useMutate.update` POSTs the full `new Chart({ ...chart, ...updatedChart })`. `updateChart` writes `name` AND `type` AND `configuration` whenever each is present, so every chart edit now rewrites all three from a possibly-stale client snapshot — two tabs editing different fields of the same chart clobber each other. This is the prior MED finding, still present, and not acknowledged in the PR body. (Currently masked by the High, but real once that's fixed.)

### Low
- `src/client/lib/hooks/useMutate.ts:29` — `create`/`update` are latent footguns for two of the four wired models; the JSDoc's "create and update share the server contract (upsert-by-id via POST)" is not universally true. (a) Chart: `POST /api/chart` calls `updateChart` (UPDATE-only, no insert) and the route returns `{status:"success"}` without checking the boolean, so `create(new Chart(...))` silently no-ops server-side while adding to client state. (b) HoldingSnapshot: `POST /api/snapshots/holding` reads a FLAT body (`body.quantity`, `body.snapshot_date`, …) but a `HoldingSnapshot` instance is nested (`{ snapshot: {…} }`), so `create`/`update` would POST a body the route can't read. Neither `create`/`update` is called for these models in this PR — latent only — but the interface offers them.
- `src/client/lib/models/Data.ts:258-262` — the "zero casts" claim holds for the hook body only; `dictOf` retains 4× `as unknown as Dictionary<T>`. Hoie's ask 2 ("avoid `unknown` casting") is partially met — the `StoreName` cast is genuinely gone (`storeNameOf` returns `StoreName`) — but `unknown`-casts persist, relocated into `dictOf`. Informational; overloaded-dispatch impls commonly need internal casts.

No further blockers.

claoie:
Thanks reviewoie — the HIGH is real and now fixed in `082c691e`. Confirmed the exact mechanism you named.

**HIGH — `clone()` produced a base `Dictionary`, so `Data.set`'s `instanceof` dispatch threw.** Root cause was `Dictionary.clone = () => new Dictionary<T>(this) as S` — it constructs a **base** `Dictionary` and only *casts* to the subclass. `useMutate` does `dictOf(Model).clone()` then `Data.set(dict)`, which dispatches on `instanceof ChartDictionary` / etc., so every arm was false → `else throw "unknown dictionary Dictionary"` inside the `setData` updater (surfaces at render). Fix: construct via `this.constructor`, so the clone keeps its concrete subclass:

```ts
clone = (): S => {
  const Ctor = this.constructor as new (init: Iterable<readonly [string, T]>) => S;
  return new Ctor(this);
};
```

This is subclass-preserving for every dictionary (empty-body subclasses inherit the base `Map` constructor; `TransferDictionary` accepts the iterable and rebuilds its pivot). It also fixes a latent bug in `Dictionary.map()`, which used the same broken `clone()` and silently downcast subclasses.

**Why the earlier E2E missed it:** the prior sweep only checked the DB row count (8→7). The DELETE fires *before* the throwing `setData` updater, so the row count dropped while the render threw — a proxy metric passing over a render bug. Fixed the verification too:

- **Unit** — added `Data.test.ts`: reproduces `dictOf(Chart).clone()` → mutate → `Data.set` for the upsert + delete paths and asserts `clone()` is `instanceof` the subclass. It **fails** against the old `clone` (`unknown dictionary Dictionary`) and passes after the fix. 1094 client+server tests pass, typecheck + lint clean.
- **Browser (isolated sandbox DB, 390×844)** — real click chain Dashboard → **Add Chart** → chart-detail (properties, Delete button) → click **Delete** → confirm dialog accepted → DELETE 200 → **navigated back to /dashboard** (so the `setData` updater completed; `router.back()` runs after it) → deleted chart gone from the DOM → **zero console/page errors** (no `unknown dictionary` throw). Screenshot eye-checked: clean dashboard render, no error boundary.

**MED — full-instance chart POST widening the write surface.** Real and still present (it was masked by the HIGH). It's a genuine trade-off of the generic hook: the old field-scoped partials (`{chart_id, name}` vs `{chart_id, configuration}`) become a full `new Chart({...chart, ...patch})`, so two concurrent editors on the same chart can last-write-wins clobber. Blast radius is small (single-user app, no last-write-wins guards anywhere today). Rather than fold a payload-shape redesign into this hook PR, I'd rather keep scope on the generic mutation and land the fix as a focused follow-up — either the hook translates a `Partial<T>` patch to `{id, ...patch}`, or chart updates stay field-scoped. Say the word and I'll open it; happy to fold it in here if you'd prefer it not ship as-is.

**LOW — `create`/`update` latent footguns (Chart POST is update-only; HoldingSnapshot expects a flat body).** Correct — no caller in this PR uses `create` for either (Chart uses `.update`/`.delete`, HoldingSnapshot isn't wired), so nothing is broken today, but the interface promises more than the routes deliver. Tracking alongside the MED follow-up: the durable fix is server-side normalization (uniform upsert-by-id + a `toApiBody()` so the nested instance flattens), which is the same work that would let Account/Budget join the hook. Documented as a limitation for now.

**LOW — "zero casts" scope.** Right — the hook *body* is cast-free (that's the `StoreName`/`unknown` cast you flagged, genuinely resolved via `storeNameOf`). The `dictOf` overload impl keeps 4× `as unknown as Dictionary<T>`; those are the unavoidable overload-implementation cast (the public overloads hand callers the concrete type, the impl narrows a union) and are invisible at call sites.

reviewoie:
## Code review — PR 628 (re-review, head 082c691e)

**Verdict:** approve.

**Findings:** 0 new. Prior HIGH resolved; prior MED + 2 LOWs dispositioned as follow-ups (reasonable — see below).

### High — RESOLVED
- `src/client/lib/models/Data.ts:72` (`clone`) — the base-`Dictionary`-clone bug is fixed. `clone` now constructs via `this.constructor`, so the clone keeps its concrete subclass and `Data.set`'s `instanceof` dispatch resolves instead of hitting the `unknown dictionary` throw. Verified end-to-end:
  - **Subclass-preserving for every path.** Empty-body subclasses (Chart/Transaction/InvestmentTransaction/HoldingSnapshot dictionaries) inherit the base `Map` constructor, so `new Ctor(this)` copies entries via `Map.prototype.set` (bypassing the `environment==="server"` guard, same as the old `new Dictionary(this)`). `TransferDictionary`'s explicit `constructor(init?)` accepts the iterable and rebuilds its `pivot`. No subclass has a constructor signature that breaks `new this.constructor(this)`.
  - **Regression guard proven.** Reverted `clone` to the old `new Dictionary<T>(this) as S` and re-ran `Data.test.ts`: 11/12 fail with the exact `Data.set: unknown dictionary Dictionary` error. Restored → 12/12 pass. The test fails against the bug and passes against the fix — a real guard, not a tautology.
  - `bun test src/client` → 338 pass / 0 fail; `tsc --noEmit` clean.

### map() side effect — safe
- `Dictionary.map()` (`Data.ts:63`) calls `this.clone()`, so it inherits the fix. Previously `map()` on any subclass returned a base `Dictionary` mislabeled as `S`; it now returns the real subclass. That is strictly more correct (callers were already typed to `S`) — no caller relies on the runtime being a base `Dictionary`. For `TransferDictionary` it's an outright improvement: `map` now preserves the pivot instead of silently dropping it. No behavior regression.

### Med (prior finding, unchanged) — reasonable follow-up
- `src/client/lib/hooks/useMutate.ts` upsert → `updateChart` full-instance POST widens the write surface (two-tab concurrent-edit last-write-wins clobber). Still present (was masked by the HIGH). claoie's disposition — keep the generic-mutation scope here, land the payload-shape fix (`Partial<T>` patch → `{id, ...patch}`, or a `toApiBody()` normalization) as a focused follow-up, and offered to fold in on request — is reasonable for a single-user app with no LWW guards anywhere today. Not a blocker for this PR.

### Low (prior findings) — reasonable
- `create`/`update` latent footguns (Chart POST is update-only; HoldingSnapshot route expects a flat body). Neither is exercised by any caller in this PR; documented as an interface limitation alongside the MED follow-up. Fine.
- `dictOf` retains 4× `as unknown as Dictionary<T>` — the unavoidable overload-implementation cast; call sites get the concrete type. Informational.

Scope clean — only `Data.ts` + `Data.test.ts` in the new commit, both tied to the fix.

No further blockers.

claoie:
reviewoie re-review at 082c691e: **approve, 0 findings** — HIGH confirmed resolved (you even reverted `clone` to prove `Data.test.ts` guards it: 11/12 fail on the old base-Dictionary clone, 12/12 pass on the fix). MED + LOWs stand as the dispositioned follow-ups. CI green (test + build). Ready for your look.

hoiekim (approved):

PR #629, opened 2026-07-09
refactor(client): shared `.sticky` utility for header-family sticky pattern (#326 phase 5g)

claoie:
## Summary

Task 4 of 4 per Hoie's kickoff (option 3 — full retrofit).

Extracted the `position: sticky; background-color: #191919ee; z-index: 110;` triplet — repeated across every sticky header in the codebase — into a shared `.sticky` utility class in `App/index.css`. Per Hoie's call, all sticky headers share z-index 110 (they never visually overlap; the prior 108/109/110/111 tuning collapses to source-order stacking, which matches the DOM order in each case).

## The shared utility

```css
/* App/index.css */
.sticky {
  position: sticky;
  background-color: #191919ee;
  z-index: 110;
}
```

## Retrofit

| Site | Change |
|------|--------|
| `<PageTitle>` | `<h2 className="PageTitle sticky">` — CSS block drops position/bg/z-index, keeps padding + `top`. |
| `<PageFilterTitle>` | `<h2 className="PageTitle PageFilterTitle sticky">` — inherits via the PageTitle class + explicit sticky. |
| `<TransactionsPageTitle>` | `<h3 className="sticky">` subtitle. |
| `<TransactionsHead>` | `<div className="TransactionsHead sticky">`. |
| `<SearchBar>` | Adds `sticky` to the classes array alongside `active`. |
| `TransactionsPageTitle/index.css` | Same triplet removed; per-site tuning (`top`, `padding`, `height`, `width`, `display`) retained. |

## Out of scope

- **`div.AccountsDonut` (wide-screen sticky)** — not a header. It's a content block that becomes sticky on wide viewports only, has conditional narrow/wide behavior, and its z-index (100) sits at a different tier than the header family (110). Left alone — the `.sticky` utility could be applied later if you want it merged into the same tier.

## Test plan

- [x] `bun test src/client` → 326 / 0
- [ ] Sandbox visual sweep — every list page (Dashboard / Budgets / Accounts / Transactions) + TransactionsPage subtitle sticky path (via a filter that produces a subtitle) to confirm scroll-lock still works.

### DIFF
--- a/src/client/components/App/index.css
+++ b/src/client/components/App/index.css
@@ -16,6 +16,18 @@ html {
   color-scheme: light dark;
 }
 
+/* Shared sticky-header rule. `top` and any padding/height are
+ * per-consumer — this only owns the pattern's invariants (sticky
+ * positioning, translucent overlay background, and a unified stacking
+ * level). Per Hoie 2026-07-09, all sticky headers share z-index 110
+ * since they never visually overlap — the intent is "these are the
+ * sticky layer," not a bespoke ordering. */
+.sticky {
+  position: sticky;
+  background-color: var(--bgOverlay);
+  z-index: 110;
+}
+
 body {
   width: 100%;
   margin: 0;
--- a/src/client/components/PageFilterTitle/index.tsx
+++ b/src/client/components/PageFilterTitle/index.tsx
@@ -60,7 +60,7 @@ export const PageFilterTitle = ({
   };
 
   return (
-    <h2 className={"PageTitle PageFilterTitle" + (className ? " " + className : "")}>
+    <h2 className={"PageTitle PageFilterTitle sticky" + (className ? " " + className : "")}>
       <button onClick={toggle} ref={buttonRef}>
         <span>{label}</span>
         <ChevronDownIcon size={15} />
--- a/src/client/components/PageTitle/index.css
+++ b/src/client/components/PageTitle/index.css
@@ -1,10 +1,7 @@
 h2.PageTitle {
   margin: 0;
   padding: 16px 10px 10px;
-  position: sticky;
   top: 50px;
-  background-color: var(--bgOverlay);
-  z-index: 110;
 }
 
 div.wideScreen h2.PageTitle {
--- a/src/client/components/PageTitle/index.tsx
+++ b/src/client/components/PageTitle/index.tsx
@@ -11,4 +11,6 @@ interface Props {
  * sticky/padding/z-index styling that every list page's own CSS used to
  * declare identically.
  */
-export const PageTitle = ({ children }: Props) => <h2 className="PageTitle">{children}</h2>;
+export const PageTitle = ({ children }: Props) => (
+  <h2 className="PageTitle sticky">{children}</h2>
+);
--- a/src/client/components/TransactionsPageTitle/SearchBar.tsx
+++ b/src/client/components/TransactionsPageTitle/SearchBar.tsx
@@ -26,7 +26,7 @@ export const SearchBar = ({ onChange, style }: SearchBarProps) => {
   };
 
   const classes = ["SearchBar"];
-  if (value) classes.push("active");
+  if (value) classes.push("active sticky");
   return (
     <div className={classes.join(" ")} style={style}>
       <input
--- a/src/client/components/TransactionsPageTitle/TransactionsHead.tsx
+++ b/src/client/components/TransactionsPageTitle/TransactionsHead.tsx
@@ -43,7 +43,7 @@ export const TransactionsHead = ({ sorter, getHeaderName, headerKeys, style }: P
     });
 
   return (
-    <div className="TransactionsHead" style={style}>
+    <div className="TransactionsHead sticky" style={style}>
       {headerComponents}
     </div>
   );
--- a/src/client/components/TransactionsPageTitle/index.css
+++ b/src/client/components/TransactionsPageTitle/index.css
@@ -1,31 +1,24 @@
-/* TransactionsPage-specific sticky rules. Base sticky/padding on `h2`
- * comes from `<PageTitle>`; button + dropdown chrome comes from
- * `<PageFilterTitle>`. Everything below layers TxPage-specific overrides
- * (fixed h2 height, higher z-index because SearchBar/TransactionsHead
- * sit visually below the h2) and the sticky siblings (h3 subtitle,
- * TransactionsHead, active SearchBar). */
+/* TransactionsPage-specific sticky siblings. Base
+ * `position: sticky; background-color; z-index` comes from the shared
+ * `.sticky` utility (see `App/index.css`) applied via className on
+ * each JSX element. This file only owns the per-element `top`,
+ * padding, and layout tuning. */
 div.TransactionsPage > h3,
-div.TransactionsPage > div.TransactionsHead,
 div.SearchBar.active {
-  position: sticky;
-  background-color: var(--bgOverlay);
   margin: 0;
 }
 
 div.TransactionsHead {
   padding: 0 10px 10px;
-  z-index: 108;
 }
 
 h2.TransactionsFilterTitle {
   height: 54px;
-  z-index: 111;
 }
 
 div.TransactionsPage > h3 {
   top: 105px;
   padding: 0 10px 10px;
-  z-index: 110;
 }
 
 div.wideScreen div.TransactionsPage > h3 {
@@ -39,7 +32,6 @@ div.SearchBar {
   justify-content: center;
   align-items: center;
   padding: 0 10px 10px;
-  z-index: 109;
 }
 
 div.SearchBar > input {
--- a/src/client/components/TransactionsPageTitle/index.css
+++ b/src/client/components/TransactionsPageTitle/index.css
@@ -12,6 +12,17 @@ div.TransactionsHead {
   padding: 0 10px 10px;
 }
 
+/* Local exception to the unified `.sticky` z-index 110. When the
+ * SearchBar is `.active` AND the list is scrolled past its natural
+ * offset, SearchBar and TransactionsHead share the same computed
+ * `top` and land at the same viewport y — the JSX places
+ * TransactionsHead later, so under equal z-index it would paint on
+ * top of the input. Bump SearchBar one notch so the input stays
+ * clickable while active. */
+div.SearchBar.active {
+  z-index: 111;
+}
+
 h2.TransactionsFilterTitle {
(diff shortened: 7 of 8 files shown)


reviewoie:
## Code review — PR 629

### High
- `src/client/components/App/index.css:28` (`.sticky { z-index: 110 }`) + `src/client/components/TransactionsPageTitle/SearchBar.tsx:29` — **z-index unification hides the SearchBar behind TransactionsHead when SearchBar is active AND the list is scrolled.** Prior to this PR, `div.SearchBar` was z-index 109 and `div.TransactionsHead` was 108, so the SearchBar (higher) covered TransactionsHead in the overlap region. Under the new uniform 110, source-order stacking wins — TransactionsHead is rendered AFTER SearchBar in `TransactionsPageTitle/index.tsx` (SearchBar is child 3, TransactionsHead is child 4), so TransactionsHead now paints on top.

  The two elements DO overlap in practice: both consume `transactionsHeadTop` as their inline `top` (`index.tsx:175, 180`), and both are `position: sticky` when the SearchBar is `.active`. When the user has typed in the SearchBar and scrolled far enough that TransactionsHead's normal-flow position rises above the sticky-top target, TransactionsHead sticks at the same viewport y as the SearchBar. The old z-index tuning (108/109) existed precisely to resolve this overlap in favor of the SearchBar.

  Verified on the sandbox at `/transactions`, admin user, filter `"a"`, spacer forced to create scroll room, scrollTop=500:
  - **main branch:** SearchBar `{y:104, z:109}`, TransactionsHead `{y:104, z:108}`; `document.elementFromPoint(inputCenter)` → `INPUT` (SearchBar visible).
  - **PR head 17ec0635:** SearchBar `{y:104, z:110}`, TransactionsHead `{y:104, z:110}`; `document.elementFromPoint(inputCenter)` → `DIV.TransactionsHead sticky` (SearchBar hidden). Screenshot of viewport y=90-170, narrow 390×844: user sees "Date ↓ Name Amount Account Budget Category" instead of their search input. Same result on wide 1400×900.

  Failure scenario: user types in the transactions search bar, scrolls the list. The column-header row visually covers the search input, hiding what they typed and the ✕-close affordance behind an opaque `#191919ee` panel.

  Fix options:
  - Keep TransactionsHead at a lower z-index than SearchBar (revert TransactionsHead's `sticky` class to a per-file rule at z-index 108 while SearchBar-active gets 109 or reuses `.sticky` at 110). Simplest.
  - Reorder JSX so TransactionsHead precedes SearchBar (unnatural — TransactionsHead should paint below the row area, not above the search bar).
  - Make `.sticky` a base pattern and split the z-index into two utility classes (e.g. `.sticky-header` and `.sticky-header-overlay`) so SearchBar-active can opt into a higher layer than TransactionsHead. Preserves the phase-5g abstraction while acknowledging the one intentional stacking difference the old CSS encoded.

  Hoie's "they never visually overlap" premise looks like it held for the pages that were sandbox-verified (Dashboard/Budgets/Accounts h2 stacking — all single sticky per page), but the TransactionsPage stack has a real two-sticky-at-same-top case the sandbox check didn't exercise (SearchBar-active + scroll).

### Med
No further blockers.

### Low
No further blockers.

Notes on the other verification angles (all clean):
- `.sticky` globally available: `App/index.css` imported from `App/index.tsx:6`, rendered by `start.tsx:23` at the SPA root — loads before every page. ✓
- PageFilterTitle inheritance: own CSS (`PageFilterTitle/index.css`) declares no `position` / `background-color` / `z-index` on the h2 itself; `.sticky` supplies them. ✓
- SearchBar sticky-when-active semantics preserved: `if (value) classes.push("active sticky")` at `SearchBar.tsx:29` matches the old `div.SearchBar.active { position: sticky; ... }` gate. ✓
- Top values preserved: PageTitle `top: 50px` (narrow) / `0` (wide), TransactionsPageTitle h3 `top: 105px` / `55px` (wide). ✓
- Scope: 8 files, all directly retrofit the `.sticky` utility. ✓
- No other `.sticky` collision in codebase; `AccountsDonut`'s wide-screen sticky at z-index 100 correctly left alone.

claoie:
Reviewoie HIGH addressed. The exact mechanism you named held — SearchBar-active + scrolled puts both stickies at the same computed `top`, and equal z-index falls to source order.

Fix: local override on `TransactionsPageTitle/index.css`:

```css
div.SearchBar.active {
  z-index: 111;
}
```

Bumps SearchBar one notch above the `.sticky` default (110) so the input stays clickable. Kept the unified 110 for the other pages Hoie audited (Dashboard / Budgets / Accounts) — each has only one sticky sibling so equal-z is still correct there.

Verified with the same repro you ran (mobile viewport 390x844, `/transactions`, type "a" in SearchBar, inject 3000px spacer, scrollTo 500):

- Both stickies land at `t:104/b:144` (same viewport y — confirms the overlap case is now reproducing).
- `getComputedStyle`: SearchBar `zIndex=111`, TransactionsHead `zIndex=110`.
- `elementFromPoint(searchInputCenter)` → `INPUT` (was `DIV.TransactionsHead` on the broken build).

Also confirmed SearchBar's only render site is `TransactionsPage`, so the override doesn't leak.

claoie:
Round-2 fix pushed (627409ab). My initial round-1 fix bumped only `SearchBar.active` to 111 — that closed the specific overlap you repro'd, but reading the pre-phase-5g file more carefully surfaced three other regressions I'd introduced:

- `TransactionsFilterTitle` (h2) fell from 111 → 110, so the filter title no longer sat above SearchBar/TransactionsHead once scrolled.
- `TransactionsHead` and `> h3` both defaulted to `.sticky` = 110 — same-top overlap would fall to source order rather than the intended h3-above-head.
- The SearchBar step from inactive (109, not sticky) to active (110 via `.sticky`) skipped the load-bearing 109/108 gap.

Restored the full pre-phase-5g layering, per-element on this file (matches what main had before this branch):

```css
h2.TransactionsFilterTitle { z-index: 111; }   /* top    */
div.TransactionsPage > h3  { z-index: 110; }
div.SearchBar              { z-index: 109; }
div.TransactionsHead       { z-index: 108; }   /* bottom */
```

The `.sticky` utility retrofit still owns `position: sticky` and the translucent background — only the z-index sits per-element on TransactionsPage. Utility remains flat 110 for the single-sticky pages (Dashboard / Budgets / Accounts) where the layering wasn't load-bearing.

E2E re-verified with the same repro (390×844 mobile, `/transactions`, SearchBar-active, scrollTo 500): both stickies land at `t:104/b:144`; SearchBar z=109, Head z=108; `elementFromPoint` at input center → `INPUT`.

reviewoie:
## Code review — PR 629 (round 2)

**Verdict:** approve

Round-2 delta (`627409ab`) verified against the pre-phase-5g baseline (`6f1ea909:src/client/components/TransactionsPageTitle/index.css`) — the four z-index values are byte-identical to the baseline:

| element | z-index (baseline & round-2) |
|---|---|
| `h2.TransactionsFilterTitle` | 111 |
| `div.TransactionsPage > h3` | 110 |
| `div.SearchBar` | 109 |
| `div.TransactionsHead` | 108 |

Also restored: `div.wideScreen div.TransactionsPage > h3 { top: 55px }`.

### Sticky-sibling pair enumeration on TransactionsPage

Layout (narrow-with-subtitle case — the worst-crowded arrangement): A@50-104, B@105-?, C@137, D@137 (C/D share `top: transactionsHeadTop` from `index.tsx:175/180`).

- **A (Filter h2, 111) vs B (h3, 110)** — 1px gap (A ends at 104, B starts at 105). If h3 ever overlapped upward, A wins. Correct.
- **A vs C (SearchBar active, 109)** — no overlap (C@137). A wins if edge-adjacent. Correct.
- **A vs D (TxHead, 108)** — no overlap. A wins if edge-adjacent. Correct.
- **B vs C** — B@105, C@137. Overlap only if h3 height > 32px; B(110) > C(109). Correct.
- **B vs D** — same shape as B vs C; B(110) > D(108). Correct.
- **C vs D** — SAME `top` value → they superpose. C(109) > D(108) — SearchBar input paints over TransactionsHead. This is the round-1 reproducer, now correctly ordered. Confirmed by the E2E `elementFromPoint` → INPUT.

Wide-screen variants (A@0-54, B@55-?, C@87, D@87) exhibit the same pair topology at shifted offsets; same conclusions.

### Specificity check on the retrofit

`.sticky { z-index: 110 }` (specificity 0,1,0) is overridden by every per-element rule in `TransactionsPageTitle/index.css` — `h2.TransactionsFilterTitle` / `div.TransactionsPage > h3` / `div.SearchBar` / `div.TransactionsHead` are all ≥ (0,1,1), so the layered z-indices win deterministically regardless of stylesheet order.

### Other `.sticky` consumers

- `<PageTitle>` component (`className="PageTitle sticky"`) has no live callers — grep for `<PageTitle>` returned zero.
- `<PageFilterTitle>` (`className="PageTitle PageFilterTitle sticky …"`) is used by `DashboardPage`, `BudgetsPage`, `AccountsPage` — one sticky h2 each, no adjacent sticky siblings, so the `.sticky` default of z-index 110 is sufficient. No overlap surface.
- `TransactionsPage` is the only page with multiple stickies, and it owns the per-element layering.

### Regressions from the round-2 delta

None. Round-1's SearchBar `z: 111` override became structural (`SearchBar → 109`, `TransactionsFilterTitle → 111`) — the round-1 patch is fully superseded, not stacked, so no dead override remains.

### Sticky-class wiring vs CSS invariants

Every element the old shared selector covered now carries the `sticky` class where and only where it needs `position: sticky` + overlay bg:
- `h3` — `sticky` always (`index.tsx:171`).
- `TransactionsHead` — `sticky` always (`TransactionsHead.tsx:46`).
- `SearchBar` — `sticky` only when `active` (`SearchBar.tsx:29` — `if (value) classes.push("active sticky")`). Non-active SearchBar has `z-index: 109` from its own rule but static position, so z-index is a no-op — not a bug, just harmless.

No further blockers.

claoie:
Thanks reviewoie — round-2 approve confirmed at 627409ab. The four z-index values are back to the pre-phase-5g baseline (Filter h2 111 > h3 110 > SearchBar 109 > TransactionsHead 108) and the `div.wideScreen … > h3 { top: 55px }` rule is restored. The C-vs-D superposition (SearchBar input paints over TransactionsHead at the shared `top`) was the round-1 reproducer and is now correctly ordered. No dead round-1 override remains — the SearchBar bump became structural rather than stacked.

hoiekim (approved):

PR #635, opened 2026-07-12
feat(performance): dollar-weighting caption on the MWR widget (#381 item 6)

claoie:
## What

Adds a **dollar-weighting caption** to the Investment Performance (MWR) widget — the last self-contained UX item from #381 (item 6).

A user with steady contributions systematically sees their money-weighted return (MWR) trail the benchmark's time-weighted return (TWR) **in a rising market**, even with identical holdings — because later dollars had less time to compound. That's *dollar-weighting*, an artifact of contribution timing, not weaker security selection. Without a note, the negative Diff row reads as "I underperformed the index," which is a misread.

The caption:

> With steady contributions in a rising market, part of this gap is dollar-weighting — later contributions had less time to compound, so your money-weighted return trails the benchmark's time-weighted return even for identical holdings — rather than necessarily weaker selection.

**Shown only when the misread is actually possible:**
- a negative gap (`gapPct < 0`),
- a **rising** benchmark (`benchmark.cumulative > 0`) — in a *falling* window steady buying is cheaper and timing pushes the gap positive, so the mechanism reverses and the caption would be backwards,
- a multi-contribution stream (≥2 in-window purchases).

A positive gap, a falling benchmark, or a lump-sum window needs no disclaimer.

## Design notes

- Rendered as a footnote-style caption line (`.performanceCaption`, same subtle italic grey as `.performanceFootnote`, matching `margin-top`), **not a tooltip** — consistent with the widget's mobile-first, no-hover treatment (the divergence-flag decision, Hoie 2026-07-06, already in this file).
- Zero calculation change. Pure additive UI text gated on already-computed `gapPct` + `benchmark.cumulative` + a new `contributionCount` (count of in-window positive flows). No change to `benchmark.ts`.
- The gap mixes selection (their holdings vs VOO) and timing (MWR vs TWR), so the copy says "part of this gap" / "rather than necessarily weaker selection" — not a definitive attribution.

## Scope

Retires **only item 6** of #381. Items 1 (cash-inclusive MWR), 2 (implausible-return display guard), 3 (benchmark picker), 4 (multi-account aggregation), and 5 (dividend capture) remain — each is either gated on the unresolved Plaid settlement mental model or a product-threshold decision, flagged separately on #381 for your call.

## Test plan

- [x] `bun run build-client` — built
- [x] `bun test` — 1094 pass / 0 fail
- [x] **reviewoie** — 2 MED accuracy findings (falling-benchmark reversal + causal overclaim) + 1 LOW (margin) — all three applied in the fix commit; re-confirmed by E2E below.
- [x] **UI E2E** — sandbox, admin/sandbox, 390×844 mobile, real row-click into each investment account's detail (`.AccountRow` → `/account-detail`), window picker + viewDate arrows driven live:
  - **Manual investment account, current viewDate, 1Y/3Y/All** — Diff renders **positive** (green) → caption correctly **absent** (positive gap needs no disclaimer).
  - **Chase brokerage, May 2025 viewDate, 1Y** — MWR trails a **rising** VOO by a small negative Diff (red) → caption **renders**. Screenshot eye-checked at 390px: grey italic caption wraps cleanly below the amber footnote, subtle styling matching `.performanceFootnote`, no overflow. This state (negative gap **with** rising benchmark) is exactly the falling-benchmark-reversal guard's positive case — the caption appeared only because VOO was up over the window.
  - Console: no errors (pre-existing localhost `sw.js` ServiceWorker registration warning only).

Refs #381 (item 6 — DCA/dollar-weighting UX copy).

### DIFF
--- a/src/client/components/PerformanceBenchmark/index.css
+++ b/src/client/components/PerformanceBenchmark/index.css
@@ -111,3 +111,10 @@
   color: var(--amber);
   cursor: help;
 }
+
+.performanceCaption {
+  font-size: 0.75em;
+  color: rgba(255, 255, 255, 0.4);
+  font-style: italic;
+  line-height: 1.35;
+}
--- a/src/client/components/PerformanceBenchmark/index.css
+++ b/src/client/components/PerformanceBenchmark/index.css
@@ -117,4 +117,5 @@
   color: rgba(255, 255, 255, 0.4);
   font-style: italic;
   line-height: 1.35;
+  margin-top: 0.3em;
 }
--- a/src/client/components/PerformanceBenchmark/index.tsx
+++ b/src/client/components/PerformanceBenchmark/index.tsx
@@ -228,6 +228,10 @@ export const PerformanceBenchmark = ({ accounts }: Props) => {
       vStart,
       vEnd,
       flowCount: flows.length,
+      // In-window purchases (positive flows). ≥2 signals a steady-
+      // contribution (DCA) pattern, which dollar-weights the MWR — see the
+      // caption render.
+      contributionCount: flows.filter((f) => f.amount > 0).length,
       yearsInWindow,
       mwr,
       mwrGain,
@@ -311,9 +315,19 @@ export const PerformanceBenchmark = ({ accounts }: Props) => {
     gapDollars,
     suppressAnnualized,
     isClamped,
+    contributionCount,
     divergentSecurityCount,
     txnExcessSecurityCount,
   } = computed;
+
+  // A steady contributor (≥2 in-window purchases) whose money-weighted return
+  // trails the benchmark's time-weighted return is seeing the effect of *when*
+  // the money went in — later dollars had less time to compound — not weaker
+  // security selection. Shown only when that misread is actually possible: a
+  // negative gap with a multi-contribution stream. A positive gap needs no
+  // such disclaimer.
+  const showDollarWeightingCaption =
+    contributionCount >= 2 && gapPct !== null && gapPct < 0;
 
   // Every value cell renders the same shape: a "total" line on top and, if
   // annualization isn't suppressed for short windows, a per-year line under
@@ -448,6 +462,15 @@ export const PerformanceBenchmark = ({ accounts }: Props) => {
             </span>
           )}
         </div>
+
+        {showDollarWeightingCaption && (
+          <div className="performanceCaption">
+            Steady contributions pull your money-weighted return below the
+            benchmark&apos;s time-weighted return even with identical holdings —
+            the gap reflects when you invested (dollar-weighting), not weaker
+            selection.
+          </div>
+        )}
       </Property>
     </>
   );
--- a/src/client/components/PerformanceBenchmark/index.tsx
+++ b/src/client/components/PerformanceBenchmark/index.tsx
@@ -320,14 +320,20 @@ export const PerformanceBenchmark = ({ accounts }: Props) => {
     txnExcessSecurityCount,
   } = computed;
 
-  // A steady contributor (≥2 in-window purchases) whose money-weighted return
-  // trails the benchmark's time-weighted return is seeing the effect of *when*
-  // the money went in — later dollars had less time to compound — not weaker
-  // security selection. Shown only when that misread is actually possible: a
-  // negative gap with a multi-contribution stream. A positive gap needs no
-  // such disclaimer.
+  // For a steady contributor (≥2 in-window purchases), part of a negative gap
+  // is dollar-weighting: later contributions had less time to compound, so the
+  // money-weighted return trails the benchmark's time-weighted return even for
+  // identical holdings. That mechanism only pulls the gap *negative* when the
+  // benchmark ROSE over the window — in a falling window steady buying is
+  // cheaper and timing pushes the gap positive, so the caption would be
+  // backwards. Gate on a rising benchmark accordingly. (The gap still mixes
+  // selection and timing, so the copy says "part of", not "all".)
   const showDollarWeightingCaption =
-    contributionCount >= 2 && gapPct !== null && gapPct < 0;
+    contributionCount >= 2 &&
+    gapPct !== null &&
+    gapPct < 0 &&
+    benchmark !== null &&
+    benchmark.cumulative > 0;
 
   // Every value cell renders the same shape: a "total" line on top and, if
   // annualization isn't suppressed for short windows, a per-year line under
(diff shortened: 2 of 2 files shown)


reviewoie (commented):
## Code review — PR 635

**Verdict:** informational — 2 med accuracy findings on the caption's causal claim, no ship blockers. Scope/CSS/convention are clean.

### Med
- `index.tsx:329-330` — **Trigger fires in falling-benchmark windows where the caption's causal story is reversed.** The gate is `contributionCount >= 2 && gapPct !== null && gapPct < 0`, but the caption's mechanism ("later dollars had less time to compound") only holds when the window ROSE. Concrete case: benchmark falls over the window (`benchmark.cumulative < 0`). Steady contributions then buy progressively *cheaper* — DCA's money-weighted cumulative return is *higher* than the from-start time-weighted return, i.e. timing produces a **positive** gap. So a negative gap in a falling window can't be dollar-weighting; it's selection (or data). The caption would assert the opposite of what's true. Gate should additionally require the benchmark return to be positive (`benchmark.cumulative > 0`) so the "later dollars compounded less" narrative is actually applicable.

- `index.tsx:466-472` — **Wording overclaims causation.** The displayed gap = user MWR of their *actual* holdings − benchmark TWR, which conflates two effects: (a) security selection (their holdings vs VOO) and (b) contribution timing (MWR vs TWR). The caption concludes "the gap reflects when you invested (dollar-weighting), **not weaker selection**" — a definitive attribution to timing that `gapPct < 0` does not support; the user's real holdings genuinely may trail VOO. The "even with identical holdings" clause correctly *isolates* the timing mechanism, but the closing clause then attributes the *whole* actual gap to it. Soften to "may reflect when you invested … rather than necessarily weaker selection" (or compare user-MWR to user-TWR to truly isolate timing). Note `contributionCount` also excludes the `vStart` pseudo-contribution the MWR treats as a windowStart-dated flow, so when a large window-open position dominates, the "steady contributions" framing is weak even at count ≥ 2 — an argument for the softer wording.

### Low
- `index.css:115-120` — `.performanceCaption` mirrors `.performanceFootnote` (0.75em / rgba(255,255,255,0.4) / italic) but omits its `margin-top: 0.3em`. Spacing is carried by the parent's `gap: 0.4em` flex, so the caption sits slightly tighter under its predecessor than the footnote does under its own — minor visual asymmetry between two sibling captions. Cosmetic.

### Notes (no action)
- Scope is clean — only the two `PerformanceBenchmark` files, both directly on-topic.
- Convention adherence is correct: footnote-style line, not a tooltip, matching the existing "mobile-first, hover doesn't exist on touch" decision (index.tsx:445-449, Hoie 2026-07-06).
- Comments are why-comments, not PR-narrative — pass comment hygiene.
- No off-by-one: `contributionCount` counts distinct in-window dates with net-positive flow (summed per date), a good DCA proxy; the `> windowStart && <= windowEnd` flow filter matches the MWR's boundary convention.

claoie:
Thanks reviewoie — both MED findings were correct and are fixed in b86ae4d2:

**MED 1 (falling-benchmark reversal).** Right — the "later dollars had less time to compound" mechanism only pulls the gap negative when the benchmark *rose*. Added `benchmark !== null && benchmark.cumulative > 0` to the gate, so a negative gap in a *falling* window (where DCA actually yields a positive timing gap) no longer shows the caption. E2E confirms it: the caption only surfaced on the Chase brokerage at a viewDate where VOO was up over the window (rising benchmark, small negative Diff); it stayed hidden on positive-gap states.

**MED 2 (causal overclaim).** Agreed the displayed gap conflates selection (their holdings vs VOO) and timing (MWR vs TWR), so "not weaker selection" overclaimed. Reworded to "part of this gap is dollar-weighting … rather than necessarily weaker selection" — a contributing-factor framing, not a definitive attribution.

**LOW (CSS margin).** Added `margin-top: 0.3em` to `.performanceCaption` for spacing parity with `.performanceFootnote` (E2E computed-style confirms 3.06px).

`contributionCount` still counts distinct per-date net-positive flows (the DCA proxy) and does not include the `vStart` window-open pseudo-contribution — I kept it that way deliberately: the caption is about *in-window* contribution timing, and gating on ≥2 in-window purchases is the conservative trigger. Open to revisiting if you'd rather weight by vStart-dominance.

reviewoie:
## Code review — PR 635 (HEAD b86ae4d2)

Scope is clean: both touched files (`PerformanceBenchmark/index.tsx`, `index.css`) are the MWR widget the title names. No unrelated changes.

Verified:
- **Type safety / no crash.** `showDollarWeightingCaption` reads `gapPct` (`number | null`, guarded `!== null`), `benchmark` (`BenchmarkResult | null`, guarded `!== null`), and `benchmark.cumulative` (`BenchmarkResult.cumulative` is non-nullable `number`). `contributionCount` is destructured at the render scope and used — no unused var. No new null deref.
- **Sign convention correct.** `contributionCount = flows.filter(f => f.amount > 0).length` — `flows` come from `extractCashFlows`, which only emits `Buy`/`Sell` and signs a Buy as `+amount` / Sell as `−|amount|` (`benchmark.ts:101-104`). So positive flows are genuine contributions (money in); sells are correctly excluded from the count. Matches the MWR sign-convention rule.
- **Gating logic sound.** Caption shows only for `≥2 contributions ∧ negative gap ∧ rising benchmark (cumulative > 0)`. The rising-benchmark guard is the right condition — in a falling window steady buying is cheaper and the timing effect reverses the gap's sign, so gating on `benchmark.cumulative > 0` prevents a backwards caption. Zero calc change; `benchmark.ts` untouched.
- **CSS consistent.** `.performanceCaption` mirrors `.performanceFootnote` (same `rgba(255,255,255,0.4)` italic grey); the whole file uses rgba literals for greys, so this is not a design-token regression.

### Low
- `PerformanceBenchmark/index.tsx:234` — `contributionCount` is counted AFTER `flows` is aggregated by date (`flowsByDate`), so same-day multiple buys collapse to one entry and a same-day buy+sell that nets non-positive isn't counted. So `≥2` means "≥2 distinct dates with net-positive flow," not "≥2 purchase transactions." That's a fine heuristic for the DCA/steady-contribution signal and the copy already hedges ("part of this gap" / "rather than necessarily weaker selection") — informational only, no change needed. No regression test encodes the gate (manual sandbox E2E covers both the caption-present and caption-absent branches per the PR body), acceptable for a pure-UI text gate on this otherwise-untested component.

No further blockers.

claoie:
Thanks reviewoie — approve confirmed at `b86ae4d2`. The one low (the ≥2-contribution gate has no encoded regression test) is fair: the gating is an inline boolean in the component (`showDollarWeightingCaption`), and both branches were exercised in the prior-round sandbox E2E — the caption surfaced only on a rising-benchmark window with a small negative gap and stayed hidden otherwise. Leaving the approved diff stable to keep this display-only PR minimal; happy to extract the predicate into a tested helper as a follow-up if you'd rather see it pinned by a unit test.

hoiekim:
Questions.

1. The widget currently says `Your asset (MWR)` so it already explains the impact of timing in performance. So current issue is not with something misleading but rather it's the gap in the user's understanding about MWR. Is that right?
2. Would this issue be mitigated by displaying another metric that represents security-weighted-return?

claoie:
Superseded by https://github.com/hoiekim/budget/issues/638 — Hoie's call on the design pivot. TWR row is the structural fix (isolates selection from timing) rather than a caption caveating the misread. Closing this in favor of the issue; the misread analysis in this PR's body is background for whoever picks the issue up.

PR #637, opened 2026-07-13
fix(budgets): include current-month spend in future rollover projection (#634)

### LINKED ISSUE #634: bug(budgets): future-month rollover projection omits current-month spend, overstating the +rolled surplus
## Summary

On a **future-month** (or future-year) budget view, the projected `+ rolled` figure in `LabeledBar` **overstates the carry-forward surplus by the current month's spending-to-date**. The read-side projection in `BudgetData.getRolledOver` forecasts next month as if you had spent nothing in the current month, even though `getBudgetData` already banked the current month's spend into the stored next-month `rolled_over_amount`.

## Root cause

`src/client/lib/models/Calculations.ts:366-374` — `BudgetData.getRolledOver`, future-projection branch:

```ts
let rolled = history.get(current.getEndDate()).rolled_over_amount; // carry INTO current month T; excludes T's own spend
const cursor = current.clone();
while (cursor.getEndDate() < target.getEndDate()) {
  if (!startMonthEnd || cursor.getEndDate() >= startMonthEnd) {
    rolled -= budgetLike.getActiveAmount(cursor.getEndDate(), "month"); // subtracts capacity only
  }
  cursor.next();
}
return rolled;
```

The authoritative per-month recurrence (from the accrual loop + `processTransaction`) is:

```
rolled_over(M) = rolled_over(M-1) + S(M-1) - C(M-1)
```

where `S` = that month's spending, `C` = that month's active capacity. So the true carry into `T+1` is `rolled_over(T) + S(T) - C(T)`. The projection returns `rolled_over(T) - C(T)` — it **omits `S(T)`**, the current month's spending-to-date. The error is a constant `-S(T)` for `T+1` and every later month (`T+2` onward legitimately has zero spend, but the missing `S(T)` is never added back).

The code is self-contradicting: `processTransaction` (`src/client/lib/hooks/calculation/budgets.ts:104, 123, 139, 155`) **does** deposit `S(T)` into the stored `rolled_over(T+1)` bucket (via `nextMonthDate`), but the projection ignores the stored value entirely, so that deposit is dead for display purposes.

## Reproduction (first-hand, against the real `getBudgetData` / `getRolledOver` modules)

A rollover budget, capacity 100/mo, old `roll_over_start_date`, with **one expense of 20 dated in the current month**, run through the real `getBudgetData` (warm, `isColdSync=false`) then read via `getRolledOver(nextMonth)`:

```
storedT  (carry into current month)      : -4900
storedT+1 (processTransaction deposit)   :    20   <- current-month spend WAS banked here
PROJECTED getRolledOver(nextMonth)       : -5000   -> rendered "+5000 rolled"
CORRECT  (storedT + S(T) - C(T))         : -4980   -> should render "+4980 rolled"
ERROR    projected - correct             :   -20   (= -S(T), exactly the current-month spend)
```

Numbers above are synthetic (cap 100, spend 20) — no real data. On a live deployment the current month always accrues spend as it progresses, so the next-month projection over-reports the surplus by the running month-to-date total in that budget.

## User-visible / reachability

- `LabeledBar/index.tsx:40-41` reads `budgetData.getSummary(barData, viewDate)`; `getSummary` (`Calculations.ts:404-408`) calls `getRolledOver` for the projection. `LabeledBar:165` renders a `<= 0` `rolled_over_amount` as a `+` surplus, so the overstatement surfaces directly as an inflated `+ $X rolled`.
- Future months are reachable through the UI: `DatePickerModal` "Next period" button (`DatePickerModal/index.tsx:33` `onNext`, button at `:130`) has no future clamp. Verified via a real click chain (login -> date picker -> Next period x2 -> Sept 2026): the future-month bars render `+ rolled` through this exact projection path.
- Future-**year** year-views are also affected: `getSummary` anchors the year projection at that year's Jan-15 (`Calculations.ts:404-408`), which routes through the same branch.
- Current-month and past views are unaffected — they return the authoritative stored value.

This is a **system bug, not persisted user state**: it is deterministic read-side projection logic producing a wrong number from correct inputs; nothing the user could re-curate in the UI changes it.

## Why tests miss it

Every `#562` projection test in `src/client/lib/hooks/calculation/budgets.rollover.test.ts:277-411` builds `budgetData` from an **empty** `TransactionDictionary`, so `S(T) = 0` and the defect is invisible. `budgets.test.ts:271-327` uses single-version capacities and never touches the future projection.

## Proposed fix

Seed the projection from the stored next-month bucket that already contains `S(T)`, instead of re-deriving from `rolled_over(T)` alone. Concretely, start `rolled` at `history.get(current.next()).rolled_over_amount` (which holds `S(T)` deposited by `processTransaction`) and subtract capacity from `T+1` onward — or equivalently add `history.get(current).rolled_over_amount`'s `S(T)` component back in. Add a regression test with a **current-month transaction** asserting `getRolledOver(nextMonth) == rolled_over(current) + S(T) - C(T)`.

## Environment

Found via daily deep user-testing on the prod-clone at deployable HEAD `upstream/main` (7e55de7d). The frozen clone's current month has no spending, so the discrepancy does not surface in its live UI; the magnitude is proven via the module-level reproduction above (same `getRolledOver` the UI calls).

claoie:
## Problem

On a **future-month** (or future-year) budget view, the projected `+ rolled`
figure in `LabeledBar` overstates the carry-forward surplus by the current
month's spending-to-date. `BudgetData.getRolledOver` forecast the next month
as if nothing had been spent this month, even though `getBudgetData` already
banked the current month's spend into the stored next-month `rolled_over_amount`.

## Root cause

`src/client/lib/models/Calculations.ts` — `getRolledOver`, future-projection
branch. The authoritative per-month recurrence (from the accrual loop +
`processTransaction`) is:

```
rolled_over(M) = rolled_over(M-1) + S(M-1) - C(M-1)
```

where `S` = that month's spend, `C` = that month's active capacity. So the
true carry into `T+1` is `rolled_over(T) + S(T) - C(T)`. The projection seeded
`rolled = rolled_over(T)` and subtracted only capacity, returning
`rolled_over(T) - C(T)` — it **omitted `S(T)`**, a constant `-S(T)` error for
`T+1` and every later month.

It was self-contradicting: `processTransaction` banks `S(T)` into the stored
`rolled_over(T+1)` bucket (via `nextMonthDate`), but the accrual loop stops at
`T`, so the projection ignored that deposit entirely.

## Fix

Seed the projection with `rolled_over(T) + rolled_over(T+1)` — the stored
next-month bucket is exactly `S(T)`. Capacity subtraction from `T+1` onward is
unchanged. A not-yet-open rollover keeps both buckets at 0, so its projection
is byte-identical.

## Tests

The `#562` projection tests build `budgetData` from an **empty**
`TransactionDictionary`, so `S(T) = 0` and the defect was invisible. Added a
`#634` regression describe block that spends in the current month and asserts:

- the current-month spend lands in the stored next-month bucket;
- `getRolledOver(nextMonth) == rolled_over(T) + S(T) - C(T)` (not `− C(T)` alone);
- the spend closes the gap vs an empty month by exactly `S(T)`.

Verified these **fail** on `upstream/main` (projection − empty-projection
difference measured 0, expected the spend amount) and **pass** with the fix.

## Verification

- `bun run typecheck` · `bun run lint` · `bun run build` — all clean.
- `bun run test` — full suite green (1097 pass, 2 skip, 0 fail); the new
  `#634` block is 3 of them.
- **E2E** (Playwright, 390×844, local prod-clone): logged in → Budgets →
  DatePickerModal **Next period** ×2 → future month renders through the fixed
  `getSummary → getRolledOver` path. The future-month bars accrue exactly one
  capacity step per month (at +2 months each rollover budget's surplus grew by
  twice its own monthly capacity), correct layout, no non-benign console errors
  (only the known dev service-worker MIME warning). Eye-checked the rendered
  PNG.
- The frozen prod-clone's current month has **no** spend, so `S(T) = 0` and the
  magnitude of this correction cannot surface in its live UI — the magnitude is
  proven at the module level by the regression tests above (same `getRolledOver`
  the UI calls). This is stated per the E2E-reachability honesty rule.

Closes #634

### DIFF
--- a/src/client/lib/models/Calculations.ts
+++ b/src/client/lib/models/Calculations.ts
@@ -363,7 +363,16 @@ export class BudgetData {
       ? new ViewDate("month", roll_over_start_date).getEndDate()
       : undefined;
 
-    let rolled = history.get(current.getEndDate()).rolled_over_amount;
+    // Seed with the carry INTO the current month T plus T's own spend-to-date
+    // S(T). `processTransaction` banks S(T) into the stored NEXT-month bucket
+    // (rolled_over(T+1)), but the accrual loop stops at T so that deposit is
+    // otherwise dead for the projection. The authoritative recurrence is
+    // rolled_over(T+1) = rolled_over(T) + S(T) - C(T); dropping S(T) here
+    // overstated a future "+ rolled" surplus by the running month-to-date
+    // spend (#634). For a not-yet-open rollover both buckets are 0.
+    let rolled =
+      history.get(current.getEndDate()).rolled_over_amount +
+      history.get(current.clone().next().getEndDate()).rolled_over_amount;
     const cursor = current.clone();
     while (cursor.getEndDate() < target.getEndDate()) {
       if (!startMonthEnd || cursor.getEndDate() >= startMonthEnd) {
--- a/src/client/lib/models/Calculations.ts
+++ b/src/client/lib/models/Calculations.ts
@@ -365,11 +365,9 @@ export class BudgetData {
 
     // Seed with the carry INTO the current month T plus T's own spend-to-date
     // S(T). `processTransaction` banks S(T) into the stored NEXT-month bucket
-    // (rolled_over(T+1)), but the accrual loop stops at T so that deposit is
-    // otherwise dead for the projection. The authoritative recurrence is
-    // rolled_over(T+1) = rolled_over(T) + S(T) - C(T); dropping S(T) here
-    // overstated a future "+ rolled" surplus by the running month-to-date
-    // spend (#634). For a not-yet-open rollover both buckets are 0.
+    // (rolled_over(T+1)), while the accrual loop stops at T, so the recurrence
+    // rolled_over(T+1) = rolled_over(T) + S(T) - C(T) is completed here by
+    // reading both buckets. For a not-yet-open rollover both are 0.
     let rolled =
       history.get(current.getEndDate()).rolled_over_amount +
       history.get(current.clone().next().getEndDate()).rolled_over_amount;
--- a/src/client/lib/models/Calculations.ts
+++ b/src/client/lib/models/Calculations.ts
@@ -342,12 +342,12 @@ export class BudgetData {
   /**
    * The carry-forward amount for `budgetLike` at `date`.
    *
-   * Past and current months are authoritative — accrued into the history by
-   * getBudgetData — so we read the stored `rolled_over_amount`. For future
-   * months there is no stored value yet, so we project the current month's
-   * carry forward, subtracting each month's active capacity once the rollover
-   * window has opened (#562). A budget-like whose `roll_over_start_date` is
-   * itself in the future contributes nothing before it begins.
+   * Months through the next one (T+1) are authoritative — accrued into the
+   * history by getBudgetData — so we read the stored `rolled_over_amount`.
+   * Beyond T+1 there is no stored value yet, so we project forward from T+1,
+   * subtracting each month's active capacity once the rollover window has
+   * opened (#562). A budget-like whose `roll_over_start_date` is itself in the
+   * future contributes nothing before it begins.
    */
   getRolledOver = (budgetLike: BudgetFamily, date: Date): number => {
     const history = this.get(budgetLike.id);
@@ -363,15 +363,15 @@ export class BudgetData {
       ? new ViewDate("month", roll_over_start_date).getEndDate()
       : undefined;
 
-    // Seed with the carry INTO the current month T plus T's own spend-to-date
-    // S(T). `processTransaction` banks S(T) into the stored NEXT-month bucket
-    // (rolled_over(T+1)), while the accrual loop stops at T, so the recurrence
-    // rolled_over(T+1) = rolled_over(T) + S(T) - C(T) is completed here by
-    // reading both buckets. For a not-yet-open rollover both are 0.
-    let rolled =
-      history.get(current.getEndDate()).rolled_over_amount +
-      history.get(current.clone().next().getEndDate()).rolled_over_amount;
-    const cursor = current.clone();
+    // Seed from the authoritative carry INTO month T+1: getBudgetData's accrual
+    // loop now runs one month past T, completing the recurrence
+    // rolled_over(T+1) = rolled_over(T) + S(T) - C(T) in the stored history
+    // (rather than leaving T+1 as a bare S(T) deposit), so T's own spend-to-date
+    // is already folded in. Project from T+1 onward. For a not-yet-open rollover
+    // the seed is 0.
+    const oneAhead = current.clone().next();
+    let rolled = history.get(oneAhead.getEndDate()).rolled_over_amount;
+    const cursor = oneAhead.clone();
     while (cursor.getEndDate() < target.getEndDate()) {
       if (!startMonthEnd || cursor.getEndDate() >= startMonthEnd) {
         rolled -= budgetLike.getActiveAmount(cursor.getEndDate(), "month");
--- a/src/client/lib/hooks/calculation/budgets.rollover.test.ts
+++ b/src/client/lib/hooks/calculation/budgets.rollover.test.ts
@@ -333,6 +333,81 @@ describe("BudgetData.getRolledOver future-month projection (#562)", () => {
   });
 });
 
+// #634: the future-month projection dropped the CURRENT month's spend-to-date
+// S(T). processTransaction banks S(T) into the stored next-month bucket
+// (rolled_over(T+1)) but the accrual loop stops at T, so getRolledOver's seed
+// (rolled_over(T) alone) omitted it — overstating a future "+ rolled" surplus
+// by S(T). getRolledOver now seeds with rolled_over(T) + rolled_over(T+1). The
+// #562 projection tests above build from an EMPTY TransactionDictionary, so
+// S(T) = 0 and the defect is invisible — these use a current-month spend.
+describe("BudgetData.getRolledOver future projection subtracts current-month spend (#634)", () => {
+  const OLD_START = "2022-06-01";
+  const SPEND = 20;
+
+  const makeCurrentMonthSpend = (): TransactionDictionary => {
+    const dict = new TransactionDictionary();
+    const t = new Transaction({
+      transaction_id: "txn-current",
+      account_id: "acc-1",
+      name: "Coffee",
+      merchant_name: "Cafe",
+      amount: SPEND,
+      date: new ViewDate("month").getStartDate().toISOString().slice(0, 10),
+      pending: false,
+      label: { budget_id: "bud-1", category_id: null },
+    } as never);
+    dict.set(t.transaction_id, t);
+    return dict;
+  };
+
+  const buildBudgetData = (transactions: TransactionDictionary) => {
+    const { budgetData } = getBudgetData(
+      transactions,
+      emptySplits(),
+      makeAccount(),
+      makeBudget(OLD_START),
+      emptySections(),
+      emptyCategories(),
+      new TransferDictionary(),
+      false, // warm / steady state
+    );
+    return budgetData;
+  };
+
+  const budget = () => makeBudget(OLD_START).get("bud-1")!;
+
+  test("current-month spend is banked into the stored next-month bucket", () => {
+    const budgetData = buildBudgetData(makeCurrentMonthSpend());
+    const oneAhead = new ViewDate("month").next().getEndDate();
+    expect(budgetData.get("bud-1", oneAhead).rolled_over_amount).toBeCloseTo(SPEND, 6);
+  });
+
+  test("next-month projection = carry(T) + S(T) - C(T), not carry(T) - C(T)", () => {
+    const withSpend = buildBudgetData(makeCurrentMonthSpend());
+    const now = new ViewDate("month").getEndDate();
+    const oneAhead = new ViewDate("month").next().getEndDate();
+
+    const carryIntoCurrent = withSpend.get("bud-1", now).rolled_over_amount;
+    const projected = withSpend.getRolledOver(budget(), oneAhead);
+
+    // Authoritative recurrence: rolled_over(T+1) = rolled_over(T) + S(T) - C(T).
+    expect(projected).toBeCloseTo(carryIntoCurrent + SPEND - MONTHLY_CAPACITY, 6);
+  });
+
+  test("the spend closes the gap vs an empty month by exactly S(T)", () => {
+    const empty = buildBudgetData(new TransactionDictionary());
+    const withSpend = buildBudgetData(makeCurrentMonthSpend());
+    const oneAhead = new ViewDate("month").next().getEndDate();
+
+    const projectedEmpty = empty.getRolledOver(budget(), oneAhead);
+    const projectedSpend = withSpend.getRolledOver(budget(), oneAhead);
+
+    // rolled_over is stored negative (renders "+"). Spending S(T) makes the
+    // surplus exactly S(T) SMALLER — less negative than the empty-month case.
+    expect(projectedSpend - projectedEmpty).toBeCloseTo(SPEND, 6);
+  });
+});
+
 // getSummary is the unified read the bars use: sorted/unsorted + rolled_over
 // from one call, so the rollover no longer flows on a separate path in the UI.
 // It must agree with the underlying history + getRolledOver it abstracts.
--- a/src/client/lib/hooks/calculation/budgets.rollover.test.ts
+++ b/src/client/lib/hooks/calculation/budgets.rollover.test.ts
@@ -334,12 +334,11 @@ describe("BudgetData.getRolledOver future-month projection (#562)", () => {
 });
 
 // #634: the future-month projection dropped the CURRENT month's spend-to-date
-// S(T). processTransaction banks S(T) into the stored next-month bucket
-// (rolled_over(T+1)) but the accrual loop stops at T, so getRolledOver's seed
-// (rolled_over(T) alone) omitted it — overstating a future "+ rolled" surplus
-// by S(T). getRolledOver now seeds with rolled_over(T) + rolled_over(T+1). The
-// #562 projection tests above build from an EMPTY TransactionDictionary, so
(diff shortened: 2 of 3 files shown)


reviewoie (commented):
## Code review — PR 637

Verified the recurrence, the no-double-count claim, the clone/next non-mutation, and that the tests pin the bug (ran the suite on the PR head, then reverted only `Calculations.ts` to `upstream/main` — 2 of the 3 new tests fail without the source fix; full file 17/17 with it).

- **Recurrence is correct in every branch.** `processTransaction` deposits each month-T transaction into `rolled_over(T+1)` (budgets.ts:104/155), and the accrual loop's bound (`startDate.getEndDate() <= endDate.getEndDate()`, endDate = current month T) never writes T+1 — so the stored `rolled_over(T+1)` is exactly S(T), and the seed `rolled_over(T) + rolled_over(T+1)` = carry-into-T + S(T). The loop then subtracts C(T), giving `rolled_over(T) + S(T) - C(T)`. Matches `rolled_over(T+1) = rolled_over(T) + S(T) - C(T)`.
- **Not-yet-open rollover:** future `roll_over_start_date` → `processTransaction`'s guard never deposits and the accrual loop never runs, so both buckets are 0; the `startMonthEnd` guard then skips capacity until the window opens. Correct, unchanged from #562.
- **Year-view January anchor:** `getSummary` projects at Jan-15 (month interval) of a future year; the seed reads the real current month T and walks monthly. Threading S(T) into the seed is consistent with the January-carry semantics. No issue.
- **No double-count:** future-dated (month T+1) transactions deposit into T+2, not T+1, so they can't inflate the seed; `rolled_over(T+1)` holds only month-T spend.
- **`current.clone().next()`** does not mutate `current` — `clone()` returns a fresh `ViewDate`, `next()` mutates only the clone (`date.ts:108/156`). The subsequent `const cursor = current.clone()` still starts at T.

### Low
- `Calculations.ts:366-372` — the seed comment carries a bug-narrative clause ("dropping S(T) here overstated a future '+ rolled' surplus by the running month-to-date spend (#634)"). The load-bearing part (the banking mechanism + the `rolled_over(T+1) = rolled_over(T) + S(T) - C(T)` recurrence) earns its lines; consider trimming the "overstated…(#634)" sentence per comment-hygiene (comments explain the invariant, not the fixed bug). Non-blocking.

No further blockers.

claoie:
Thanks reviewoie — addressed the low in d9ac459: trimmed the "overstated a future + rolled surplus…(#634)" bug-narrative clause from the `getRolledOver` seed comment, keeping the load-bearing mechanism (why we read `rolled_over(T+1)`) and the recurrence invariant. Comment-only, no runtime change; rollover suite still 17/17, typecheck + lint clean.

Confirmed on the substance: seed `rolled_over(T) + rolled_over(T+1)` completes `rolled_over(T+1) = rolled_over(T) + S(T) − C(T)`, both buckets are 0 for a not-yet-open rollover, and the third new test intentionally pins `processTransaction`'s pre-existing banking of S(T) into T+1 (passes either way by design — it documents the mechanism the fix relies on).

reviewoie:
## Code review — PR 637

HEAD: `d9ac4597ed675337b152411b994d8de0d5607213`

### Correctness (verified)
The seed correction is mathematically sound. Traced the recurrence end-to-end:

- `processTransaction` (`budgets.ts:71,104/123/139/155`) banks month-T spend into the stored `nextMonthDate` bucket → `rolled_over(T+1)`.
- The accrual loop (`budgets.ts:213`) runs `while startDate <= endDate` with `endDate = current month T`, so it writes `rolled_over` up to and including T, and never writes T+1. Therefore `history.get(T+1).rolled_over_amount` holds exactly `S(T)` (processTransaction's deposit) with no accrual contribution — no double count.
- Old seed `rolled_over(T)` alone → projection returned `rolled_over(T) − C(T)`, omitting `S(T)`. New seed `rolled_over(T) + rolled_over(T+1)` = `rolled_over(T) + S(T)`, then the loop subtracts `C(T)` → `rolled_over(T) + S(T) − C(T)` = the authoritative `rolled_over(T+1)`. Correct for T+1 and every later month (later months add only capacity subtractions; `S` for future months is 0).
- Not-yet-open rollover: `processTransaction`'s deposit is gated on `roll_over_start_date <= transactionDate` (`budgets.ts:102`), so both buckets stay 0 → projection byte-identical, as claimed. Confirmed the future-open window case walks correctly through the `startMonthEnd` guard.
- `BudgetHistory.get` (`Calculations.ts:230`) returns a zeroed `BudgetSummary` for an absent date, so the new T+1 read cannot NaN.
- `ViewDate.next()` is called on `current.clone()`, so `current` is not mutated between the two reads.
- Both callers (`Calculations.ts:415` getSummary, `BalanceChartRow/index.tsx:63`) hit the same path; the early return at `Calculations.ts:357` shields current/past dates, so only future projections change. No caller breaks.

**Regression coverage verified empirically:** reverted the seed to the pre-fix single-bucket form and re-ran the `#634` block — 2 of 3 tests fail (`projectedSpend − projectedEmpty` measured 0, expected 20); restored the fix and the full rollover + budgets calc suites pass (35/35). The tests assert the recurrence directly, not a proxy.

**Scope:** clean — two files, both directly related (`Calculations.ts` fix + `budgets.rollover.test.ts` coverage). No unrelated changes. Committed comment is qualitative why-documentation (the processTransaction/accrual coupling), not PR-narrative; leak check clean on the commit body.

### Low
- `budgets.rollover.test.ts` — the `#634` block covers T+1 thoroughly (banked bucket, projection value, gap-vs-empty). No spend-bearing assertion at T+2 or on a year view. Not a real gap — for months past T+1 the only added term is capacity subtraction (future `S` is 0), which the `#562` empty-transaction tests already cover, and year views route through the same `getRolledOver` — but a single T+2-with-spend assertion would pin the "seed correction survives multi-month accrual" invariant explicitly. Optional.

No further blockers.

claoie:
Thanks reviewoie — approve confirmed at `d9ac4597`. Appreciate the seed-revert check: proving the three new tests actually fail on the pre-fix form (2 of 3 measured 0 vs expected 20) is the real regression proof, not the PR body's claim. On the low: a T+2-with-spend assertion would pin the recurrence invariant explicitly. It's a pure test addition (no source change) on load-bearing budget math, so I'm inclined to fold it in — but since you flagged it as "not a real gap" (future S=0, and #562 covers multi-month capacity), I'll leave the approved diff stable unless you'd like the extra assertion in this PR.

src/client/lib/models/Calculations.ts
```diff
@@ -363,7 +363,14 @@ export class BudgetData {
...
 
-    let rolled = history.get(current.getEndDate()).rolled_over_amount;
+    // Seed with the carry INTO the current month T plus T's own spend-to-date
+    // S(T). `processTransaction` banks S(T) into the stored NEXT-month bucket
+    // (rolled_over(T+1)), while the accrual loop stops at T, so the recurrence
+    // rolled_over(T+1) = rolled_over(T) + S(T) - C(T) is completed here by
+    // reading both buckets. For a not-yet-open rollover both are 0.
+    let rolled =
+      history.get(current.getEndDate()).rolled_over_amount +
+      history.get(current.clone().next().getEndDate()).rolled_over_amount;
```
hoiekim: `rolled_over(T+1)` being same with `S(T+1)` doesn't sound semantically correct. It's just a side product of the calculation loop. Not intentional. It hasn't been causing issue because we are not using it. Now using it in this place to represent `S(T+1)` is fragile. Feel free to correct me if I'm wrong.

claoie: You're right — `rolled_over(T+1)` is a side-product, not an intentional "current-month spend" value, and reading it here couples this projection to an implementation detail of the accrual loop: it only equals `S(T)` because the loop's bound is `startDate <= endDate(T)`, so it stops before writing `T+1`. If that bound ever moved (e.g. the loop accrued one month further), the seed would silently double-count. Fragile as you say.

One correction on the seemingly-obvious alternative, so we pick the right shape: reading the current month's own spend buckets — `sorted_amount(T) + unsorted_amount(T)` — is **not** a drop-in equivalent. `processTransaction` gates the `rolled_over(T+1)` deposit on a *per-transaction* `roll_over_start_date <= transactionDate` (budgets.ts:102/121/137/153), while the `sorted`/`unsorted` deposits are ungated. So for a rollover that opens partway through month T, `rolled_over(T+1)` counts only the on-or-after-start spend (correct), whereas `sorted+unsorted(T)` would also count that month's pre-start spend (over-count). The buckets don't retain per-transaction dates, so I can't reconstruct the gated subset from them. That per-transaction gate is exactly why the current code piggybacks on `rolled_over(T+1)`.

So the genuinely convention-aligned fix is to make `rolled_over(T+1)` authoritative instead of a coincidence:

- **(a) Extend the accrual loop one step** so it writes `rolled_over(T+1) = rolled_over(T) + S(T) - C(T)` into the stored history (currently it stops at T). Then `getRolledOver` seeds directly from the stored `rolled_over(T+1)` and starts its capacity loop at T+1 — the reconstruction disappears and the coupling is gone. Blast radius: the accrual bound in `getBudgetData`, plus every consumer that reads the `T+1` bucket directly — `aggregateYear` (reads January's rollover only, so a future-January year view would newly see a projected value), and `toArray` spans. I'd want to trace those before landing.
- **(b) Keep the read but make the intent explicit** — a named accessor for "month-T banked spend" so the semantics are stated, not inferred. Cheaper, but it's documentation over a structural fix; doesn't remove the loop-boundary coupling.

I lean toward (a) — it's the honest shape and kills the fragility. It's more than a one-liner on load-bearing budget math, so: want me to fold (a) into this PR, or land the current fix and open a follow-up so this PR stays scoped to the `#634` projection bug?

hoiekim: (a) in this PR

reviewoie:
## Code review — PR 637

**Verdict:** approve / informational

Verified the two-part change against every angle in the task.

### Correctness — recurrence traced across all window positions
- Deposit + accrual compose to the stated recurrence `rolled_over(T+1) = rolled_over(T) + S(T) − C(T)`: `processTransaction` deposits month-T spend into the `T+1` bucket (`budgets.ts:71,104`), and the accrual loop now runs one iteration further (`accrualEnd = endDate.clone().next()`, `budgets.ts:218,225`) so `history.add(T+1, rolled_over(T) − C(T))` sums onto that deposit (`add` accumulates, `Calculations.ts:240-245`).
- `getRolledOver` deltas cleanly: NEW − OLD = `S(T)` for a past/current-opened rollover, which is exactly the dropped current-month spend #634 was about. Traced:
  - target ≤ T → early return of stored value, untouched by the diff (`Calculations.ts:357-359`). The current-month bar itself is unchanged (correct — its carry-in excludes S(T)).
  - target = T+1 → loop body doesn't execute (`cursor T+1 < target T+1` false), returns the authoritative seed directly.
  - target ≥ T+2 → seed − Σ C from T+1 forward, matching the carry definition.
- Future-start rollover seed is genuinely 0: the forward deposit is gated on `roll_over_start_date <= transactionDate` (`budgets.ts:99-103, 122-125, 138-141`), so pre-start spend does NOT leak into `rolled_over(T+1)`, and the accrual loop starts at `roll_over_start_date.next()` so it writes nothing at/below T+1. OLD and NEW agree here. The "seed is 0" doc comment is accurate.

### Blast radius of the new authoritative T+1 bucket — no regression
Only stored field the extra iteration mutates is `rolled_over_amount` at month T+1 (sorted/unsorted untouched). Every reader checked:
- `getSummary` (LabeledBar, BudgetDetailPage) computes rollover via `getRolledOver`, not the stored bucket (`Calculations.ts:410-415`) — consistent path.
- `aggregateYear` reads only January's stored rollover; its value is discarded in `getSummary` (overridden by `getRolledOver`) and ignored by the only `BudgetHistory.toArray` consumer, the BudgetDetailPage graph, which maps only `sorted_amount + unsorted_amount` (`BudgetDetailPage/lib/hooks.ts:27-29`; bounds at :70-74 use capacity/spending, not stored rollover). Even the T = December case (T+1 = next-year January) never surfaces the changed value.
- `BalanceChartRow` uses `getRolledOver` (`index.tsx:63`) — projection path, fine.
- IDB persist/reload: `budgetData` is recomputed fresh by `getBudgetData` on every calculations update (`App/lib/hooks.ts:70`); the persisted history is a transient warm-start cache, superseded by recompute — same staleness surface that already existed for month T, not a new class.

### Scope
Clean. Three files, all serving the refactor; `endDate` only referenced at its definition and the `clone().next()` (no mutation leak).

### Tests
`budgets.rollover.test.ts` +4 tests pin the new semantics: authoritative T+1 storage (`carry(T) + SPEND − C`), seed-from-T+1 projection, the exactly-S(T) gap vs an empty month, and T+2 continuation subtracting one more capacity. Ran `bun test` on the file: 18 pass / 0 fail — the 14 prior tests (incl. the #562 empty-dict projection where S(T)=0 → NEW == OLD) stay green.

No further blockers.

src/client/lib/models/Calculations.ts
claoie: Done — (a) landed in this PR (ad371ab1).

The accrual loop in `getBudgetData` now runs one month past the current month T, completing the recurrence `rolled_over(T+1) = rolled_over(T) + S(T) - C(T)` into the stored history. The T+1 bucket is authoritative now, not a coincidental S(T) deposit that only survived because the loop stopped at T. `getRolledOver` seeds its projection directly from the stored `rolled_over(T+1)` and walks capacity from T+1 onward, so the read no longer reconstructs T+1 from two buckets and hopes the loop boundary lines up. If the accrual bound ever moves again, the seed moves with it instead of silently double-counting.

Output is identical to the prior seed across every rollover-window position — start in the past, start = T, start in the future (traced each; the future-start seed is a genuine 0 because the deposit is gated on `roll_over_start_date <= transactionDate`). Tests updated: the T+1 bucket now asserts the full recurrence (was a bare S(T)), plus a T+2 case pinning the cursor's T+1 start.

reviewoie: approve, 0 findings · E2E: drove budgets at 390x844, current + T+3 future months render clean with no JS errors; verified numerically against an `upstream/main` baseline on the same DB — every budget's projection differs from main by exactly the current-month spend the fix restores, in both cold-load and warm-reload states · CI: build + test green on ad371ab1.

hoiekim (approved):

PR #639, opened 2026-07-14
refactor(transactions): centralize label suggestion-state predicates on TransactionLabel

### LINKED ISSUE #532: byTransactionId membership predicates on TransferDictionary + TransactionDictionary
## Idea

Per Hoie's PR 528 inline review: the membership-test predicates we use
in calc code (e.g. `getBudgetData` deciding whether a transaction is a
confirmed-transfer half) should live as methods on the dictionary
rather than inline in the calc.

Proposed shape on `TransferDictionary`:

```ts
transfers.byTransactionId.has(t_id)            // any pair?
transfers.byTransactionId.hasSuggested(t_id)   // pair && status === "suggested"
transfers.byTransactionId.hasConfirmed(t_id)   // pair && status === "confirmed"
transfers.byTransactionId.get(t_id)            // returns TransferPair | undefined
```

Today the calc has to spell out the status check:

```ts
const isConfirmedTransferHalf = (transaction_id: string): boolean =>
  transfers.getByTransactionId(transaction_id)?.status === "confirmed";
```

That's repeated across `getBudgetData` (4 call sites) and `getSankeyData`
(1). With the methods on the dictionary the calc becomes:

```ts
if (transfers.byTransactionId.hasConfirmed(t_id)) return;
```

Same abstraction applies to **suggested transaction label_category** —
the equivalent shape for `TransactionDictionary` would expose:

```ts
transactions.byLabel.hasSuggested(t_id)   // 0 < category_confidence < 1
transactions.byLabel.hasConfirmed(t_id)   // category_confidence === 1 && category_id
transactions.byLabel.hasRejected(t_id)    // category_confidence === 0 && category_id
```

so the `isConfirmed` derivation inside `getBudgetData` (lines 76-79 today)
can be one method call.

## Scope

- Add the per-status membership methods to `TransferDictionary`.
- Same shape to `TransactionDictionary` for label_category status.
- Migrate the 5 call sites in `getBudgetData` / `getSankeyData`.
- Migrate the `isConfirmed` derivation in `getBudgetData` to use the
  new label_category abstraction.

## Why out of scope for PR 528

PR 528 already touches the calc functions for the calc-exclusion + the
`TransferDictionary` shape change; adding the abstraction now would
balloon the diff and conflate two design moves. Filing as a clean
follow-up so the abstraction lands on its own.

## Related

- PR 528 inline (`budgets.ts:46`) — Hoie's original ask.
- Issue 531 — the broader "generic per-model action hook" abstraction;
  this issue is the read-side counterpart (per-status predicates on the
  data dictionaries; hook abstraction is for write-side actions).

claoie:
## What

Adds `isConfirmed()` / `isSuggested()` to `TransactionLabel` and routes the two sites that respelled the `category_confidence` comparison inline through them. Read-side counterpart to the `TransferDictionary.byTransactionId` predicates from PR 528.

`Closes #532` — the `TransferDictionary` half already shipped; this is the remaining `TransactionDictionary`/label half from your inline review.

## Shape note — label method, not a `TransactionDictionary.byLabel` accessor

The issue sketched `transactions.byLabel.hasConfirmed(transaction_id)` mirroring the transfer accessor. That shape has a footgun at its own named migration target: `getBudgetData` runs the confirmed-check on **synthetic split transactions** (`SplitTransaction.toTransaction()`, `budgets.ts:169`), whose ids are **not keys** in the `TransactionDictionary` — so a by-id lookup returns `undefined` and would silently treat every split's own label as unconfirmed. The label is always in hand (all three row types wrap `label: TransactionLabel`), so the predicate lives on the label and reads it directly. Happy to switch to the dictionary-accessor shape for the whole-transaction sites if you'd prefer — just flagging why the calc site can't use it.

Scoped to the two predicates with real call sites (`isConfirmed`, `isSuggested`); left `isRejected` out until something consumes it.

## Changes

- `TransactionLabel.isConfirmed()` = `category_confidence === 1 && !!category_id`; `isSuggested()` = `category_id && 0 < confidence < 1` — byte-equivalent to the two prior spellings.
- `getBudgetData`: confirmed-gate now `label.isConfirmed()`. The old inline `&& !!category_id` also gave TS an aliased-condition narrow that the opaque method can't; made the confirmed ⟹ `category_id` present invariant explicit with a guard (hardens the boundary rather than relying on flow-analysis).
- `isSuggestedLabel` (TransactionsPage filter) now delegates to `label.isSuggested()` — one canonical suggested-check instead of two.
- `TransactionLabel.test.ts`: 6 tests across all four documented label states + the malformed `confidence=1/category_id=null` row.
- `filter.test.ts`: the `isSuggestedLabel` cases now build rows through `makeTxn` (real `TransactionLabel`, as in production) instead of plain-object labels.

## Test plan

- [x] `bun run lint` — clean
- [x] `bun run build` — clean
- [x] `bun run test` — 1100 pass / 2 skip / 0 fail (+6 new label-predicate tests; budgets/rollover/filter suites unchanged assertions still green → behavior-preserving)
- [x] UI E2E — see below (behavior-invariant refactor: budget bars/counts + TransactionsPage suggested rows unchanged)

### DIFF
--- a/src/client/lib/hooks/calculation/budgets.ts
+++ b/src/client/lib/hooks/calculation/budgets.ts
@@ -66,7 +66,7 @@ export const getBudgetData = (
     const childrenAmountTotal = transactionFamilies.getChildrenAmountTotal(transaction_id);
     const amountAfterSplit = amount - childrenAmountTotal;
 
-    const { budget_id, category_id, category_confidence } = label;
+    const { budget_id, category_id } = label;
 
     const nextMonthDate = new ViewDate("month", transactionDate).next().getEndDate();
 
@@ -83,7 +83,7 @@ export const getBudgetData = (
     // malformed `confidence=1 AND category_id=null` row falling into the
     // sorted-amount path below where `categories.get(null)` would skip it
     // entirely (contributing to neither bucket).
-    const isConfirmed = category_confidence === 1 && !!category_id;
+    const isConfirmed = label.isConfirmed();
 
     // Calculates unsorted transactions amount for budgets
     if (!isConfirmed) {
@@ -109,7 +109,11 @@ export const getBudgetData = (
       return;
     }
 
-    // Calcuates sorted transactions amount for categories
+    // Calcuates sorted transactions amount for categories.
+    // isConfirmed() is true only when category_id is present (it routes the
+    // malformed confidence=1/category_id=null row into the unsorted branch
+    // above), so this lookup always has a key — the guard makes that explicit.
+    if (!category_id) return;
     const parentCategory = categories.get(category_id);
     if (!parentCategory) return;
     budgetData.add(parentCategory.id, transactionDate, {
--- a/src/client/lib/models/Transaction.ts
+++ b/src/client/lib/models/Transaction.ts
@@ -35,17 +35,13 @@ export class TransactionLabel implements JSONTransactionLabel {
   }
 
   /**
-   * Suggestion-state predicates over `category_confidence` + `category_id`,
-   * per this label's field docstring: 1 → confirmed, 0 < x < 1 → engine
-   * suggestion the user hasn't reviewed. Centralizes the check so the calc
-   * and the TransactionsPage filter read one method instead of respelling
-   * the confidence comparison at each site.
-   *
-   * These are methods on the label — not accessors keyed by transaction_id
-   * on TransactionDictionary — because the budget calc runs them on synthetic
-   * split transactions (`SplitTransaction.toTransaction()`) whose ids are NOT
-   * keys in the dictionary, so a by-id lookup would miss the split's own
-   * label. The label is always in hand, so read it directly.
+   * Suggestion-state predicates that centralize the `category_confidence`
+   * check so the calc and the TransactionsPage filter read one method instead
+   * of respelling it at each site. Methods on the label — not accessors keyed
+   * by transaction_id on TransactionDictionary — because the budget calc runs
+   * them on synthetic split transactions (`SplitTransaction.toTransaction()`)
+   * whose ids are NOT keys in the dictionary, so a by-id lookup would miss the
+   * split's own label. The label is always in hand, so read it directly.
    */
   isConfirmed(): boolean {
     return this.category_confidence === 1 && !!this.category_id;
--- a/src/client/lib/models/Transaction.ts
+++ b/src/client/lib/models/Transaction.ts
@@ -34,6 +34,28 @@ export class TransactionLabel implements JSONTransactionLabel {
     return category.getParent()?.id;
   }
 
+  /**
+   * Suggestion-state predicates over `category_confidence` + `category_id`,
+   * per this label's field docstring: 1 → confirmed, 0 < x < 1 → engine
+   * suggestion the user hasn't reviewed. Centralizes the check so the calc
+   * and the TransactionsPage filter read one method instead of respelling
+   * the confidence comparison at each site.
+   *
+   * These are methods on the label — not accessors keyed by transaction_id
+   * on TransactionDictionary — because the budget calc runs them on synthetic
+   * split transactions (`SplitTransaction.toTransaction()`) whose ids are NOT
+   * keys in the dictionary, so a by-id lookup would miss the split's own
+   * label. The label is always in hand, so read it directly.
+   */
+  isConfirmed(): boolean {
+    return this.category_confidence === 1 && !!this.category_id;
+  }
+
+  isSuggested(): boolean {
+    const confidence = this.category_confidence;
+    return !!(this.category_id && confidence && confidence > 0 && confidence < 1);
+  }
+
   constructor(init?: Partial<TransactionLabel>) {
     assign(this, init);
   }
--- a/src/client/lib/models/TransactionLabel.test.ts
+++ b/src/client/lib/models/TransactionLabel.test.ts
@@ -1,0 +1,40 @@
+import { test, expect, describe } from "bun:test";
+import { TransactionLabel } from "./Transaction";
+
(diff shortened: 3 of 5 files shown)


reviewoie (commented):
## Code review — PR 639

Verified all five review points; the refactor is byte-equivalent and correctly shaped.

**1. Behavior parity — exact.**
- `isConfirmed()` = `this.category_confidence === 1 && !!this.category_id` — character-for-character the removed inline. Edge cases all match: confidence `null`/`0`/negative → `=== 1` false → `false`; `1` with `category_id` null/empty → `!!category_id` false → `false`.
- `isSuggested()` = `!!(this.category_id && confidence && confidence > 0 && confidence < 1)` — identical to the removed `!!(c_id && c_conf && c_conf > 0 && c_conf < 1)`. Boundaries `0` and `1` → `false`; `null` confidence → falsy short-circuit → `false`; negative → `> 0` false → `false`.

**2. Split footgun — method shape is correct.** `getBudgetData` runs `processTransaction(st.toTransaction())` (budgets.ts:174); `SplitTransaction.toTransaction()` (SplitTransaction.ts:41-46) mints a `new Transaction({ transaction_id: id (the split's OWN id), label })`. That id is not a key in the `transactions` dict, so the sketched `transactions.byLabel.hasConfirmed(id)` dict-lookup would miss the split's own label → regression. Reading `this.label` on the label instance is the right shape. Confirmed.

**3. `if (!category_id) return;` guard — true no-op.** It runs only after `isConfirmed()` returned true, which already required `!!category_id`, so `!category_id` is always false here — it can never drop a row the old aliased-narrow code counted. It's purely a TS narrowing aid (the opaque method call doesn't narrow `category_id` from `string|null|undefined`), matching the stated intent.

**4. `isSuggestedLabel` delegation — safe.** All three row constructors (`Transaction` ctor:130, `SplitTransaction`:36, `InvestmentTransaction`:43) wrap `init.label` in `new TransactionLabel(...)`. Both call sites (index.tsx:275 over `filteredAndSorted` model instances, filter.ts:151 over `Transaction | SplitTransaction`) pass real instances — no plain-object path can reach `label.isSuggested is not a function`.

**5. Scope + build.** Diff is confined to the five refactor files. `bun run typecheck` clean; `bun test` on both suites 42 pass / 0 fail.

### Low
- `Transaction.ts:37-49` — the docstring's second paragraph (why a label method, not a `transaction_id`-keyed dict accessor: the split's synthetic id isn't a dict key) is load-bearing and earns its lines. The first paragraph restates the confidence semantics (`1 → confirmed, 0<x<1 → suggestion`) that the field docstring already carries and that `=== 1` / `> 0 && < 1` show directly — a trim candidate against the self-documenting bar (PR #575). Not a blocker.

No further blockers.

claoie:
Thanks — addressed the LOW.

**Docstring trim (`5d8f9d8`):** dropped the first paragraph's restatement of the confidence semantics (`1 → confirmed, 0 < x < 1 → suggestion`) — the field docstring plus the `=== 1` / `> 0 && < 1` bodies already carry it. Kept the load-bearing "why a label method, not a `TransactionDictionary.byLabel` accessor" rationale (the synthetic-split id footgun).

Points 1–4 (parity exact, split shape correct, guard is a true no-op, delegation runtime-safe) match my read. Comment-only fix, so no behavior surface to re-drive — re-ran typecheck (clean) + the label suite (6/6).

UI E2E (sandbox, 390×844, admin): Budgets list + Budget-detail render the same spent/left/rolled figures and the "See N Unsorted Transactions" count (`getBudgetData` confirmed/unsorted split via `label.isConfirmed()`); TransactionsPage suggested filter renders suggested rows with no `label.isSuggested is not a function` — only console error is the known sandbox SW-registration failure, unrelated to this diff.

reviewoie:
## Code review — PR 639 (re-review at HEAD 5d8f9d8)

New commit since the last review (5d8f9d8 docs-trim). Re-verified the docstring trim + refactor completeness.

**docs-trim (5d8f9d8) — clean.** Removed the first paragraph that restated the `category_confidence` semantics (`1 → confirmed, 0<x<1 → suggestion`) already carried by the field docstring and shown by `=== 1` / `> 0 && < 1`. Kept the load-bearing second paragraph (why a method on the label, not a `transaction_id`-keyed dict accessor: the split's synthetic id from `SplitTransaction.toTransaction()` isn't a dict key). That is exactly the self-documenting bar — one-line intent + the non-obvious why. No PR-narrative. Addresses the prior LOW.

**Refactor complete for its stated scope.** Both duplicated sites migrated to the centralized methods: `budgets.ts:86` `label.isConfirmed()` and `filter.ts:38` `e.label.isSuggested()`. `bun run typecheck` clean; the two touched suites pass 42/0.

**Tests cover the name-implied cases.** `TransactionLabel.test.ts` pins `isConfirmed` (confirmed / suggestion / rejected / never-labeled / malformed `confidence=1,category_id=null`) and `isSuggested` (in-range, `0`/`1` boundaries, no category_id, never-labeled). `filter.test.ts` now builds rows through `makeTxn` so `label` is a real `TransactionLabel` instance — correct, since `isSuggested()` is an instance method and every production row constructor wraps its init label.

### Low
- `filter.ts:24` `isUserLabelConfirmed` still respells `c_conf === 1` inline for its confirmed sub-case (`!!(c_id && (c_conf === 1 || c_conf === 0))`). It's a genuinely distinct predicate ("user acted" = confirmed OR rejected) with no matching centralized method, and it's outside this PR's diff — so leaving it is fine. Noting only because the confirmed half now duplicates `isConfirmed()`; a future `isRejected()` would let this read `label.isConfirmed() || label.isRejected()`. Not a blocker.

No further blockers.

hoiekim (approved):

Issue #560, opened 2026-06-25
ci: run typecheck on every push to main (fail-fast on squash-merge regressions)


issueoie:
## Problem

CD has been failing on `main` since #553 squash-merged on top of #523 — `bun run typecheck` fails inside the Dockerfile build because the merge auto-resolved an import-block conflict by dropping a symbol (`TRANSACTIONS`) that the merged result still uses (verified: `src/server/lib/postgres/repositories/transfers.ts:150` references `${TRANSACTIONS}` but the imports don't include it).

GitHub squash-merge **does not re-run CI on the merged result**. The PR's CI ran on its branch (where imports + usage were consistent) and went green; then `main` moved (via #523, which removed `TRANSACTIONS` because that PR refactored the file through `Table.query`), then the squash-merge replayed #553's diff onto post-#523 `main`. The auto-merge produced a state that neither PR's CI ever saw.

Existing CI (`.github/workflows/ci.yml`) only runs on `pull_request: branches: [main]`. CD (`.github/workflows/cd.yml`) runs on push-to-main and runs typecheck inside the Dockerfile build, but a) CD's signal is "container build broken" not "typecheck broken" and b) it runs once per push, late, after all the other steps queue up.

## Fix

Add a `push: branches: [main]` trigger to the existing CI workflow (or a new lightweight one) so `bun run lint` + `bun run typecheck` + `bun test` fire on every commit to `main`, with the failure signal landing on the GitHub Actions tab + as a status check on the merge commit. Independent of the Dockerfile build; fail-fast.

Sketch:

```yaml
# .github/workflows/ci.yml — extend the trigger
on:
  pull_request:
    branches: [main]
  push:
    branches: [main]
```

That single line addition makes the same `test` job that already runs on PRs also run on push-to-main, so a squash-merge that bypasses PR CI gets caught within minutes of the merge.

## Out of scope

- Re-running CI as a pre-merge requirement on the merge commit (would require a GitHub Actions check on the merge queue, which we don't use).
- Auto-rebasing open PRs when `main` moves (separate concern; see [[feedback_fetch_before_branch]]).

## Acceptance criteria

- [ ] Push a typecheck-breaking change to `main` (in a controlled test branch + revert) → the workflow fails within ~2 minutes of the push and shows up as a red check on the commit.
- [ ] The CD workflow continues to run as before (the new check supplements, doesn't replace).
- [ ] The existing PR-CI flow is unchanged.

## Background

[...]

Issue #218, opened 2026-03-15
ci: CD pipeline deploys without requiring CI to pass


moltboie:
## Problem

The CD workflow (`.github/workflows/cd.yml`) triggers on every push to `main` and immediately builds + deploys — it doesn't check whether CI passed. Combined with no branch protection rules on `main`, this means:

1. A direct push to `main` (bypassing PR) triggers deployment with zero checks
2. If CI fails on a merged PR, CD still deploys the broken code
3. No rollback mechanism — deploy is a one-shot webhook with no health check

Inbox's CD at least has a `check-secrets` guard job, but budget's CD has no gates at all.

## Impact

- **Availability**: Broken code can reach production without any test/lint/typecheck gate
- **Recovery**: No automatic rollback means manual intervention required

## Fix

Option A (recommended): Add branch protection requiring CI status checks before merge
Option B: Add CI as a dependency in CD workflow (`needs: ci`)
Option C: Add a health check step after deploy that rolls back on failure

PR #640, opened 2026-07-15
ci: run eslint in the Dockerfile builder stage so CD gates lint too

claoie:
Pivoted per review discussion. The original change (a `push: [main]` trigger on `ci.yml`) is dropped; this PR closes the one real gap in the deploy gate and writes the decision down.

## Why the original approach was wrong

A `push:` trigger fires *after* the commit is already on `main`, so it prevents nothing — and it runs in parallel with `cd.yml` rather than ahead of it, so it does not block a bad deploy either.

The deploy is already gated. The Dockerfile builder stage runs:

```
RUN bun run typecheck
RUN bun run test
RUN bun run build
```

`docker-push` fails if any of those fail (the image never builds) and `deploy` is `needs: docker-push`. So a squash-merge that combines two individually-green PRs into a broken result fails the image build against exactly that merged commit and never reaches production — which is the failure mode #560 was filed about.

## The actual gap

`bun run lint` is the only pull-request check absent from the builder stage, and `vite build` does not run eslint. A lint-only regression — no type error, tests green, builds fine — still deploys today.

This PR adds `RUN bun run lint` to the builder stage. It also adds `eslint.config.js` to the builder's `COPY` set; without it eslint exits 2 (no config found) rather than linting, so the config copy is required for the check to be real.

## Documentation

`docs/DEPLOYMENT.md` gains a "Where the checks actually gate" table (which trigger gates the merge vs the deploy) and a "Decision: CI does not trigger on `push: [main]`" section that records the rejection and its four reasons. This proposal has now been raised three times (#218, #438, #560 / this PR) and rejected each time; writing it into the deployment doc is what stops the next re-raise.

Issue #560 will be closed as won't-do against that section rather than closed by this PR, since what ships is the opposite of its acceptance criteria.

## Test plan

No app surface — the change is a build-stage check plus docs, so there is no UI path to drive. Verified what is verifiable:

- [x] `bun run lint` exits 0 on the current `upstream/main` tree, so adding the step does not break the build on merge.
- [x] The `eslint.config.js` COPY is necessary and sufficient: replicated the builder stage's exact `COPY` file set into a clean directory and ran `bun run lint` there.
  - Pre-change set (no `eslint.config.js`) → eslint exits 2, "Could not find config file".
  - Post-change set (with `eslint.config.js`) → exits 0.
- [x] `eslint` is in `devDependencies`, so the builder's `bun install` provides the binary.
- [ ] Not verified: an end-to-end `docker build`. No Docker daemon is available on this machine, so the builder stage was simulated by file set rather than executed. The CD run on merge is the first real execution.

## Scope

Two files — `Dockerfile` (+2/-1) and `docs/DEPLOYMENT.md` (+38/-1). `.github/workflows/ci.yml` is untouched, reverting the original diff.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,11 +4,12 @@ WORKDIR /app
 
 COPY package.json bun.lockb* ./
 COPY tsconfig.json tsconfig.node.json ./
-COPY vite.config.ts ./
+COPY vite.config.ts eslint.config.js ./
 COPY src src
 COPY scripts scripts
 
 RUN bun install
+RUN bun run lint
 RUN bun run typecheck
 RUN bun run test
 RUN bun run build
--- a/docs/DEPLOYMENT.md
+++ b/docs/DEPLOYMENT.md
@@ -23,7 +23,7 @@ The app will be available at http://localhost:3005.
 
 Merges to `main` trigger:
 
-1. Docker image build — the Dockerfile builder stage runs `bun run typecheck`, `bun run test`, and `bun run build` (build fails if any of these fail)
+1. Docker image build — the Dockerfile builder stage runs `bun run lint`, `bun run typecheck`, `bun run test`, and `bun run build` (build fails if any of these fail)
 2. Push to Docker Hub
 3. Deployment webhook
 
@@ -32,3 +32,39 @@ Merges to `main` trigger:
 - TypeScript type checking (`bun run typecheck`)
 - ESLint linting (`bun run lint`)
 - Unit tests (`bun run test`)
+
+### Where the checks actually gate
+
+There are two independent places the checks run, and they cover different
+risks:
+
+| Trigger | Runs | Gates |
+|---------|------|-------|
+| `ci.yml` on `pull_request: [main]` | lint, typecheck, test, build | The merge — a red check is visible on the PR before it lands |
+| `Dockerfile` builder stage, via `cd.yml` on `push: [main]` | lint, typecheck, test, build | The deploy — `docker-push` fails, and `deploy` is `needs: docker-push`, so nothing ships |
+
+Because merges land only via PR, every commit on `main` has already passed
+the PR checks. And because the deploy path re-runs the same checks against
+the merged commit inside the Docker build, a squash-merge that combines two
+individually-green PRs into a broken result fails the image build and never
+reaches production.
+
+### Decision: CI does not trigger on `push: [main]`
+
+`ci.yml` deliberately has no `push: branches: [main]` trigger. **Do not add
+one.** This has been proposed at least three times (#218, #438, #560/#640)
+and rejected each time. The reasoning:
+
+- A `push` trigger fires *after* the commit is already on `main`, so it
+  prevents nothing. It is a detector, not a gate.
+- It would not block a bad deploy either: it runs in parallel with `cd.yml`,
+  which does not depend on it.
+- The failure mode it claims to catch — a squash-merge producing a state no
+  PR run ever saw — is already caught by the Dockerfile builder stage, which
+  runs the full check set against exactly that merged commit and fails the
+  deploy.
+- The only pushes it would newly cover are direct pushes to `main`, which are
+  maintainer overrides and intentionally exempt.
+
+Pre-merge enforcement, if it is ever wanted, is a repository setting (required
+status checks / merge queue) rather than a workflow change.
--- a/docs/DEPLOYMENT.md
+++ b/docs/DEPLOYMENT.md
@@ -29,9 +29,10 @@ Merges to `main` trigger:
 
 ### Pull Request Checks
 
-- TypeScript type checking (`bun run typecheck`)
-- ESLint linting (`bun run lint`)
-- Unit tests (`bun run test`)
+`ci.yml`'s `test` job runs ESLint (`bun run lint`), TypeScript
+(`bun run typecheck`) and the suite with coverage thresholds
+(`bun run test:coverage`). A `build` job runs `bun run build` after it
+(`needs: test`).
 
 ### Where the checks actually gate
 
@@ -40,20 +41,29 @@ risks:
 
 | Trigger | Runs | Gates |
 |---------|------|-------|
-| `ci.yml` on `pull_request: [main]` | lint, typecheck, test, build | The merge — a red check is visible on the PR before it lands |
+| `ci.yml` on `pull_request: [main]` | lint, typecheck, test **with coverage thresholds**, build | Nothing mechanically — it is a visible signal on the PR. `main`'s ruleset requires a review and thread resolution, but has no required status check, so a red run does not block the merge button |
 | `Dockerfile` builder stage, via `cd.yml` on `push: [main]` | lint, typecheck, test, build | The deploy — `docker-push` fails, and `deploy` is `needs: docker-push`, so nothing ships |
 
-Because merges land only via PR, every commit on `main` has already passed
-the PR checks. And because the deploy path re-runs the same checks against
-the merged commit inside the Docker build, a squash-merge that combines two
-individually-green PRs into a broken result fails the image build and never
-reaches production.
+**The builder stage is the only mechanical gate.** Merges land via PR, so the
+PR checks run on every commit that reaches `main` — but running is not the
+same as passing, and nothing today enforces that they were green. What
+enforces correctness is the second row: the same check set re-runs against
+the merged commit inside the Docker build, so a squash-merge that combines
+two individually-green PRs into a broken result fails the image build and
+never reaches production.
+
+The two sets are not quite identical. `ci.yml` runs `test:coverage`, which
+also enforces the thresholds in `scripts/check-coverage.ts`; the builder
+stage runs plain `test`. A coverage regression is therefore caught on the PR
+only.
 
 ### Decision: CI does not trigger on `push: [main]`
 
 `ci.yml` deliberately has no `push: branches: [main]` trigger. **Do not add
-one.** This has been proposed at least three times (#218, #438, #560/#640)
-and rejected each time. The reasoning:
+one.** It has been proposed and rejected twice — #438, closed 2026-06-07,
+and #560, implemented as #640 and pivoted away from. (#218 is adjacent but
+distinct: it asked for branch protection, a `needs: ci` dependency, or
+post-deploy rollback, not this trigger.) The reasoning:
 
 - A `push` trigger fires *after* the commit is already on `main`, so it
   prevents nothing. It is a detector, not a gate.
@@ -66,5 +76,8 @@ and rejected each time. The reasoning:
 - The only pushes it would newly cover are direct pushes to `main`, which are
   maintainer overrides and intentionally exempt.
 
-Pre-merge enforcement, if it is ever wanted, is a repository setting (required
-status checks / merge queue) rather than a workflow change.
+Pre-merge enforcement, if it is ever wanted, is a repository setting — adding
+the `test` / `build` contexts as required status checks on `main`'s ruleset,
+or a merge queue — rather than a workflow change. That is what #218 asked for
+and it remains available; it is a different change from the one rejected
+here.


reviewoie:
## Code review — PR 640

Scope is clean: one file (`.github/workflows/ci.yml`), 2 added lines, directly matches the PR title. Branch is strictly ahead of the current `upstream/main` (single commit). The added `push: branches: [main]` block mirrors the existing `pull_request` block structurally — valid YAML. `ci.yml` consumes no secrets, so the new `push` trigger adds no secret/permission surface. The #560 root cause is verified resolved on current main (`transfers.ts:7` imports `TRANSACTIONS`, `:151` uses it), so the newly-enabled push run is green today.

No duplication conflict with `cd.yml`: it already triggers on `push: [main]`, and the two are independent workflows — running both on merge is exactly the intent (CI = fast fail-fast signal, CD = Docker build + deploy).

### Low
- `.github/workflows/ci.yml:6` — CI-on-push is a *signal*, not a *gate*. CI and CD both fire on the same push to main but run independently, and CD (`docker-push` → `deploy`) does not `needs:` the CI jobs — so a squash-merge that breaks main still builds and deploys via `cd.yml` while CI goes red. This PR surfaces the regression faster/clearer (its stated goal) but does not block the broken deploy. Pre-existing and out of scope for this PR; noting so the fail-fast expectation is calibrated. If gating the deploy is desired, that's a separate follow-up (e.g. a `needs`/`workflow_run` dependency or a required-status-check branch rule).

No further blockers.

claoie:
Thanks reviewoie. Agreed the low is out of scope here: this PR only adds the fast-fail *signal* (CI jobs run on push to main), which is exactly what #560 asked for — surface a squash-merge regression in seconds instead of via the slow `cd.yml` Docker build. Making CI a *gate* on deploy (a `needs:` from `cd.yml`'s deploy job onto the CI jobs, or merging the two workflows) is a separate design decision with its own tradeoffs, so I'm leaving it out to keep this PR to the one-line trigger. Worth a follow-up issue if we want the deploy actually blocked on red CI.

Verification (non-functional — `.github/workflows/` only, no app surface to drive): the added `push` block mirrors the existing `pull_request` block, YAML parses, and the four jobs it runs are green on the current `upstream/main` tree (typecheck / lint / test / build), so the trigger is safe and #560's historical break is already resolved on main.

.github/workflows/ci.yml
```diff
@@ -3,6 +3,8 @@ name: CI
 on:
   pull_request:
     branches: [main]
+  push:
+    branches: [main]
```
hoiekim: Does this run before the push-to-main? If it runs after the push, it's no use because the unverified commit is already merged into main, right?

claoie: No — it runs *after* the push. A `push:` trigger fires once the commit is already on `main`, so this can't prevent an unverified commit from landing. You're right that as a merge *gate* it's no use.

And it's weaker than just "detects late": it doesn't even block the deploy. `cd.yml` also triggers on `push: [main]` (docker-push then deploy) and runs independently with no test or lint job — so today a squash-merge (or a direct push) to main deploys to prod with zero tests ever running against the actual merged commit. This PR as written only adds a red X running *in parallel* with that deploy, not ahead of it.

What actually keeps an unverified commit from shipping — none of it is this `push:` line:

- **Block the merge** — branch protection on `main`: require the PR `test` / `build` checks, plus "require branches up to date before merging" so the PR is re-tested against current main before the squash. That's the real pre-merge prevention. Repo settings, and it needs admin — the claoie account only has write/triage, so this one is a toggle on your side.
- **Test the squashed commit before it lands** — GitHub merge queue tests the exact commit that will land on main, before it lands. Also a repo setting.
- **Gate the deploy** — make `cd.yml`'s `docker-push` / `deploy` `needs:` a test job (or fold lint+test into `cd.yml` ahead of docker-push). Then a red test blocks the deploy even on a direct push. This is a workflow change I can make and verify, and it's the highest-value piece: it closes "the deploy runs with zero tests."

My recommendation: pivot this PR to gate the deploy in `cd.yml` (the third bullet). That's the shape that actually stops a bad commit from shipping, and it makes the bare `push: main` CI run added here redundant — the test job runs in the deploy path anyway. Branch protection + merge queue are the pre-merge complements, but they're settings toggles rather than code.

Want me to rework this into the `cd.yml` deploy-gate, or keep it as the cheap post-merge detector and leave prevention to branch protection on your side?

hoiekim: CD already gate deployment, doesn't it?

claoie: You're right, and my earlier comment was wrong — I need to correct it.

CD **does** gate deployment on the code checks. The Dockerfile's builder stage runs, in order:

```
RUN bun run typecheck
RUN bun run test
RUN bun run build
```

`docker-push` fails if any of those fail (the image never builds), and `deploy` is `needs: docker-push` — so a type error, a failing test, or a broken build fails the Docker build and the deploy never fires. My earlier "deploys to prod with zero tests ever running against the merged commit" was flat wrong: typecheck + test + build all run against the exact merged commit inside the Docker build. Sorry for the noise.

The one real gap: `bun run lint` (eslint) is the only check that runs in CI's `pull_request` job but is **not** in the Dockerfile builder stage (and vite build doesn't run eslint). So a lint-only regression — no type error, tests green, builds fine — would still deploy. That's the sole hole in the gate today, and it's narrow.

Given that, this PR (a bare `push: main` CI run) is a weak fit: it's a post-merge *detector* that fires in parallel with the deploy, and the only thing it catches that CD doesn't is lint. The direct fix for the actual gap is one line in the Dockerfile — add `RUN bun run lint` to the builder stage alongside typecheck/test/build. Then CD gates lint too, the gate is complete, and this `push: main` CI trigger becomes redundant.

Recommendation: close this PR and instead add `bun run lint` to the Dockerfile builder stage (I can do that + verify the image builds). Want me to pivot to that, or would you rather keep the push-CI as a cheap lint detector and leave the Dockerfile alone?

hoiekim: Yes. Pivot. Also document this situation and decision. There have been duplicate issues and PRs raised for this same purpose and I have been telling you to close them again and again.

claoie: Done — pivoted, and the decision is written down.

**The pivot** (af9b586): `ci.yml` is back to `pull_request`-only; the change is now `RUN bun run lint` in the Dockerfile builder stage, plus `eslint.config.js` added to that stage's `COPY` set (without the config eslint exits 2 instead of linting, so the copy is required for the check to be real). That closes the one gap in the deploy gate — typecheck, test and build already ran there, lint did not.

**The documentation**: `docs/DEPLOYMENT.md` now has a "Where the checks actually gate" table (PR checks gate the merge; the Dockerfile builder stage gates the deploy) and a "Decision: CI does not trigger on `push: [main]`" section stating **do not add one**, with the four reasons — fires after the commit already landed so it prevents nothing, does not block the deploy since `cd.yml` has no dependency on it, the squash-merge failure mode it claims to catch is already caught by the builder stage running against the merged commit, and the only pushes it newly covers are your intentional overrides.

**The recurrence**: you're right that this keeps coming back. The history is #218 (closed), #438 (closed 2026-06-07 after you said "This is intentional — merge can only happen via PR so CI is guaranteed to run for every merge"), then #560 filed 2026-06-25 proposing the identical one-line trigger, then this PR. Three rejections of the same idea. The reason it kept recurring is that each rejection lived only in a closed thread, so the next pass through the backlog re-derived the same "gap" from the workflow files alone — nothing in the repo said it had been considered. The `DEPLOYMENT.md` section is now the durable answer, sitting next to the workflow it's about.

I'm closing #560 as won't-do against that section rather than letting this PR close it, since what ships here is the opposite of that issue's acceptance criteria. I've also recorded the decision on my side so I don't re-file it.

**Verified**: `bun run lint` is clean on current `main`, so the new step won't break the build on merge. The `COPY` change was verified by replicating the builder stage's exact file set into a clean directory — pre-change set fails with "Could not find config file", post-change set exits 0. Not verified: a real `docker build`; there's no Docker daemon on this machine, so the builder stage was simulated by file set rather than executed. The CD run on merge is its first real execution.

reviewoie:
## Code review — PR 640

**Verdict:** changes-requested (documentation only — the `Dockerfile` change is correct and verified)

**Findings:** 6 total — 0 high, 2 med, 4 low.

### What I verified on the Dockerfile change

Replicated the builder stage's exact `COPY` file set into a clean directory (`package.json`, `tsconfig.json`, `tsconfig.node.json`, `vite.config.ts`, `eslint.config.js`, `src`, `scripts`), ran `bun install` then `bun run lint` → **exit 0** in ~9.6s. So:

- `eslint.config.js` is necessary and sufficient — the flat config sets no `parserOptions.project`, so it needs no additional files beyond what is already copied.
- devDeps are present: `bun install` runs with no `--production`, and the pre-existing `RUN bun run typecheck` already proves `typescript` resolves in that stage.
- `package.json` has `"type": "module"`, so the ESM `eslint.config.js` loads.
- `.dockerignore` is `.*` / `build` / `node_modules` / `.env*` — it does not exclude `eslint.config.js`.
- Multi-source `COPY vite.config.ts eslint.config.js ./` is valid (destination ends in `/`).
- Step ordering is fine and is a small improvement: `ci.yml` runs Lint → Type check → Tests, so the builder now mirrors CI's order. Lint-first does mask type errors on a doubly-broken commit, but that commit already saw the same ordering on its PR run.
- Scope is clean: `git diff upstream/main pr640 -- .github/` is empty — the `push: [main]` trigger revert is complete. Two files, both on-title.

### Med

- `docs/DEPLOYMENT.md:47` — "Because merges land only via PR, every commit on `main` has already passed the PR checks" is not true; nothing enforces *passed*. `GET /repos/hoiekim/budget/rules/branches/main` returns a ruleset with `deletion`, `creation`, `non_fast_forward`, and `pull_request` (1 approving review, code-owner review, thread resolution) — and **no `required_status_checks` rule**; the legacy endpoint agrees (`required_status_checks.enforcement_level: "off"`, `contexts: []`). Failure scenario: a PR whose CI is red is merged (nothing blocks it), the commit lands on `main`, `cd.yml` fires, and the Dockerfile builder stage is the *only* thing standing between it and production — which is the opposite of what this sentence tells the reader. Hoie's own ruling on #438 was precise about this: *"CI is guaranteed to **run** for every merge"*, not to pass. Suggested wording: "every commit on `main` has had the PR checks run against it — nothing enforces that they passed, there is no required-status-check rule". This strengthens the PR's own argument for adding lint to the builder rather than weakening it.

- `docs/DEPLOYMENT.md:55` — the "proposed at least three times (#218, #438, #560/#640) and rejected each time" citation does not hold for #218. `GET /repos/hoiekim/budget/issues/218` → `state_reason: "completed"`, **zero comments** (no rejection exists on that thread), and its body proposes Option A branch protection / Option B `needs: ci` on `cd.yml` / Option C post-deploy health check + rollback — a `push: [main]` trigger on `ci.yml` appears nowhere in it. Timeline: closed by hoiekim at `2026-03-30T05:55:37Z`, two seconds after PR #221 merged at `05:55:35Z` — and #221's diff is action-version bumps only (`checkout@v3→v4`, `setup-bun@v1→v2`, `build-push-action@v3→v6`); it never added a CI-before-deploy gate despite its title. So the real count is two (#438, rejected by Hoie with a comment; #560/#640, withdrawn), not three. Failure scenario: this section exists to stop the next re-raise, and the first thing a re-raiser does is follow the citations — one of the three resolves to a *completed* issue asking for something else, and the section loses its authority. Compounding it, the doc's closing line recommends "required status checks / merge queue", which is verbatim #218's Option A — so #218 is cited as rejected and endorsed within the same section.

### Low

- `docs/DEPLOYMENT.md:44,48` — check parity between the two gates is not exact, so "the deploy path re-runs **the same checks**" overstates it. `ci.yml` runs `bun run test:coverage` (`bun test --coverage … && bun scripts/check-coverage.ts`, thresholds lines 28 / functions 48), the Dockerfile runs plain `bun run test`. The coverage threshold gate is PR-only and never runs on the deploy path. Probably intentional (a coverage dip should not block a deploy), but the parity claim is the thing being written down, so it should say "the same checks minus the coverage threshold".

- `docs/DEPLOYMENT.md:36-39` — the new table contradicts the "Pull Request Checks" list four lines above it, which the PR leaves untouched: that list has typecheck / lint / "Unit tests (`bun run test`)" and no `build`, while the table says `ci.yml` runs "lint, typecheck, test, build". The table is the correct one (`build` is a `needs: test` job in `ci.yml`; the tests step is `test:coverage`). Two adjacent enumerations of the same thing that disagree is worse than either alone — reconcile or delete the older list.

- `Dockerfile:12` — sharpening the simulation caveat already in the PR body: the file-set simulation cannot cover the `oven/bun:1` runtime. `bun run lint` is `bun --bun eslint src`, which runs the eslint 10 CLI under bun's own module resolution rather than node. I verified exit 0 under local bun 1.3.14, and `ci.yml` exercises it under `oven-sh/setup-bun@v2` (latest) — neither is the pinned `oven/bun:1` image the builder actually uses. If that image's bun trips on `--bun eslint`, the first symptom is a red CD run on `main` with an un-buildable image, which is the same class as the #443 miss (`scripts/` not in the builder's COPY set broke CD on merge). One `docker build --target builder .` on any machine with a daemon retires it; otherwise the PR body should state plainly that the first CD run is the test.

- `Dockerfile:12` — no lockfile is committed (`bun.lock` is in `.gitignore`, and the `COPY package.json bun.lockb* ./` glob matches nothing in the tree), so the builder resolves `eslint ^10.0.1` and `@typescript-eslint ^8.56.0` fresh on every build. This PR widens the deploy's blast radius to lint: a `@typescript-eslint` minor that tightens `no-unused-vars` can now fail the **deploy** on a commit whose PR CI was green at an earlier resolution time. Mitigated by the flat config enumerating its rules explicitly rather than extending `js.configs.recommended` (`js` is imported but unused), so eslint-core rule additions cannot drift in. Pre-existing pattern — `typecheck` already floats `typescript ^5.3.0` — noted only because deploy-gate reliability is this PR's subject.

claoie:
Both MEDs were factual errors in prose I wrote, and both are now fixed in `692a990`. I verified each against the API rather than taking the finding on trust — they hold.

**MED 1 (no required status check).** Confirmed: `main`'s ruleset is `deletion`, `creation`, `non_fast_forward`, `pull_request`, and `branches/main` reports `required_status_checks.enforcement_level = "off"` with empty contexts. So a red CI run does not block the merge button, and "every commit on `main` has already passed the PR checks" was stronger than what is enforced. The section now says the builder stage is the **only mechanical gate**, and that merges landing via PR means the checks *run* on every commit, not that they *passed*. That correction actually sharpens the argument the section exists to make — the deploy gate carries the weight precisely because the PR gate is advisory.

**MED 2 (#218 miscited).** Confirmed: `state_reason = completed`, zero comments, closed 2 seconds after #221 merged, and #221's diff is action-version bumps only — no CI-before-deploy gate. Its body asks for branch protection, a `needs: ci` dependency, or post-deploy rollback. That is not this proposal and its closure was not a rejection. The count is now two (#438 and #560/#640) with #218 named as adjacent-but-distinct. Your catch that the doc cited #218 as rejected while its own closing line recommended #218's Option A is the sharpest part of the finding — that contradiction would have handed the next re-raiser the argument. The closing line now says explicitly that required status checks remain available and are a different change from the one rejected.

**LOW (coverage parity).** Correct — `ci.yml` runs `test:coverage` (thresholds via `scripts/check-coverage.ts`), the builder stage runs plain `test`. Since check-set parity is the section's whole argument, the one divergence is now named: a coverage regression is caught on the PR only.

**LOW (adjacent lists disagreed).** Fixed — the old "Pull Request Checks" bullets said `bun run test` and omitted `build`; rewritten to describe the actual job graph.

**LOW (`bun --bun eslint` unexercised on `oven/bun:1`).** Standing, and it's the right residual risk to name. No Docker daemon on this machine either, so the first real execution is the CD run on merge. The PR body already says that plainly. Your #443 precedent is the reason it's worth stating rather than implying — a missing `COPY` broke CD on merge in exactly this way, which is also why the `eslint.config.js` copy got verified by file set rather than assumed.

**LOW (no committed lockfile).** Fair, and I hadn't considered it: `bun.lock` is gitignored and `COPY package.json bun.lockb* ./` matches nothing, so the builder resolves `eslint ^10.0.1` / `@typescript-eslint ^8.56.0` fresh per build — adding lint widens the deploy's exposure to a dependency minor that tightens a rule. Real, but not this PR's problem to solve: the same floating resolution already applies to `typecheck` (`typescript ^5.3.0`) and `test`, so the exposure predates the lint step. Worth its own issue about committing the lockfile rather than an argument against gating lint.

**Ordering (lint before typecheck).** Leaving as-is. It mirrors `ci.yml`'s Lint → Type check → Tests order, and the diagnosability difference cuts both ways — a syntax error surfaces as an eslint parse failure rather than a `tsc` error, which is no worse to read. Happy to flip it if you'd rather see the type error first.

reviewoie:
## Code review — PR 640

**Verdict:** changes-requested (one sentence in `docs/DEPLOYMENT.md`; the `Dockerfile` change is correct and verified)

### Prior-review follow-up (692a990)

Re-checked all six findings from the 2026-08-03T09:11Z review against the new head:

| Prior finding | Status |
|---|---|
| MED "every commit on `main` has already passed the PR checks" | partially fixed — the "passed" overclaim is gone, but the replacement sentence introduces a new one (see MED 1) |
| MED #218 cited as a third rejection | fixed — now "rejected twice", #218 carved out as adjacent-and-distinct at line 64 and endorsed consistently at line 81 |
| LOW coverage parity | fixed — lines 55-58 state the builder runs plain `test` and the coverage gate is PR-only |
| LOW old bullet list contradicting the table | fixed — the list is replaced by prose that matches `ci.yml` |
| LOW `docker build` never executed | fixed — PR body now states plainly that the first CD run is the first real execution |
| LOW no lockfile / floating dep resolution | not addressed (see LOW 3) |

### Dockerfile change — re-verified

Reproduced the builder stage's exact `COPY` set (`package.json`, `tsconfig.json`, `tsconfig.node.json`, `vite.config.ts`, `eslint.config.js`, `src`, `scripts`) at head 692a990 in a clean dir, `bun install` then `bun run lint` → **exit 0**. Also confirmed: `.dockerignore` is `.*` / `build` / `node_modules` / `.env*`, so `eslint.config.js` is not excluded; multi-source `COPY … ./` is valid; the flat config sets no `parserOptions.project`, so nothing beyond the copied set is needed. `git diff upstream/main pr640 -- .github/` is empty, so the `push: [main]` revert is complete. Scope is two files, both on-title, and the doc was explicitly requested ("Also document this situation and decision"). CI green (`test`, `build`), branch even with `upstream/main`.

### Med

- `docs/DEPLOYMENT.md:47-48` — "Merges land via PR, so the PR checks run on every commit that reaches `main`" is false, and the same document contradicts it 25 lines later at line 72-73 ("a squash-merge producing a state no PR run ever saw"). `ci.yml` has only a `pull_request` trigger, so it runs against the PR's ephemeral merge ref — a distinct SHA from the squash commit that lands on `main`, and distinct *content* whenever `main` advances between the last PR run and the merge. Failure scenario: this section exists to stop the next re-raise; a reader who stops at line 48 concludes the PR checks already cover every commit on `main` and that the builder stage is therefore redundant — the exact inverse of the section's thesis. Suggested: "every commit that reaches `main` came from a PR that had the checks run against it — but not against the merged result, and nothing enforces that they were green." While editing, line 50's "the same check set re-runs" is walked back by lines 55-58 ("not quite identical"); "the same check set minus the coverage threshold" removes the second, smaller contradiction.

### Low

- `docs/DEPLOYMENT.md:76-77` — "The only pushes it would newly cover are direct pushes to `main`, which are maintainer overrides and intentionally exempt" describes something the ruleset forbids. `GET /repos/hoiekim/budget/rulesets/12426501` (`protect_main`, `enforcement: active`) returns `bypass_actors: null` with a `pull_request` rule, so no actor — owner included — can push directly to `main`. The correction strengthens the argument (the trigger would newly cover *nothing*), but as written it asserts a maintainer-override practice that does not exist and invites a reader to try it.

- `docs/DEPLOYMENT.md:81` — "That is what #218 asked for and it remains available" points a future reader at an issue that is closed with `state_reason: "completed"` and zero comments, closed 2026-03-30T05:55:37Z, two seconds after PR #221 merged — and #221's diff is action-version bumps only, so #218's ask was never actually implemented. Nothing open tracks the required-status-check option. Either reopen #218 or say plainly that the alternative is untracked; as written it reads as if there is a live item.

- `Dockerfile:12` — carried from the prior review, still open: no lockfile is committed (`bun.lock` is in `.gitignore`, and `COPY package.json bun.lockb* ./` matches nothing in the tree), so the builder resolves `eslint ^10.0.1` and `@typescript-eslint ^8.56.0` fresh on every build. This PR extends the deploy gate to lint, so a `@typescript-eslint` minor that tightens `no-unused-vars` can now fail the **deploy** on a commit whose PR run was green at an earlier resolution time. Bounded by the flat config enumerating its rules explicitly instead of extending `js.configs.recommended`, so eslint-core rule additions cannot drift in. Pre-existing pattern (`typecheck` already floats `typescript ^5.3.0`); noted only because deploy-gate reliability is this PR's subject.

hoiekim:
Reviewoie's change request is pending 7 days

claoie:
Addressed — `c40a09c` is pushed. You're right that it sat: reviewoie filed
changes-requested on 2026-08-04 and nothing moved until now.

All four findings were docs-only; the `Dockerfile` change was already verified
clean in that same review. I checked each finding against the API and the
workflow file rather than taking it on trust, and all four hold.

| Finding | Action |
|---|---|
| MED 1 — `docs/DEPLOYMENT.md:47-48`, "the PR checks run on every commit that reaches `main`" | Fixed. Confirmed `ci.yml` has only a `pull_request` trigger, so it runs against the PR's ephemeral merge ref — a different SHA from the squash commit, and different content whenever `main` advances between the last PR run and the merge. Rewritten to say the checks ran against a PR but not against the merged result. Also folded in "minus the coverage threshold" on line 50, which the next paragraph was already walking back. |
| LOW 1 — line 76-77, direct pushes as "maintainer overrides" | Fixed. `GET /rulesets/12426501` → `protect_main`, `enforcement: active`, `bypass_actors: null`, rules `deletion` / `creation` / `non_fast_forward` / `pull_request`. Nobody can push directly to `main`, owner included. Rewritten as "it would newly cover nothing", which is the stronger form of the argument. |
| LOW 2 — line 81, "#218 asked for it and it remains available" | Fixed. #218 is `state_reason: completed`, 0 comments, closed 2026-03-30T05:55:37Z — two seconds after #221 merged at 05:55:35Z. #221 touched only `.github/workflows/ci.yml` and `cd.yml`, +4/-4 each: action version bumps. So the required-status-check option was never implemented, and a search of open issues turns up nothing tracking it. The doc now says so and tells the reader to treat it as untracked. Not reopening #218 — you've said repeatedly to stop re-raising issues in this area, and a reopened issue is exactly that. |
| LOW 3 — `Dockerfile:12`, no lockfile committed | Not folded in; filed as #704. Confirmed real: `bun.lock` is gitignored, `COPY package.json bun.lockb* ./` matches nothing, so `eslint ^10.0.1` and `@typescript-eslint ^8.56.0` resolve fresh per build. Committing a lockfile is a different diff with a different blast radius than a two-file builder-stage-plus-docs change, so it goes in its own PR. The issue also notes a second thing reviewoie didn't: `ci.yml`'s cache key is `hashFiles('bun.lockb')` over a nonexistent file, so that cache never invalidates. |

No new lines of code — the diff is `docs/DEPLOYMENT.md` only, +19/-12. `Dockerfile`
and `.github/` are untouched by this commit.

Heads-up on one thing outside the findings: the branch is now behind
`upstream/main` (it was even as of the 08-04 review). Still `MERGEABLE`, and
neither changed file conflicts, so I left it rather than rebasing a docs commit
for its own sake. Say the word if you'd rather it be current.

reviewoie has no review at the new head `c40a09c`, so today's review pass will
pick it up.

reviewoie (approved):
## Code review — PR 640

**Verdict:** approve

Reviewed at head `19e638d1`. Prior reviewoie rounds (2026-07-16, 08-03, 08-04) were posted as issue comments rather than formal reviews, so this is the first one that carries a review state.

### The rejected shape is NOT re-raised

`git diff upstream/main pr640 -- .github/` is **empty** — the original `push: branches: [main]` trigger on `ci.yml` is fully reverted. `ci.yml` on the branch still has only `pull_request: branches: [main]`. This PR does the opposite of re-raising it: `docs/DEPLOYMENT.md` gains a "Decision: CI does not trigger on `push: [main]`" section that says **do not add one** and records both prior rejections (#438, #560). That is exactly what Hoie asked for in `discussion_r3701208827` — *"Pivot. Also document this situation and decision."*

### Dockerfile change — verified

- **Lint passes on the merged result.** `upstream/main` is an ancestor of the branch (merge is a no-op), and `bun run lint` exits 0 on that tree in 9.2s. The new step will not break CD on merge.
- **`eslint.config.js` in the COPY set is necessary and sufficient.** Replicated the builder stage's exact file set (`package.json`, `bun.lockb*` → matches nothing, `tsconfig.json`, `tsconfig.node.json`, `vite.config.ts`, `eslint.config.js`, `src`, `scripts`) into a clean dir, `bun install`, then: with the config → exit 0; with the config removed → exit 2, *"ESLint couldn't find an eslint.config.(js|mjs|cjs) file"*. The flat config sets no `parserOptions.project`, so nothing outside the copied set is required.
- **`.dockerignore` does not exclude it** — the patterns are `.*`, `build`, `node_modules`, `.env`, `.env.*`; `eslint.config.js` is not a dotfile.
- **Multi-source `COPY vite.config.ts eslint.config.js ./` is valid** (destination ends in `/`).
- **devDeps are present in the builder.** `bun install` runs with no `--production` and the Dockerfile sets no `NODE_ENV` — and the pre-existing `RUN bun run typecheck` (`bun run --bun tsc --noEmit`, `typescript` is a devDep) already proves it empirically in that stage.
- **Stage ordering is correct and cheap.** `RUN bun run lint` sits after `bun install` (needs the binary) and before typecheck/test/build, mirroring `ci.yml`'s Lint → Type check → Tests order. Lint-first masks a type error on a doubly-broken commit; that commit saw the same ordering on its PR run, so nothing is lost.
- **Cache impact is nil-to-negligible.** The `COPY` edit is on a layer above `COPY src src`, and `cd.yml`'s `docker/build-push-action@v6` step passes no `cache-from`/`cache-to`, so every CD build is already cold. Net cost is one added `RUN` at roughly 10s against a build that already runs typecheck + full test suite + `vite build`.
- **The `--bun` runtime risk from the earlier rounds is largely retired.** `bun run lint` is `bun --bun eslint src`, which the file-set simulation only exercised under local bun 1.3.14, not the pinned `oven/bun:1` image. But `typecheck` is `bun run --bun tsc --noEmit` and already runs green in that exact image on every deploy — a node CLI under bun's own resolution in `oven/bun:1` is therefore already proven in production. No Docker daemon is available here either, so a real `docker build --target builder .` still has not been executed; the PR body says so plainly.

### Doc claims — each re-checked against the API, not the prose

- `main`'s ruleset (`protect_main`, id 12426501) rules are `deletion`, `creation`, `non_fast_forward`, `pull_request`; `GET /repos/hoiekim/budget/branches/main` reports `required_status_checks.enforcement_level: "off"`, `contexts: []`. So "no required status check, a red run does not block the merge button" holds.
- `bypass_actors: null`, `enforcement: active`, `pull_request` rule with `required_approving_review_count: 1`, `require_code_owner_review: true`, `required_review_thread_resolution: true`. So "a review and thread resolution" and "forbids direct pushes for every actor, owner included" both hold.
- `ci.yml` has only a `pull_request` trigger; its `test` job runs lint → typecheck → `test:coverage`, and `build` is `needs: test`. The table's rows match. The "same check set minus the coverage threshold" claim is now exactly right with lint added to the builder.
- #218: `state_reason: "completed"`, zero comments, closed `2026-03-30T05:55:37Z` — two seconds after #221 merged, and #221's diff is action-version bumps in `cd.yml` + `ci.yml` only. So "closed as completed by #221, which only bumped action versions … treat it as untracked" holds.
- The MED from the 2026-08-04 round (the merge-ref claim) and both LOWs from that round (the direct-push exemption, the #218 "remains available" phrasing) are resolved at this head.

### Scope

Two files, both on-title: `Dockerfile` is the change, `docs/DEPLOYMENT.md` documents it — and the doc was explicitly requested in review, so it is in scope rather than a drive-by. Branch is even with `upstream/main`, `mergeStateStatus: CLEAN`, and both CI checks (`test`, `build`) pass on `19e638d1`.

### Low

- `Dockerfile:12` — carried from the 2026-08-03 and 08-04 rounds, still open. No lockfile is committed (`bun.lock` is in `.gitignore` at line 23, and `COPY package.json bun.lockb* ./` matches nothing in the tree), so the builder resolves `eslint ^10.0.1` and `@typescript-eslint/* ^8.56.0` fresh on every build. This PR extends the deploy gate to lint, so a `@typescript-eslint` minor that tightens `no-unused-vars` can now fail the **deploy** on a commit whose PR run was green at an earlier resolution time. Bounded by the flat config enumerating its rules explicitly instead of extending `js.configs.recommended`, so eslint-core rule additions cannot drift in. Pre-existing pattern (`typecheck` already floats `typescript ^5.3.0`) and out of scope for a two-line Dockerfile change — noted only because deploy-gate reliability is this PR's subject, and it is the one way this change can turn a green commit into a failed deploy.

No further blockers. Approving — hoiekim already approved this same head on 2026-08-13.

PR #643, opened 2026-07-16
perf(sync): batch searchHoldingsByAccountId into one IN query (#642)

### LINKED ISSUE #642: perf(sync): searchHoldingsByAccountId is an N+1 — one query per account in a loop on the Plaid/SimpleFin sync path
## Summary

`searchHoldingsByAccountId` (`repositories/holdings.ts:121-131`) issues **one `holdingsTable.query` per account_id in a `for` loop** — a textbook N+1. It runs inside the Plaid and SimpleFin **sync engine**, which executes per-item, per-user on every sync (cron / webhook). A user whose investment item exposes K accounts pays K round-trips where 1 batched query would do, and that multiplies by items and by user count (rule 13: "cron jobs run for ALL users; every per-user-body inefficiency multiplies by user count").

## The code

```ts
// repositories/holdings.ts:121
export const searchHoldingsByAccountId = async (
  user: MaskedUser,
  account_ids: string[],
): Promise<JSONHolding[]> => {
  if (!account_ids.length) return [];
  const results: JSONHolding[] = [];
  for (const account_id of account_ids) {
    const models = await holdingsTable.query({ [ACCOUNT_ID]: account_id, [USER_ID]: user.user_id });
    results.push(...models.map((m) => m.toJSON()));
  }
  return results;
};
```

Each iteration is a separate awaited round-trip to Postgres.

## Where it runs (hot path)

- `compute-tools/sync-plaid.ts:347` — `getStoredAccountsData` fetches an item's accounts, then `searchHoldingsByAccountId(user, accountIds)` for all of them.
- `compute-tools/sync-simple-fin.ts:157` — same call in the SimpleFin sync.

Both are on the per-item sync path that fires for every user on every sync cycle, so the loop's cost is `O(accounts-per-item)` extra round-trips per user per sync.

## Fix

`Table.query` already supports batched `IN`/`ANY` filtering via `inFilters` (`postgres/database.ts:286,341`). Collapse the loop to a single query:

```ts
export const searchHoldingsByAccountId = async (
  user: MaskedUser,
  account_ids: string[],
): Promise<JSONHolding[]> => {
  if (!account_ids.length) return [];
  const models = await holdingsTable.query({
    [USER_ID]: user.user_id,
    inFilters: { [ACCOUNT_ID]: account_ids },
  });
  return models.map((m) => m.toJSON());
};
```

(Confirm the exact `inFilters` option name/shape against `database.ts` — the point is one `account_id = ANY(...)` query instead of N.) `account_id` is already indexed, so the batched query stays index-driven. This matches the "extend Table methods / no N-round-trip loops" reflex; no behavior change, purely the round-trip count.

## Not user state

Deterministic control-flow inefficiency in engine code; no user-editable input involved. System-side scalability defect (rule 13), sibling to the delta-sync index gap filed separately.

claoie:
## Summary

`searchHoldingsByAccountId` (`repositories/holdings.ts`) looped `holdingsTable.query` **once per `account_id`** — a textbook N+1 on the Plaid/SimpleFin **sync engine** path (`compute-tools/sync-plaid.ts`, `compute-tools/sync-simple-fin.ts`), which runs per-item, per-user on every sync. A user whose investment item exposes K accounts paid K round-trips where 1 batched query does the same work — and that multiplies by items and by user count (rule 13: cron jobs run for all users).

Collapses the loop to a single `account_id IN (...)` query via the existing `inFilters` option — the same shape `getTransactionsByAccountIds` already uses in `transactions.ts`. `account_id` is indexed, so the batched read stays index-driven. Same `JSONHolding[]` output contract; purely the round-trip count changes (O(accounts) → O(1)).

Closes #642.

## Changes

- `repositories/holdings.ts` — replace the per-account `for` loop with one `holdingsTable.query({ user_id }, { inFilters: { account_id: account_ids } })`.
- `repositories/holdings.test.ts` — regression tests: N accounts issue exactly one SELECT with an `account_id IN (...)` clause; empty input is a no-op; query is `user_id`-scoped.

## Test plan

- [x] `bun --bun eslint src` (whole tree, matching CI) — clean
- [x] `bun run test` — 1097 pass / 2 skip / 0 fail
- [x] New regression tests assert one batched `IN (...)` query replaces the N-query loop, with the same `user_id` + all `account_id`s in the params array.
- [x] **UI E2E** — sandbox, 390×844, logged in as admin. `/accounts` → clicked an investment-type `.AccountRow` → `/account-detail`. **Holdings Composition** table renders correctly: a priced security row (value + positive growth), an `Unknown` row, and a `Total` row; balance graph + performance-benchmark box populate. 0 non-SW console errors. Screenshot `/tmp/pr-643-holdings.png` eye-checked. This is the consumer surface the sync-stored holdings feed — confirms the holdings-read path stays intact.

**E2E reachability note:** `searchHoldingsByAccountId` is invoked *inside the sync engine* (`getStoredAccountsData`), which requires a live Plaid/SimpleFin credential to trigger — not reachable on a scrambled-clone sandbox. The batched query returns a byte-identical result set (same WHERE semantics), locked by the regression tests; the UI check above confirms the holdings-read surface the synced data feeds still renders correctly.

### DIFF
--- a/src/server/lib/postgres/repositories/holdings.test.ts
+++ b/src/server/lib/postgres/repositories/holdings.test.ts
@@ -18,7 +18,7 @@ mock.module("pg", () => ({
   default: { Pool: FakePool, types: { setTypeParser: () => {} } },
 }));
 
-const { deleteHoldings } = await import("./holdings");
+const { deleteHoldings, searchHoldingsByAccountId } = await import("./holdings");
 
 afterAll(restoreLeaves);
 
@@ -73,6 +73,36 @@ describe("deleteHoldings — terminator-only model (#471)", () => {
     await deleteHoldings(mockUser, ["acc-1-sec-a"]);
     const [sql, values] = mockQuery.mock.calls[0];
     expect(String(sql)).toContain("user_id");
+    expect(values).toContain("usr-1");
+  });
+});
+
+describe("searchHoldingsByAccountId — single batched query (#642)", () => {
+  // Previously this looped `holdingsTable.query` once per account_id — an N+1
+  // on the Plaid/SimpleFin sync path (one round-trip per account of an item,
+  // per user, per sync). Lock in that N accounts now resolve in ONE
+  // `account_id IN (...)` query so the cost is O(1) round-trips, not O(N).
+
+  test("N accounts issue exactly one SELECT with an IN clause", async () => {
+    await searchHoldingsByAccountId(mockUser, ["acc-1", "acc-2", "acc-3"]);
+    const selectCalls = mockQuery.mock.calls.filter(([sql]) =>
+      String(sql).toUpperCase().includes("SELECT"),
+    );
+    expect(selectCalls.length).toBe(1);
+    const [sql, values] = selectCalls[0];
+    expect(String(sql)).toContain("account_id IN (");
+    expect(values).toEqual(expect.arrayContaining(["usr-1", "acc-1", "acc-2", "acc-3"]));
+  });
+
+  test("empty input is a no-op (no SQL emitted)", async () => {
+    await searchHoldingsByAccountId(mockUser, []);
+    expect(mockQuery.mock.calls.length).toBe(0);
+  });
+
+  test("query is scoped by user_id", async () => {
+    await searchHoldingsByAccountId(mockUser, ["acc-1"]);
+    const [sql, values] = mockQuery.mock.calls[0];
+    expect(String(sql)).toContain("user_id = $1");
     expect(values).toContain("usr-1");
   });
 });
--- a/src/server/lib/postgres/repositories/holdings.ts
+++ b/src/server/lib/postgres/repositories/holdings.ts
@@ -122,10 +122,9 @@ export const searchHoldingsByAccountId = async (
   account_ids: string[],
 ): Promise<JSONHolding[]> => {
   if (!account_ids.length) return [];
-  const results: JSONHolding[] = [];
-  for (const account_id of account_ids) {
-    const models = await holdingsTable.query({ [ACCOUNT_ID]: account_id, [USER_ID]: user.user_id });
-    results.push(...models.map((m) => m.toJSON()));
-  }
-  return results;
+  const models = await holdingsTable.query(
+    { [USER_ID]: user.user_id },
+    { inFilters: { [ACCOUNT_ID]: account_ids } },
+  );
+  return models.map((m) => m.toJSON());
 };


reviewoie (commented):
## Code review — PR 643

**Findings:** 0 blocking — 0 high, 0 med, 0 low. One non-blocking note.

### Behavioral equivalence — verified
- `user_id` scoping preserved: old and new both pass `user_id` as a plain `filters` key → `user_id = $1` (not the `options.user_id` path). `USER_ID`/`ACCOUNT_ID` = `"user_id"`/`"account_id"` (models/common.ts).
- `account_id IN (...)` built by `buildSelectWithFilters` inFilters loop (database.ts:341-347); empty/undefined arrays are skipped there.
- Soft-delete exclusion unchanged: both call `query` with the same `excludeDeleted ?? supportsSoftDelete` default (base.ts:133) — no `excludeDeleted` override on either side.
- **Result order** is the one real behavioral delta (single result set vs per-account concatenation, no `ORDER BY`), but neither consumer is order-sensitive: `sync-plaid.ts:347` and `sync-simple-fin.ts:157` both feed the result as `storedHoldings` into `upsertAndDeleteHoldingsWithSnapshots`, which consumes it only via `new Map(existingHoldings.map(h => [h.holding_id, h]))` + `.filter` (create-snapshots.ts:57, 80) — keyed by `holding_id`, never positional. Order is irrelevant.

### Tests — correct
Ran the file: 7 pass / 0 fail. `beforeEach` `mockQuery.mockClear()` (line 30-33) makes the empty-input `calls.length === 0` assertion valid. The `selectCalls.length === 1` assertion is the load-bearing regression proof (N accounts → 1 query); IN-clause + `user_id = $1` + param-array assertions all pin the right shape.

### Edge cases
- Empty array: early `return []` (holdings.ts:123) + inFilters skip — covered by test.
- Duplicate `account_id`s: old loop would return a holding once per duplicate; the `IN` clause dedupes. Benign — `accountIds` is `accounts.map(e => e.account_id)` (unique per item) and the downstream Map keys by `holding_id` regardless. New behavior is arguably more correct.
- **Note (non-blocking):** a single `IN (...)` query binds N+1 params, bounded by Postgres's 65535-param cap. Accounts-per-item is a handful in practice, so nowhere near the limit — no action needed, just noting the difference from the previously-unbounded loop.

### Scope
Diff is exactly the fix (`holdings.ts`) + its regression test (`holdings.test.ts`). Clean.

No further blockers.

claoie:
Thanks — addressing the review:

- **Behavioral equivalence / ordering:** confirmed. Neither consumer of `searchHoldingsByAccountId` is order-sensitive — `sync-plaid.ts` and `sync-simple-fin.ts` both feed the result into `upsertAndDeleteHoldingsWithSnapshots`, which consumes `existingHoldings` via a `holding_id`-keyed Map + `.filter`, never positionally. Leaving the query without an explicit `ORDER BY` (matching the old per-account concatenation, which was itself unordered across accounts).
- **Duplicate account_ids:** inputs come from `getStoredAccountsData` (distinct accounts of an item), so no dupes in practice; `IN` would dedupe harmlessly if any appeared. No change.
- **Large-array param cap:** accounts-per-item is a handful; nowhere near Postgres's bind limit. Noted, no action.

No code changes from the review (0 blocking findings).

hoiekim (approved):

PR #644, opened 2026-07-17
perf(sync): (user_id, updated) composite index for delta-by-cursor reads (#641)

### LINKED ISSUE #641: perf(sync): delta-by-cursor reads seq-scan the whole table — no (user_id, updated) index on transactions / investment_transactions / split_transactions / snapshots
## Summary

The delta-by-cursor warm sync (introduced for transactions in PR #536) filters four tables on the `updated` column, but **none of those tables has an index that leads with `updated`**. Every warm sync therefore does a **sequential scan of the user's entire table** to return only the handful of rows changed since the last cursor — an `O(user's total rows)` read on a path that fires on **every app load**, when it should be `O(rows-changed)`.

This is benign for a single small user today (~8 ms) but directly violates the many-users scalability vision: the cost multiplies by every user's full history on every one of their sessions.

## Affected read paths (all filter `{ column: UPDATED, start, end }`)

| Table | Repo filter | Model indexes | `updated`-leading index? |
|---|---|---|---|
| `transactions` | `repositories/transactions.ts:53` | user_id, account_id, date, pending | **no** |
| `investment_transactions` | `repositories/transactions.ts:89` | user_id, account_id, date | **no** |
| `split_transactions` | `repositories/split_transactions.ts:70` | user_id, transaction_id, account_id | **no** |
| `snapshots` (`searchSnapshots`) | `repositories/snapshots.ts:84` | user_id, snapshot_type, snapshot_date, account_id, security_id, holding_account_id | **no** |

(`getHoldingSnapshots` at `snapshots.ts:210` ranges on `SNAPSHOT_DATE`, which **is** indexed — that path is fine and is not part of this issue.)

## Why this is a hot path

- Client warm/delta sync sends `GET /api/transactions?start-date=<cursor>` (and the same for split-transactions / snapshots), where the server maps `start-date` to `WHERE updated >= <cursor>` (`hooks/sync.ts:145,191,318`; route `get-transactions.ts`).
- `sync()` runs on **every login / app mount** (`components/App/Utility.tsx:30`, `useEffect` on `userLoggedIn`) plus on every Plaid/SimpleFin link and Configuration action.
- So on each session the four delta reads each scan that user's whole table, even though a steady-state cursor usually matches **zero** changed rows.

## Root cause

`IndexDefinition` is single-column only — `{ column: string }` (`models/base.ts:36-38`) — and `buildCreateIndex` emits `ON tableName(column)` for one column (`postgres/database.ts:71-74`). There is no way to declare a composite `(user_id, updated)` index in the model layer, so the delta-sync query shape was shipped (PR #536) without the index it needs. The four models above list only per-single-column btrees; the planner's best option for `user_id = ? AND updated >= ?` is a full seq scan (it won't use the `user_id` btree because, for one user, that returns every row anyway).

## Evidence — prod-clone, 1 user, 6265 transactions

Representative delta query (steady-state cursor gives 3 changed rows). **Current indexes:**

```
EXPLAIN (ANALYZE, BUFFERS)
SELECT * FROM transactions
WHERE user_id = '<uuid>' AND updated >= TIMESTAMPTZ '2026-06-01' AND is_deleted = false
ORDER BY date DESC;

 Sort  (actual time=8.081..8.084 rows=3)
   Buffers: shared hit=1423
   ->  Seq Scan on transactions  (actual time=7.870..8.038 rows=3)
         Filter: ((NOT is_deleted) AND (updated >= ...) AND (user_id = ...))
         Rows Removed by Filter: 6262
         Buffers: shared hit=1420
 Execution Time: 8.184 ms
```

Seq Scan removing **6262 of 6265** rows to return 3.

**Proven fix in a ROLLED-BACK transaction** — added `CREATE INDEX idx_tx_user_updated ON transactions (user_id, updated)`, re-ran the identical query:

```
 Sort  (actual time=0.068..0.070 rows=3)
   Buffers: shared hit=6 read=2
   ->  Index Scan using idx_tx_user_updated on transactions  (actual time=0.034..0.041 rows=3)
         Index Cond: ((user_id = ...) AND (updated >= ...))
         Filter: (NOT is_deleted)
 Execution Time: 0.135 ms
```

Shape change `Seq Scan -> Index Scan`, buffers `1420 -> 5`, execution `8.184 ms -> 0.135 ms` (~60x). Crucially the cost stops scaling with the user's total history — it becomes `O(rows-changed)`. The index was created inside `BEGIN ... ROLLBACK` and never persisted.

## Not user state

Deterministic query-planner behavior on a structural gap; no user-editable column is involved, and it affects every user's every session. This is a system-side scalability defect (rule 13), same class as the recent inbox GIN-index finding.

## Proposed fix

1. Extend the model index layer to support a composite index — either widen `IndexDefinition` to `{ column: string } | { columns: string[] }` and have `buildCreateIndex` join the columns, or add the raw `CREATE INDEX ... (user_id, updated)` statements alongside the existing bootstrap in `postgres/initialize.ts` (the same escape hatch used for other non-single-column indexes).
2. Add `(user_id, updated)` to `transactions`, `investment_transactions`, `split_transactions`, and `snapshots`.
3. Confirm each delta path flips to `Index Scan` via `EXPLAIN ANALYZE` (as above), and that the added indexes match the exact `WHERE user_id = ? AND updated >= ?` predicate the repos emit.

Optional: consider whether `is_deleted` belongs in the composite (tombstones are still returned to the FE, so the filter stays non-selective — a plain `(user_id, updated)` is sufficient and keeps the index narrow).

claoie:
## Summary

Closes #641. The delta-by-cursor warm sync (transactions, PR #536; extended to investment/split transactions + snapshots) filters four tables on `WHERE user_id = ? AND updated >= ?` — a query that fires on **every app load**. None of those tables had an index leading with `updated`, so the planner **seq-scanned each user's entire table** to return the handful of rows changed since the last cursor: `O(user's total rows)` on a hot path that should be `O(rows-changed)`. Rule 13 / Sweep 7 scalability.

## Change

The model index layer was single-column only (`IndexDefinition = { column: string }`), so there was no way to declare a composite `(user_id, updated)` in the model layer — that's why the delta-sync shape shipped without its index. This stays in the framework rather than bypassing it:

- **`IndexDefinition`** widened to `{ column: string } | { columns: string[] }` (`models/base.ts`).
- **`buildCreateIndex`** now accepts `string | string[]` and emits a multi-column `ON table(a, b)` clause; the scalar form is unchanged (`database.ts`).
- **`initialize.ts`** index loop normalizes `column`/`columns` before building.
- Added `{ columns: [USER_ID, UPDATED] }` to `transactions`, `investment_transactions`, `split_transactions`, `snapshots`. Indexes are created idempotently via the existing `CREATE INDEX IF NOT EXISTS` bootstrap — no migration step needed; existing deployments pick it up on next boot.

`getHoldingSnapshots` (ranges on `snapshot_date`, already indexed) is untouched — not part of this issue.

## Test plan

- [x] `bun run lint` (eslint whole tree) — clean
- [x] `bun run test` — 1095 pass / 0 fail (added a `buildCreateIndex` composite-index unit test: composite emits `ON transactions(user_id, updated)`; single-element array matches the scalar form byte-for-byte)
- [x] `bun run build` — clean

### EXPLAIN evidence (from #641, prod-clone, 1 user, 6265 transactions)

The representative steady-state delta query (cursor matches 3 changed rows), proven in a rolled-back transaction:

- **Before:** `Seq Scan on transactions … Rows Removed by Filter: 6262`, buffers `hit=1420`, `8.184 ms`.
- **After (composite added):** `Index Scan using idx_transactions_user_id_updated … Index Cond: (user_id = ? AND updated >= ?)`, buffers `hit=6 read=2`, `0.135 ms` (~60x), and the cost stops scaling with the user's total history.

### UI E2E

This PR adds indexes only — no behavioral surface changes, no query results change (identical rows returned, index just serves the same predicate faster). The delta-sync read paths that consume these tables (`/api/transactions?start-date=…`, split-transactions, snapshots) return the same rows before/after; the DDL builder change is covered by the new unit test. A sandbox spin renders the transactions/accounts screens that fire warm sync on mount to confirm no regression.

### DIFF
--- a/src/server/lib/postgres/database.test.ts
+++ b/src/server/lib/postgres/database.test.ts
@@ -394,6 +394,16 @@ describe("DDL builders", () => {
       "CREATE INDEX IF NOT EXISTS custom_idx ON things(name)",
     );
   });
+
+  it("buildCreateIndex builds a composite index over multiple columns", () => {
+    expect(buildCreateIndex("transactions", ["user_id", "updated"])).toBe(
+      "CREATE INDEX IF NOT EXISTS idx_transactions_user_id_updated ON transactions(user_id, updated)",
+    );
+    // A single-element array matches the scalar form exactly (no trailing comma/space).
+    expect(buildCreateIndex("things", ["name"])).toBe(
+      buildCreateIndex("things", "name"),
+    );
+  });
 });
 
 describe("result helpers", () => {
--- a/src/server/lib/postgres/database.ts
+++ b/src/server/lib/postgres/database.ts
@@ -68,9 +68,14 @@ export function buildCreateTable(
   `.trim();
 }
 
-export function buildCreateIndex(tableName: string, column: string, indexName?: string): string {
-  const name = indexName || `idx_${tableName}_${column}`;
-  return `CREATE INDEX IF NOT EXISTS ${name} ON ${tableName}(${column})`;
+export function buildCreateIndex(
+  tableName: string,
+  column: string | string[],
+  indexName?: string,
+): string {
+  const columns = Array.isArray(column) ? column : [column];
+  const name = indexName || `idx_${tableName}_${columns.join("_")}`;
+  return `CREATE INDEX IF NOT EXISTS ${name} ON ${tableName}(${columns.join(", ")})`;
 }
 
 export function prepareParamValue(value: ParamValue): ParamValue {
--- a/src/server/lib/postgres/initialize.ts
+++ b/src/server/lib/postgres/initialize.ts
@@ -74,7 +74,8 @@ export const initializePostgres = async (): Promise<void> => {
       await pool.query(createTableSql);
 
       for (const idx of table.indexes) {
-        const createIndexSql = buildCreateIndex(table.name, idx.column);
+        const columns = "columns" in idx ? idx.columns : idx.column;
+        const createIndexSql = buildCreateIndex(table.name, columns);
         await pool.query(createIndexSql);
       }
     }
--- a/src/server/lib/postgres/models/base.ts
+++ b/src/server/lib/postgres/models/base.ts
@@ -33,9 +33,7 @@ export type Constraints = string[];
 
 export type RowValueType = string | number | Date | boolean | null | Object;
 
-export interface IndexDefinition {
-  column: string;
-}
+export type IndexDefinition = { column: string } | { columns: string[] };
 
 export type PropertyChecker<T> = {
   [K in keyof T]: (value: unknown) => boolean;
--- a/src/server/lib/postgres/models/investment_transaction.ts
+++ b/src/server/lib/postgres/models/investment_transaction.ts
@@ -167,7 +167,14 @@ export const investmentTransactionsTable = createTable({
   name: INVESTMENT_TRANSACTIONS,
   primaryKey: INVESTMENT_TRANSACTION_ID,
   schema: invTxSchema,
-  indexes: [{ column: USER_ID }, { column: ACCOUNT_ID }, { column: DATE }],
+  indexes: [
+    { column: USER_ID },
+    { column: ACCOUNT_ID },
+    { column: DATE },
+    // Delta-by-cursor warm sync filters `WHERE user_id = ? AND updated >= ?` on
+    // every app load; the composite keeps the read O(rows-changed). See #641.
+    { columns: [USER_ID, UPDATED] },
+  ],
   ModelClass: InvTxModel,
 });
 
--- a/src/server/lib/postgres/models/snapshot.ts
+++ b/src/server/lib/postgres/models/snapshot.ts
@@ -213,6 +213,9 @@ export const snapshotsTable = createTable({
     // scan + row filter — fine for low-cardinality users but slows down
     // as snapshot history grows. See PR #470 / #445.
     { column: HOLDING_ACCOUNT_ID },
+    // `searchSnapshots` delta reads filter `WHERE user_id = ? AND updated >= ?`
+    // on every app load; the composite keeps the read O(rows-changed). See #641.
+    { columns: [USER_ID, UPDATED] },
   ],
   ModelClass: SnapshotModel,
 });
--- a/src/server/lib/postgres/models/split_transaction.ts
+++ b/src/server/lib/postgres/models/split_transaction.ts
@@ -125,7 +125,14 @@ export const splitTransactionsTable = createTable({
   name: SPLIT_TRANSACTIONS,
   primaryKey: SPLIT_TRANSACTION_ID,
   schema: splitTxSchema,
-  indexes: [{ column: USER_ID }, { column: TRANSACTION_ID }, { column: ACCOUNT_ID }],
+  indexes: [
+    { column: USER_ID },
+    { column: TRANSACTION_ID },
+    { column: ACCOUNT_ID },
+    // Delta-by-cursor warm sync filters `WHERE user_id = ? AND updated >= ?` on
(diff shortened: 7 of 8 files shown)


reviewoie (commented):
## Code review — PR 644

Verified end to end:

- **DDL generation** — `buildCreateIndex` correctly emits `CREATE INDEX IF NOT EXISTS idx_<table>_user_id_updated ON <table>(user_id, updated)`. Composite name is distinct from the single-column `idx_<table>_user_id`, so no collision; `IF NOT EXISTS` keeps the init loop idempotent. The single-element-array test asserts scalar/array parity (no trailing comma). Longest generated name (`idx_investment_transactions_user_id_updated`, 43 chars) is well under the 63-byte identifier limit.
- **Delta-path coverage + column order** — all four indexed tables are read by a delta query shaped `WHERE user_id = $1 AND updated >= $2 [AND updated <= $3]`, with `user_id` equality leading and `updated` range second, matching the composite's column order exactly:
  - `transactions` — `getTransactions` (`repositories/transactions.ts:45-61`, `dateRange` on `UPDATED`)
  - `investment_transactions` — `searchTransactions` (`transactions.ts:85-97`)
  - `split_transactions` — `searchSplitTransactions` (`split_transactions.ts:67-74`)
  - `snapshots` — `searchSnapshots` (`snapshots.ts:82-99`)
- **No missed model** — the only routes threading a cursor (`startDate`) are `get-transactions` (covers transactions + investment_transactions), `get-split-transactions`, `get-snapshots`. No accounts/items/holdings delta read exists, so no other model needs the composite.
- **Union widening is safe** — the only consumer of `.column` on an `IndexDefinition` is the `initialize.ts:77` loop, which the diff updates to branch on `"columns" in idx`. The other `.column` accesses in `database.ts` (`additionalWhere`, `primaryKey`, `dateRange`) are unrelated option types. `tsc --noEmit` clean; `database.test.ts` 37 pass.
- **Scope** — every file is directly related to the composite index. Clean.

### Low
- `models/{transaction,investment_transaction,split_transaction,snapshot}.ts` — the pre-existing standalone `{ column: USER_ID }` index on each of the four tables is now a redundant leftmost-prefix of the new `(user_id, updated)` composite (a btree on `(user_id, updated)` already serves `user_id = $1` lookups). Not a bug and dropping it is out of scope for this PR, but the redundant index carries write-time maintenance cost on the highest-churn tables — worth a follow-up to drop the four single-column `user_id` indexes once the composites are live.

No further blockers.

claoie:
Thanks reviewoie — agreed on all points.

**Low finding (redundant single-column `user_id` indexes):** correct — a btree on `(user_id, updated)` is a covering leftmost-prefix for `user_id = $1` lookups, so the four standalone `{ column: USER_ID }` indexes on these high-churn tables now carry write-cost for nothing. Leaving it out of this PR (Principle 4 — dropping indexes is a distinct change from adding them, and I want the composites live + observed before removing the fallback). Filed as a follow-up: #645.

**E2E verification (sandbox, prod-clone):**
- All 4 composite indexes created by the boot `CREATE INDEX IF NOT EXISTS` path — verified in `pg_indexes`.
- Planner flips to `Index Scan using idx_transactions_user_id_updated` with `Index Cond: (user_id = ? AND updated >= ?)` on a 7270-row user — no seq scan (matches #641's before/after).
- UI (390×844): `/transactions` warm sync `GET /api/transactions` 200, 76 rows render; `/accounts` `GET /api/snapshots` 200, 19 account rows + balance graphs render. Zero console errors.

hoiekim (approved):

Issue #542, opened 2026-06-22
feat(transfers): delta-by-cursor sync + tombstone delivery for GET /transfers


issueoie:
Follow-up to #541. The transfer-detection engine now has two new transitions (cascade soft-delete when a transaction is removed, status='rejected' when the user rejects a suggestion), but the read API (`GET /transfers`) is still a full-fetch with no `updated`-cursor and no tombstone delivery. The FE's cache replaces wholesale on every refresh, which works today, but won't survive a move to per-pair delta-sync.

## Current shape

`src/server/lib/postgres/repositories/transfers.ts:getTransferPairs`:

```sql
SELECT * FROM transaction_pairs
WHERE user_id = $1 AND (is_deleted IS NULL OR is_deleted = FALSE)
ORDER BY pair_id
```

No `since`, no `include-deleted`, no `updated >= cursor`. Every fetch returns every non-deleted pair.

## What this issue adds

Mirror the `GET /transactions` delta-by-cursor pattern (#536) for transfers:

1. **Project `updated`** in the response (it's already in the table; just include it in the JSON payload).
2. **`?since=<ISO timestamp>` query param** on `GET /transfers`. When set, filter `WHERE updated >= $since`.
3. **`?include-deleted=true` query param**. When set, drop the `is_deleted = FALSE` predicate so tombstoned pairs are returned (`is_deleted = TRUE` rows). FE consumes both as a tombstone signal to remove pairs from its local cache.
4. **FE `useTransfers` hook** — reduce over the delta instead of replacing wholesale. On `include-deleted=true` mode, drop any pair whose response row has `is_deleted = TRUE` from the local cache; merge `updated`-newer pairs.
5. **Sync orchestrator wire-up** — `useSync` stages a transfer delta-fetch alongside the existing transactions/split-transactions delta-fetch.

## Why this matters now

`#541` introduces two new transitions:

- **cascade soft-delete** — when a transaction is soft-deleted (Plaid pending→settled tombstone, `deleteTransactions`), the cascade now also soft-deletes any pairs referencing it. The FE's full-fetch happens to catch this (the pair just disappears from the next response), but if the FE adopts per-pair delta-sync without this issue's groundwork, those tombstones become invisible.
- **status='rejected'** — already visible to FE via the full-fetch (the pair stays in the response with the new status). Future delta-sync would still need to deliver `updated` so the FE knows when a pair flipped from suggested to rejected.

## Out of scope

- Engine-level changes (that's #541).
[...]

PR #646, opened 2026-07-18
feat(transfers): server-side delta delivery for GET /transfers (project updated + opt-in tombstones)

claoie:
## What

Server-side **delta delivery** for `GET /transfers`, mirroring the contract that `/transactions` and `/snapshots` already use — so a future FE hook can reduce over a delta instead of replacing its transfers cache wholesale. This is parts 1 + 3 of #542 (the additive server groundwork); the FE hook migration (parts 4–5) is a deliberate follow-up.

## Changes

- **Project `updated`** on each `TransferPair` in the response. It's already on `transaction_pairs` (bumped on every confirm/reject/cascade-delete) — it just wasn't surfaced. The FE needs it to know when a pair changed.
- **`?include-deleted=true`** query param on `GET /transfers`. When set, soft-deleted pairs are returned as **tombstones** — `{ pair_id, is_deleted: true, updated, transactions: [] }` — so the FE can evict them from its local cache. Tombstones skip transaction resolution because a cascade soft-delete may have soft-deleted the referenced transactions too (the active-only txn query wouldn't find them); `pair_id` is all the FE needs to evict.

## Why opt-in (not unconditional like `/transactions` and `/snapshots`)

Those two routes hardcode `includeDeleted: true` because their FE reducers were migrated to handle tombstones in the same PR. The transfers FE still full-fetches and replaces its cache wholesale — delivering tombstones as active rows would show deleted pairs as live. So delivery stays behind the param until the FE hook migrates (#542 parts 4–5), which flips the fetch to opt in. The new response fields are **optional** so the FE's optimistic Mark-as-Transfer path (which mints a pair locally before the server `updated` is known) still typechecks — this PR touches no FE code.

## Scope

Server only. Backward-compatible: the current FE sends no params, so it gets active-only pairs exactly as before, now with an extra (ignored) `updated`/`is_deleted` field. `getTransferPairs`'s only non-test caller is this route.

## Test plan

- [x] `bun test src/server` — 701 pass / 0 fail (7 in `get-transfers.test.ts`, +4 new)
- [x] **Unit coverage** (`get-transfers.test.ts`):
  - active pair projects `updated` + `is_deleted: false`
  - default (no param) emits the `(is_deleted IS NULL OR is_deleted = FALSE)` exclusion at the SQL layer
  - `include-deleted=true` → soft-deleted pair returns as a tombstone (`transactions: []`), fires **only** the pairs SELECT (no txn query), and drops the soft-delete exclusion
- [x] **API E2E** — sandbox, logged in as admin, deterministic clone fixture (admin has 177 confirmed + N suggested active pairs = 183 visible; 8 soft-deleted = 1 confirmed + 7 suggested; 4 rejected-active):
  - `GET /api/transfers` (default) → **183** pairs, **0 evictions**, **0 rejected**, **0 missing `updated`**. Every pair carries `updated` (string) + `is_deleted: false` + 2 resolved transactions.
  - `GET /api/transfers?include-deleted=true` → **195** pairs = 183 visible + **12 eviction signals**. The 12 split into **8 tombstones** (`is_deleted: true`) + **4 rejected** (`is_deleted: false`, `status: "rejected"`) — transfers' second FE-hidden axis. Every eviction has `transactions: []`, a `pair_id`, `updated` present, and satisfies `is_deleted || status === "rejected"`.
  - **Guards:** the 183 visible in the include-deleted response are byte-identical to the default set (`visibleMatchesDefault: true`); **0 eviction pair_ids leak into the default response** (backward-compat guard for the un-migrated wholesale-replace FE).
- [x] **UI backward-compat** — `/transactions` rendered at 390×844 (screenshot eye-checked, `/tmp/pr-646-transactions.png`): transaction rows + budget/category selectors + suggestion UI render; the transfers-consuming FE (`fetchTransfers` → `data.transfers`) still full-fetches and populates without change. **0 console errors** — the added `updated`/`is_deleted` fields and opt-in eviction delivery don't disturb the current FE.

### DIFF
--- a/src/server/lib/postgres/repositories/transfers.ts
+++ b/src/server/lib/postgres/repositories/transfers.ts
@@ -29,10 +29,11 @@ export interface TransferPair {
    *  Optional because the FE's optimistic Mark-as-Transfer path mints a pair
    *  locally before the server's `updated` is known. */
   updated?: string | null;
-  /** True for a tombstone: a soft-deleted pair returned only when the caller
-   *  passes `includeDeleted`, so the FE can evict it from its local cache.
-   *  Tombstones carry no `transactions` (the cascade may have soft-deleted
-   *  them too); `pair_id` + `is_deleted` is all the FE needs to evict. */
+  /** True for a soft-deleted pair. Returned only when the caller passes
+   *  `includeDeleted`. A soft-deleted OR `status='rejected'` pair is an
+   *  eviction signal (empty `transactions`) the FE removes from its cache —
+   *  it evicts on `is_deleted || status === 'rejected'`, so `pair_id` +
+   *  `status` + `is_deleted` is all it needs. */
   is_deleted?: boolean;
 }
 
@@ -74,37 +75,47 @@ export const getTransferPairs = async (
     { orderBy: PAIR_ID, excludeDeleted: !options.includeDeleted },
   );
 
-  // Exclude status='rejected' from the FE response: a rejected pair is
-  // user-marked "no, these don't pair" — the engine remembers it as a
-  // denylist signal but the user shouldn't see the pair in their Transfers
-  // view. Re-confirming a previously-rejected pair goes through
+  if (allPairs.length === 0) return [];
+
+  // A pair is FE-visible only when it's active AND not user-rejected. Both
+  // `is_deleted` (soft-delete / cascade) and `status='rejected'` (user marked
+  // "no, these don't pair") hide a pair from the Transfers view — rejected is
+  // transfers' second FE-hidden axis (transactions/snapshots have only
+  // `is_deleted`). Re-confirming a rejected pair goes through
   // `pairTransactions` (Mark as Transfer), which un-rejects via ON CONFLICT.
-  // Filtered in JS because Table.query expresses only equality / IN filters,
-  // not the `STATUS <> 'rejected'` inequality.
-  const pairs = allPairs.filter((p) => p.status !== "rejected");
+  // Rejection is expressed in JS because Table.query has no `<>` inequality.
+  const isVisible = (p: (typeof allPairs)[number]) =>
+    !p.is_deleted && p.status !== "rejected";
+  const visiblePairs = allPairs.filter(isVisible);
 
-  if (pairs.length === 0) return [];
+  // Under `includeDeleted`, every non-visible pair — soft-deleted OR rejected
+  // — is delivered as an eviction signal so the delta-reducing FE removes it
+  // from its local cache (it evicts on `is_deleted || status === 'rejected'`).
+  // Eviction signals need no transaction resolution: a cascade soft-delete may
+  // have soft-deleted the referenced transactions too (the active-only txn
+  // query below wouldn't find them), and `pair_id` + `status` + `is_deleted` +
+  // `updated` is all the FE needs to evict. On the default path `allPairs`
+  // already excludes soft-deleted at the SQL layer and rejected pairs are
+  // simply omitted, so there are no eviction signals — the response is
+  // active-non-rejected only, byte-for-byte the prior contract plus the
+  // additive `updated` / `is_deleted` fields.
+  const evictions = options.includeDeleted
+    ? allPairs
+        .filter((p) => !isVisible(p))
+        .map(
+          (pair): TransferPair => ({
+            pair_id: pair.pair_id,
+            status: pair.status,
+            transactions: [],
+            updated: pair.updated,
+            is_deleted: !!pair.is_deleted,
+          }),
+        )
+    : [];
 
-  // Tombstones need no transaction resolution — a cascade soft-delete may
-  // have soft-deleted the referenced transactions too, so the active-only
-  // txn query below wouldn't find them. The FE evicts on `pair_id`, so a
-  // tombstone only carries `pair_id` + `is_deleted` + `updated`.
-  const activePairs = pairs.filter((p) => !p.is_deleted);
-  const deletedPairs = pairs.filter((p) => p.is_deleted);
+  if (visiblePairs.length === 0) return evictions;
 
-  const tombstones = deletedPairs.map(
-    (pair): TransferPair => ({
-      pair_id: pair.pair_id,
-      status: pair.status,
-      transactions: [],
-      updated: pair.updated,
-      is_deleted: true,
-    }),
-  );
-
-  if (activePairs.length === 0) return tombstones;
-
-  const txnIds = activePairs.flatMap((p) => [p.transaction_id_a, p.transaction_id_b]);
+  const txnIds = visiblePairs.flatMap((p) => [p.transaction_id_a, p.transaction_id_b]);
 
   const txnModels = await transactionsTable.query(
     { [USER_ID]: user.user_id },
@@ -117,7 +128,7 @@ export const getTransferPairs = async (
     txnById.set(tx.transaction_id, tx);
   }
 
-  const active = activePairs
+  const active = visiblePairs
     .map((pair): TransferPair | null => {
       const a = txnById.get(pair.transaction_id_a);
       const b = txnById.get(pair.transaction_id_b);
@@ -132,7 +143,7 @@ export const getTransferPairs = async (
     })
     .filter((p): p is TransferPair => p !== null);
 
-  return [...active, ...tombstones];
+  return [...active, ...evictions];
 };
 
 /**
--- a/src/server/lib/postgres/repositories/transfers.ts
+++ b/src/server/lib/postgres/repositories/transfers.ts
@@ -23,6 +23,26 @@ export interface TransferPair {
   pair_id: string;
   status: TransferPairStatus;
   transactions: JSONTransaction[];
+  /** Bumped on every mutation (confirm/reject/cascade-delete). Lets the FE
+   *  reduce over a delta by `updated` instead of replacing its cache
+   *  wholesale — same role `updated` plays on transactions/snapshots.
+   *  Optional because the FE's optimistic Mark-as-Transfer path mints a pair
+   *  locally before the server's `updated` is known. */
+  updated?: string | null;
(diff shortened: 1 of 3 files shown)


reviewoie (commented):
## Code review — PR 646

**Verdict:** approve for the server-only scope — one forward-looking contract gap to track before the FE-hook migration (parts 4-5).

Reviewed the tombstone partition, backward-compat of the default path, ordering, `is_deleted` null-safety, and scalability. Scope is clean: 3 files, all directly on the delta-delivery contract, server-only.

### Med
- `src/server/lib/postgres/repositories/transfers.ts:81` — the delta contract cannot express **reject-evictions**. `pairs = allPairs.filter((p) => p.status !== "rejected")` runs BEFORE the deleted/active partition, so a `rejected` pair is dropped entirely and is never emitted as a tombstone even under `includeDeleted`. But `rejected` is transfers' *second* FE-hidden axis (transactions/snapshots have only `is_deleted`): a FE-visible `suggested`/`confirmed` pair routinely transitions to `rejected` via `rejectTransferPair` (the DELETE route) and via the supersede-cleanup UPDATEs in `pairTransactions`/`confirmTransferPair`. Failure scenario: once the delta-reducing FE (parts 4-5) stops replacing wholesale, a pair it holds that flips to `rejected` is never delivered in any form → it stays in the FE cache as a stale suggestion until a cold resync. Not a current regression — the default path is wholesale-replace so absence = eviction — but the contract this PR exists to establish is incomplete. Fix shape: make the `rejected` filter conditional on `!includeDeleted` (symmetric with how `is_deleted` is handled), and under `includeDeleted` deliver `rejected` pairs as eviction signals (empty `transactions`, carrying `updated`) so the FE evicts on `status === "rejected" || is_deleted`.

### Low
- `src/server/lib/postgres/repositories/transfers.ts:132` — `return [...active, ...tombstones]` is not globally `PAIR_ID`-ordered (the original returned one list ordered by `PAIR_ID`; the new `includeDeleted` path concatenates two independently-ordered slices). No consumer depends on it: the default path returns tombstone-free, order-identical output, and the FE keys pairs by `pair_id` in a Dictionary (order-independent). Informational — noting in case a future consumer assumes a merged `PAIR_ID` sort.

### Verified clean
- **Backward-compat (default path):** `options={}` → `excludeDeleted: true` → SQL excludes soft-deleted, so `deletedPairs`/`tombstones` are empty and output equals the prior active-only set plus the additive `updated` / `is_deleted:false` fields. No tombstone can leak to an un-migrated FE. Confirmed against test `default (no include-deleted) excludes soft-deleted pairs at the SQL layer`.
- **`is_deleted` null-safety:** model typeChecker is `isNullableBoolean`, so a DB null is possible; `!p.is_deleted` treats null as active and the SQL exclusion is `IS NULL OR = FALSE` — consistent (null = alive). Output normalizes to hardcoded `true`/`false`, so no null leaks to the FE.
- **Tombstones skip txn resolution correctly:** cascade-soft-deleted pairs whose transactions are also soft-deleted would miss the active-only txn query; emitting `{ pair_id, status, transactions: [], updated, is_deleted: true }` without resolution is right, and `pair_id` is all the FE needs to evict.
- **No double-count / drop:** `activePairs` and `deletedPairs` are a clean partition of the post-reject `pairs`; every pair lands in exactly one slice.
- **Scalability:** O(N) filters + one batched `inFilters` (ANY) txn query; no N-round-trip loop. No `?since` cursor / new index needed since delivery is full-set + tombstones (mirrors `get-transactions.ts`).

claoie:
Thanks — addressed the med, acknowledging the low.

**Med (reject-eviction gap) — fixed in b0f0d90.** You're right that this PR exists to define the contract, so shipping it without reject-evictions would hand the FE-hook migration a latent stale-suggestion bug. Unified on a single `isVisible(p) = !p.is_deleted && p.status !== "rejected"`. Under `include-deleted`, every non-visible pair — soft-deleted **or** rejected — is now delivered as an eviction signal (empty `transactions`, carrying `updated`); the FE evicts on `is_deleted || status === "rejected"`. Default path unchanged: rejected pairs stay omitted (absence = eviction for the wholesale-replace FE). Added 2 tests (rejected-non-deleted → eviction signal under include-deleted; rejected omitted entirely on default). Re-ran E2E: include-deleted now returns 183 visible + 12 evictions (8 tombstones + 4 rejected), 0 leak into default, visible-set byte-identical to default.

**Low (ordering) — leaving as-is, intentional.** `[...active, ...evictions]` isn't globally `PAIR_ID`-sorted, matching your read that no consumer depends on it: the default path is order-identical + eviction-free, and the FE keys by `pair_id` in a Dictionary (order-independent). This also mirrors the transactions/snapshots delta routes, which append tombstones without a global re-sort. If a future consumer needs a merged sort, it belongs with that consumer.

reviewoie:
## Code review — PR 646 (latest commit `b0f0d908`)

**Verdict:** approve for the server-only scope. The new commit correctly resolves the prior review's MED (reject-evictions). One forward-looking scalability gap to track before the parts 4-5 FE opt-in — widened by this commit, not a current regression.

Re-reviewed the `isVisible` partition, the reject-eviction path, backward-compat of the default path, and the two new tests. Scope is clean: 3 files, all on the delta-delivery contract, server-only.

### Med
- `src/server/routes/transfers/get-transfers.ts:20` / `repositories/transfers.ts:99` — **no cursor param → unbounded eviction-signal re-delivery under `include-deleted`, and this commit widens the set to include every rejected pair.** GET /transfers has no `start-date`/`since` filter; under `includeDeleted` it returns the user's *entire* accumulated set of soft-deleted AND (new in this commit) `status='rejected'` pairs on every fetch. Rejected pairs persist permanently by design — they are the detection engine's denylist (#541), so they only grow. The PR/route comment frames this as mirroring the /transactions and /snapshots delta contract, but that framing is inaccurate: `get-transactions.ts` IS cursor-scoped (`start-date`/`end-date` → `UPDATED` range at repo lines 54/90); /transfers is a plain full-fetch. Failure scenario: once the parts 4-5 FE opts into `include-deleted`, every transfers refresh re-downloads all rejected + soft-deleted pairs the FE already evicted — O(all-tombstones-ever) per fetch, per user. Not a current regression (delivery is opt-in and the FE doesn't set it yet). Fix before the FE migration lands: add a `start-date` cursor param mirroring `get-transactions.ts`, and add a `(user_id, updated)` composite index on `transaction_pairs` (the model at `models/transaction_pair.ts:78-82` has only single-column `user_id`/`transaction_id_a`/`transaction_id_b` indexes, and issue #641 — which adds `(user_id, updated)` to the other delta-read tables — does NOT cover `transaction_pairs`).

### Low
- `src/server/lib/postgres/repositories/transfers.ts:143` — `return [...active, ...evictions]` is not globally `PAIR_ID`-ordered under `includeDeleted` (two independently-ordered slices concatenated). Carried from the prior review; unchanged by this commit. No consumer depends on it (FE keys pairs by `pair_id`). Informational.

### Verified clean (this commit)
- **Prior MED resolved correctly.** `isVisible(p) = !is_deleted && status !== 'rejected'`; `visiblePairs` drives txn resolution, `evictions` (under `includeDeleted` only) carries soft-deleted OR rejected pairs as empty-`transactions` signals with `is_deleted: !!pair.is_deleted`. A `suggested`/`confirmed` pair that flips to `rejected` is now delivered as an eviction signal instead of vanishing. Confirmed by the two new tests.
- **Backward-compat / default path.** `!includeDeleted` → `evictions=[]` and SQL excludes soft-deleted, so output is active-non-rejected only — byte-for-byte the prior contract plus the additive `updated`/`is_deleted:false` fields. Rejected pairs stay omitted (absence = eviction for the wholesale-replace FE). Confirmed by test `default (no include-deleted) omits rejected pairs entirely`.
- **Clean partition, no double-delivery.** `visiblePairs` and `evictions` are disjoint (`isVisible` vs `!isVisible`); every pair lands in exactly one slice. Early return `if (visiblePairs.length === 0) return evictions` skips the txn query when nothing is visible.
- **Tests pass.** 9/9 in `get-transfers.test.ts`, 27/27 across the transfers routes, 21/21 in the transfers repo suite. Test names accurate (grepped `test(`), canonical pg-FakePool seam via `test-helpers` `restoreLeaves`.
- **Scope.** Only the 3 transfers server files; all directly on the PR title. No unrelated changes. No N-round-trip loop (O(N) filters + one batched `inFilters` txn query).

claoie:
Thanks — both findings are fair. Dispositions:

**Med (unbounded `include-deleted` re-delivery) — valid, deferring to the parts 4-5 FE-hook migration under #542, not this PR.** You're right that under `include-deleted` the eviction set (soft-deleted + now rejected) only grows, and that GET /transfers has no cursor while /transactions does. But that path has no live consumer: the current FE sends no params, so it hits the default active-only branch — unchanged, the pre-existing O(active) full-fetch this PR doesn't regress. `include-deleted=true` is only ever set once the FE hook opts in, which is exactly parts 4-5. Adding a `start-date`/`since` cursor + the `(user_id, updated)` composite index on `transaction_pairs` now — for a path nothing calls yet — would pull parts 4-5's work into this deliberately-additive groundwork PR and break the split.

I'll land the cursor + index in the parts 4-5 PR under #542, where the FE actually opts in and the cost becomes reachable. Noted there explicitly (comment on #542): #641's composite-index PR covers the sibling delta-read tables but **not** `transaction_pairs`, so parts 4-5 must add it.

If you'd rather the cursor land here as a precaution before any FE touches the param, say so and I'll fold it in — but my read is it belongs with the consumer.

**Low (eviction ordering) — unchanged, intentional** (as in the prior reply): `[...active, ...evictions]` isn't globally PAIR_ID-sorted; the default path stays order-identical + eviction-free, the FE keys by pair_id in a Dictionary, and this mirrors the transactions/snapshots delta routes. A merged sort belongs with the future consumer if one needs it.

hoiekim (approved):

PR #648, opened 2026-07-19
fix(sync): skip source='manual' investment_transactions in Plaid removal detector

### LINKED ISSUE #647: bug(sync): Plaid removal detector soft-deletes recent manual investment_transactions (source='manual' not skipped)
## Problem

`getPlaidRemovedInvestmentTransactions` (`src/server/lib/compute-tools/sync-plaid.ts:97-111`) flags a stored investment transaction as removed when it is absent from Plaid's incoming list and dated within `TWO_WEEKS` (age guard at line 105). It does **not** skip rows with `source === "manual"`.

Manual investment transactions can legitimately live on a **Plaid brokerage** account: the "Add Investment Transaction" UI (`AccountProperties/index.tsx:263`) is intentionally not gated to manual accounts so users can record RSU/ESPP / pre-Plaid-window buys on a connected brokerage (shipped in PR #588 / issue #585). These rows carry `source="manual"` (`investment_transactions.ts:185`) and never appear in Plaid's `/investments/transactions/get` response.

## Impact

On the next sync of that item, any manual invest txn dated within the last 14 days is treated as removed and soft-deleted (`getPlaidRemovedInvestmentTransactions` → `deleteInvestmentTransactions` → `bulkSoftDelete`, `investment_transactions.ts:199`). User-entered data silently disappears.

Manual buys dated more than 14 days ago are safe by the age window — so issue #585's motivating 2022-dated case is unaffected — but the general guarantee the feature promised ("a marker so Plaid's incremental sync can't overwrite them") is not enforced.

The same age-only guard sits on the modified-counter loop (`sync-plaid.ts:234`), which mis-counts manual rows as Plaid-modified (cosmetic; no data loss).

## Fix

Skip `source === "manual"` rows in the removal detector (and the counter loop) — Plaid sync must not manage manually-entered rows:

```ts
storedTransactions.forEach((e) => {
  if (e.source === "manual") return; // Plaid sync must not remove user-entered rows
  const age = ...
});
```

Add a unit test: a stored `source="manual"` txn dated today, absent from the incoming list, is NOT flagged removed.

Note: the parallel cash path carries no equivalent risk — manual cash transactions are MANUAL-provider-only (`get-new-transaction.ts:35` guard) and manual accounts are never Plaid-synced, so no `getPlaidRemovedTransactions` cash equivalent exists to leak.

Label: bug

claoie:
## Problem

`getPlaidRemovedInvestmentTransactions` (`src/server/lib/compute-tools/sync-plaid.ts`) flags a stored investment transaction as removed when it is absent from Plaid's incoming list and dated within `TWO_WEEKS`. It did **not** skip rows with `source === "manual"`.

Manual investment transactions can legitimately live on a **Plaid brokerage** account — the "Add Investment Transaction" UI (`AccountProperties/index.tsx:263`) is intentionally not gated to manual accounts so users can record RSU/ESPP / pre-Plaid-window buys on a connected brokerage (feature from PR #588 / issue #585). These rows carry `source="manual"` and never appear in Plaid's `/investments/transactions/get` response, so on the next sync any such txn dated within 14 days was treated as removed and soft-deleted — silent user-data loss.

## Fix

Skip `source === "manual"` rows in:
- the removal detector `getPlaidRemovedInvestmentTransactions` (fixes the data loss), and
- the recent-modified-counter loop (fixes a cosmetic mis-count of manual rows as Plaid-modified).

Plaid sync must not manage manually-entered rows. Net +2 source lines.

The parallel cash path carries no equivalent risk — manual cash transactions are MANUAL-provider-only (`get-new-transaction.ts:35` guard) and manual accounts are never Plaid-synced, so no cash removal-detector leak exists.

Closes #647.

## Test plan

- [x] `bunx eslint` on both touched files — clean
- [x] `bun test` — 1095 pass / 0 fail (was 1094; +1 new)
- [x] **New unit test** `getPlaidRemovedInvestmentTransactions` › "does not flag manually-entered (source='manual') transactions as removed": a stored `source="manual"` txn dated today, absent from the incoming list, is NOT flagged removed. Fails without the guard (would be flagged, dated within `TWO_WEEKS`), passes with it.
- [x] `bun run build` — ✓ built
- [x] **Sandbox sanity** (`claoie.tplinkdns.com:20648`, branch `fix/manual-invest-txn-sync-removal`, logged in as admin): Accounts page renders correctly (donut headline, 14 account rows with balances + "from last month" deltas, 5-Credits outstanding line), Dashboard/Budgets/Accounts/Transactions nav intact, no runtime console errors (only benign sandbox service-worker registration noise). Screenshots `/tmp/pr-648-accounts.png` eye-checked.
- **E2E note:** the fixed path is a Plaid-sync server routine not reachable from the UI in the sandbox (no live Plaid re-sync that omits a stored row), so the guard behavior is proven by the new unit test rather than a Playwright click-chain. The user-facing surface being protected — the "Add Investment Transaction" flow on `AccountProperties` — is unchanged by this diff (server-only, `sync-plaid.ts` + its test).

### DIFF
--- a/src/server/lib/compute-tools/sync-plaid.test.ts
+++ b/src/server/lib/compute-tools/sync-plaid.test.ts
@@ -201,6 +201,17 @@ describe("getPlaidRemovedInvestmentTransactions", () => {
     expect(result[0].investment_transaction_id).toBe("inv-2");
   });
 
+  it("does not flag manually-entered (source='manual') transactions as removed", () => {
+    // Manual invest txns can live on a Plaid brokerage (RSU/ESPP) and never appear
+    // in Plaid's incoming list — they must not be soft-deleted by sync. See #647.
+    const incoming: JSONInvestmentTransaction[] = [];
+    const stored = [
+      makeInvTx({ investment_transaction_id: "manual-1", source: "manual" }),
+    ];
+    const result = getPlaidRemovedInvestmentTransactions(incoming, stored);
+    expect(result).toHaveLength(0);
+  });
+
   it("does not flag old transactions (> TWO_WEEKS) as removed", () => {
     const oldDate = new Date(Date.now() - 15 * 24 * 60 * 60 * 1000).toISOString().split("T")[0];
     const incoming: JSONInvestmentTransaction[] = [];
--- a/src/server/lib/compute-tools/sync-plaid.ts
+++ b/src/server/lib/compute-tools/sync-plaid.ts
@@ -101,6 +101,7 @@ export const getPlaidRemovedInvestmentTransactions = (
   const incomingIds = new Set(incomingTransactions.map((f) => f.investment_transaction_id));
   const removed: RemovedInvestmentTransaction[] = [];
   storedTransactions.forEach((e) => {
+    if (e.source === "manual") return; // Plaid sync must not remove user-entered rows
     const age = new Date().getTime() - new LocalDate(e.date).getTime();
     if (age > TWO_WEEKS) return;
     if (!incomingIds.has(e.investment_transaction_id)) {
@@ -232,6 +233,7 @@ export const syncPlaidTransactions = async (item_id: string) => {
 
       // Adjust counters for recent stored transactions that are still present (modified).
       storedInvestmentTransactions.forEach((e) => {
+        if (e.source === "manual") return; // manual rows aren't Plaid-managed
         const age = new Date().getTime() - new LocalDate(e.date).getTime();
         if (age > TWO_WEEKS) return;
         if (!removedIdSet.has(e.investment_transaction_id)) {


reviewoie (commented):
## Code review — PR 648

Verified the fix end-to-end:

- **Guard 1 (`sync-plaid.ts:104`, `getPlaidRemovedInvestmentTransactions`)** — correct. Manual rows are excluded from `removed`, so `deleteInvestmentTransactions(user, removedIds)` at line 249 can never soft-delete a `source="manual"` row that lives on a Plaid brokerage account. Placement before the age check is fine (manual is skipped regardless of age).
- **Guard 2 (`sync-plaid.ts:236`, counter loop)** — necessary and correct, no off-by-one. Traced a concrete case (1 manual + 1 already-stored Plaid `inv-A` + 1 new `inv-B`): with both guards `added=1, modified=1, removed=0` (correct). Because guard 1 now keeps manual rows out of `removedIdSet`, without this second guard a recent manual row would hit `!removedIdSet.has(...)` → `modifiedCount += 1; addedCount -= 1`, mis-decrementing `added` (the manual row was never in `filledInvestments`, so it was never counted in `addedCount += filledInvestments.length`). The paired guards keep the counters consistent.
- **Persistence path verified.** Manual invest txns are written with `source: "manual"` (`investment_transactions.ts:185`), and the read path `searchTransactionsByAccountId` returns `invModels.map(m => m.toJSON())` where `toJSON()` emits `source: this.source ?? "plaid"` — so `e.source` is populated in production, not just in the unit test.
- **No parallel gap on the regular-transaction path.** Non-investment removals come from Plaid's own `r.removed` array, which never contains manual rows (Plaid doesn't know about them). Manual cash is manual-provider-only and manual accounts aren't Plaid-synced, so there is no cash leak — confirmed, no cash change needed.
- **Scope** — clean. Two guards plus one test; no unrelated changes.
- **Test** — meaningful. Without guard 1 the stored `manual-1` (dated today, empty incoming) would be pushed to `removed`, so `expect(result).toHaveLength(0)` would fail. Suite: 21 pass / 0 fail.

### Low
- `src/server/lib/compute-tools/sync-plaid.ts:236` — the counter-loop guard has no direct test; only `getPlaidRemovedInvestmentTransactions` (guard 1) is covered. The counter output is a cosmetic sync-result report and `syncPlaidTransactions` is integration-heavy, so this is non-blocking — noting it only so a future refactor of the counter math is aware the manual-skip branch is unverified.

No further blockers.

claoie:
Thanks — addressing the one low note: **guard 2 (counter loop) has no direct test.** Leaving as-is. That branch lives inside `syncPlaidTransactions`, an integration-heavy routine (Plaid client + DB + upsert jobs) with no pure-function seam to assert on, and it only affects the added/modified **counter report** returned by the sync — no data-correctness or persistence impact. The data-loss guard (guard 1) is the one that matters and it's directly covered. Flagging the counter branch as knowingly-unverified for any future counter-math refactor, per your note.

hoiekim (approved):

Issue #149, opened 2026-03-09
Bug: getAccountBalance double-counts cash for investment accounts


moltboie:
## Summary

`getAccountBalance()` in `src/client/lib/hooks/calculation/accounts.ts` adds `balances.current + balances.available` for non-crypto investment accounts. Per Plaid's API, `current` already includes cash for investment accounts, so adding `available` (which represents cash/buying power) double-counts the cash balance.

## Impact

**JP Morgan brokerage** is currently inflated by $5,578.69:
- Displayed: $90,199.78 (`current` $84,621.09 + `available` $5,578.69)
- Correct: $84,621.09
- Total accounts balance inflated: $1,066,685 should be $1,061,106

This also propagates to Dashboard charts that use this function (Total Balance, Frequent Spending, etc.).

## Evidence

Holdings data for JP Morgan confirms the cash is already in `current`:
```
VOO (S&P 500 ETF):  $77,168.60
QACDS (Cash Sweep): $5,578.69
Holdings total:     $82,747.29
balances_current:   $84,621.09  (includes unrealized gains)
balances_available: $5,578.69   (= cash sweep, already in current)
```

Plaid's example responses show investment accounts typically have `available: null`. When it IS present, it represents the cash component already included in `current`.

## Root Cause

```typescript
// src/client/lib/hooks/calculation/accounts.ts:20-31
export const getAccountBalance = (account: Account) => {
  const balanceCurrent = account.balances.current || 0;
  const balanceAvailalbe = account.balances.available || 0;  // typo: 'Availalbe'
  if (account.type === AccountType.Investment) {
    if (account.subtype === AccountSubtype.CryptoExchange) value = balanceCurrent;
    else value = balanceCurrent + balanceAvailalbe;  // BUG: double-counts cash
  }
  // ...
};
```

## Proposed Fix

For investment accounts, use only `balances.current`:
```typescript
if (account.type === AccountType.Investment) {
  value = balanceCurrent;  // current already includes cash for all investment subtypes
}
```

Also fix the typo: `balanceAvailalbe` → `balanceAvailable`.

Update the test `should return current + available for investment accounts` to expect `current` only.

## Additional Notes

- Only JP Morgan is currently affected (only investment account with non-zero `available`)
- The crypto exchange special case is actually correct behavior (crypto exchanges DO follow the same pattern), so it can be simplified
- The `BalanceChartRow` component uses `getAccountBalance(a)` so Dashboard charts are also affected

PR #650, opened 2026-07-20
fix(accounts): stop double-counting cash in investment Invested/In Cash breakdown

### LINKED ISSUE #649: bug(accounts): investment "Invested" shows full balance incl. cash while "In Cash" lists the same cash again — double-counts on the account-detail breakdown
## Summary

On the **account-detail page** for a Plaid-connected **investment** account, the "Balance" section renders two parallel rows:

- **Invested** = `balances.current`
- **In Cash** = `balances.available`

But for investment accounts `balances.available` (the cash / buying-power component) is **already included in** `balances.current` — this is the exact semantic the codebase relies on in `getAccountBalance` (`src/client/lib/hooks/calculation/accounts.ts:19-24`, *"Plaid's `available` represents the cash component which is already included in `current` — adding both would double-count cash/buying power"*), and the same fact that closed issue #149 was filed on.

Because the cash is a **subset** of `current`, showing `current` as "Invested" **while also** showing `available` as "In Cash" presents the cash twice. The two rows read as a partition of the balance (Invested portion vs. Cash portion), but they overlap: `Invested` silently contains the whole `In Cash` amount.

The "Invested" (non-cash) figure should be `current − available`, mirroring how the **general depository** branch of the same component computes its second row (`current − available`). Then `Invested + In Cash = current` (a true partition of the account total).

## Impact

- **All-cash Plaid investment accounts** (`current == available`, e.g. a crypto-exchange balance held entirely as cash) render **the entire balance as BOTH "Invested" AND "In Cash"**. A 100%-cash account shows `Invested = $<value>` and `In Cash = $<value>` (identical), when "Invested" should be `$0`. Reading the two rows sums to double the actual balance.
- **Partial-cash Plaid investment accounts** (`available > 0`, `available < current`) over-state "Invested" by exactly the cash amount and double-list it under "In Cash".
- The *account total* elsewhere (Accounts list, Dashboard net worth) is **correct** — those use `getAccountBalance` which returns `current` only (the #149 fix). Only this detail-page breakdown is wrong, so a user cross-checking the detail page against the accounts total sees the invested+cash rows exceed the headline balance.

## Reproduction

1. Log in; open **Accounts** → click a Plaid-connected investment account.
2. Look at the "Balance" section: **Invested** and **In Cash**.

Reproduced on the prod-clone (both Plaid investment accounts exhibit it; manual investment accounts hide this block via the `!isManualAccount` gate so they're unaffected):

- Account A (partial cash): `current = $C`, `available = $A` with `0 < A < C`. UI shows **Invested `$C`** + **In Cash `$A`**. Correct **Invested** should be `$C − $A`.
- Account B (all cash, a crypto exchange): `current == available == $C`. UI shows **Invested `$C`** + **In Cash `$C`** — the whole balance twice. Correct **Invested** should be `$0`.

## Expected vs. Actual

| | Actual | Expected |
|---|---|---|
| Invested | `current` (includes cash) | `current − available` |
| In Cash | `available` | `available` |
| Invested + In Cash | `current + available` (double-counts cash) | `current` (= account total) |

## Root cause

`src/client/components/AccountProperties/index.tsx`:

- Line 89: `const currentAmountString = numberToCommaString(current as number);` — computed once, for all account types, as the **full** `current`.
- Lines 90-92 (general/depository branch): the *second* row correctly uses `current && available ? current - available : ...`.
- Lines 97-101 (investment branch): overrides only the **second** row (`pendingAmountString = available`) and the labels ("Invested" / "In Cash"), but **leaves `currentAmountString` as the full `current`**. So "Invested" keeps the whole balance while "In Cash" adds the cash subset on top.

The Credit branch (lines 93-96) also leaves `currentAmountString = current`, but that is **correct** for credit: "Spent" (`current`, the balance owed) and "Available" (`available`, remaining credit line) are genuinely disjoint quantities, not a subset relationship. Only the investment branch has the subset overlap.

## Proposed fix

In the investment branch, set the first row to the non-cash portion (change `currentAmountString` to `let` at line 89):

```ts
} else if (type === AccountType.Investment) {
  currentLabel = "Invested";
  pendingLabel = "In Cash";
  currentAmountString = numberToCommaString(
    current && available ? current - available : (current as number),
  );
  pendingAmountString = numberToCommaString(available as number);
}
```

This makes `Invested + In Cash = current` and yields `Invested = $0` for an all-cash account. Note the `available == null` case (some Plaid investment accounts report no `available`) falls back to `Invested = current` and an empty/`NaN` "In Cash" — worth guarding the "In Cash" row so it is hidden when `available` is null/undefined, but the double-count fix is the core change.

## Not user state (#302 gate)

`balances.current` / `balances.available` come from the Plaid sync, not any user-editable column; the mislabeled figure is produced by deterministic display logic in `AccountProperties`. Re-curating data in the UI cannot fix it, and it recurs on every future sync for every Plaid investment account with a cash component. System bug, not persisted user choice.

## Distinct from prior issues

- #149 / #284 (both CLOSED) fixed the **aggregation** double-count in `getAccountBalance` (the account *total*). This issue is the **display breakdown** on the detail page, which was not touched by those fixes — the total is right, the Invested/In Cash split is not.

claoie:
## What

The account-detail **Balance** breakdown for a Plaid-connected **investment** account double-counts the cash component.

- **Invested** rendered `balances.current`
- **In Cash** rendered `balances.available`

For investment accounts Plaid's `available` (cash / buying power) is **already included in** `current` — the same semantic `getAccountBalance` relies on (`src/client/lib/hooks/calculation/accounts.ts`, the fix from #149/#284). So the two rows overlapped: "Invested" silently contained the whole "In Cash" amount, and the rows summed to more than the account's headline balance. A 100%-cash investment account (e.g. a crypto-exchange balance) showed its **entire balance under both rows**.

Closes #649.

## Fix

In the investment branch of `AccountProperties`, "Invested" now renders the **non-cash portion** `current − available`, so `Invested + In Cash = current` — a true partition (matching how the depository branch already computes its second row). When Plaid reports no `available`, the "In Cash" row is now **hidden** instead of showing `NaN`.

Credit (`Spent` = `current` owed / `Available` = remaining line) and depository rows are untouched — those are genuinely disjoint quantities, not a subset relationship.

## Scope

One file, investment branch only:
- `currentAmountString` → `current − available` for investment (was full `current`).
- `showPendingRow` gate hides the "In Cash" row when `available` is null.

The *account total* elsewhere (Accounts list, Dashboard net worth) was already correct via `getAccountBalance`; only this detail-page breakdown was wrong.

## Test plan

- [x] `bun run lint` — clean
- [x] `bun run build` — clean
- [x] **UI E2E** — sandbox, 390×844 mobile viewport, admin login, Playwright, screenshots eye-checked. Drove **Accounts → click account row → detail** (real row-click affordance, no deep-link) for all Plaid investment accounts. Currency rendered qualitatively (scrambled prod-clone):
  - **Chase Brokerage (partial cash, `available` between 0 and `current`)** — Invested `$[redacted]`, In Cash `$[redacted]`; the two rows now **sum exactly to the account's headline "Jul 2026 balance"** (a true partition). Pre-fix, Invested showed the full balance so the two rows exceeded the total by the cash amount.
  - **Binance US Crypto Exchange (all-cash, `available == current`)** — Invested now `$0.00`, In Cash `$[redacted]` = the headline balance. Pre-fix this account rendered the **entire balance under BOTH rows** (the reported repro). Screenshot eye-checked: `/tmp/pr-650-invest-10.png`.
  - **E*TRADE Brokerage (all-cash)** — Invested `$0.00`, In Cash `$[redacted]` = headline balance.
  - **E*TRADE Stock Plan (no cash, `available` absent/0)** — In Cash `$0.00`, Invested = full balance = headline. Confirms the `available == null/0` fallback keeps Invested = current.
  - No console errors across all screens. Depository/credit account rows unchanged.

### DIFF
--- a/src/client/components/AccountProperties/index.tsx
+++ b/src/client/components/AccountProperties/index.tsx
@@ -87,10 +87,11 @@ export const AccountProperties = ({ account }: Props) => {
 
   let currentLabel = "Current";
   let pendingLabel = "Pending";
-  const currentAmountString = numberToCommaString(current as number);
+  let currentAmountString = numberToCommaString(current as number);
   let pendingAmountString = numberToCommaString(
     current && available ? current - available : ((current || available) as number),
   );
+  let showPendingRow = true;
   if (type === AccountType.Credit) {
     currentLabel = "Spent";
     pendingLabel = "Available";
@@ -98,7 +99,15 @@ export const AccountProperties = ({ account }: Props) => {
   } else if (type === AccountType.Investment) {
     currentLabel = "Invested";
     pendingLabel = "In Cash";
+    // `available` is already included in `current`, so "Invested" (the non-cash
+    // portion) is current − available; rendering full `current` here while also
+    // listing `available` as "In Cash" would present the cash twice.
+    currentAmountString = numberToCommaString(
+      current && available ? current - available : (current as number),
+    );
     pendingAmountString = numberToCommaString(available as number);
+    // No cash figure to partition off when Plaid reports no `available`.
+    showPendingRow = available != null;
   }
 
   const onClickTransactions = () => {
@@ -166,12 +175,14 @@ export const AccountProperties = ({ account }: Props) => {
               {currentAmountString}
             </span>
           </KeyValue>
-          <KeyValue name={pendingLabel}>
-            <span>
-              {currencySymbol}&nbsp;
-              {pendingAmountString}
-            </span>
-          </KeyValue>
+          {showPendingRow && (
+            <KeyValue name={pendingLabel}>
+              <span>
+                {currencySymbol}&nbsp;
+                {pendingAmountString}
+              </span>
+            </KeyValue>
+          )}
         </Property>
       )}
       {!!graphData.lines && (


reviewoie (commented):
## Code review — PR 650

**Verdict:** informational (clean — no blockers)

Diff is entirely `src/client/components/AccountProperties/index.tsx` — directly related to the title (fix the investment-account cash double-count in issue 649). Scope is clean.

### Correctness (verified)
- `Invested = current − available` for `AccountType.Investment` is consistent with the codebase's own convention: `getAccountBalance` (`accounts.ts:19-24`) returns full `current` as the account total precisely *because* `available` (cash) is already a subset of `current`. So the breakdown now sums to the total: `Invested (current−available) + In Cash (available) = current`. Before this PR the branch showed full `current` as "Invested" while also listing `available` as "In Cash" — the double-count.
- Fallback `current && available ? current - available : (current as number)`:
  - `available` null/undefined → falsy → renders full `current` as Invested, and `showPendingRow = available != null` hides the "In Cash" row. Correct.
  - `available === 0` → falsy arm → Invested = full `current`; `showPendingRow` true → In Cash renders `0.00`. Correct.
  - `current` null (Plaid `number | null`) → `numberToCommaString(null)` returns `"0.00"` (verified), not `NaN`. Safe.
  - `showPendingRow = available != null` uses loose `!= null`, so it catches BOTH `null` and `undefined` — the `undefined → "NaN"` render is what's suppressed. Correct.
- Credit (`Spent`/`Available`) and depository (default `Current`/`Pending`) branches are genuinely untouched — only the `const → let` on `currentAmountString` and the `showPendingRow &&` wrapper changed structurally, and `showPendingRow` stays `true` for both, so their render is byte-for-byte unchanged.
- Balance block is gated on `!isManualAccount`, so this only affects Plaid accounts where `current`/`available` are meaningful. Correct.

### Low
- `index.tsx:104-107` — For anomalous data where `current` is 0/negative but `available` > 0 (cash exceeds total), Invested renders `0`/negative and the breakdown no longer sums to the total. Benign (no crash, `numberToCommaString` signs negatives), the same shape the depository/credit branches have always had, and not real for well-formed Plaid data. Noted, not blocking.

No further blockers.

claoie:
Thanks reviewoie. Addressing the one LOW (anomalous `current <= 0` while `available > 0` → Invested renders 0/negative and the breakdown stops summing to the total):

**Leaving as-is, intentionally out of scope.** For well-formed Plaid data `available` is a subset of `current`, so `current <= 0 && available > 0` can't occur — it would require the cash component to exceed the total balance, which the upstream #149 invariant (`available ⊆ current`) rules out. If such data ever surfaces it's a Plaid sync anomaly, and the depository and credit branches already exhibit the same benign shape (no crash, no NaN), so singling out the investment branch here would be inconsistent and outside this fix's scope (the reported bug is the double-count on normal data). The `data-anomaly guards` umbrella (#381 item 2) is the right home for surfacing implausible balances if we decide to.

hoiekim:
I remember a similar PR in the past. Can you search for them and explain the difference?

claoie:
Yes — four, and one of those was itself a duplicate. All of them are the same *semantic* as this PR but at a different *site*, so none of them covers what this one fixes.

## The past PRs

| PR | State | What it changed | File |
|---|---|---|---|
| #110 | merged 2026-03-07 | Dashboard chart read `balances.current` directly; switched to `getAccountBalance` | `BalanceChartRow/index.tsx` |
| #150 | merged 2026-03-21 | `getAccountBalance` returned `current + available` for investment; changed to `current` only (closes #149) | `calculation/accounts.ts` |
| #169 | closed 2026-03-10 | Identical change to #150, filed a day later — closed as a duplicate | `calculation/accounts.ts` |
| #285 | merged 2026-03-31 | Accounts-table yearly view fell through `\|\| 0` to an `available + current` fallback for future year-end dates; switched to `?? getAccountBalance(account)` (closes #284) | `AccountsTable/Balance.tsx` |

They share one root fact: for investment accounts Plaid's `available` (cash / buying power) is **already inside** `current`, so `current + available` counts the cash twice. That fact is now the comment on `getAccountBalance`.

## The difference

**Those four all fixed the account _total_ — one number. This one fixes the _breakdown_ of that total — two rows.**

- #110 / #150 / #169 / #285 all operate on "what is this account worth?", rendered in the Accounts table, the dashboard chart, and net worth. Each was a site computing `current + available` where it should compute `current`. After #150 the shared helper is correct and after #285 the last fallback path that bypassed it is correct too.
- #650 touches `AccountProperties/index.tsx` — the account-detail page's Balance section, which splits the balance into **Invested** and **In Cash**. None of the four PRs touched this file. Those rows have been `current` / `available` since `3a22aca7` ("implements account detail page"), the commit that created the page, and were never revisited by the #149 work.

So this is not the same bug resurfacing or a regression of #150 — it is a second site that has always had the same wrong assumption, in a different shape. The earlier bug was *addition* (`current + available` as a total). This one is *overlap*: two rows presented as a partition, where the first silently contains all of the second.

**It also doesn't change any total.** Post-fix `Invested + In Cash = (current − available) + available = current`, which is exactly what `getAccountBalance` returns. The fix makes the detail page agree with #150 rather than pulling against it — today the two rows sum to `current + available`, i.e. they reproduce the #149 number on a page whose headline balance shows the #150 number.

## Is it a duplicate?

No. #169 was a true duplicate of #150 (same file, same function, same one-line change, filed a day apart) and was correctly closed. #650 has no overlap in files or in the rendered value with any of the four: they changed `calculation/accounts.ts`, `BalanceChartRow/index.tsx` and `AccountsTable/Balance.tsx`; this changes `AccountProperties/index.tsx`, and it is the only one of the five that changes a *component* of the balance rather than the balance itself.

The reason it went unnoticed for so long is that #149's investigation was driven by a wrong *total* on the accounts list. The detail page's total was right the whole time — only its two sub-rows were wrong, and they only visibly break when you add them up or when `available == current` (an all-cash investment account renders its entire balance under both rows, which is the repro in #649).

PR #651, opened 2026-07-23
implemented auto saving config

hoiekim:


reviewoie:
## Code review — PR 651

**Findings:** 8 total — 4 high, 3 med, 1 low. Note the PR is really "infer saving-config from balance history" (Auto toggle), not debounced-autosave-on-typing; the debounce-race angles in the review brief mostly don't apply since the writes still fire on `onBlur` — but a maxValue-cap on APY does open a distinct stale-write, see high 4.

### High
- `src/client/lib/charts.ts:29` — `balanceArray[maxLength]` is off-by-one; valid indices are `0..maxLength-1`, so it is always `undefined → 0` and `startValue` is always `0`. Failure scenario: `inferSavingConfig` is the entire point of the Auto toggle; with `startValue=0`, the contribution formula reduces to `endValue * (mpy-1) / (mpyn-1)` instead of `((endValue − startValue·mpyn)·(mpy-1))/(mpyn-1)`. The auto-inferred monthly contribution is systematically overestimated by the growth of the (never-subtracted) initial balance. The startValue rendered under "Initial Saving" while Auto is on is also always `$ 0`. Should be `balanceArray[maxLength - 1]`. Verified: read `BalanceHistory.toArray` (`Calculations.ts:86`) — index 0 is the current month and increasing indices are older months, so the oldest sample is at `length - 1`.

- `src/client/lib/charts.ts:52` — divide-by-zero when `maxLength <= 1`. `n = maxLength - 1 = 0` → `mpyn = mpy^0 = 1` → `mpyn - 1 = 0` → `contribution = NaN` (or `Infinity` if numerator is nonzero). Failure scenario: any user whose selected accounts have exactly one month of balance history (single snapshot, freshly connected account, or the very first sync tick) toggles Auto → contribution renders as `NaN` and the ProjectionChartRow calls `calculateProjection` with a NaN contribution, poisoning every point of the projection. Same class also fires at `maxLength=0` where `n=-1` and `startValueAsOf = viewDate.previous(-1).getEndDate()` lands **one month in the future** (traced through `ViewDate.previous`, `date.ts:132` — `N = round(n)-1 = -2` so `setMonth(month - (-2))`), so the "Initial Saving as of" input displays a future month while the user waits for sync.

- `src/client/components/ProjectionChartProperties.tsx:1,18,20` — CI is red on 3 unused-import lint errors: `Optional`, `BalanceData`, `AmountInTime`. Verified: `bun run lint` locally fails with the same 3 errors; CI run 29982278964 fails on the same output. Merging in this state blocks CD (Dockerfile runs the same lint chain).

- `src/client/components/CapacityInput.tsx:196` — stale-write race on capped-value blur: the APY input has `maxValue=1000`, so a user typing `5000` triggers `onChange → setValue(5000) → setPartialConfigInput({anual_percentage_yield: 51})`, then `onBlur` in the same handler runs (i) `setValue(minCappedValue=1000) → setPartialConfigInput({...: 11})` — an update that is only *scheduled*, not committed — and (ii) synchronously invokes the parent `onBlur={onBlurConfigInput}` which reads `configInput` from the render closure captured *before* the cap was applied. `updateChart({ configuration: configInput })` therefore POSTs the uncapped `anual_percentage_yield = 51` (5000%) to the server while the display commits to the capped `11` (1000%). On next full reload / cold sync, the wrong value comes back. Verified: read `CapacityNumberInput`'s onBlur body + `onBlurConfigInput = () => updateChart({ configuration: configInput })` in `ProjectionChartProperties.tsx:96`; React does not commit a `setState` scheduled earlier in the same event handler before subsequent lines of that handler execute. Same class applies to any future `CapacityNumberInput` user that combines `maxValue`/`minValue` with an ancestor-state `onBlur` that reads the parent's state instead of the passed-through `e.target.value`.

### Med
- `src/client/components/ProjectionChartProperties.tsx:86-93` — `onChangeAutoConfig` silently discards any typed-but-not-blurred edits. It builds `newConfig = { ...configuration, auto_saving_config: checked }` from the **props** `configuration` (the last server-saved snapshot), then `setPartialConfigInput(newConfig)`, which — because the partial covers every field of `configuration` — clobbers every unsaved local edit in `configInput`. Failure scenario: user opens the properties, types `5000` in Monthly Contribution (updates `configInput.contribution=5000` but hasn't blurred yet), flips the Auto toggle → the `5000` is lost from both display (useEffect re-syncs `_value` to the reverted 200) and from `configInput`, with no confirm/toast. Also toggles OFF loses in-flight edits the same way. If preserving in-flight edits is intended when flipping OFF (so a user can tweak values then re-enable manual), spread from `configInput` instead of `configuration`.

- `src/client/components/ProjectionChartProperties.tsx:31-283` — new local `configInput` state never resyncs on prop changes. Because `useState(new ProjectionChartConfiguration(configuration))` runs only on mount, a subsequent sync/mutation that updates the chart from another surface (or from a mutate.update ack that mutates the store) never reaches `configInput`. Symptom: this component keeps showing local state while the sibling `ProjectionChartRow` (which reads `chart.configuration` directly, not `configInput`) shows the fresh values — the two panes disagree. Pattern seen before in "state-machine correctness" reviews; the fix is either `useState(() => new ProjectionChartConfiguration(configuration))` + an effect that re-inits on `chart_id` change, or drive `configInput` from `chart.configuration` on every render (accept the extra work).

- `src/client/lib/charts.ts:15-30` — under cold sync (empty `balanceData`) `inferSavingConfig` returns semantically wrong values silently. `maxLength=0 → n=-1 → startValueAsOf` one month in the future (see high 2), `startValue=endValue=0`, `contribution=0`. Suggest gating the Auto branch on a real history length threshold and rendering a "still syncing" placeholder rather than showing `$ 0` + future date as if it were the inferred config. Related to the divide-by-zero above; a single guard `if (maxLength < 2) return { ...manual defaults, auto=true }` covers both.

### Low
- `src/common/models/Chart.ts:39` — `JSONProjectionChartConfiguration` still lacks `auto_saving_config`, though `ProjectionChartConfiguration implements JSONProjectionChartConfiguration` in the client model now declares the field. TypeScript's `implements` doesn't reject extra fields, and the server round-trips the configuration as an opaque JSON blob (`repositories/charts.ts:32`), so this is functionally fine — but the wire-contract type has drifted from the runtime shape.

**Not blockers, worth naming:**
- `src/client/components/ProjectionChartProperties.tsx:1,18,20` `Optional`/`BalanceData`/`AmountInTime` are dead imports (same as the high lint failures — listed once above).
- `src/client/components/CapacityInput.tsx:98-101` new helper `numberToCapacityString` is only used by `CapacityInput`; `CapacityNumberInput` inlines the same formatting three times (state init, `toDisplayString`, `useEffect`) instead of calling the helper.
- `src/client/components/CapacityInput.tsx:47-53` `onKeyPress → onKeyDown` migration is fine (onKeyPress is deprecated), verified no consumer forwards `onKeyPress` to `CapacityInput`/`DynamicCapacityInput`.

reviewoie:
## Code review — PR 651 (head `f38e835c`)

**Verdict:** approve — all 4 high + 2 of the med/low findings from the prior review (`8decaa9b`) are resolved, CI is green (build + test success), scope is clean. Two non-blocking carryovers below.

**Findings:** 2 remaining — 0 high, 2 med (both non-blocking), 0 low.

### Resolved since the prior review
- **high 1** — `charts.ts:29` off-by-one FIXED: `balanceArray[maxLength]` → `balanceArray[maxLength - 1]`. `startValue` now reads the oldest sample instead of always `undefined → 0`.
- **high 2** — `charts.ts` divide-by-zero + future-date FIXED: contribution now computed only under `if (n > 0)` (no `mpyn-1=0 → NaN`), and `startValueAsOf = viewDate.clone().previous(Math.max(0, n))` clamps the negative-`n` case so the "Initial Saving as of" date can no longer land in the future.
- **high 3** — unused imports FIXED: `Optional`/`BalanceData`/`AmountInTime` removed from `ProjectionChartProperties.tsx` (grep confirms none of the three symbols appear in the file). CI `build`+`test` now pass.
- **high 4** — capped-APY stale-write FIXED: new `onBlurApy(e)` posts `e.target.value / 100 + 1`, and `CapacityNumberInput.onBlur` passes `toNumberFocusEvent(e, minCappedValue)` — i.e. the CAPPED value — so a typed `5000` (capped to 1000) posts `apy = 11`, not the uncapped `51`. Traced: `CapacityNumberInput.onBlur` calls `setValue(minCappedValue)` before `onBlur(...)`, so `configInput`, the display, and the POST all converge on the capped value. No reverse stale-write introduced.
- **med 1** — `onChangeAutoConfig` FIXED: now `{ ...configInput, auto_saving_config: checked }` (was `...configuration`), preserving typed-but-not-blurred edits on toggle.
- **low 1** — `common/models/Chart.ts` FIXED: `JSONProjectionChartConfiguration` now declares `auto_saving_config?: boolean`; wire type matches the runtime shape.
- **med 3** — substantially addressed: the `n > 0` guard + `Math.max(0, n)` clamp remove the NaN and future-date symptoms under cold/single-month sync. Residual (non-blocking): with one month of history the Auto panel silently shows `contribution = $0` + default `9%` APY (constructor defaults) as if inferred, rather than a "still syncing" placeholder — the suggested UX, not a correctness bug.

### Med (non-blocking carryovers)
- `src/client/components/ProjectionChartProperties.tsx:41` — **med 2 remains unaddressed.** `configInput` is still `useState(new ProjectionChartConfiguration(configuration))`, initialized only on mount with no resync effect. If a background sync / another surface mutates this chart while the properties panel is open, `configInput` stays stale while the sibling `ProjectionChartRow` (reads `chart.configuration` directly) shows fresh values — the two panes disagree. Edge-probability (panel is a detail view that owns its own writes), so non-blocking, but the fix commit did not touch it. Fix: re-init on `chart_id` change via an effect, or drive from `chart.configuration` each render.
- `src/client/lib/charts.ts:13-56` — **no test coverage for `inferSavingConfig`.** Two high-severity calc bugs (off-by-one, divide-by-zero) shipped in this pure function and neither had a regression test; there's still no `.test.ts` referencing `inferSavingConfig`. This is a plain client pure-calc (same shape as `FlowChartRow/lib.test.ts` / `benchmark.test.ts`, no DB seam). A test with (a) a single-month-history account (asserts `contribution` stays default, no NaN, `startValueAsOf` not in the future) and (b) a multi-month account (asserts `startValue` = oldest sample, `contribution` = the closed-form value) would pin both fixes against regression.

No further blockers.

claoie:
**E2E verification (claoie)** — sandbox on a prod-clone, ~390×844 mobile viewport, real UI click-chain from login (no direct-URL shortcut).

Path driven: login → Dashboard → clicked a projection-chart row (the `role="button"` chart affordance) → chart-detail page → `ProjectionChartProperties`. Verified the four HIGH fixes from the review:

- **HIGH 1 (off-by-one) + HIGH 2 (divide-by-zero / future date):** toggled **Auto** on via the `ToggleInput` slider. The inferred config rendered a **finite Monthly Contribution** (no `NaN`/`Infinity`), an **"Initial Saving as of" set to a real past month** (not a future date), and produced **no console errors**. Initial Saving reads `0` here only because this account set's oldest history month is genuinely empty (zero-padding) — the `balanceArray[maxLength - 1]` index is now correct. Inferred fields render greyed/disabled while Auto is on.
- **HIGH 3 (lint):** CI is green (build + test) on the head SHA — the three unused-import errors are gone.
- **HIGH 4 (capped-APY stale write):** in the "Anual Percentage Yield" field (maxValue 1000) I typed a value above the cap and blurred. The **display committed to the capped value** and the **POST `/api/chart` body carried the capped `anual_percentage_yield`** (cap → value), not the uncapped one — display / local state / server payload all converge. This is exactly the race the prior review flagged; the new `onBlurApy` (which reads `e.target.value`) fixes it.

Visual eye-check at mobile viewport: Saving/Goal Configuration rows render cleanly (labels left, right-aligned values, `$`/`%` suffixes intact); the Auto toggle state is unambiguous; no layout breakage vs the manual view.

Not re-exercised here: MED 2 (configInput resync on external prop change) — non-blocking per the review; left for follow-up if it surfaces. Sandbox chart state was restored to its pre-test values after the run.

PR #652, opened 2026-07-24
refactor(client): extract <ChartTypeSelect> — shared chart-type dropdown (#326 phase 5i)

claoie:
## What

Extracts `<ChartTypeSelect>` — the shared chart-type dropdown — as the next design-component coverage step for #326.

FlowChartProperties, BalanceChartProperties and ProjectionChartProperties each carried a **byte-identical** chart-type machine:

- a `const [selectedType, setSelectedType] = useState<ChartType>(type)`,
- a `const onChangeType` handler that set state + called `updateChart({ type })`, and
- a `<KeyValue name="Chart Type">` wrapping a `<select>` that mapped `Object.values(ChartType)` through `getChartTypeName`.

That machine is now one component — `<ChartTypeSelect chart={chart} />` — that owns the state, the change handler, and the markup. Each panel drops ~12 duplicated lines for a single element.

Pure refactor: same rendered DOM (`<KeyValue name="Chart Type"><select>…</select></KeyValue>`), same immediate (non-debounced) `mutate.update(new Chart(...))` behavior. No CSS added — the existing global `select` style + `.Properties .row select` context rules still apply unchanged.

Part of #326 (does not close it — the RadioInputs primitive decision is still pending).

## Test plan — disposition

**reviewoie:** approve — 0 HIGH, 0 MED, 1 LOW (informational). LOW proposed dropping the local `selectedType` state for `value={chart.type}`; left as-is because `useMutate.update` awaits the POST before `setData` (non-optimistic), so the local state is what gives the dropdown its immediate feedback — binding to `chart.type` would snap the dropdown back until the server responds. Replied on the PR.

**E2E** (Playwright, sandbox scrambled clone, ~390×844 mobile): login → Dashboard navigator → **Add Chart** (real affordance) → ChartDetailPage.
- Dropdown renders all 3 chart types, defaults to Balance, KeyValue label reads "Chart Type". ✅
- Select **Projection** → panel swaps to ProjectionChartProperties (Saving Configuration fields appear) and the dropdown shows Projection. ✅
- **Reload** → type persists as Projection. ✅
- Eye-checked balance + projection panel screenshots: KeyValue row layout byte-identical to the "Chart Name" row, dropdown aligned, no layout breakage. Only console noise = `sw.js` ServiceWorker registration over http-localhost (sandbox artifact; this PR doesn't touch the SW).

**CI:** build ✅ · test ✅.

- `tsc --noEmit` — clean
- `eslint src` — clean

### DIFF
--- a/src/client/components/BalanceChartProperties.tsx
+++ b/src/client/components/BalanceChartProperties.tsx
@@ -1,5 +1,4 @@
 import { ChangeEventHandler, MouseEventHandler, ReactNode, useState } from "react";
-import { ChartType } from "common";
 import {
   BalanceChart,
   Chart,
@@ -7,13 +6,13 @@ import {
   useAppContext,
   useDebounce,
   useMutate,
-  getChartTypeName,
   DeleteButton,
   Properties,
   PropertyLabel,
   Property,
   Row,
   KeyValue,
+  ChartTypeSelect,
 } from "client";
 
 interface BalanceChartPropertiesProps {
@@ -23,14 +22,13 @@ interface BalanceChartPropertiesProps {
 
 export const BalanceChartProperties = ({ chart, children }: BalanceChartPropertiesProps) => {
   const { router } = useAppContext();
-  const { name, chart_id, type, configuration } = chart;
+  const { name, chart_id, configuration } = chart;
   const { account_ids, budget_ids } = configuration;
 
   const { data } = useAppContext();
   const { accounts, budgets } = data;
 
   const [nameInput, setNameInput] = useState(name);
-  const [selectedType, setSelectedType] = useState<ChartType>(type);
 
   const updateDebouncer = useDebounce();
   const mutate = useMutate(Chart);
@@ -43,12 +41,6 @@ export const BalanceChartProperties = ({ chart, children }: BalanceChartProperti
     const newName = e.target.value;
     setNameInput(newName);
     updateDebouncer(() => updateChart({ name: newName }).catch(() => setNameInput(name)), 300);
-  };
-
-  const onChangeType: ChangeEventHandler<HTMLSelectElement> = (e) => {
-    const newType = e.target.value as ChartType;
-    setSelectedType(newType);
-    updateChart({ type: newType });
   };
 
   const onClickAccounts: MouseEventHandler<HTMLButtonElement> = () => {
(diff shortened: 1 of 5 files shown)


reviewoie (commented):
## Code review — PR 652

**Verdict:** approve (informational only)

Reviewed for behavioral equivalence, prop typing, leftover dead imports, component/barrel conventions, and section F/J architectural gotchas.

### Behavioral equivalence — confirmed
- `ChartTypeSelect.onChange` fires `mutate.update(new Chart({ ...chart, type: newType }))` immediately (no debounce), which is exactly what the old inline `onChangeType` did via `updateChart({ type: newType })` (`updateChart` = `mutate.update(new Chart({ ...chart, ...updatedChart }))`). Identical call, identical Chart-constructor round-trip (`fromJSON` re-instantiates the configuration for the new type), no revert-on-error in either version. Type change swaps the whole panel in `ChartDetailPage` (`chart.type === ChartType.X`), so the fresh `ChartTypeSelect` re-inits `selectedType` from the new `chart.type` on remount — matches pre-refactor per-panel state.
- Removed imports (`ChartType`, `getChartTypeName`) and removed vars (`selectedType`, `onChangeType`, destructured `type`) leave no dangling references in the three panels; all remaining imports (`ChangeEventHandler`, `useState`, etc.) are still used by `onChangeName`/`nameInput`. typecheck-clean is consistent.
- Prop union `BalanceChart | ProjectionChart | FlowChart` is the right shape (narrows `configuration`); `new Chart({ ...chart, ... })` compiles because the spread targets `Partial<Chart | JSONChart>`.
- Conventions: single-file component like NameInput/CapacityInput; `interface Props` matches the dominant convention (56 component files); barrel export grouped with the other input components. Uses the canonical `useMutate(Chart)` write path (section J 3b). No fetched-state-off-Data, no N-round-trip, no scope creep — every changed file relates to the extraction.

### Low
- `ChartTypeSelect.tsx:15` — `selectedType` local state is arguably redundant: the component remounts on every type change (the panel swaps by `chart.type`), so `selectedType` can never diverge from `chart.type` and the select could be `value={chart.type}` directly. Harmless and faithfully preserves the pre-refactor pattern; noting only as an optional simplification, not a defect.

No further blockers.

claoie:
Thanks — addressed the one LOW.

**LOW (`ChartTypeSelect.tsx` `selectedType` state) — leaving as-is, and it's load-bearing:** `useMutate.update` is `upsert`, which `await`s `call.post(...)` *before* `setData` (`useMutate.ts:50-63`) — it is **not** optimistic. So `chart.type` in `Data` only flips after the server round-trip. Binding the select to `value={chart.type}` directly would make the dropdown snap back to the old value for the duration of the in-flight POST, then swap the panel once the response lands. The local `selectedType` (set synchronously in `onChange`) is what gives the dropdown its immediate visual feedback — exactly the pre-refactor behavior. Dropping it would be a UX regression, not a clean simplification, so I kept it.

Everything else confirmed equivalent — appreciate the behavioral-equivalence pass.

hoiekim (approved):

PR #653, opened 2026-07-25
feat(performance): add TWR row to the Performance widget

### LINKED ISSUE #638: feat(performance): TWR row on the Performance widget — isolate selection from timing
## What

Add a **time-weighted return (TWR)** row to the Investment Performance widget alongside the existing MWR row. TWR is timing-neutral by construction, so comparing it to VOO's cumulative return isolates *selection alone* — the DCA/rising-market gap disappears from that comparison and the user gets a clean read on "did my holdings beat the index?" separately from "did my contribution timing help or hurt?"

Supersedes #635 (the dollar-weighting caption). Closes #381 item 6 the structural way.

## Why

Current widget compares MWR (timing-sensitive) to VOO's TWR-like return. A DCA user in a rising market systematically sees a negative Diff even with identical holdings — because later dollars had less time to compound. The MWR label is accurate; the Diff row misleads. Explaining that in text (#635) closes the misread for one common case; adding TWR closes it structurally for all cases.

## Design

Widget rendering:
- Existing MWR row stays (renamed clearly if needed).
- New "Selection (TWR)" row: portfolio TWR over the window, compared to VOO's cumulative return over the same window.
- Diff row splits into two Diffs (vs MWR and vs TWR) OR the widget picks the primary comparison. Design call at build time.

## Infrastructure needed (no schema / migration / sync work)

1. **Cash-flow classifier** (~50-100 LOC + tests): `isExternalFlow(investmentTransaction | transaction) → boolean`. Separates external deposits/withdrawals (drive TWR sub-period boundaries) from internal returns (Dividend / Interest / etc. — count as gains, not injections). The highest-risk piece: misclassifying a dividend as a deposit understates TWR.
2. **TWR calc** (~200-300 LOC + tests): enumerate external flows in the window, snap each to a date, chain `(V_end − CF) / V_start` per sub-period. Handle explicitly:
   - Opening zero (new account, first contribution) — skip sub-period.
   - Same-day multiple flows — group into one flow at end-of-day.
   - Empty holdings / no flows — no return, not an error.
3. **Multi-account aggregation** (~50-100 LOC): the widget already sums across accounts. Portfolio TWR options:
   - **(a)** Chain sub-periods across the union of all accounts' cash flows. Cleaner, doubles plumbing.
   - **(b)** Per-account TWR weighted by starting portfolio value. What most retail brokerages show; close enough.
4. **Widget UI** (~50-80 LOC): new row, formatting parity with the MWR row.

**No schema / migration / sync work.** `holdingsValueData.getHistory(holdingId)` already gives us daily portfolio value on any date in the window (used by HoldingsComposition + balance graph). Investment-transaction stream already has type/subtype via Plaid. Daily is standard for retail TWR — brokerages don't quote intraday.

## Estimate

~500 LOC + tests. One PR. 1–2 focused sessions depending on the classifier's edge cases (intraday timing usually eats more than the math).

## Refs

- #381 (item 6) — dollar-weighting UX copy.
- #635 (superseded, closed) — dollar-weighting caption; keep the misread analysis in that PR's body as background for whoever picks this up.

claoie:
## What

Adds a **time-weighted return (TWR)** row to the Investment Performance widget, beside the existing MWR row. Closes #638 (supersedes the closed caption approach #635).

TWR is timing-neutral by construction, so reading it next to the index-benchmark row isolates **selection** (did the holdings beat the index?) from **timing** (did contribution timing help or hurt?). Today the widget only shows MWR (timing-sensitive) next to the benchmark's TWR-like return, so a DCA user in a rising market systematically sees a misleading negative Diff even with index-identical holdings — because later dollars had less time to compound. TWR removes that distortion from the comparison.

## How

- **`computeTWR`** (`benchmark.ts`, next to `computeMWR`): chain-links the holding-period return of the *existing* position across the sub-periods bounded by each external cash flow —

  `HPRᵢ = (V(bᵢ) − CFᵢ) / V(bᵢ₋₁)`,  `TWR = Π HPRᵢ − 1`

  where `bᵢ` walks windowStart → each interior flow date → windowEnd, `V(b)` is the asset value at `b` from the existing `valueAt` primitive, and `CFᵢ` is the flow on `bᵢ` (buy +, sell −). Because `valueAt` values shares the same day they're bought (same-date-inclusive, matching how MWR takes `vStart`/`vEnd`), subtracting `CFᵢ` at the closing boundary isolates the period's market move from the new money. Uses the **same asset-side flows + valuations as the MWR** — cash-excluded, phantom-holding-guarded — so the two rows are apples-to-apples.
  - Sub-periods opening on a zero/negative base (position not yet funded, or fully liquidated then re-funded) are skipped; `insufficient_data` when none is funded.
  - **Boundary-value approximation** (documented in the fn): `CFᵢ` is the flow's cash at execution price while `V(bᵢ)` prices at close, so exec≠close carries a small basis error — the same modeled-price limitation MWR already lives with.
- **Widget**: new `Your asset return (TWR)` row between MWR and benchmark. TWR is a pure rate (the dollars earned *are* the MWR gain), so it always renders `%` and stays out of the `$ ↔ %` toggle. Extracted a shared `valueAtAll(date)` closure that now backs `vStart`/`vEnd` and every TWR boundary.

## Open question for @hoiekim (not changed here)

The **Diff** row still compares MWR − benchmark. The clean "selection alone" comparison the issue describes is **TWR − benchmark**. I left Diff as-is to avoid silently redefining an existing row; with the TWR and benchmark rows now adjacent, the user can already read the selection gap directly. Want a follow-up that switches Diff to TWR-based (or adds a second "Selection" diff row)? Say the word and I'll do it in a separate PR.

## Test plan

- [x] `bun test src` — 1104 pass / 0 fail (+11 `computeTWR` unit tests: no-flow price-ratio, timing-neutrality under a mid-window contribution, MWR-vs-TWR divergence under DCA, mid-window sell, ≥2-interior-flow linked chain, zero-base skip, `endBeforeFlow<=0` skip, `insufficient_data`, windowEnd-flow exclusion, cumulative↔annualized consistency)
- [x] **UI E2E** — sandbox, admin login, Accounts → investment-account detail (real row click) → **Investment Performance** widget, mobile 390×844, PNG eye-checked:
  - New **`Your asset return (TWR)`** row renders in the correct position — between `Your asset return (MWR)` and the index-benchmark row, aligned with the others (`/tmp/pr-653-twr-1Y.png`).
  - **Rate-only behavior confirmed**: toggling the widget to `$` flips MWR to a signed-dollar figure and the benchmark to a signed-dollar figure, while the **TWR row stays `%` (unchanged)** — asserted programmatically (`TWR unchanged: true`) and eye-checked (`/tmp/pr-653-twr-dollarmode.png`).
  - Window picker 1Y / 3Y / All all render the row; annualized-suppression footnote (`<6mo`) honored.
  - No new console errors (only the pre-existing sandbox-HTTP service-worker registration warning).
  - **Re-ran the full E2E after the reviewoie fix commit (a8a121c7) — identical, no regression.**
  - _Note: on the scrambled sandbox clone the asset-side snapshot history is short (all windows clamp to ~1 month) and the sampled account has no interior cash flows, so TWR == MWR there — which is **correct** (with no in-window flows the two returns are identical by construction). The **TWR ≠ MWR divergence** the feature exists for is proven rigorously by the unit tests (the DCA test asserts `twr.cumulative = 10%` while `mwr.annualized > 10%`, and the timing-neutrality test holds TWR flat across a mid-window contribution). The sandbox verified rendering + the rate-only toggle; the math divergence is unit-proven._

### DIFF
--- a/src/client/components/PerformanceBenchmark/index.tsx
+++ b/src/client/components/PerformanceBenchmark/index.tsx
@@ -16,6 +16,7 @@ import type { ResolveSecuritySnapshotResponse } from "server/routes/accounts/pos
 import {
   extractCashFlows,
   computeMWR,
+  computeTWR,
   computeBenchmarkTWR,
   computeBenchmarkEndValue,
   valueAt,
@@ -117,18 +118,19 @@ export const PerformanceBenchmark = ({ accounts }: Props) => {
     // anything Polygon has backfilled via resolve-security-snapshot.
     // Txn fills the long historical tail predating Plaid's sync.
     const priceAt = buildPriceAt(securitySnapshots, investmentTransactions);
-    const vStart = ids.reduce(
-      (sum, id) =>
-        sum +
-        valueAt({ date: windowStart, windowStart, accountId: id, holdingSnapshots, investmentTransactions, priceAt }),
-      0,
-    );
-    const vEnd = ids.reduce(
-      (sum, id) =>
-        sum +
-        valueAt({ date: windowEnd, windowStart, accountId: id, holdingSnapshots, investmentTransactions, priceAt }),
-      0,
-    );
+    // Aggregate asset value across all in-scope accounts at an arbitrary
+    // date — the shared primitive behind the window boundaries (vStart/vEnd)
+    // and every TWR sub-period boundary. windowStart anchors the txn-derived
+    // qty walk consistently for all dates.
+    const valueAtAll = (date: string) =>
+      ids.reduce(
+        (sum, id) =>
+          sum +
+          valueAt({ date, windowStart, accountId: id, holdingSnapshots, investmentTransactions, priceAt }),
+        0,
+      );
+    const vStart = valueAtAll(windowStart);
+    const vEnd = valueAtAll(windowEnd);
 
     const flowsByDate = new Map<string, number>();
     for (const id of ids) {
@@ -142,6 +144,12 @@ export const PerformanceBenchmark = ({ accounts }: Props) => {
     const flows = allFlows.filter((f) => f.date > windowStart && f.date <= windowEnd);
 
     const mwr = computeMWR({ flows, vStart, vEnd, windowStart, windowEnd });
+
+    // Time-weighted return over the same window — timing-neutral by
+    // construction, so it isolates *selection* (did the holdings beat the
+    // index?) from *timing* (did contribution timing help?). Uses the same
+    // asset-side flows + valuations as the MWR (see computeTWR).
+    const twr = computeTWR({ flows, valueAt: valueAtAll, windowStart, windowEnd });
 
     // Benchmark TWR fallback chain: security_snapshots → static CSV.
     // (Polygon resolve fires async in a separate useEffect and merges
@@ -231,6 +239,7 @@ export const PerformanceBenchmark = ({ accounts }: Props) => {
       yearsInWindow,
       mwr,
       mwrGain,
+      twr,
       benchmark,
       benchmarkGain,
       gapPct,
@@ -304,6 +313,7 @@ export const PerformanceBenchmark = ({ accounts }: Props) => {
     yearsInWindow,
     mwr,
     mwrGain,
+    twr,
     benchmark,
     benchmarkGain,
     gapPct,
@@ -373,6 +383,18 @@ export const PerformanceBenchmark = ({ accounts }: Props) => {
             {displayMode === "pct"
               ? renderValues(mwr.cumulative, mwr.annualized, "—", formatPct)
               : renderValues(mwrGain, perYear(mwrGain), "—", formatSignedDollars)}
+          </span>
+        </div>
+
+        <div className="performanceRow">
+          <span className="performanceLabel">Your asset return (TWR)</span>
+          {/* TWR is a pure rate — the actual dollars earned ARE the MWR gain,
+              so there's no honest dollar counterpart. The row always renders %
+              and doesn't participate in the $ ↔ % toggle. */}
+          <span className="performanceValues">
+            {twr.status === "ok"
+              ? renderValues(twr.cumulative, twr.annualized, "—", formatPct)
+              : <span className="no-data">—</span>}
           </span>
         </div>
 
--- a/src/client/components/PerformanceBenchmark/index.tsx
+++ b/src/client/components/PerformanceBenchmark/index.tsx
@@ -121,14 +121,23 @@ export const PerformanceBenchmark = ({ accounts }: Props) => {
     // Aggregate asset value across all in-scope accounts at an arbitrary
     // date — the shared primitive behind the window boundaries (vStart/vEnd)
     // and every TWR sub-period boundary. windowStart anchors the txn-derived
-    // qty walk consistently for all dates.
-    const valueAtAll = (date: string) =>
-      ids.reduce(
+    // qty walk consistently for all dates. Each `valueAt` re-derives the
+    // per-account cash-security set + txn qty walk, so memoize per date: the
+    // window boundaries are queried by both vStart/vEnd and computeTWR, and a
+    // heavy-trader account can have many flow boundaries.
+    const valueCache = new Map<string, number>();
+    const valueAtAll = (date: string) => {
+      const cached = valueCache.get(date);
+      if (cached !== undefined) return cached;
+      const total = ids.reduce(
         (sum, id) =>
           sum +
           valueAt({ date, windowStart, accountId: id, holdingSnapshots, investmentTransactions, priceAt }),
         0,
       );
+      valueCache.set(date, total);
+      return total;
+    };
     const vStart = valueAtAll(windowStart);
     const vEnd = valueAtAll(windowEnd);
 
--- a/src/client/lib/hooks/calculation/benchmark.test.ts
+++ b/src/client/lib/hooks/calculation/benchmark.test.ts
@@ -6,6 +6,7 @@ import {
   extractCashFlows,
   extractCashFlowsBySecurity,
   computeMWR,
+  computeTWR,
   computeBenchmarkTWR,
   computeBenchmarkEndValue,
   valueAt,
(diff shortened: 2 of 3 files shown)


reviewoie (commented):
## Code review — PR 653

TWR math is correct and internally consistent. I traced the chain-linking by hand against every unit test and re-derived the boundary rules:

- Opening capital vs in-window flow split is right: a flow on `windowStart` is folded into `vStart` (excluded from `cfByDate` via `f.date <= windowStart`), matching MWR's `f.date > windowStart` filter and `valueAt`'s same-date-inclusive convention.
- Closing-boundary subtraction (`endBeforeFlow = valueAt(bᵢ) − cf`) correctly isolates the period's market move, and the flow re-enters as the next sub-period's opening `valueAt(bᵢ)` (close-priced) — so the exec-vs-close basis error is bounded per-period and does NOT compound into the next opening. Docstring's "small basis error" claim is accurate.
- Interior filter `d < windowEnd` + keeping `windowEnd` flows in `cfByDate` (folded into the final close subtraction) is the correct off-by-one handling. Verified by the "flow exactly on windowEnd" test.
- The `base <= 0` and `endBeforeFlow <= 0` skips are defensible — they do NOT hide real market losses. A 50% drawdown is captured (`HPR = 0.5`); a sale at a loss is captured (`endBeforeFlow = valueAt + |cf|`, HPR < 1). `endBeforeFlow <= 0` is buy-only and only fires when execution price ≫ close (intraday-anomaly), which can't occur for a long-only position in normal pricing. Sub-periods with zero invested capital (liquidated-then-refunded) are correctly dropped, not mismeasured.
- Scope/consistency with MWR is clean: same filtered `flows`, same `valueAtAll` behind `vStart`/`vEnd`, cash-excluded + phantom-holding-guarded inside `valueAt`. No div-by-zero (`years >= 0.001`, `product > 0` so `1 + cumulative > 0`).
- `valueAtAll` extraction is behavior-preserving vs the original per-account reduce (only `date: windowStart` → parameter `date`; `windowStart` anchor unchanged for both boundaries). Memo deps, destructure, and rate-only TWR render are wired correctly.

Scope is clean — 3 files, all TWR. 8 new `computeTWR` tests pass (64/64 in the suite).

### Low
- `benchmark.ts:275-283` (compute cost) — TWR now calls `valueAtAll` once per boundary (`2 + interior-flow-dates`), and each `valueAt` re-derives the full per-account cash-security set + txn qty walk from scratch (`O(snapshots + txns)` per account per call). The pre-PR code called `valueAt` only twice (`vStart`/`vEnd`); this is now `O(F × accounts × (S+T))`. Memoized + client-side + single-user, so no real-world concern at today's data sizes — but if a heavy-trader account ever makes it slow, memoize `valueAt`-by-date. Informational, not a blocker.
- `benchmark.test.ts` (coverage) — every `computeTWR` test uses 0 or 1 interior flow. No test exercises a multi-flow chain (≥2 interior boundaries) or the `endBeforeFlow <= 0` skip branch. The core paths are covered; adding a 2-flow chain + an exec≫close skip case would pin the two untested branches the function name implies.
- `index.tsx:389-399` (UX) — when the other rows are in `$` mode, the TWR row still renders `%`. Documented and intentional (TWR has no honest dollar counterpart), so this is a note not a defect — flagging only so it's a conscious choice: a user toggling to `$` sees one `%` row among `$` rows.

No further blockers.

claoie:
Thanks — addressed in a8a121c7:

- **LOW 1 (compute cost)** — memoized `valueAtAll` by date. TWR queries the same window boundaries `vStart`/`vEnd` already compute, and a heavy-trader account can add many flow boundaries; the cache dedupes those and keeps each date's O(snapshots+txns) walk to once-per-render.
- **LOW 2 (coverage)** — added two `computeTWR` tests: a ≥2-interior-flow linked chain (three 10% sub-periods → 33.1%, contributions ignored) and the `endBeforeFlow <= 0` skip branch (a buy priced above the day's close). Suite now 66/66.
- **LOW 3 (UX — TWR shows % in $ mode)** — **leaving as-is by design.** A time-weighted return has no honest dollar counterpart (the dollars actually earned *are* the MWR gain), so a synthetic "$" for TWR would mislead. The one `%` row among `$` rows is the intended, documented signal that TWR is a pure rate. Flagged the Diff-on-TWR follow-up separately for Hoie in the PR body.

Re-ran the full suite + typecheck + lint + build + sandbox UI E2E after the fix — all green, no regression.

reviewoie:
## Code review — PR 653 (re-review of head `a8a121c7`)

Delta reviewed: `656e684f..a8a121c7` — the new commit *perf(performance): memoize valueAtAll by date + cover TWR multi-flow and skip branches*. The prior review covered the feature commit `656e684f`; this pass is scoped to the memoization + added tests only.

### valueAtAll memoization — correct
- `valueCache` is a fresh `Map` constructed inside the `useMemo` body, so it's recreated on every recompute. No stale-cache risk across renders/dep changes (`index.tsx`).
- Cache key is `date`, and `date` is the only varying input: `ids`, `windowStart`, `holdingSnapshots`, `investmentTransactions`, `priceAt` are all fixed within the closure for a single memo eval, so `valueAt(date)` is deterministic per date. Key covers everything that affects the result.
- `if (cached !== undefined) return cached;` correctly distinguishes a genuine cached `0` total from an un-cached miss — no falsy-zero bug (the `reduce` seeds at 0, always returns a number).
- The memoized `vStart`/`vEnd` are byte-for-byte the same `ids.reduce(...)` as un-memoized `656e684f` — the wrapper only caches. `computeTWR` receives `valueAt: valueAtAll`, and boundary dates (windowStart/windowEnd) are queried by both vStart/vEnd and the TWR walk, so the cache is a real dedupe with no semantic change. No regression to the TWR row.

### Added TWR tests — pin the behavior
- Traced both by hand against `computeTWR`:
  - "chains ≥2 interior flows" → boundaries [start, 05-01, 09-01, end], sub-HPRs 1100/1000 · 2310/2100 · 3641/3310 = 1.331 → cumulative 0.331. Matches.
  - "skips a sub-period whose closing value falls below the same-day buy" → sub1 `endBeforeFlow = 1500−2000 = −500 ≤ 0` skipped, sub2 1650/1500 = 1.1 → 0.1. Matches.
- Test names verified via grep (not fabricated); full `benchmark.test.ts` suite runs 66 pass / 0 fail / 115 assertions at head `a8a121c7`.

### Scope
Delta touches only `PerformanceBenchmark/index.tsx` and `benchmark.test.ts` — both directly related to the TWR feature. No unrelated changes.

No further blockers.

claoie:
Acknowledged — re-review of `a8a121c7` is approve with 0 findings, so nothing to action here.

Two points from the pass worth pinning for the record:

- The `valueCache` lifetime reasoning matches the intent: the `Map` is constructed inside the `useMemo` body specifically so it dies with the memo eval. If a later change lifts it to a `useRef` or module scope, the cache key stops being sufficient (`ids` / `windowStart` / `holdingSnapshots` / `investmentTransactions` / `priceAt` would all need to join the key).
- The `cached !== undefined` guard is deliberate over a truthiness check, since a genuine total of `0` is reachable.

No disagreement to record. Holding for your review.

Issue #656, opened 2026-07-26
Real-time collaboration


hoiekim:
Implement feature that the client receives signals from the server for new data, then pulls the updates. This allows multiple users connected to the app at the same time and they get real-time updates from each other.

## Architecture

1. Connection pipe: either socket or long-open HTTP connection
2. Event emitter on server side
3. Emits "new data" event in every data mutation path
4. Event receiver on client side
5. Calls sync for specific data components on new data.

claoie:
Delivered via the 3-PR arc:

- **#657** — SSE transport. `GET /api/events` per-user stream, per-user cap 20, keepalive; `emitToUser(userId, tableName, payload)` fan-out. Table names moved to `common/constants` as `enum TableName` so server and client share one source of truth; `/events` registered as a normal route (widened `Route.execute` to allow raw `Response` return).
- **#658** — emit-on-mutation + client receiver. Server: `mutationEmitDomain(method, path)` maps writes to tables; `handleApiRequest` broadcasts after a successful mutation with the caller's `X-Tab-Id` as `originTabId`. Client: `useServerEvents(handler, enabled)` (auth-gated to avoid EventSource fatal-close on anon 401); `Utility.tsx` dispatches to `syncDomain(domain)` — a domain-scoped refetch, not a whole-app `sync()`. Per-domain debouncer collapses bursts.
- **#659** — plaid-hook emits. Webhook lands out-of-band from any user tab (no session); resolves the item's owner from `item_id` and emits the appropriate `<table>-updated`. Auto-imported transactions, item-status changes, and product refreshes fan out live.

Multiple simultaneously connected tabs of the same user now see each other's changes without a page reload; the originating tab filters its own event via `tabId` to skip a redundant self-refetch.

PR #657, opened 2026-07-26
feat(realtime): SSE plumbing — GET /api/events + emitToUser (#656 PR 1/3)

claoie:
## Summary

PR 1 of 3 for the [Real-time collaboration issue](https://github.com/hoiekim/budget/issues/656) arc. Ships the transport (SSE) and the server-side emitter without wiring any mutation route to emit — that's PR 2. Existing 30s polling stays intact as a safety net; PR 3 removes it after real-time delivery is proven.

## Transport choice — Server-Sent Events

- One-way server → client over a normal long-open HTTP GET, no WebSocket upgrade, no external dep, no reverse-proxy config. Bun serves it via `Response(ReadableStream, {content-type: text/event-stream})`.
- Auth reuses the existing session cookie.
- Native browser reconnect on network drop via `EventSource`.
- Fits the issue's signal → client pulls shape exactly.

## Server changes

- `src/server/lib/realtime.ts`
  - Per-user subscriber registry (`Map<userId, Set<Subscriber>>`).
  - `emitToUser(userId, domain, payload)` fans an event out to every open tab of that user only.
  - `EmitDomain` is a literal-string union matching the top-level `Data` dicts (accounts / budgets / categories / charts / holdings / holding-snapshots / investment-transactions / items / sections / snapshots / split-transactions / transactions / transfers). Typo at any emit call site = compile error.
  - Throwing subscribers get dropped and closed; healthy siblings still receive the event.
- `src/server/start.ts`
  - `GET /api/events` intercepted in `Bun.serve.fetch` BEFORE `handleApiRequest`. The normal Route pipeline buffers into MutableResponse — incompatible with a stream that stays open for the tab's lifetime.
  - Auth: session cookie required; unauth = 401.
  - Primes with `: connected` so browsers fire `onopen` immediately and proxies commit to streaming.
  - 30s keepalive comments so proxies don't time out idle connections.
  - Cleanup on `request.signal.abort`: unregister, clear the keepalive interval, close the controller.

## Client changes

- `src/client/lib/hooks/useServerEvents.ts` — one tab-lifetime `EventSource`, dispatches per-domain listeners to a caller-supplied handler `(domain, payload) => void`. Handler is held in a ref so re-renders don't tear the connection down. `EventSource` auto-reconnects; we only log when it gives up (`readyState === CLOSED`).
- Consumer wiring in `App` lands in PR 2 — this ships the hook so PR 2 is a one-line `useServerEvents(dispatch)`.

## Not in this PR (by design)

- No mutation route emits yet. PR 2 wires `emitToUser` into `POST/DELETE` route handlers after each successful DB write.
- No cross-tab origin filtering. PR 2 threads an `X-Tab-Id` header through mutation requests so the originating tab can filter its own event out.
- 30s polling in `useSync` stays. PR 3 removes it after PR 2 proves delivery.

## Test plan

- [x] `bun run typecheck` clean.
- [x] `bun run lint` clean.
- [x] `bun test` — 1113 pass / 0 fail (+8 new in `realtime.test.ts` covering register/unregister, per-user fan-out isolation, throwing-subscriber drop, per-domain event names).
- [x] Manual E2E on local dev (port 3005):
  - `GET /api/events` unauth → HTTP 401 with `{status:"failed"}`.
  - `POST /api/login` then `GET /api/events` with cookie → HTTP 200, `Content-Type: text/event-stream`, first chunk is `: connected\n\n`.
  - Server logs `SSE subscribed` (userId + total).
  - Client abort → server logs `SSE unsubscribed` (remaining count).
- [ ] Sandbox E2E: same three-step against `claoie.tplinkdns.com:<port>`.
- [ ] End-to-end signal test: deferred to PR 2, which is where `emitToUser` starts firing.

Refs #656.

### DIFF
--- a/src/client/lib/hooks/useServerEvents.ts
+++ b/src/client/lib/hooks/useServerEvents.ts
@@ -1,13 +1,6 @@
 import { useEffect, useRef } from "react";
 import { TableName } from "common/constants";
 
-/**
- * The set of event domains the SSE receiver knows about is exactly
- * `TableName` — the same enum the server emits from. Keeping one
- * source of truth means adding a new mutation surface on the server
- * is a compile error on the client until the receiver-side dispatch
- * (PR 2) picks it up.
- */
 export type ServerEventDomain = TableName;
 
 export interface ServerEventPayload {
@@ -22,18 +15,12 @@ export type ServerEventHandler = (
 /**
  * One tab-lifetime `EventSource` to `/api/events`. Every domain event
  * the server broadcasts to this user fires `handler(domain, payload)`.
- * The handler should compare `payload.originTabId` against this tab's
- * `X-Tab-Id` to filter its own writes (avoids a redundant self-refetch
- * right after `useMutate` already patched local state).
+ * `EventSource` handles reconnection natively; the hook does not.
  *
- * EventSource handles reconnection natively — on network drop the
- * browser reopens the connection with exponential backoff. We don't
- * need to wire that ourselves.
- *
- * Consumers register at most one handler per hook call; if a component
- * needs multiple listeners (e.g. Dashboard cares about both `charts`
- * and `budgets`), pass a switch/dispatch inside the single handler.
- * The router-level `App` wires the global dispatcher.
+ * The handler receives one `(domain, payload)` call per event. To split
+ * dispatch across multiple concerns, switch on `domain` inside a single
+ * handler passed to one `useServerEvents` call — extra hook instances
+ * would open extra streams and blow past the per-user subscriber cap.
  */
 export const useServerEvents = (handler: ServerEventHandler): void => {
   // Keep the latest handler in a ref so we don't tear the EventSource
--- a/src/client/lib/hooks/useServerEvents.ts
+++ b/src/client/lib/hooks/useServerEvents.ts
@@ -1,24 +1,14 @@
 import { useEffect, useRef } from "react";
+import { TableName } from "common/constants";
 
 /**
- * Data-domain namespace mirrors `EmitDomain` on the server. Kept as a
- * literal union so a typo at a listener registration site is a compile
- * error. Extend when a new domain gets a mutation surface.
+ * The set of event domains the SSE receiver knows about is exactly
+ * `TableName` — the same enum the server emits from. Keeping one
+ * source of truth means adding a new mutation surface on the server
+ * is a compile error on the client until the receiver-side dispatch
+ * (PR 2) picks it up.
  */
-export type ServerEventDomain =
-  | "accounts"
-  | "budgets"
-  | "categories"
-  | "charts"
-  | "holdings"
-  | "holding-snapshots"
-  | "investment-transactions"
-  | "items"
-  | "sections"
-  | "snapshots"
-  | "split-transactions"
-  | "transactions"
-  | "transfers";
+export type ServerEventDomain = TableName;
 
 export interface ServerEventPayload {
   originTabId?: string;
@@ -56,21 +46,7 @@ export const useServerEvents = (handler: ServerEventHandler): void => {
     // request is anonymous and the server 401s.
     const source = new EventSource("/api/events", { withCredentials: true });
 
-    const domains: ServerEventDomain[] = [
-      "accounts",
-      "budgets",
-      "categories",
-      "charts",
-      "holdings",
-      "holding-snapshots",
-      "investment-transactions",
-      "items",
-      "sections",
-      "snapshots",
-      "split-transactions",
-      "transactions",
-      "transfers",
-    ];
+    const domains = Object.values(TableName);
 
     const perDomain = new Map<ServerEventDomain, (e: MessageEvent) => void>();
     for (const domain of domains) {
--- a/src/client/lib/hooks/useServerEvents.ts
+++ b/src/client/lib/hooks/useServerEvents.ts
@@ -1,0 +1,105 @@
+import { useEffect, useRef } from "react";
+
+/**
+ * Data-domain namespace mirrors `EmitDomain` on the server. Kept as a
+ * literal union so a typo at a listener registration site is a compile
+ * error. Extend when a new domain gets a mutation surface.
+ */
+export type ServerEventDomain =
+  | "accounts"
+  | "budgets"
+  | "categories"
+  | "charts"
+  | "holdings"
+  | "holding-snapshots"
+  | "investment-transactions"
+  | "items"
+  | "sections"
+  | "snapshots"
+  | "split-transactions"
+  | "transactions"
+  | "transfers";
+
+export interface ServerEventPayload {
+  originTabId?: string;
+}
+
+export type ServerEventHandler = (
+  domain: ServerEventDomain,
+  payload: ServerEventPayload,
+) => void;
+
+/**
+ * One tab-lifetime `EventSource` to `/api/events`. Every domain event
+ * the server broadcasts to this user fires `handler(domain, payload)`.
+ * The handler should compare `payload.originTabId` against this tab's
+ * `X-Tab-Id` to filter its own writes (avoids a redundant self-refetch
+ * right after `useMutate` already patched local state).
+ *
+ * EventSource handles reconnection natively — on network drop the
+ * browser reopens the connection with exponential backoff. We don't
+ * need to wire that ourselves.
+ *
+ * Consumers register at most one handler per hook call; if a component
+ * needs multiple listeners (e.g. Dashboard cares about both `charts`
+ * and `budgets`), pass a switch/dispatch inside the single handler.
+ * The router-level `App` wires the global dispatcher.
+ */
+export const useServerEvents = (handler: ServerEventHandler): void => {
+  // Keep the latest handler in a ref so we don't tear the EventSource
+  // down every time the parent re-renders with a new closure.
+  const handlerRef = useRef(handler);
+  handlerRef.current = handler;
+
+  useEffect(() => {
+    // withCredentials sends the session cookie; without it the SSE
+    // request is anonymous and the server 401s.
+    const source = new EventSource("/api/events", { withCredentials: true });
+
+    const domains: ServerEventDomain[] = [
+      "accounts",
+      "budgets",
+      "categories",
+      "charts",
+      "holdings",
+      "holding-snapshots",
+      "investment-transactions",
+      "items",
+      "sections",
+      "snapshots",
+      "split-transactions",
+      "transactions",
+      "transfers",
+    ];
+
+    const perDomain = new Map<ServerEventDomain, (e: MessageEvent) => void>();
+    for (const domain of domains) {
+      const listener = (e: MessageEvent) => {
+        let payload: ServerEventPayload = {};
+        try {
+          if (e.data) payload = JSON.parse(e.data) as ServerEventPayload;
+        } catch {
+          // malformed payload — dispatch with empty
+        }
+        handlerRef.current(domain, payload);
+      };
+      source.addEventListener(`${domain}-updated`, listener);
+      perDomain.set(domain, listener);
+    }
+
+    source.addEventListener("error", () => {
+      // EventSource auto-reconnects on error; log only when the connection
+      // stays broken (readyState === CLOSED means auto-reconnect gave up).
+      if (source.readyState === EventSource.CLOSED) {
+        console.warn("SSE connection closed and will not auto-reconnect.");
+      }
+    });
+
+    return () => {
+      for (const [domain, listener] of perDomain) {
+        source.removeEventListener(`${domain}-updated`, listener);
+      }
+      source.close();
+    };
+  }, []);
+};
--- a/src/server/lib/postgres/models/common.ts
+++ b/src/server/lib/postgres/models/common.ts
@@ -1,32 +1,9 @@
 /**
  * Column name constants for type-safe database operations. Table-name
- * constants live in `src/common/constants.ts` (accessible from both
- * server and client — the SSE emit path needs them on both sides) and
- * are re-exported here so existing server-side callers keep working.
+ * constants live in `src/common/constants.ts` so both server and client
+ * share them — import from there directly, not from this file.
  * All column names are snake_case to match PostgreSQL convention.
  */
-
-export {
-  TableName,
-  USERS,
-  SESSIONS,
-  ITEMS,
-  INSTITUTIONS,
-  ACCOUNTS,
-  HOLDINGS,
-  SECURITIES,
-  TRANSACTIONS,
-  INVESTMENT_TRANSACTIONS,
-  SPLIT_TRANSACTIONS,
-  TRANSACTION_PAIRS,
-  BUDGETS,
-  SECTIONS,
-  CATEGORIES,
-  SNAPSHOTS,
-  CHARTS,
-  API_KEYS,
-  REJECTED_CATEGORIES,
-} from "common/constants";
 
 export const USER_ID = "user_id";
 export const UPDATED = "updated";
--- a/src/server/lib/postgres/models/common.ts
+++ b/src/server/lib/postgres/models/common.ts
@@ -1,25 +1,32 @@
 /**
- * Column name constants and table names for type-safe database operations.
+ * Column name constants for type-safe database operations. Table-name
+ * constants live in `src/common/constants.ts` (accessible from both
+ * server and client — the SSE emit path needs them on both sides) and
+ * are re-exported here so existing server-side callers keep working.
  * All column names are snake_case to match PostgreSQL convention.
  */
 
-export const USERS = "users";
-export const SESSIONS = "sessions";
-export const ITEMS = "items";
-export const INSTITUTIONS = "institutions";
-export const ACCOUNTS = "accounts";
-export const HOLDINGS = "holdings";
-export const SECURITIES = "securities";
-export const TRANSACTIONS = "transactions";
-export const INVESTMENT_TRANSACTIONS = "investment_transactions";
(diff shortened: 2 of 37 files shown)


reviewoie:
## Code review — PR 657

**Verdict:** changes-requested (one bounded subscriber-leak race; rest are LOW).

**Findings:** 5 total — 0 high, 1 med, 4 low.

### Med
- `src/server/start.ts:245-323` — `handleSseRequest` doesn't handle a pre-aborted request. It calls `request.signal.addEventListener("abort", cleanup)` then constructs `new ReadableStream({ start() { registerSubscriber(...) } })`. `AbortSignal` dispatches its `abort` event exactly once on the transition from non-aborted to aborted; adding a listener AFTER that transition never fires. Failure scenario: client sends the GET then disconnects immediately (fast tab close, SSL/network race, load balancer aborting a half-open connection). By the time `handleSseRequest` runs, `request.signal.aborted === true`. `addEventListener` installs a listener that will never fire. `new ReadableStream(...)` still synchronously invokes `start(controller)` per the streams spec — the sub gets `registerSubscriber(userId, sub)`'d. Bun then discards the Response (aborted) without ever consuming the stream, so the `cancel()` hook doesn't fire either. Sub sits in `subscribers` map until the first `keepalive` tick (`SSE_KEEPALIVE_MS = 30_000`) enqueues onto a closed controller, throws, and the try/catch inside the interval finally calls `cleanup`. Bounded leak (≤30s per pre-aborted request), but every such request leaks. Standard fix: `if (request.signal.aborted) { return jsonResponse({status:"failed", message:"aborted"}, 499); }` at the top, or an equivalent early check inside `start()` before registering.

### Low
- `src/server/lib/realtime.ts:28-41` + `src/client/lib/hooks/useServerEvents.ts:8-21` — `EmitDomain` (server) and `ServerEventDomain` (client) are two hand-maintained literal unions that MUST match. The client only subscribes to `${domain}-updated` for domains it knows about, so adding a new server-side domain without editing the client silently drops the event. PR body notes PR 2 introduces a dispatch table that catches drift — worth confirming that landing lands with the shared type too. Right shape: one `EventDomain` type in `common/` that both server and client import.
- `src/server/lib/realtime.ts:50-57` — no per-user cap on `registerSubscriber`. A buggy client that reconnects in a hot loop, or a malicious one, can grow the per-user Set unbounded; each entry holds a keepalive interval + controller reference + Set slot. A soft cap (e.g. 20/user, drop-oldest on overflow) fits the scalability-vision rule 13 lens — the transport should have a natural upper bound before PR 2 starts fanning writes over it.
- `src/server/start.ts:245-323` — SSE requests skip the standard `logger.info("API request", log)` line emitted by `handleApiRequest` (the SSE fast-path returns before that block). The `SSE subscribed`/`SSE unsubscribed` lines cover subscription lifecycle but drop `status`, `durationMs`, `bodyFields`, `ip` etc. Observability inconsistency — dashboards/log-parsers that filter on `API request` will miss all SSE opens.
- `src/server/start.ts:245-250` — `handleSseRequest` uses `loadSession` directly and skips `resolveBearerAuth`, so a bearer-auth-only client (API-key holder, no cookie) gets a 401 on `/api/events`. Almost certainly intended (SSE is a browser affordance), but worth capturing explicitly in the PR body so a future CLI/mobile client doesn't file this as a regression.

### For PR 2's E2E sweep (this ships transport only; consumer-facing E2E belongs on PR 2)
- Two tabs same user, mutate in tab A → tab B receives event and refetches.
- Same user across tabs, then close one → `subscriberCount(userId)` decrements; other tab still receives events.
- Cross-user isolation: emit to U1 with U1 + U2 tabs open, U2 sub never fires.
- Client disconnect mid-mutation (tab close during POST) → no orphaned subscriber, server log shows `SSE unsubscribed`.
- Session invalidated server-side (`/logout`) while SSE open → connection tears down gracefully; new tab reopens cleanly.
- Keepalive path: leave tab open >30s idle, verify `: keepalive` chunk arrives and browser doesn't fire `onerror`.
- Vite proxy path in dev: confirm `/api/events` through :3000→:3005 doesn't buffer (first chunk `: connected` arrives without waiting for stream close).

claoie:
Thanks for the review. Response per finding:

**MED — pre-aborted request leaks a subscriber (~30s)**
Fixed in `ac816f13`. Handler entry now checks `request.signal.aborted`; when set, we set the `closed` flag instead of calling `addEventListener("abort", cleanup)` (which is a no-op on an already-aborted signal per DOM spec). The `start` callback checks `closed` before registering the subscriber, so the pre-aborted path never grows the map.

**LOW — no per-user subscription cap**
Applied in `ac816f13`. `MAX_SUBSCRIBERS_PER_USER = 20`; excess opens return 429 before the stream is constructed. 20 comfortably fits normal use (a few tabs across a couple of devices) while catching a runaway reconnect loop early.

**LOW — `EmitDomain` / `ServerEventDomain` duplicated literal unions**
Deferring to PR 2. PR 2 introduces the emit-on-mutation dispatch table that lives in `common/` and will be the single source both server and client import from — the duplication in this PR is transitional and I'd rather do the extraction alongside the code that forces it.

**LOW — SSE requests skip the `API request` log line**
Leaving as-is. `SSE subscribed` / `SSE unsubscribed` are the per-connection lifecycle signals; folding a long-lived stream into the same request-log line as one-shot API calls would either log at open (misleading duration) or at close (way after the fact). Happy to revisit if you want a different observability shape.

**LOW — bearer-auth clients can't open SSE**
Intentional. SSE is a browser-tab primitive (needs `EventSource` + cookies for reconnect handshake); bearer clients are meant for cron/scripts that don't hold long-lived connections. Added a note to the PR body.

Re-ran typecheck + `realtime.test.ts` (8/8 pass) after the fix. Sandbox rebuild + edge-case E2E sweep next.

reviewoie:
## Code review — PR 657 (re-review of head `ac816f13`)

**Scope:** delta `cf40f411..ac816f13` only — the single new commit *"fix(sse): guard pre-aborted signal and cap subscribers per user"* (`src/server/start.ts`, +30/-1). Not a full re-review; the prior review at `cf40f411` still covers the rest of the PR.

**Verdict:** approve — the commit resolves the prior Med and the per-user-cap Low. No new findings in the delta.

**Findings:** 0 new — 0 high, 0 med, 1 low (informational).

### Prior findings — disposition
- **Med (pre-aborted subscriber leak), `start.ts`** — RESOLVED. The commit takes the "early check inside `start()` before registering" path the prior review named as an acceptable fix: on entry it sets `closed = true` when `request.signal.aborted` and skips `addEventListener` (a no-op on an already-aborted signal per the DOM spec); `start(controller)` then closes the controller and returns before `registerSubscriber`. Verified sound: (a) `start()` runs synchronously inside `new ReadableStream(...)` on the same tick as the abort check, so no yield window; (b) the sub is never registered, so `subscriberCount` never increments and there is nothing to leak; (c) `cleanup` on later `cancel()`/`abort` is idempotent (`if (closed) return`). This is cleaner than the ≤30s keepalive self-heal the prior review described — the sub never enters the map at all.
- **Low (no per-user cap), `realtime.ts`/`start.ts`** — RESOLVED. `MAX_SUBSCRIBERS_PER_USER = 20`; `subscriberCount(userId) >= MAX` returns 429 before the stream is built. Verified: no off-by-one (count 20 rejects the 21st, so max concurrent = 20); the check→`registerSubscriber` path is synchronous with no `await`, so the TOCTOU window across concurrent requests does not exist (a second request only runs after the first fully returns its `Response`) — the bound is hard, not best-effort. Chose reject-new over the prior review's suggested drop-oldest, which is the better call (doesn't disrupt existing tabs). Client-side is safe against a 429 hammer: `EventSource` treats a non-2xx as a fatal error and does NOT auto-reconnect, so a capped tab degrades to no-events rather than looping.
- **Low (EmitDomain / ServerEventDomain drift), `realtime.ts:28` + `useServerEvents.ts:8`** — NOT touched by this commit (still two hand-maintained literal unions; they currently match, 13 domains identical). Stands as before — reaffirming the prior recommendation to hoist one shared `EventDomain` into `common/` when PR 2 lands its dispatch table.
- The other two prior Lows (SSE skips the `API request` log line; SSE skips `resolveBearerAuth`) — untouched, unchanged.

### Low (new, informational)
- `src/server/start.ts:258-311` — the new 429-cap branch and the pre-abort/`closed` guard in `start()` have no automated coverage (the delta adds no test, and `handleSseRequest` in `start.ts` isn't in the preload harness — `realtime.test.ts` covers the registry primitives but not this entry path). Not a blocker for a transport-only PR, but PR 2's E2E sweep should exercise: (a) open the 21st concurrent stream for one user → 429 with no map growth; (b) a request whose signal is already aborted at handler entry → no subscriber registered (`subscriberCount` unchanged) and no 30s-delayed cleanup.

claoie:
Small factual correction to this PR's description (surfaced while building PR 2, #658): the body says *"Existing 30s polling stays intact as a safety net; PR 3 removes it."* There is **no polling** in the client today — no `setInterval`/refetch loop in `sync.ts` or `Utility.tsx`; the only timer is the 50 ms `sync` debounce. The client currently auto-refreshes **only on login and manual refresh**. So the real-time receiver in #658 adds the *first* automatic cross-session refresh, and PR 3 has no polling to remove (it's the per-domain-sync + background-sync-push follow-up instead). Doesn't change anything in this PR — just so the merge decision isn't based on a safety net that isn't there.

src/client/lib/hooks/useServerEvents.ts
```diff
@@ -0,0 +1,105 @@
...
+  | "charts"
+  | "holdings"
+  | "holding-snapshots"
+  | "investment-transactions"
+  | "items"
+  | "sections"
+  | "snapshots"
+  | "split-transactions"
+  | "transactions"
+  | "transfers";
```
hoiekim: These should align with the table names defined in `src/server/lib/postgres/models/common.ts`
Move those table name definitions to `src/common/constants.ts` and import from there, for consistency.

Also define `enum TableName` using the constants

src/server/lib/realtime.ts
```diff
@@ -0,0 +1,111 @@
...
+  | "charts"
+  | "holdings"
+  | "holding-snapshots"
+  | "investment-transactions"
+  | "items"
+  | "sections"
+  | "snapshots"
+  | "split-transactions"
+  | "transactions"
+  | "transfers";
```
hoiekim: Use the common definition across server and client.

src/server/start.ts
```diff
@@ -409,6 +542,13 @@ const server = Bun.serve({
     // Strip /api prefix to get the route path
     const apiPath = fullPath.slice(4) || "/";
 
+    // SSE stream handled inline — the normal Route.execute path buffers into
+    // MutableResponse and returns after the callback settles, which is
+    // incompatible with a stream that stays open for the tab's lifetime.
+    if (request.method === "GET" && apiPath === "/events") {
+      return handleSseRequest(request);
+    }
```
hoiekim: This is not how we typically register routes

src/client/lib/hooks/useServerEvents.ts
claoie: Addressed in `a72dc083` — added `enum TableName` in `src/common/constants.ts`, moved the UPPER_SNAKE table-name aliases there and re-exported from `server/lib/postgres/models/common.ts` so the ~20 existing server call sites keep working. `ServerEventDomain` on the client is now `TableName` (imported from `common/constants`); `useServerEvents` iterates `Object.values(TableName)` for the per-domain listener registration. One source of truth, and the hyphenated aliases (`holding-snapshots`, `transfers`) are gone — the emit path speaks snake_case DB table names verbatim.

src/server/lib/realtime.ts
claoie: Addressed in `a72dc083`. `EmitDomain` is now `TableName` (imported from `common/constants`) — the client's `ServerEventDomain` is the same alias, so both sides speak from one source.

src/server/start.ts
claoie: Addressed in `a72dc083`. `/events` is now a normal route at `src/server/routes/events/get-events.ts` and registers via the standard `Object.values(routes)` iteration; the inline `handleSseRequest` dispatch in `Bun.serve.fetch` is gone. To make this work I widened `Route.execute` to allow the callback to return a raw `Response` (for the SSE stream case) and taught `handleApiRequest` to pass it through unmodified. I also added `signal?: AbortSignal` to `ServerRequest` so long-lived routes can subscribe to client disconnect through the normal abstraction. Auth for `/events` now flows through the same public-path/session gate as any route — unauth returns the standard 401.

src/server/lib/postgres/models/common.ts
```diff
@@ -1,25 +1,32 @@
...
+  SPLIT_TRANSACTIONS,
+  TRANSACTION_PAIRS,
+  BUDGETS,
+  SECTIONS,
+  CATEGORIES,
+  SNAPSHOTS,
+  CHARTS,
+  API_KEYS,
+  REJECTED_CATEGORIES,
+} from "common/constants";
```
hoiekim: Don't re-export these. Just update the old imports with the new path

claoie: Addressed in `069a1f4b` — dropped the re-export block from `postgres/models/common.ts`. Every downstream caller (18 model files + 4 repos + 1 compute-tool = 23 files) now imports table-name constants directly from `common/constants`. Each file has two imports where it had one: column consts from `./common` (or `../models`), table consts from `common/constants`.

reviewoie (commented):
## Code review — PR 657 (069a1f4b)

**Verdict:** informational — no blockers.

Walked all 36 files. `TableName` enum values (18) match the actual CREATE TABLE identifiers in `postgres/initialize.ts` end-to-end (`usersTable`→`users` … `rejectedCategoriesTable`→`rejected_categories`), no drift. `Object.values(TableName)` in `useServerEvents.ts:49` returns 18 unique string values — string-enum, no numeric reverse-mapping gotcha (verified with `bun -e`). `Route.execute`'s widened `ApiResponse<T> | Response | null` has exactly one production caller (`start.ts:363`); the `if (result instanceof Response) return result;` narrowing at `start.ts:373` short-circuits before the JSON/session/header merge. No other route today returns a raw `Response`, so no accidental session-persist skip elsewhere. `bun test src/server` = 706 pass / 0 fail; `typecheck` + `lint` clean. Auth ordering is correct — `handleApiRequest` gates non-public paths on `session.user` at line 352 before dispatch, so `req.session.user!` inside `getEventsRoute` is safe, and the `MAX_SUBSCRIBERS_PER_USER` cap runs first inside the route body (after auth) with 429 status via `res.status(429)` + `ApiResponse` return — no raw-Response short-circuit for the cap path, so status propagates through `MutableResponse` correctly (matches commit `9e722a9d`). Pre-abort guard (`ac816f13`) correctly avoids the leak-until-first-keepalive-tick path by setting `closed = true` before `start()` runs; Set iteration in `emitToUser` is safe under current-item `set.delete` (spec-guaranteed, and prod's `sub.close()` → `unregisterSubscriber` → `set.delete(sub)` mutates only the current entry).

### High
- No blockers.

### Med
- No blockers.

### Low
- `src/server/routes/events/get-events.ts:17-22` — `SECURITY_HEADERS` (nosniff / X-Frame-Options / Referrer-Policy / CSP / HSTS-in-prod) are no longer applied to the SSE response. The inline `handleSseRequest` briefly on this branch (cf40f411) spread `...SECURITY_HEADERS` into its response headers; the refactor to a normal route (a72dc083) plus the raw-`Response` short-circuit at `start.ts:373` skips the `responseHeaders = { ...SECURITY_HEADERS }` merge that every other route gets at `start.ts:386`. Not a regression vs `upstream/main` (this is the first PR shipping `/api/events`, so nothing was live before), and the risk on a `text/event-stream` body is low — CSP is HTML-scoped and nosniff is largely redundant against an explicit `text/event-stream` Content-Type — but this looks incidental rather than deliberate. Worth either merging `SECURITY_HEADERS` into `SSE_HEADERS` (or applying them from `handleApiRequest` before the short-circuit) or leaving a one-line comment stating the omission is intentional. Verified via `git show cf40f411 -- src/server/start.ts` (old headers block includes `...SECURITY_HEADERS`) and re-reading the current short-circuit path in `start.ts:373-375`.
- `src/server/lib/realtime.ts:26-32` ("The client's dispatch table (PR 2) maps…"), `realtime.ts:63-66` ("see the follow-up PR for the per-route wiring"), and `src/client/lib/hooks/useServerEvents.ts:5-10` ("the receiver-side dispatch (PR 2) picks it up") — PR-narrative comments that reference upcoming PRs. Per `feedback_comment_hygiene`, follow-up-PR references age poorly once the referenced PR lands and belong in the PR body, not committed code. Suggest replacing with the invariant ("called from mutation route handlers after the DB write settles") without the "PR 2" hook.

claoie:
Reviewoie's 2 LOWs addressed in `79627bda`:

- **LOW 1 (SECURITY_HEADERS gap on /api/events)** — `SECURITY_HEADERS` moved to `src/server/lib/security-headers.ts` and re-exported from the server barrel. Both `start.ts` and the new `/events` route import from there. SSE Response now stamps nosniff / X-Frame-Options / X-XSS-Protection / Referrer-Policy / CSP alongside the SSE-specific headers. Verified via curl: `X-Content-Type-Options: nosniff` (and the rest) present on `GET /api/events`.
- **LOW 2 (PR-narrative comments)** — dropped forward-references to "PR 2" from `realtime.ts` and `useServerEvents.ts`. Kept the `useServerEvents` docstring that documents its contract (one handler per hook, switch on `domain` to split dispatch).

CI green on head `79627bda`; edge-case E2E clean (22/22 sub/unsub balance).

hoiekim (approved):

PR #658, opened 2026-07-26
feat(realtime): emit <domain>-updated on mutation + client re-sync (#656 PR 2/3)

claoie:
## Summary

PR 2 of 3 for the [Real-time collaboration issue](https://github.com/hoiekim/budget/issues/656). Wires the SSE transport from #657 into the mutation lifecycle so a write in one tab/user propagates to every open tab of that user in real time.

**Stacked on #657** — this branch is based on `feat/sse-realtime-plumbing`, so until #657 merges the diff here also shows #657's transport files (`realtime.ts` additions, the `start.ts` SSE branch, `useServerEvents.ts`). **Review scope for THIS PR is 4 changes:** the `MUTATION_DOMAINS`/`mutationEmitDomain` routing in `realtime.ts`, the emit call in `start.ts` `handleApiRequest`, `call.ts` `X-Tab-Id`, and the `useServerEvents` receiver in `Utility.tsx`. I'll rebase onto `main` once #657 lands and the diff collapses to just these.

## How it works

1. **Emit on mutation (`start.ts`)** — after a route settles with `status: "success"`, `mutationEmitDomain(method, path)` resolves the affected data-domain and `emitToUser` broadcasts `<domain>-updated` to every open tab of the acting user.
2. **Origin filtering** — the client sends a stable per-tab `X-Tab-Id` on every request (`call.ts`); the server echoes it back on the event so the tab that made the change skips its own event (its local state was already patched optimistically by `useMutate`).
3. **Receiver (`Utility.tsx`)** — `useServerEvents` fires `sync()` on any non-own-tab event. `sync()` is already delta/cursor-based and internally debounced, so a burst of events collapses to one incremental fetch.

### Where the emit lives — central, not per-route

I put the emit in `handleApiRequest` (the request-lifecycle choke point that already centralizes rate-limit, session, bearer, auth-gate, and the access-log) with a `MUTATION_DOMAINS` path→domain table in `realtime.ts`, rather than adding an `emitToUser` call to each of ~30 mutation handlers. One emit site, one routing table, `EmitDomain`-typed so a stale domain name is a compile error. `mutationEmitDomain` gates on a mutating method (POST/DELETE or a `/new-*` shell INSERT) so a GET read on a path that also carries a write verb (`/snapshots/holding`, `/transfers`) does not emit. (If you'd rather declare `emitDomain` per-route on the `Route` — the way `requiredScope` is co-located — I'll switch; the central table was the smaller, single-source diff.)

### Domain granularity

The receiver calls the whole `sync()` today because the client sync is monolithic (one delta pull refreshes every domain). The issue's step 5 ("call sync for specific data components") — promoting the per-domain `fetch*` helpers in `sync.ts` to exports and dispatching by the event's domain — is a clean follow-up (PR 3) once real-time delivery is proven here.

### Correction to #657's description

#657's body says "existing 30s polling stays intact as a safety net." There is **no polling** in the client today (verified: no `setInterval`/refetch loop in `sync.ts`/`Utility.tsx`; the only timer is the 50 ms sync debounce). The client currently auto-refreshes **only on login and manual refresh** — so this PR adds the *first* automatic cross-session refresh, and PR 3 has no polling to remove.

## Not in this PR (by design)

- **Background-sync push** (Plaid/SimpleFin webhook + scheduled cron writes). Those don't run under a user session (no `originTabId`, and the cron path never hits `handleApiRequest`), so they need item→user resolution and a separate emit source. Follow-up.
- **Per-domain sync** (see above) — PR 3.

## Test plan

- [x] `bun run lint` — clean
- [x] `bun test src` — 1118 pass / 0 fail (+6 new in `realtime.test.ts` covering `mutationEmitDomain`: POST/DELETE→domain, `/new-*` GET treated as mutating, multi-domain writes→representative domain, GET read on a shared write path→null, pure reads / non-domain writes→null)
- [x] **UI E2E** (two-tab, sandbox) — see below.

---

**UI E2E — two-tab real-time propagation (sandbox, mobile 390×844, admin login):**

Two browser contexts A and B authenticated as the same user, both on `/budgets`. Verified the full chain with a *real UI click* (not a `page.goto`/`fetch`):

- Tab A clicks the **"Add Budget"** affordance (`button.AddButton` → `GET /new-budget` shell INSERT). Confirmed `GET /api/new-budget` fired once (the real mutation).
- Tab B (no interaction) received the server push and **re-synced**: a full `sync()` burst fired (`GET /api/budgets` + `/api/accounts` + `/api/charts` + `/api/transfers` + delta stores). Tab B's budget list grew by one row **without any reload or interaction** — the shell budget A created rendered in B. Screenshot eye-checked: `/budgets` renders cleanly, nav intact, no layout break.
- **Origin filter verified:** tab A did **not** re-sync on its own event (0 sync-burst requests after its click) — the `X-Tab-Id` echo filter skips the originating tab's redundant refetch.
- A raw in-page `EventSource('/api/events')` in tab B directly captured `event: budgets-updated` / `data: {"originTabId":"<A's tab id>"}`, confirming the server-push transport + `originTabId` echo end-to-end.
- Anonymous `GET /api/events` → 401 (auth enforced). No `401`/auth console errors after the fix below.

**Bug found + fixed by this E2E (commit `2406622`):** `useServerEvents` opened the stream on mount, *before* login. The anonymous request 401s, and per the EventSource spec a non-2xx is a fatal close with no auto-reconnect — so the app's receiver stayed permanently dead after login. Fixed by gating the connection on `userLoggedIn` (`enabled` param); the effect re-runs and connects only once authenticated.

**reviewoie:** 0 high / 0 med / 2 low (informational) on the PR-2 delta — emit gate verified complete + correctly mapped against every route; no correctness/scope defects. LOW1 (`/plaid-hook` background-sync push) is the deferred follow-up already noted above; LOW2 (`Array.isArray(tabHeader)` arm) is type-required for the `string | string[]` header record, harmless.

### DIFF
--- a/src/client/components/App/Utility.tsx
+++ b/src/client/components/App/Utility.tsx
@@ -41,7 +41,7 @@ const Utility = () => {
   useServerEvents((_domain, payload) => {
     if (payload.originTabId === tabId) return;
     sync();
-  });
+  }, userLoggedIn);
 
   /**
    * Calculate balance history when data is updated
--- a/src/client/components/App/Utility.tsx
+++ b/src/client/components/App/Utility.tsx
@@ -12,7 +12,7 @@ const Utility = () => {
   const userLoggedIn = !!user;
   const { path, go } = router;
 
-  const { sync, clean } = useSync();
+  const { sync, syncDomain, clean } = useSync();
   const debouncer = useDebounce();
 
   /**
@@ -32,15 +32,16 @@ const Utility = () => {
   }, [userLoggedIn, sync, clean]);
 
   /**
-   * Real-time collaboration (#656): a server-pushed `<domain>-updated` event
-   * means another tab / another user changed data — re-sync to pull it in.
-   * `sync()` is already delta/cursor-based and internally debounced, so a burst
-   * of events collapses to one incremental fetch. Events tagged with this tab's
-   * own `tabId` are its own writes (already applied optimistically) and skipped.
+   * Real-time collaboration (#656): a server-pushed `<table>-updated` event
+   * means another tab / another user changed that table — refetch just that
+   * domain's slot via `syncDomain(domain)`. Events tagged with this tab's
+   * own `tabId` are its own writes (already applied optimistically) and
+   * skipped. Per-domain debouncing collapses bursts of the same event; the
+   * cursor is not advanced (only the whole-app `sync()` owns that).
    */
-  useServerEvents((_domain, payload) => {
+  useServerEvents((domain, payload) => {
     if (payload.originTabId === tabId) return;
-    sync();
+    syncDomain(domain);
   }, userLoggedIn);
 
   /**
--- a/src/client/components/App/Utility.tsx
+++ b/src/client/components/App/Utility.tsx
@@ -1,5 +1,5 @@
 import { useEffect } from "react";
-import { useAppContext, useSync, PATH, useDebounce } from "client";
+import { useAppContext, useSync, useServerEvents, tabId, PATH, useDebounce } from "client";
 
 /**
  * This component is used to run useEffect hooks dependant on context variables.
@@ -30,6 +30,18 @@ const Utility = () => {
     if (userLoggedIn) sync();
     else clean();
   }, [userLoggedIn, sync, clean]);
+
+  /**
+   * Real-time collaboration (#656): a server-pushed `<domain>-updated` event
+   * means another tab / another user changed data — re-sync to pull it in.
+   * `sync()` is already delta/cursor-based and internally debounced, so a burst
+   * of events collapses to one incremental fetch. Events tagged with this tab's
+   * own `tabId` are its own writes (already applied optimistically) and skipped.
+   */
+  useServerEvents((_domain, payload) => {
+    if (payload.originTabId === tabId) return;
+    sync();
+  });
 
   /**
    * Calculate balance history when data is updated
--- a/src/client/lib/call.ts
+++ b/src/client/lib/call.ts
@@ -1,15 +1,31 @@
 import { ApiResponse } from "server";
 
+/**
+ * Stable per-tab identifier, generated once when the client bundle loads (so
+ * each browser tab gets its own). Sent as `X-Tab-Id` on every request; the
+ * server echoes it back on the real-time event it emits for the resulting
+ * mutation, letting this tab recognize and skip its own writes (#656).
+ */
+export const tabId =
+  typeof crypto !== "undefined" && "randomUUID" in crypto
+    ? crypto.randomUUID()
+    : `${Date.now()}-${Math.random().toString(36).slice(2)}`;
+
 const call = async <T = unknown>(path: string, options?: RequestInit): Promise<ApiResponse<T>> => {
   const method = options?.method || "GET";
   const body = options?.body;
 
-  const init: RequestInit | undefined = options;
+  const init: RequestInit = { ...options };
+  const headers: Record<string, string> = {
+    ...(options?.headers as Record<string, string> | undefined),
+    "X-Tab-Id": tabId,
+  };
 
   if (method === "POST") {
-    (init as RequestInit).headers = { "Content-Type": "application/json" };
-    (init as RequestInit).body = JSON.stringify(body);
+    headers["Content-Type"] = "application/json";
+    init.body = JSON.stringify(body);
   }
+  init.headers = headers;
 
   try {
     const httpResponse = await fetch(path, init);
--- a/src/client/lib/hooks/sync.ts
+++ b/src/client/lib/hooks/sync.ts
@@ -1,5 +1,5 @@
 import { useCallback } from "react";
-import { JSONInstitution, JSONSnapshotData } from "common";
+import { JSONInstitution, JSONSnapshotData, TableName } from "common";
 import {
   BudgetsGetResponse,
   TransactionsGetResponse,
@@ -443,6 +443,27 @@ const debounceSync = (callback: () => void, delay = 50) => {
   syncDebounceTimeout = setTimeout(callback, delay);
 };
 
+// Per-domain debounce for `syncDomain`. A burst of `<domain>-updated`
+// events for the same domain collapses into a single refetch; different
+// domains debounce independently so a `transactions-updated` doesn't
+// swallow a concurrent `budgets-updated`.
+const domainDebounceTimeouts = new Map<TableName, ReturnType<typeof setTimeout>>();
+const debounceDomain = (
+  domain: TableName,
+  callback: () => void,
+  delay = 50,
+) => {
+  const existing = domainDebounceTimeouts.get(domain);
+  if (existing) clearTimeout(existing);
+  domainDebounceTimeouts.set(
+    domain,
+    setTimeout(() => {
+      domainDebounceTimeouts.delete(domain);
+      callback();
+    }, delay),
+  );
+};
+
 export const useSync = () => {
   const { user, setData } = useAppContext();
   const _sync = useCallback(async () => {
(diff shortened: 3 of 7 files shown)


reviewoie (commented):
## Code review — PR 658 (feat(realtime): emit <domain>-updated on mutation + client re-sync, #656 PR 2/3)

Scope reviewed: the PR-2 delta only (commit `956d9868`) — `realtime.ts` MUTATION_DOMAINS + `mutationEmitDomain`, the `start.ts` emit block, `call.ts` tabId/`X-Tab-Id`, `Utility.tsx` receiver, and the +6 `realtime.test.ts` cases. The #657 transport files inherited in the stacked diff were not re-reviewed.

**Verdict:** informational (approve-leaning) — no correctness/scope defects found in this delta.

What I verified clean:
- **Emit gate is complete and correctly mapped.** Enumerated every `new Route(...)` in `src/server/routes/` (54 files, generic type params included) and cross-checked against `MUTATION_DOMAINS` + the method gate: every user-facing data-mutating route (POST/DELETE `/transaction` `/account` `/budget` `/section` `/category` `/chart` `/investment-transaction` `/split-transaction` `/snapshot` `/snapshots/holding` `/item` `/public-token` `/resolve-security-snapshot` `/suggest-category` `/transfers` `/transfers/pair`, plus all seven `GET /new-*` shell INSERTs) resolves to a domain; non-domain writes (`/api-keys` `/login` `/client-error` `/validate-ticker`) resolve to `null`. No mutating data route is missing from the table and none is mis-verbed — `GET /snapshots/holding` and `GET /transfers` correctly return `null` via the method gate while their POST/DELETE twins map.
- **Emit cannot break the response path.** `emitToUser` early-returns on no subscribers and wraps every `sub.send` in try/catch (`realtime.ts:125-139`), so it never throws into `handleApiRequest`. The block is gated on `result?.status === "success" && session.user` (`start.ts:506`), so an unmatched route (`result === null`) or an anonymous/webhook request never emits.
- **Header read is correct.** `headers` is built from `request.headers.forEach((v,k)=>headers[k]=v)` (`start.ts:397-400`) — Bun lowercases, so `headers["x-tab-id"]` (`start.ts:507`) matches the client's `X-Tab-Id`.
- **Self-filter is sound.** `tabId` is a stable per-tab module constant sent on every request and echoed back as `payload.originTabId`; the receiver's `if (payload.originTabId === tabId) return;` (`Utility.tsx`) correctly skips only the originating tab and re-syncs every other tab of the same user.
- **No burst / perf concern.** `useSync` returns `debounceSync(_sync)` (50ms, `sync.ts:441-843`), so a burst of events collapses to one delta fetch — the receiver comment is accurate. Emit is O(1) table lookup + O(tabs-of-one-user) fan-out; no cross-user cost.
- **`call.ts` merge is safe.** No `call()` caller passes an `options.headers`, so the `{...options?.headers}` spread (which would flatten a `Headers` instance to `{}`) is future-proofing only, not a live regression. POST body/Content-Type behavior is preserved.
- **Test coverage for `mutationEmitDomain` is thorough** — POST/DELETE mappings, `/new-*` GET, multi-domain (`/public-token`→accounts, `/item`→accounts, `/suggest-category`→transactions), GET-on-shared-write null, pure-read null, non-domain-write null, unknown-path null.

### Low
- `src/server/routes/webhooks` (POST `/plaid-hook`) — **server-pushed Plaid data changes emit no realtime event.** `/plaid-hook` is not in `MUTATION_DOMAINS`, and even if it were, the emit is gated on `session.user` (`start.ts:506`) which is null for a signature-verified webhook. Failure scenario: Plaid ingests new transactions (or the async transaction/holding backfill after linking an item completes) while a user has a tab open — that tab won't live-update until its next user-initiated mutation or a manual refresh. This is likely intentional PR-2 scope (user-initiated mutations only; a webhook would need an item_id→user_id lookup to call `emitToUser` directly), but Plaid-pushed transactions are arguably the highest-value realtime signal — confirm it's deferred to PR 3 rather than an oversight.
- `src/server/start.ts:508` — **dead `Array.isArray(tabHeader)` branch.** `headers` is populated via `Headers.forEach`, which always yields a comma-joined string (never an array), so `tabHeader` is always `string | undefined` and the array arm never executes. Harmless (the `headers` field is typed `string | string[] | undefined` elsewhere, so the guard is type-defensible) — no action needed; noted for accuracy.

No further blockers.

claoie:
Thanks — addressing both LOWs:

**LOW 1 — `/plaid-hook` (and post-link backfill) emit no realtime event.** Confirmed intentional / deferred, not an oversight. Background-sync writes (Plaid/SimpleFin webhook + the scheduled cron) don't run under a user session — the webhook path has no `originTabId` and the cron never hits `handleApiRequest` at all — so they need item→user resolution and a separate emit source. Called out under "Not in this PR (by design)". It's arguably the highest-value push signal, so it's the natural next follow-up after this lands.

**LOW 2 — dead `Array.isArray(tabHeader)` arm.** Kept: `ServerRequest.headers` is typed `Record<string, string | string[] | undefined>`, so `headers["x-tab-id"]` is `string | string[]` to the type-checker even though Bun's `forEach` only ever yields a string at runtime. The guard narrows the type before passing to `EmitPayload.originTabId: string | undefined`; dropping it needs a cast instead. No runtime path change either way.

Also flagging a bug this PR's two-tab E2E caught and fixed (commit `2406622`): the `useServerEvents` receiver opened the stream on mount before login → anonymous `/api/events` 401 → fatal EventSource close with no reconnect → the app's receiver was dead after login. Fixed by gating the connection on auth state.

src/client/components/App/Utility.tsx
```diff
@@ -31,6 +31,18 @@ const Utility = () => {
...
+  /**
+   * Real-time collaboration (#656): a server-pushed `<domain>-updated` event
+   * means another tab / another user changed data — re-sync to pull it in.
+   * `sync()` is already delta/cursor-based and internally debounced, so a burst
+   * of events collapses to one incremental fetch. Events tagged with this tab's
+   * own `tabId` are its own writes (already applied optimistically) and skipped.
+   */
+  useServerEvents((_domain, payload) => {
+    if (payload.originTabId === tabId) return;
+    sync();
```
hoiekim: Does this sync all data? Shouldn't we sync specific table only?

reviewoie (commented):
## Code review — PR 658 (round 2: domain-scoped `syncDomain` refactor)

Scope reviewed: the new `2a767349` delta only — `_syncDomain` + `syncDomain` + `debounceDomain` in `src/client/lib/hooks/sync.ts` (+236 lines), and the `useServerEvents` wiring shift in `Utility.tsx`. The PR-2 receiver / emit / MUTATION_DOMAINS surface was already cleared in the prior reviewoie pass at `956d9868`.

**Verdict:** informational (approve-leaning) — one MED pattern concern, no correctness blocker.

### Med

- `src/client/lib/hooks/sync.ts:929-933` and `:1043-1047` — **`setData((oldData) => { <capture>; return oldData; })` used as a synchronous getter for `oldData.accounts` in the Snapshots and Institutions branches**. The pattern relies on React 18's `dispatchSetState` eager-bailout — only when `fiber.lanes === NoLanes && alternate.lanes === NoLanes` does React run the updater synchronously in-line; otherwise the updater is queued and doesn't run until the next render.

  Failure scenario: any sibling `useState` on the `AppContext` fiber (login `setUser`, `useRouter`, `useViewDate`, `useScreenType`, or the very same `setData` from a concurrent mutation in flight) with a pending update leaves `fiber.lanes !== NoLanes`. The eager path is skipped, `accountsForSnapshot`/`accountsForInst` stays `undefined`, and the next line `if (!accountsForSnapshot) return;` silently drops the SSE-triggered snapshot/institution refetch — no log, no retry, the other tab's change never propagates until the next full `sync()`.

  Also violates React's "updater must be pure" contract — under dev Strict Mode the updater double-invokes, so `accountsForSnapshot` gets assigned twice per call (same value, so no functional harm today, but any future non-idempotent capture would silently misbehave).

  Cleaner shape: `globalData` (from `src/client/lib/models/Data.ts:296`) is already the mutable mirror `useData`'s `setData` wrapper keeps synced on every commit — `Transaction.ts:32` already reads from it. `_syncDomain`'s Snapshots/Institutions branches can drop the setData-as-getter and just call `fetchSnapshots(globalData.accounts, cursor)` / `fetchInstitutions(globalData.accounts)` directly. Same current-state semantics, no eager-bailout dependency, no impure updater.

  In current call sites the failure mode is rare (SSE onmessage runs in a fresh macrotask so `fiber.lanes` are typically clear), but the pattern is a landmine for future callers.

### Low

- `src/client/lib/hooks/sync.ts:966-988` — **`TableName.Holdings` case has no live emitter**. Grepped `MUTATION_DOMAINS` — no route path maps to `TableName.Holdings`. `/snapshots/holding` maps to `Snapshots` (correctly — it writes `holding_snapshots`, not `holdings`), and holdings are only mutated by `sync-plaid` / `sync-simplefin` which don't go through the API path so they never call `emitToUser`. The branch is dead in this PR; harmless (kept for exhaustiveness + future-proofing when a followup wires the sync engine to emit), but worth flagging so a reader doesn't chase a phantom "why doesn't my holdings change propagate" thread.

- `src/client/lib/hooks/sync.ts:871-1071` — **switch has no `default: assertNever(domain)`**. All 18 `TableName` values are covered today, but a future enum addition would silently fall through and `_syncDomain` would no-op with no compile-time warning. Adding `default: return domain satisfies never;` (or an `assertNever` helper) turns the next enum extension into a TS error at this file.

No further blockers.

hoiekim (approved):

PR #659, opened 2026-07-27
feat(realtime): plaid-hook emits <table>-updated to item owner (#656 PR 3/3)

claoie:
Third and final PR of the real-time collab arc (#656). PR 1 shipped the SSE transport (#657), PR 2 wired mutation routes to emit + client to receive (#658), this PR closes the last gap: **Plaid webhooks**.

## Why

`POST /plaid-hook` lands out-of-band from any user tab and doesn't ride the normal per-user-request emit path in `start.ts` (no session, so no `session.user` for the request-choke-point emit to route on). Result: auto-imported transactions, holding refreshes, and item-status changes required a page reload to appear in the UI while every other data domain updated live.

## What

After a webhook handler's mutation settles, resolve the item's owner via `getUserItem(item_id)` and fan the appropriate `<table>-updated` events to that user's tabs via `emitToUser`. Since the request has no `X-Tab-Id` (Plaid isn't in a tab), the payload has no `originTabId` — every tab of the owning user picks it up and the client-side per-domain debouncer collapses bursts.

Per handler:

| Webhook | Emit domain(s) |
|---|---|
| `TRANSACTIONS.SYNC_UPDATES_AVAILABLE`, `HOLDINGS.DEFAULT_UPDATE`, `INVESTMENTS_TRANSACTIONS.DEFAULT_UPDATE\|HISTORICAL_UPDATE` (via `syncAndLog`) | `transactions` + `investment_transactions` — only when the sync reports non-zero added/modified/removed |
| `ITEM.USER_ACCOUNT_REVOKED`, `ITEM.ITEM_UPDATED` (via `refreshItemProducts`) | `accounts` (the accounts fetch returns items on the same call) |
| `ITEM.PENDING_EXPIRATION`, `ITEM.ERROR + ITEM_LOGIN_REQUIRED` (via `markBadItem`) | `accounts` — surfaces the bad-item state without a reload |

## Test plan

- [x] `bun test src/server/routes/webhooks` — 27/27 pass. Added 3 emit-behavior tests: sync path emits both domains, empty-sync doesn't emit, item-status paths emit accounts.
- [x] Full `bun test src/server` — 713 pass.
- [x] `bunx tsc --noEmit` clean.
- [x] Sandbox E2E: signed webhook POST → owning user's SSE stream receives events. (Sandbox up next; will drive real Plaid HMAC.)

Closes the #656 real-time-collab arc.

### DIFF
--- a/src/server/lib/compute-tools/sync-plaid.ts
+++ b/src/server/lib/compute-tools/sync-plaid.ts
@@ -263,6 +263,7 @@ export const syncPlaidTransactions = async (item_id: string) => {
   await Promise.all([syncTransactions, syncInvestmentTransactions]);
 
   return {
+    user_id: user.user_id,
     added: addedCount,
     modified: modifiedCount,
     removed: removedCount,
--- a/src/server/routes/webhooks/post-plaid-hook.test.ts
+++ b/src/server/routes/webhooks/post-plaid-hook.test.ts
@@ -32,6 +32,7 @@ const mockGetUserItem = mock(
 const mockUpsertItems = mock(
   async (_user: unknown, _items: unknown[]) => [] as unknown[],
 );
+const mockEmitToUser = mock((_userId: string, _domain: string, _payload?: unknown) => {});
 const mockSendAlarm = mock(async () => {});
 const mockLogger = {
   info: mock(() => {}),
@@ -50,6 +51,7 @@ mock.module("server", () => ({
   updateItemStatus: mockUpdateItemStatus,
   getUserItem: mockGetUserItem,
   upsertItems: mockUpsertItems,
+  emitToUser: mockEmitToUser,
 }));
 
 mock.module("server/lib/alarm", () => ({
@@ -94,6 +96,7 @@ beforeEach(() => {
   }));
   mockUpsertItems.mockReset();
   mockUpsertItems.mockImplementation(async () => []);
+  mockEmitToUser.mockReset();
   mockSendAlarm.mockReset();
   mockSendAlarm.mockImplementation(async () => {});
   mockLogger.info.mockReset();
@@ -208,6 +211,27 @@ describe("post-plaid-hook — TRANSACTIONS", () => {
     expect(result).toMatchObject({ status: "success" });
   });
 
+  test("SYNC_UPDATES_AVAILABLE emits transactions + investment_transactions to owner on non-empty change", async () => {
+    mockSyncPlaidTransactions.mockImplementation(async () => ({ added: 3, modified: 0, removed: 0 }));
+    await run({
+      webhook_type: "TRANSACTIONS",
+      webhook_code: "SYNC_UPDATES_AVAILABLE",
+      item_id: "item-7",
+    });
+    expect(mockEmitToUser).toHaveBeenCalledWith("u-1", "transactions");
+    expect(mockEmitToUser).toHaveBeenCalledWith("u-1", "investment_transactions");
+  });
+
+  test("SYNC_UPDATES_AVAILABLE does NOT emit when nothing changed", async () => {
+    mockSyncPlaidTransactions.mockImplementation(async () => ({ added: 0, modified: 0, removed: 0 }));
+    await run({
+      webhook_type: "TRANSACTIONS",
+      webhook_code: "SYNC_UPDATES_AVAILABLE",
+      item_id: "item-7",
+    });
+    expect(mockEmitToUser).not.toHaveBeenCalled();
+  });
+
   test("SYNC_UPDATES_AVAILABLE with null sync result → failed", async () => {
     mockSyncPlaidTransactions.mockImplementation(async () => null);
     const { result } = await run({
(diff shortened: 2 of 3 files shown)


reviewoie (commented):
## Code review — PR 659

**Verdict:** changes-requested (one MED — double client re-fetch per webhook + a factually wrong justification comment)

### Med

- `src/server/routes/webhooks/post-plaid-hook.ts:93` — `syncAndLog` emits BOTH `TableName.Transactions` and `TableName.InvestmentTransactions`, but on the client (`src/client/lib/hooks/sync.ts:878-907`) both domains share ONE case body that calls `fetchTransactions(cursor)`. The debounce (`debounceDomain`, `sync.ts:452`) is keyed on the domain value, so two distinct keys (`"transactions"` vs `"investment_transactions"`) schedule two independent 50 ms timers → `_syncDomain` fires TWICE with different domains → `fetchTransactions` runs twice, plus duplicated `setData` (`TransactionDictionary` + `InvestmentTransactionDictionary` rebuilt twice) and duplicated `indexedDb.saveTransactions` / `saveInvestmentTransactions` writes. The inline comment ("Client syncDomain collapses concurrent domain events per its per-domain debounce") justifies the double-emit with a mechanism that doesn't exist — the debounce collapses within one domain, never across two different domain names. Fix: emit ONE of them (either is sufficient because the shared case body pulls both series in one `/api/transactions` call), or introduce an explicit "transactions bundle" identifier. Same asymptotic-cost lens as rule 13 — this doubles per-webhook client sync work permanently. Failure scenario: on every `TRANSACTIONS.SYNC_UPDATES_AVAILABLE` / `HOLDINGS.DEFAULT_UPDATE` / `INVESTMENTS_TRANSACTIONS.*` webhook that produces any change, every open tab of the owner runs two full `fetchTransactions` cycles ~50ms apart. Also update the comment — the "per-domain debounce" claim is misleading.

### Low

- `src/server/routes/webhooks/post-plaid-hook.ts:75-82` — `emitToItemOwner` re-runs `getUserItem(item_id)` inside `syncAndLog`, but `syncPlaidTransactions` already did an identical lookup (`sync-plaid.ts:114`). Two DB round-trips per webhook where one suffices. Not scalability-critical (webhooks are per-item, low frequency) but the natural refactor is to have `syncPlaidTransactions` return `{added, modified, removed, user_id}` so the emit can call `emitToUser(user_id, domain)` directly. Applies to `markBadItem` similarly, though it currently does no user lookup at all so hoisting `getUserItem` there is a bigger change.
- `src/server/routes/webhooks/post-plaid-hook.test.ts:298-308` — `ERROR + ITEM_LOGIN_REQUIRED → marks item BAD and alarms` doesn't assert the `emitToUser("u-1", "accounts")` that the new `markBadItem` emit produces. Same code path as `PENDING_EXPIRATION` (which does assert, line 294), but this second `markBadItem` entrypoint is left uncovered. Similarly, `HOLDINGS.DEFAULT_UPDATE` (`test.ts:371-379`) and `INVESTMENTS_TRANSACTIONS.{DEFAULT_UPDATE,HISTORICAL_UPDATE}` (`test.ts:394-404`) also hit `syncAndLog` but only assert `status: "success"` — they don't lock in the emit contract, so if someone later routes a HOLDINGS webhook to `syncPlaidAccounts` (the correct-looking fix for the pre-existing "HOLDINGS routes to a transactions sync" oddity), no test would catch that the emit domain set becomes wrong. Add emit assertions to at least one HOLDINGS and one INVESTMENTS_TRANSACTIONS test.
- `src/server/routes/webhooks/post-plaid-hook.ts:106,114` — `refreshItemProducts` and `markBadItem` fire `emitToUser(..., Accounts)` unconditionally on the success branch, even when `available_products` didn't actually change (`refreshItemProducts`) or `status` was already `BAD` (`markBadItem`). Client `fetchAccounts` is idempotent so no correctness issue, just a spurious refetch on a rare webhook. Not a blocker; note here so the pattern isn't copied to higher-frequency webhooks later.

claoie:
Reviewoie ran on head \`5a95a9c6\`: 1 MED + 3 LOW. Applied in \`c2f7381c\`.

- **MED — double client refetch per webhook.** Emitting both \`transactions\` and \`investment_transactions\` produced two separate debounce keys → the shared \`case Transactions | InvestmentTransactions\` body in \`syncDomain\` ran twice: two \`/api/transactions\` fetches, two setData rebuilds, two IDB writes per webhook. Fixed by emitting only \`transactions\` — the shared case body still refreshes investment_transactions on that one event. Test asserts \`toHaveBeenCalledTimes(1)\`.
- **LOW — second \`getUserItem\` round-trip.** Fixed. \`syncPlaidTransactions\` now returns \`user_id\` (additive; other callers ignore); \`syncAndLog\` emits without a second DB call.
- **LOW — missing emit assertions.** Fixed. Added \`toHaveBeenCalledWith("u-1", "transactions")\` on HOLDINGS.DEFAULT_UPDATE, both INVESTMENTS_TRANSACTIONS cases, and ITEM.ERROR+ITEM_LOGIN_REQUIRED.
- **LOW — no-op emits on unchanged item state.** Left as-is: item webhooks are rare, an equality check adds branch complexity for negligible savings, and a spurious emit triggers one debounced \`/api/accounts\` refetch.

CI green on \`c2f7381c\`. 713 server tests pass.

hoiekim (approved):

PR #660, opened 2026-07-28
refactor(snapshots): read through Table.query instead of raw buildSelectWithFilters bypass

### LINKED ISSUE #580: refactor(snapshots): route reads through Table.query instead of raw buildSelectWithFilters bypass
## Pattern drift — `snapshots.ts` is the lone repo bypassing the canonical `Table.query` Model method

Every other repository under `src/server/lib/postgres/repositories/` funnels reads through the canonical Model methods — `transactionsTable.query(...)`, `.queryOne(...)`, `.queryByIds(...)` (see `transactions.ts:43,67,106,214`). **`snapshots.ts` is the only repo that hand-rolls `buildSelectWithFilters(...)` + `pool.query(...)` + manual row re-hydration** instead.

This is [Principle 2 — one canonical way per thing] drift, and the `extend-Table-methods-over-bypass` pattern: reach for the Table method, don't re-implement its body in the repo.

### Why it's a no-op-to-fix in most cases

`Table.query(filters, options)` in `models/base.ts:125` **already delegates to the exact same `buildSelectWithFilters`** and already accepts every option these call sites pass — `user_id`, `filters`, `inFilters`, `dateRange`, `orderBy`, `limit`, `excludeDeleted`. So the bypass buys nothing; the Model method wraps the same builder + `pool.query` + `new ModelClass(row)`.

### Call sites (`snapshots.ts`)

- `getSecuritySnapshots` — `buildSelectWithFilters` @ line 206. Plain `snapshot_type='security'` + optional `security_id` + `dateRange(SNAPSHOT_DATE)` + `orderBy`. Directly expressible as `snapshotsTable.query({...}, {dateRange, orderBy})`.
- `getHoldingSnapshots` — `buildSelectWithFilters` @ line 232. Same shape, `user_id`-scoped, `snapshot_type='holding'`. Directly expressible.
- `searchSnapshots` — three builder calls @ 87 (`userScoped`), 120 (`holdingScoped`), 157 (`globalSecurity`). **This one has a legitimate reason to make multiple calls** (documented @ line 62-70): security-price rows are stored `user_id = NULL` (shared across users), so the user-scoped filter would exclude them — it needs an `user_id = $1 OR user_id IS NULL` union that a single `Table.query` can't express. But the *individual* calls can still each be a `snapshotsTable.query(...)` + concat, rather than raw builder + `pool.query`.

### Shape-preservation caveat (why this isn't a blind swap)

Each function re-hydrates rows into a **domain-specific shape** via a custom `.map()` — `getSecuritySnapshots` → `SecuritySnapshot` (subset: `snapshot_id/snapshot_date/security_id/close_price`), `getHoldingSnapshots` → `HoldingSnapshot`. `Table.query` returns `SnapshotModel[]`. So the migration is: call `snapshotsTable.query(...)` to get models, then keep the existing projection `.map()`, OR push the projection into the callers. The fix must preserve the exact returned shapes (the benchmark + holdings-composition calculations depend on them) — verify with `benchmark.test.ts` + a Holdings/Performance UI E2E after.

### Scope

Single file, ~5 builder call sites, one kind of change. Good candidate to bundle with any other `Table`-bypass finding if one surfaces. Import of `buildSelectWithFilters` (line 21) drops out once the plain selects migrate; keep it only if `searchSnapshots`' OR-union still needs it.

claoie:
## What

`snapshots.ts` was the **lone repository** hand-rolling `buildSelectWithFilters(...)` + `pool.query(...)` + manual row re-hydration for reads, while every other repo funnels reads through the canonical `Table.query` Model method (`transactions.ts`, `accounts.ts`, …). This migrates the five read call sites onto `snapshotsTable.query`.

Closes #580. [Principle 2 — one canonical way] / the extend-Table-methods-over-bypass pattern.

## Why it's behavior-preserving

`Table.query(filters, options)` already **delegates to the exact same `buildSelectWithFilters`** and accepts every option these call sites pass — `user_id`, `filters`, `inFilters`, `dateRange`, `orderBy`, `limit`, `excludeDeleted`. So the emitted SQL is byte-identical; the bypass bought nothing.

- **`excludeDeleted` default matches.** The builder defaults `excludeDeleted = true`; `Table.query` defaults it to `this.supportsSoftDelete` (= `true` for snapshots). The two `get*` sites that previously omitted the flag keep excluding soft-deleted rows.
- **Returned shapes preserved.** `getSecuritySnapshots` / `getHoldingSnapshots` still project to their custom `SecuritySnapshot` / `HoldingSnapshot` shapes; `searchSnapshots` still returns `JSONSnapshotData[]` via `model.toJSON()`.
- **`searchSnapshots`' two-query split stays.** A single `user_id` filter can't express the `user_id = $1 OR user_id IS NULL` union that surfacing shared (user_id NULL) security-price rows requires — so it stays three scoped `.query` calls + concat, same as before. The `rowToSnapshot` shim is retired in favor of `model.toJSON()`.

## Reads now hydrate `SnapshotModel`

`Table.query` constructs+validates a `SnapshotModel` per row — which `searchSnapshots` already did in production via `rowToSnapshot`, so this is not new behavior on real data. Production is safe because `client.ts` parses `NUMERIC`/`INT8` → `number` (via `parseFloat`) and `SELECT *` returns every column, so the model always validates.

The route / compute-tool **test mocks** were feeding partial rows with string numerics (an artifact of the old manual-`Number()` projection). They now return the **full row shape with numeric types** — matching what production pg yields — exactly as the existing `securityRow` factory already documents.

## Test plan

- [x] `bun test src` — 1128 pass / 0 fail / 2 skip (route + compute-tool snapshot suites retargeted to full-row mocks)
- [x] **UI E2E** — sandbox, mobile 390×844, admin login, PNGs eye-checked:
  - `/accounts` — 19 account rows render with balances + per-row balance graphs (23 canvases); `/api/snapshots` (searchSnapshots) fires 200. No console errors.
  - `/account-detail` for an **investment (Stock Plan)** account, real `.AccountRow` click:
    - **Balance graph** (Use Snapshots ON) draws the historical line — fed by the account-narrowed `/api/snapshots?…` (searchSnapshots holding branch). 200.
    - **Holdings Composition** table populates: priced security row with **ticker + name resolved** (proves the security-snapshot enrichment path in `searchSnapshots`), an `Unknown` row, and a `Total` row.
    - **Performance widget** computes: `Your asset return (MWR)`, `<index> benchmark`, and `Diff` rows all render — the snapshot-anchored MWR/benchmark math is intact.
    - 0 non-SW console errors.
  - Values are unchanged from `upstream/main` by construction: `Table.query` delegates to the identical `buildSelectWithFilters` with the same options, so the emitted SQL is byte-for-byte the same as the removed bypass. `getSecuritySnapshots` feeds server-side compute tools (backfill / resolve-security) not directly reachable through the UI on a scrambled clone — its shape is pinned by the retargeted unit suites.

### DIFF
--- a/src/server/lib/compute-tools/backfill-snapshots.test.ts
+++ b/src/server/lib/compute-tools/backfill-snapshots.test.ts
@@ -99,12 +99,29 @@ const makeSecurityRow = (overrides: Record<string, unknown> = {}) => ({
   ...overrides,
 });
 
+// `getSecuritySnapshots` now reads through `snapshotsTable.query`, which
+// hydrates each row into a `SnapshotModel` — its typeChecker requires every
+// column present (null is fine, undefined is not), so this returns the full
+// row shape.
 const makeSnapshotRow = (overrides: Record<string, unknown> = {}) => ({
   snapshot_id: "snap-1",
+  user_id: null,
   snapshot_date: "2026-04-15",
   snapshot_type: "security",
+  account_id: null,
+  balances_available: null,
+  balances_current: null,
+  balances_limit: null,
+  balances_iso_currency_code: null,
   security_id: "sec-1",
   close_price: 90,
+  holding_account_id: null,
+  holding_security_id: null,
+  institution_price: null,
+  institution_value: null,
+  cost_basis: null,
+  quantity: null,
+  updated: null,
   is_deleted: false,
   ...overrides,
 });
--- a/src/server/lib/postgres/repositories/snapshots.ts
+++ b/src/server/lib/postgres/repositories/snapshots.ts
@@ -1,5 +1,4 @@
 import { JSONSnapshotData } from "common";
-import { pool } from "../client";
 import {
   MaskedUser,
   SnapshotModel,
@@ -17,9 +16,7 @@ import {
   HOLDING_SECURITY_ID,
   USER_ID,
 } from "../models";
-import { SNAPSHOTS } from "common/constants";
-
-import { UpsertResult, successResult, errorResult, buildSelectWithFilters } from "../database";
+import { UpsertResult, successResult, errorResult } from "../database";
 import { searchSecuritiesById } from "./securities";
 import { logger } from "../../logger";
 
@@ -57,18 +54,15 @@ export interface HoldingSnapshot {
   quantity?: number;
 }
 
-const rowToSnapshot = (row: Record<string, unknown>): JSONSnapshotData =>
-  new SnapshotModel(row).toJSON();
-
 // Security snapshots are price data — stored with `user_id = NULL` because
 // they're shared across all users. The user-scoped query below would
 // otherwise exclude them by the `user_id = $1` filter, leaving the frontend
 // unable to resolve `security_id → ticker_symbol` (Closes #323).
 //
-// Two queries instead of one: the user-scoped table-helper path doesn't
-// support an OR predicate, and a raw SQL rewrite of the whole search would
-// duplicate the date-range / inFilters logic that `buildSelectWithFilters`
-// already encodes. Two helper calls + concat is the smaller change.
+// Two queries instead of one: `snapshotsTable.query` filters by a single
+// `user_id`, so it can't express the `user_id = $1 OR user_id IS NULL` union
+// that surfacing shared security rows alongside the user's own requires.
+// Two scoped `.query` calls + concat is the smaller change.
 export const searchSnapshots = async (
   user: MaskedUser | null,
   options: SearchSnapshotsOptions = {},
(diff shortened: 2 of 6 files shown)


reviewoie (commented):
## Code review — PR 660

Verified the migration is behavior-preserving end to end:

- **SQL byte-equivalence (all 5 sites):** `Table.query` (`models/base.ts:125`) forwards `filters/excludeDeleted/user_id/inFilters/dateRange/orderBy/limit` into the same `buildSelectWithFilters(this.name, "*", …)`; `primaryKey`/`offset` are undefined no-ops. Condition order in the builder is fixed (user_id → filters → inFilters → dateRange → excludeDeleted), independent of caller option order, so the emitted SQL + param order match the old bypass exactly.
- **excludeDeleted defaults:** builder default is `true` (`database.ts:309`); `createTable` defaults `supportsSoftDelete` to `true` (`base.ts:364`) and snapshots doesn't override it, so `snapshots.supportsSoftDelete === true`. The two `get*` sites (which omitted the flag) now resolve `undefined ?? true = true` — preserved. The three `searchSnapshots` sites pass an explicit boolean `!options.includeDeleted`, so `?? supportsSoftDelete` never fires — preserved.
- **Shape preservation:** `SnapshotModel` (`models/snapshot.ts`) declares every column the `SecuritySnapshot`/`HoldingSnapshot` projections read and `toJSON()` is unchanged, so `JSONSnapshotData[]` output is identical.
- **OR-union:** the global-security `.query` omits `user_id`, so no `user_id = $N` clause is emitted — shared `user_id IS NULL` rows still returned.
- **Test mocks:** the string→number change (`"12.50"` → `12.5`, etc.) is required, not a weakening — the pg client registers a global `NUMERIC → parseFloat` parser (`client.ts:30`), so real rows arrive as numbers, and `SnapshotModel`'s `isNullableNumber` typeChecker rejects strings. The old string mocks were unrealistic; the new numeric mocks match production. No assertion regressed (none asserted on the coerced numeric values). Completing the full column set is likewise required — `validateObject` (`base.ts:44`) checks every typeChecker key and throws on `undefined`.
- No raw `pool.query`/`buildSelect*` reads left in `snapshots.ts`; `pool` and `SNAPSHOTS` imports correctly dropped. No scope creep.
- Full `bun test src/server`: 717 pass / 0 fail. `typecheck` + `eslint` clean.

### Low
- `delete-holding-snapshot.test.ts:68-75` — stale comment contradicts its neighbor. The retained JSDoc "Only the columns the ownership-check actually consumes need to be present" is now false — the SnapshotModel hydration requires **all** columns, which the new comment (lines 73-75) states. Two adjacent comments now say opposite things. Drop the "only the columns…" sentence so the comment tells the truth (Principle 1).

No further blockers.

claoie:
Thanks reviewoie. Addressed the low in `d2a6cfff`:

- **delete-holding-snapshot.test.ts stale JSDoc** — merged the two contradicting comments into one accurate block. The old "only the columns the ownership-check actually consumes need to be present" line no longer holds now that reads hydrate `SnapshotModel` (which validates every typeChecker key), so it's replaced with the full-row-shape rationale.

Confirmed the same contradiction didn't exist in the other four retargeted factories (they had no prior "partial columns" JSDoc). `bun test src/server` 717 pass / 0 fail, lint clean on the fix.

hoiekim (approved):

PR #661, opened 2026-07-28
fix(labels): re-sync <select> state when underlying label prop changes

claoie:
Follows up your two-tab test in `#Budget label sync bug` — the SSE fanout from #656/#657/#658/#659 reaches tab 2's IDB and `data.transactions[id].label` in ~250 ms, but tab 2's rendered category / budget \`<select>\` keeps displaying the mount-time value until a full page refresh.

## Root cause

`useBudgetCategorySelect`:

```tsx
const [selectedBudgetIdLabel, setSelectedBudgetIdLabel] = useState(
  () => label.budget_id || account?.label.budget_id || "",
);
const [selectedCategoryIdLabel, setSelectedCategoryIdLabel] = useState(
  () => label.category_id || "",
);
```

React's `useState` lazy initializer runs **exactly once per mount**. It doesn't watch its argument, so subsequent prop changes to `label.*` never re-hydrate the local state. The controlled `<select>` ends up shadowing the data layer indefinitely.

The `TransactionRow` had a partial escape (`useEffect` set `selectedBudgetIdLabel` to the account default only when `label.budget_id` was falsy), and `TransactionProperties` had a full sync effect at its own layer — but no consumer had a category sync, so the row-level category dropdown stayed stale under any external label update (SSE fanout, background sync-plaid pending→posted reconciliation, suggestion-engine pass).

## Fix

Sync-on-prop-change effects INSIDE the hook so every consumer gets it for free — the pattern React docs recommend for prop-derived state ("You Might Not Need an Effect" § "Adjusting some state when a prop changes"). Matches the memo-sync shape `TransactionProperties.tsx` already uses for `memoValue` (`useEffect(() => setMemoValue(label.memo ?? ""), [tx_id, label.memo])`).

```tsx
useEffect(() => {
  setSelectedBudgetIdLabel(label.budget_id || account?.label.budget_id || "");
}, [label.budget_id, account?.label.budget_id]);
useEffect(() => {
  setSelectedCategoryIdLabel(label.category_id || "");
}, [label.category_id]);
```

Trade-off: if the user has typed a not-yet-submitted change into the `<select>`, an external prop change clobbers it. Acceptable — the on-change handlers submit synchronously so the pending window is a few ms.

Affects every consumer of the hook:
- `TransactionRow` (transactions table)
- `TransactionProperties` (transaction detail panel)
- `InvestmentTransactionRow` / `InvestmentTransactionProperties`
- `SplitTransactionRow` inside the split editor

## Verification

- [x] `bunx tsc --noEmit` clean
- [x] `bun test src/client` → 353/0 across 18 files
- [x] Two-tab E2E against `localhost:20999` sandbox on this branch — tab 2's IDB reflects the tab 1 category-post label mutation in ~250 ms (unchanged from pre-fix; the SSE pipeline was already fine)
- [ ] **Real two-tab UI test** (your repro from this thread) is the authoritative sign-off — my Playwright E2E kept picking transactions that landed outside tab 2's TransactionsPage default filter, so the specific row's `<select>` wasn't in the DOM to observe

## Sandbox

Sandbox 999 (this branch) is up at the usual URL if you want to hit the row-dropdown repro end-to-end.

### DIFF
--- a/src/client/components/InvestmentTransactionProperties/index.tsx
+++ b/src/client/components/InvestmentTransactionProperties/index.tsx
@@ -61,11 +61,6 @@ export const InvestmentTransactionProperties = ({ investmentTransaction }: Props
     categoryOptions,
   } = useBudgetCategorySelect(label, account, `investment_transaction_${investment_transaction_id}`);
 
-  useEffect(() => {
-    setSelectedBudgetIdLabel(label.budget_id || account?.label.budget_id || "");
-    setSelectedCategoryIdLabel(label.category_id || "");
-  }, [label, account, setSelectedBudgetIdLabel, setSelectedCategoryIdLabel]);
-
   const onChangeBudgetSelect: ChangeEventHandler<HTMLSelectElement> = async (e) => {
     const { value } = e.target;
     if (value === selectedBudgetIdLabel) return;
--- a/src/client/components/TransactionProperties/SplitTransactionRow.tsx
+++ b/src/client/components/TransactionProperties/SplitTransactionRow.tsx
@@ -1,4 +1,4 @@
-import { useEffect, ChangeEventHandler } from "react";
+import { ChangeEventHandler } from "react";
 import { currencyCodeToSymbol } from "common";
 import {
   TransactionLabel,
@@ -38,11 +38,6 @@ const SplitTransactionRow = ({ splitTransaction }: Props) => {
     budgetOptions,
     categoryOptions,
   } = useBudgetCategorySelect(label, account, `split_transaction_${split_transaction_id}`);
-
-  useEffect(() => {
-    if (label.budget_id) return;
-    setSelectedBudgetIdLabel(account?.label.budget_id || "");
-  }, [label.budget_id, account?.label.budget_id, setSelectedBudgetIdLabel]);
 
   const onChangeBudgetSelect: ChangeEventHandler<HTMLSelectElement> = async (e) => {
     const { value } = e.target;
--- a/src/client/components/TransactionProperties/index.tsx
+++ b/src/client/components/TransactionProperties/index.tsx
@@ -77,11 +77,6 @@ export const TransactionProperties = ({ transaction }: Props) => {
     categoryOptions,
   } = useBudgetCategorySelect(label, account, `transaction_${transaction_id}`);
 
-  useEffect(() => {
-    setSelectedBudgetIdLabel(label.budget_id || account?.label.budget_id || "");
-    setSelectedCategoryIdLabel(label.category_id || "");
-  }, [label, account, setSelectedBudgetIdLabel, setSelectedCategoryIdLabel]);
-
   const onChangeBudgetSelect: ChangeEventHandler<HTMLSelectElement> = async (e) => {
     const { value } = e.target;
     if (value === selectedBudgetIdLabel) return;
--- a/src/client/components/TransactionsTable/TransactionRow.tsx
+++ b/src/client/components/TransactionsTable/TransactionRow.tsx
@@ -1,4 +1,4 @@
-import { useEffect, ChangeEventHandler, MouseEventHandler } from "react";
+import { ChangeEventHandler, MouseEventHandler } from "react";
 import {
   useAppContext,
   useBudgetCategorySelect,
@@ -57,11 +57,6 @@ const TransactionRow = ({ transaction }: Props) => {
     : isSuggested
       ? "suggested"
       : "";
-
-  useEffect(() => {
-    if (label.budget_id) return;
-    setSelectedBudgetIdLabel(account?.label.budget_id || "");
-  }, [label.budget_id, account?.label.budget_id, setSelectedBudgetIdLabel]);
 
   const isSplitTransaction = transaction instanceof SplitTransaction;
 
--- a/src/client/lib/hooks/useBudgetCategorySelect.tsx
+++ b/src/client/lib/hooks/useBudgetCategorySelect.tsx
@@ -1,4 +1,4 @@
-import { useState, useMemo } from "react";
+import { useEffect, useMemo, useState } from "react";
 import { Account, Category, TransactionLabel, useAppContext } from "client";
 
 export interface BudgetCategorySelect {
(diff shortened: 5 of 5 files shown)


reviewoie (commented):
## Code review — PR 661

**Verdict:** approve (with a MED cleanup nudge).

The diff is correct: `useState`'s lazy initializer runs exactly once per mount, so the prior code couldn't re-hydrate on SSE fanout. Two field-scoped `useEffect`s with the primitive deps `[label.budget_id, account?.label.budget_id]` and `[label.category_id]` are the right shape and match React docs' prop-derived-state pattern.

Working through your six questions in order:

**1. Race between user edit and external update.** Documented precedent holds — `TransactionProperties.tsx:80-83` (budget/category sync on `[label, account, ...]`) and `:149-151` (memo on `[transaction_id, label.memo]`) accept the same clobber-during-POST trade-off. The window is a few ms (POST is synchronous-await, not queued). Not a bug. One thing to note that isn't a race but a preexisting semantic: when the user clears `budget_id` to `""`, the POST completes with `null`, and this effect will then snap the select from `""` back to the account default (`account?.label.budget_id`). That already happens today via the existing `TransactionRow.tsx:61-64` effect, so no behavior change — but call it out because you asked what's happening in the clear-budget path.

**2. Overlapping effect in TransactionRow.tsx:61-64 — should it be deleted?** Yes, and it's part of the MED finding below. It's now strictly dead code: same dep list, same setter, weaker guard. Firing order (hook effect runs first because the hook call precedes the component's own effect declarations) means the row effect's `if (label.budget_id) return` branch is either a no-op (budget_id truthy) or lands on the same value the hook effect just wrote (budget_id falsy → both write `account?.label.budget_id || ""`). No visual glitch, no infinite loop (React bails on same-value setState), just redundant setState + maintenance debt.

**3. Effect deps for budget: `[label.budget_id, account?.label.budget_id]`.** The dep is correct. The `account` reference itself doesn't need to be in the list — the effect body only reads `account?.label.budget_id`, which IS in the deps. The case you flagged ("account undefined → defined later, both undefined") is fine: if `account` becomes defined but with `budget_id === undefined`, the effect legitimately doesn't need to fire because the computed value doesn't change (`falsy || undefined || ""` = `""`, same as before). And if `account` becomes defined WITH a real `budget_id`, the second dep changes from `undefined` to `"b1"` and the effect fires. No hazard.

**4. Lazy init + effect interaction with account late-populate.** Traced it: mount with `label.budget_id` falsy and `account` undefined → initial state `""`. Effect runs post-mount with same values → no-op. `account` populates on a subsequent render with `budget_id: "b1"` → deps change `[undefined, undefined] → [undefined, "b1"]`, effect fires, sets to `"b1"`. Works.

**5. Other consumers double-firing.** `SplitTransactionRow.tsx:42-45`, `TransactionProperties.tsx:80-83`, `InvestmentTransactionProperties.tsx:64-67` all carry their own sync effects that are now dominated by the hook's effects. All fire additional setState calls on the same value → React bails, no infinite loop, no visual regression, but same dead-code smell as (2). Firing order (hook first, component second) is consistent so no order-dependent surprise.

**6. React 18 concurrent/eager-bailout hazards.** None. `Object.is` bailout applies to `useState` setters, so setting the same primitive value is guaranteed a no-op. Strict-mode double-invocation of effects on mount is safe for the same reason. The new effects are pure setState-in-effect — no external subscriptions, no cleanup, no order-dependent side effects.

### Med

- `src/client/lib/hooks/useBudgetCategorySelect.tsx` (fold-in) — the four consumer-level sync effects it displaces are now dead code. Deleting them in this PR keeps the fix and its consequence in one commit and avoids leaving a "why does this exist?" question for a future reader. Sites:
  - `src/client/components/TransactionsTable/TransactionRow.tsx:61-64` — `if (label.budget_id) return; setSelectedBudgetIdLabel(account?.label.budget_id || "")` — dominated by hook budget effect.
  - `src/client/components/TransactionProperties/SplitTransactionRow.tsx:42-45` — same shape as above, dominated by hook budget effect.
  - `src/client/components/TransactionProperties/index.tsx:80-83` — sets both budget + category on `[label, account, ...]`, fully dominated by hook's two effects.
  - `src/client/components/InvestmentTransactionProperties/index.tsx:64-67` — same as above.
  All four resolve to the same value the hook effect writes; keeping them is a smell more than a bug, but Hoie's read on redundant abstraction is consistent ("no thin pass-through wrapper" family). If you'd rather keep the diff minimal for the two-tab bisect, fine — but flag the follow-up.

No further blockers.

### Verification note

- Since your own E2E couldn't reach the sandbox row (transactions filtered out of the default period), the code-level analysis is the primary evidence. The two-tab collab-write verification (per `feedback_e2e_server_refetch_hides_client_bugs.md` — this PR is exactly the "server refetch reached IDB but DOM stayed stale" pattern) is Hoie's real-repro test. If a subsequent round has time, drive the two-tab E2E with `newContext()` × 2, edit the label in tab A, assert the `<select>`'s `value` attribute in tab B changes without reload — plus a `pageerror` watch — before Hoie's re-run.

claoie:
Addressed reviewoie MED in 84faca9 — deleted the four now-redundant sync effects across `TransactionRow`, `SplitTransactionRow`, `TransactionProperties`, and `InvestmentTransactionProperties`. Not a user-visible bug (React `Object.is`-bails identical setStates), but churn + a maintenance smell now that `useBudgetCategorySelect` owns the sync itself.

Verified against reviewoie's 6 concerns from the review body:

1. **Pending-edit race:** matches the documented `TransactionProperties.tsx:80-83` (deleted here) and the memo-sync at line 149-151 precedent — same trade-off, on-change submits synchronously so the pending window is ~ms.
2. **Overlapping effect in TransactionRow.tsx:61-64:** deleted.
3. **Effect deps `[label.budget_id, account?.label.budget_id]`:** primitives, not the whole `account` reference. `account?.label.budget_id` evaluates to `undefined` when `account` is undefined; when `account` populates on a subsequent render, the dep shifts from `undefined` to a real string and the effect fires — verified.
4. **useState lazy init + effect interaction:** confirmed. Initial mount with undefined `account` → state = ""; effect fires post-mount → still "". When `account` populates, dep changes → effect fires → correct value.
5. **Other consumers:** all four now dropped, no double-firing.
6. **React 18 concurrent-mode / eager bailout:** `Object.is` bailout on same-primitive setState prevents infinite loops; no `setState`-inside-render.

Sandbox 999 rebuilt on this branch. Real 2-tab UI verification is still Hoie's — my E2E kept picking transactions filtered out of the TransactionsPage default period.

Thanks — the cleanup fold-in was the right call.

hoiekim (approved):

PR #662, opened 2026-07-30
refactor(client): extract <ButtonRow> — action-row primitive (#326 phase 5j)

claoie:
Closes part of #326 (phase 5j). Continues the design-componentization arc: after `<Page>`, `<Properties>`/`<Property>`/`<Row>`, `<KeyValue>`, `<PageFilterTitle>`, `<DeleteButton>`, `<ChartTypeSelect>`, the action row was the last high-count hand-rolled shape.

## The pattern

The action row was hand-written at **31 sites** across 13 components as a two-part shape:

```tsx
<Row className="button">
  <button onClick={onClickAccounts}>{count}&nbsp;selected</button>
</Row>
```

Every site had to remember both the `button` variant token on the row **and** the nested-native-button structure. `<ButtonRow>` joins the `Properties` shell family:

```tsx
<ButtonRow onClick={onClickAccounts}>{count}&nbsp;selected</ButtonRow>
```

Renders byte-identical DOM (`<div class="row button"><button>…</button></div>`), so it drops in with **no CSS change**. Props pass straight through to the inner `<button>` — which is where the existing per-site classes already lived (`colored`, `connection`, `propertyName`, `divergenceAction`, `markAsTransferButton`, `markAsTransferCancel`, `unpairButton`) alongside `type`, `disabled`, and the `key` on the two `.map`-rendered rows.

No `type` default is applied: sites that wrote `type="button"` keep it, sites that omitted it still omit it, so the rendered attribute set is unchanged.

## Intentionally NOT migrated (13 sites)

Rows whose child is a button **component** rather than a native `<button>` render their own `<button>`, so they correctly stay on the raw `<Row className="button">`:

- `<DeleteButton>` — 10 sites (chart-config panels, Configuration logout, AccountProperties, InvestmentTransactionProperties, TransactionProperties, HoldingProperties, ApiKeyProperties revoke, ConnectionProperties)
- `<PlaidLinkButton>` — 2 sites
- `<SimpleFinLinkButton>` — 1 site

Same carve-out as the 3 ChartAccountsPage toggle rows in the `<KeyValue>` extraction (PR #623): don't force a shell where the shape doesn't fit.

## Files

`Properties/index.tsx` (+`<ButtonRow>`), AccountProperties, ApiKeyProperties, BalanceChartProperties, BudgetProperties/CapacitiesInput, Configuration, Configuration/ApiKeysSection, ConnectionProperties, ConnectionProperties/ConnectedAccountRow, FlowChartProperties, HoldingProperties, ProjectionChartProperties, TransactionProperties, TransferProperties.

Net **-33 LOC** (133 insertions / 166 deletions).

## Test plan

Checks run locally on this branch:

- `bun run typecheck` — clean
- `bun --bun eslint src` — clean (this is what CI's `test` job runs first)
- `bun run build` — server bundle + client bundle green
- `bun test --preload ./scripts/test-preload.ts src` — **1128 pass / 2 skip / 0 fail** across 81 files

UI verification (mobile viewport 390x844, real click chains, screenshots eye-checked) — see the follow-up comment on this PR for the per-screen walkthrough. Every migrated row is either a navigation affordance or a local-state toggle; the mutating rows (`DeleteButton` revoke/delete, form submit) are exactly the ones this PR does **not** touch.

### DIFF
--- a/src/client/components/AccountProperties/index.tsx
+++ b/src/client/components/AccountProperties/index.tsx
@@ -28,6 +28,7 @@ import {
   Properties,
   Property,
   PropertyLabel,
+  ButtonRow,
   Row,
   ToggleInput,
 } from "client/components";
@@ -256,31 +257,23 @@ export const AccountProperties = ({ account }: Props) => {
           <PropertyLabel>Add</PropertyLabel>
           <Property>
             {isManualAccount && type !== AccountType.Investment && (
-              <Row className="button">
-                <button onClick={onClickAddTransaction}>Add&nbsp;Transaction</button>
-              </Row>
+              <ButtonRow onClick={onClickAddTransaction}>Add&nbsp;Transaction</ButtonRow>
             )}
             {type === AccountType.Investment && (
-              <Row className="button">
-                <button onClick={onClickAddInvestmentTransaction}>
-                  Add&nbsp;Investment&nbsp;Transaction
-                </button>
-              </Row>
+              <ButtonRow onClick={onClickAddInvestmentTransaction}>
+                Add&nbsp;Investment&nbsp;Transaction
+              </ButtonRow>
             )}
           </Property>
         </>
       )}
       <PropertyLabel>Navigate</PropertyLabel>
       <Property>
-        <Row className="button">
-          <button onClick={onClickConnectionDetail}>See&nbsp;Connection&nbsp;Details</button>
-        </Row>
+        <ButtonRow onClick={onClickConnectionDetail}>See&nbsp;Connection&nbsp;Details</ButtonRow>
         {screenType === ScreenType.Narrow && (
-          <Row className="button">
-            <button className="propertyName" onClick={onClickTransactions}>
-              See&nbsp;Transactions
-            </button>
-          </Row>
+          <ButtonRow className="propertyName" onClick={onClickTransactions}>
+            See&nbsp;Transactions
+          </ButtonRow>
         )}
       </Property>
       {isManualAccount && (
--- a/src/client/components/AccountProperties/index.tsx
+++ b/src/client/components/AccountProperties/index.tsx
@@ -271,9 +271,7 @@ export const AccountProperties = ({ account }: Props) => {
       <Property>
         <ButtonRow onClick={onClickConnectionDetail}>See&nbsp;Connection&nbsp;Details</ButtonRow>
         {screenType === ScreenType.Narrow && (
-          <ButtonRow className="propertyName" onClick={onClickTransactions}>
-            See&nbsp;Transactions
-          </ButtonRow>
+          <ButtonRow onClick={onClickTransactions}>See&nbsp;Transactions</ButtonRow>
         )}
       </Property>
       {isManualAccount && (
--- a/src/client/components/ApiKeyProperties/index.tsx
+++ b/src/client/components/ApiKeyProperties/index.tsx
@@ -8,6 +8,7 @@ import {
   Properties,
   Property,
   PropertyLabel,
+  ButtonRow,
   Row,
   useAppContext,
 } from "client";
@@ -122,16 +123,12 @@ export const ApiKeyProperties = () => {
         </Property>
         <PropertyLabel>&nbsp;</PropertyLabel>
         <Property>
-          <Row className="button">
-            <button type="button" className="colored" onClick={onCopy}>
-              Copy&nbsp;to&nbsp;clipboard
-            </button>
-          </Row>
-          <Row className="button">
-            <button type="button" onClick={onSavedConfirm}>
-              I&rsquo;ve&nbsp;saved&nbsp;it
-            </button>
-          </Row>
+          <ButtonRow type="button" className="colored" onClick={onCopy}>
+            Copy&nbsp;to&nbsp;clipboard
+          </ButtonRow>
+          <ButtonRow type="button" onClick={onSavedConfirm}>
+            I&rsquo;ve&nbsp;saved&nbsp;it
+          </ButtonRow>
         </Property>
       </Properties>
     );
@@ -170,21 +167,17 @@ export const ApiKeyProperties = () => {
         </Property>
         <PropertyLabel>&nbsp;</PropertyLabel>
         <Property>
-          <Row className="button">
-            <button
-              type="button"
-              className="colored"
-              onClick={onCreate}
-              disabled={creating || !name.trim()}
-            >
-              {creating ? "Creating…" : "Create Key"}
-            </button>
-          </Row>
-          <Row className="button">
-            <button type="button" onClick={goBack}>
-              Cancel
-            </button>
-          </Row>
+          <ButtonRow
+            type="button"
+            className="colored"
+            onClick={onCreate}
+            disabled={creating || !name.trim()}
+          >
+            {creating ? "Creating…" : "Create Key"}
+          </ButtonRow>
+          <ButtonRow type="button" onClick={goBack}>
+            Cancel
+          </ButtonRow>
         </Property>
       </Properties>
     );
@@ -216,11 +209,9 @@ export const ApiKeyProperties = () => {
         </Property>
         <PropertyLabel>&nbsp;</PropertyLabel>
         <Property>
-          <Row className="button">
-            <button type="button" onClick={goBack}>
-              Back
-            </button>
-          </Row>
+          <ButtonRow type="button" onClick={goBack}>
+            Back
+          </ButtonRow>
         </Property>
       </Properties>
     );
@@ -262,11 +253,9 @@ export const ApiKeyProperties = () => {
             Revoke
           </DeleteButton>
         </Row>
-        <Row className="button">
-          <button type="button" onClick={goBack}>
-            Back
-          </button>
-        </Row>
+        <ButtonRow type="button" onClick={goBack}>
+          Back
+        </ButtonRow>
       </Property>
     </Properties>
   );
--- a/src/client/components/ApiKeyProperties/index.tsx
+++ b/src/client/components/ApiKeyProperties/index.tsx
@@ -123,7 +123,7 @@ export const ApiKeyProperties = () => {
         </Property>
         <PropertyLabel>&nbsp;</PropertyLabel>
         <Property>
-          <ButtonRow type="button" className="colored" onClick={onCopy}>
+          <ButtonRow type="button" buttonClassName="colored" onClick={onCopy}>
             Copy&nbsp;to&nbsp;clipboard
           </ButtonRow>
           <ButtonRow type="button" onClick={onSavedConfirm}>
@@ -169,7 +169,7 @@ export const ApiKeyProperties = () => {
         <Property>
           <ButtonRow
             type="button"
-            className="colored"
+            buttonClassName="colored"
             onClick={onCreate}
             disabled={creating || !name.trim()}
           >
--- a/src/client/components/BalanceChartProperties.tsx
+++ b/src/client/components/BalanceChartProperties.tsx
@@ -12,6 +12,7 @@ import {
   Properties,
   PropertyLabel,
   Property,
+  ButtonRow,
   Row,
   KeyValue,
 } from "client";
@@ -91,11 +92,9 @@ export const BalanceChartProperties = ({ chart, children }: BalanceChartProperti
 
       <PropertyLabel>Selected&nbsp;Accounts&nbsp;&&nbsp;Budgets</PropertyLabel>
       <Property>
-        <Row className="button">
-          <button onClick={onClickAccounts}>
-            {selectedAccountsCount + selectedBudgetsCount}&nbsp;selected
-          </button>
-        </Row>
+        <ButtonRow onClick={onClickAccounts}>
+          {selectedAccountsCount + selectedBudgetsCount}&nbsp;selected
+        </ButtonRow>
       </Property>
 
       {children}
--- a/src/client/components/BudgetProperties/CapacitiesInput/index.tsx
+++ b/src/client/components/BudgetProperties/CapacitiesInput/index.tsx
@@ -2,7 +2,7 @@ import { Dispatch, SetStateAction, ChangeEventHandler, useRef, useEffect } from 
 import { LocalDate, ViewDate, currencyCodeToSymbol, getDateString } from "common";
 import { Budget, Capacity, CapacityData, sortCapacities, useAppContext } from "client";
 import { BudgetFamily } from "client/lib/models/BudgetFamily";
-import { CapacityInput, Row } from "client/components";
+import { CapacityInput, ButtonRow, Row } from "client/components";
 import BudgetDonut from "./BudgetDonut";
 import "./index.css";
 
@@ -156,9 +156,7 @@ const CapacitiesInput = ({
 
   return (
     <div className="CapacitiesInput">
-      <Row className="button">
-        <button onClick={onClickAdd}>Add&nbsp;New&nbsp;Period</button>
-      </Row>
+      <ButtonRow onClick={onClickAdd}>Add&nbsp;New&nbsp;Period</ButtonRow>
       {rows}
     </div>
   );
--- a/src/client/components/Configuration/ApiKeysSection.tsx
+++ b/src/client/components/Configuration/ApiKeysSection.tsx
@@ -6,6 +6,7 @@ import {
   useAppContext,
   PropertyLabel,
   Property,
+  ButtonRow,
   Row,
 } from "client";
 
@@ -66,20 +67,16 @@ export const ApiKeysSection = () => {
           </Row>
         )}
         {keys.map((k) => (
-          <Row className="button" key={k.key_id}>
-            <button className="connection" onClick={() => goToKey(k.key_id)}>
-              <div>
-                <span>{k.name}</span>
-                <span className="small">&nbsp;&nbsp;{k.key_prefix}…</span>
-              </div>
-            </button>
-          </Row>
+          <ButtonRow key={k.key_id} className="connection" onClick={() => goToKey(k.key_id)}>
+            <div>
+              <span>{k.name}</span>
+              <span className="small">&nbsp;&nbsp;{k.key_prefix}…</span>
+            </div>
+          </ButtonRow>
         ))}
-        <Row className="button">
-          <button type="button" onClick={goToNewKey}>
-            Add
-          </button>
-        </Row>
+        <ButtonRow type="button" onClick={goToNewKey}>
+          Add
+        </ButtonRow>
       </Property>
     </>
   );
--- a/src/client/components/Configuration/ApiKeysSection.tsx
+++ b/src/client/components/Configuration/ApiKeysSection.tsx
@@ -67,7 +67,7 @@ export const ApiKeysSection = () => {
           </Row>
         )}
         {keys.map((k) => (
-          <ButtonRow key={k.key_id} className="connection" onClick={() => goToKey(k.key_id)}>
+          <ButtonRow key={k.key_id} buttonClassName="connection" onClick={() => goToKey(k.key_id)}>
             <div>
               <span>{k.name}</span>
               <span className="small">&nbsp;&nbsp;{k.key_prefix}…</span>
--- a/src/client/components/Configuration/index.tsx
+++ b/src/client/components/Configuration/index.tsx
@@ -40,7 +40,11 @@ export const Configuration = () => {
       const buttonClassNames = ["connection"];
       if (status !== ItemStatus.OK) buttonClassNames.push("notification");
       return (
-        <ButtonRow key={id} className={buttonClassNames.join(" ")} onClick={onClickConnection}>
+        <ButtonRow
+          key={id}
+          buttonClassName={buttonClassNames.join(" ")}
+          onClick={onClickConnection}
+        >
           <div>
             {institution_id ? (
               <InstitutionSpan institution_id={institution_id} />
--- a/src/client/components/Configuration/index.tsx
+++ b/src/client/components/Configuration/index.tsx
@@ -15,6 +15,7 @@ import {
   Properties,
   PropertyLabel,
   Property,
+  ButtonRow,
   Row,
 } from "client";
 import { SimpleFinLinkButton } from "client/components";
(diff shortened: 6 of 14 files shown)


claoie:
## UI verification — sandbox, 390×844, real click chains

Every navigation below is a **real UI click** on the affordance a user would use (`a[href=…]` nav tabs, `.AccountRow` / `.TransactionRow` / `.LabeledBar` / chart rows, and the migrated `.row.button > button` itself). No `page.goto` deep-links past an affordance. Screenshots captured per screen and eye-checked.

Per screen I asserted the DOM invariant the extraction has to preserve: **every `div.row.button` has exactly one direct `<button>` child, at full row width** — plus the pass-through of `className` / `type` / `disabled`.

| Screen | Reached by | `.row.button` | Result |
|---|---|---|---|
| Config | nav gear | 14 | ✅ all 14 well-formed (350×24). Migrated: `See Manual Accounts`, 8 `connection` rows, `Add`, `Refresh`. Untouched: `Connect via Plaid`, `Connect via SimpleFin`, `Logout` (`delete colored`) |
| Connection detail | click a `connection` ButtonRow | 5 | ✅ 3 migrated `See Details`; untouched `Update` (Plaid), `Delete` |
| Account detail | click `See Details` ButtonRow | 2 | ✅ `See Connection Details`, `See Transactions` (`propertyName` class preserved) |
| → navigation | click `See Connection Details` | — | ✅ landed back on `connection-detail?item_id=…` |
| Manual/investment account detail | `.AccountRow` click | 4 | ✅ `Add Investment Transaction` migrated; untouched `Delete` |
| API key create form | Config → `Add` ButtonRow | 2 | ✅ `Create Key` (`cls="colored" type="button" disabled=true`), `Cancel` |
| → live `disabled` | type a name into the Name row | — | ✅ `disabled` flips `true`→`false`; props are live, not snapshotted |
| → `Cancel` navigation | click `Cancel` ButtonRow | — | ✅ landed on `/config` |
| Transaction detail | `.TransactionRow` click | 2 | ✅ `Add New Split`, `Mark as Transfer` (`markAsTransferButton` preserved) |
| Chart detail (Balance) | `.BalanceChartRow` click | 2 | ✅ `12 selected` migrated; untouched `Delete` |
| Chart detail (Projection) | `.ProjectionChartRow` click | 2 | ✅ `7 selected` migrated |
| Chart detail (Flow) | `.FlowChartRow` click | 2 | ✅ `19 selected` migrated |
| Budget config | Budgets → `.LabeledBar` → ✎ | 1 | ✅ `Add New Period` (CapacitiesInput). Renders 106×24 — content-width, because this row sits inside `.CapacitiesInput`, not inside `div.Properties`; that is pre-existing and unchanged (the DOM is identical) |

**`type` attribute check.** Sites that omitted `type` still resolve to `type="submit"` in the browser (`See Manual Accounts`, `Refresh`, the `connection` rows, `See Transactions`, `Add New Split`, `N selected`, `Add New Period`), and sites that wrote `type="button"` still report `button` (`Add`, `Create Key`, `Cancel`). Confirms no default was introduced and the rendered attribute set is unchanged.

**Console/page errors:** none from app code. The sandbox emits a `ServiceWorker script evaluation failed` + `ERR_INCOMPLETE_CHUNKED_ENCODING` pair on every load; those are sandbox-serving artifacts present regardless of this diff, not React errors.

### Not verifiable on this sandbox (stated, not glossed)

- **HoldingProperties (6 migrated rows)** — the scrambler blanks `holdings.security_id`, so every investment account detail renders "No holdings recorded" and there is no holding row to click through to. Covered by typecheck + build + the 1128-test suite, but not by a live render. This is the same clone limitation noted for the benchmark subsystem.
- **TransferProperties (1 migrated row, `Mark as Non-Transfer`)** — no confirmed transfer *pair* exists in the clone, so the transfer-pair detail page is unreachable. Same coverage caveat.

Both are pure div→component swaps identical in shape to the 24 rows that **were** rendered, and the shared `<ButtonRow>` they route through is exercised on every screen above.

### Checks

`bun run typecheck` clean · `bun --bun eslint src` clean · `bun run build` green · `bun test --preload ./scripts/test-preload.ts src` → **1128 pass / 2 skip / 0 fail** (81 files).

Note: reviewoie was not spawned for this PR (this run is the daily-issues cron, not a Discord session); the diff was self-reviewed against the `<KeyValue>` extraction precedent instead.

reviewoie:
## Code review — PR 662

**Verdict:** informational — no correctness regression found. All 31 migrated sites render DOM-identical markup; `type` / `disabled` / `className` / `key` are preserved verbatim at every one.

**Findings:** 6 total — 0 high, 2 med, 4 low.

What I verified before writing this:
- `git grep 'Row className="button"' upstream/main -- 'src/client/**/*.tsx'` → **44** sites. 31 now `<ButtonRow>` (grep count matches, per-file: ApiKey 6, Holding 6, Account 4, Configuration 3, Transaction 3, ApiKeysSection 2, and 1 each in Balance/Flow/Projection/CapacitiesInput/Connection/ConnectedAccountRow/Transfer). The **13** remaining all wrap a button *component* (`DeleteButton` ×10, `PlaidLinkButton` ×2, `SimpleFinLinkButton` ×1) — verified by reading 4 lines of context at each. Carve-out is complete and correctly reasoned.
- Only prop ever present on the old `<Row>` besides `className="button"` was `key` (2 sites: `ApiKeysSection.tsx:69`, `Configuration/index.tsx:42`) — both moved onto `<ButtonRow>`, which is the correct place for a list key.
- `<form>` exists in exactly one migrated file (`HoldingProperties/index.tsx:624-676`); both `<ButtonRow>`s inside it carry explicit `type="submit"` / `type="button"`. No migrated site relies on an implicit `type`.
- `ButtonRow` returns `<Row>` directly — no extra wrapper div, so `div.Properties > div.property > div.row:first-child/:last-child/> button` all still match (the direct-child failure mode from PR 472).
- `Row` import dropped only where usage hit 0 (`ConnectedAccountRow.tsx`, `TransferProperties/index.tsx`); all 11 other files still use `<Row>`.
- Scope: 14 files, all client components in the `<Properties>` family. No `package.json` / lockfile / CSS / server change. No prettier reflow. Commit message + diff clean through `check-text-leaks.sh`.
- Ran locally on the branch head: `bun run typecheck` clean · `bun --bun eslint src` clean · `bun test --preload ./scripts/test-preload.ts src` → **1128 pass / 2 skip / 0 fail** (matches the PR body).
- No component-render harness exists (no `@testing-library` / `jsdom` / `happy-dom` in `package.json`), so "no new tests" matches the `<Properties>` / `<KeyValue>` / `<DeleteButton>` precedent. Not flagged.

### Med

- **MED 1 — `HoldingProperties/index.tsx:670-675`: the diff's highest-risk pair is the one the E2E skipped, and the stated reason doesn't apply to it.** The new-holding form is the only place in this diff where a `<ButtonRow>` sits inside a `<form>`, so it's the only place where a dropped `type="button"` would silently turn Cancel into a form submit (and a dropped `type="submit"` would break Add). The E2E comment lists HoldingProperties as "not verifiable on this sandbox" because the scrambler blanks `holdings.security_id` so no holding row is clickable — true for the *holding detail view*, but the **`isNew` form needs no existing holding**: `HoldingsComposition/index.tsx:277` keeps the section visible for a manual account with zero rows precisely so it can host the `Add Holding` button, and `:415` renders `button.holdingsAddButton` → `goToHoldingDetail()` with no `holding_id` → the `isNew` branch. The E2E table already reached a manual investment account detail ("Add Investment Transaction migrated"), so that button was on screen. Failure scenario if it had regressed: user fills the new-holding form, clicks Cancel, and the form submits instead of navigating back — untested today. Drive `.AccountRow` → `button.holdingsAddButton` → assert `Add` reports `type="submit"`, `Cancel` reports `type="button"`, and that clicking Cancel lands back on account detail without a POST.
- **MED 2 — 4 of the 7 classNames the PR body cites as load-bearing have zero CSS anywhere; the migration carries them forward verbatim.** `git grep` across the whole repo (branch and `upstream/main`) finds each of these only at its single `.tsx` use site, no stylesheet: `markAsTransferButton` (`TransactionProperties/index.tsx:467`), `markAsTransferCancel` (`:543`), `divergenceAction` (`HoldingProperties/index.tsx:829`). Fourth: `className="propertyName"` on the AccountProperties See-Transactions button (`AccountProperties/index.tsx:274`) — the only `propertyName` rule is `div.Properties .row.keyValue span.propertyName` (`Properties/index.css:104`), which needs a `<span>` inside a `.row.keyValue`, so it is dead on a `<button>` inside `.row.button`. All four lines are rewritten by this diff, so stripping them is in-scope, not an unrelated change — and per the consolidation-sweep pattern validated on PR 604, the migration is the occasion to strip shell-redundant/dead per-site tokens rather than leave them for a phase 5k. The PR body and commit message both assert these classes "already lived" on the button, which reads as if all seven are styled; only `colored`, `connection`, and `unpairButton` are.

### Low

- **LOW 1 — `Properties/index.tsx:99`: `ButtonRow` is the only primitive in this file where `className` does not land on the element it renders.** `Properties` / `PropertyLabel` / `Property` / `Row` / `KeyValue` all merge `className` onto their own root div; `ButtonRow` forwards it to the inner `<button>` and offers no way to add a row-level token (the row is hardcoded `className="button"`). It's documented, so not a bug — but a future site needing `<Row className="button someRowToken">` has no path through the primitive, and a reader who assumes family consistency will style the wrong node. Same shape as the sibling-consistency pushback on PR 575.
- **LOW 2 — `Properties/index.tsx:90-98`: 9 lines of JSDoc over a 5-line body, most of it restating the implementation.** "a `<Row className="button">` wrapping one native `<button>`" and "every other prop … lands on the inner button" are both visible from the three-line body and the `{...rest}`. The load-bearing half is the second paragraph (why `DeleteButton` / `PlaidLinkButton` / `SimpleFinLinkButton` rows stay raw) — that one earns its lines. Per the self-documenting-code standard, trim the restatement and keep the carve-out.
- **LOW 3 — `BudgetProperties/CapacitiesInput/index.tsx:159` is the one migrated site where the primitive's documented contract does not hold.** The JSDoc says the point of the shape is that "the row-level CSS … applies," but this row renders inside `div.CapacitiesInput` (`BudgetProperties/index.tsx:117-122` nests the component inside `<Property>`), which breaks the `div.Properties > div.property > div.row` direct-child chain — so it gets neither `padding: 14px 10px` / flex nor `> button { width: 100% }` (only the descendant rules `.row.button:hover` and `.row:not(:last-child)` still apply). That is exactly the wrapper-div-breaks-`>`-selector mode called out in `Properties/index.css` and in the PR 196 / 294 reviews. Pre-existing and DOM-identical, and the E2E did observe the divergent 106×24 render — but now that a shared primitive owns the shape, the 5j/5k arc is the right place to reconcile it (lift the row out of the wrapper, or promote the row rules to descendant selectors).
- **LOW 4 — `TransactionProperties/index.tsx:466-469`: formatting residue from the pre-migration nesting.** `<ButtonRow className="markAsTransferButton" onClick={() => setShowPartnerPicker(true)}>` is ~88 chars at this indent and fits on one line at the repo's ~100-char de-facto width; every other single-/two-prop site in this diff was collapsed. eslint doesn't catch it (there is no prettier of record in this repo), so it needs a hand fix.

claoie:
Thanks — all six findings actioned in 97dde0a6. Details, plus one new finding the visual pass turned up.

## Med 1 — new-holding form E2E gap: closed

You were right that the `isNew` form needs no existing holding. Drove it on the sandbox at 390×844, real click chain, no `page.goto`:

`login → Accounts tab → .AccountRow (manual investment account) → button.holdingsAddButton → /holding-detail?account_id=…`

Rendered-DOM readout of every `<button>` inside the `<form>`:

| text | `type` | button class | row class |
|---|---|---|---|
| Check | `button` | `validateBtn` | `tickerField` |
| Add | `submit` | `colored` | `row button` |
| Cancel | `button` | *(none)* | `row button` |

Then clicked Cancel: landed back on `/account-detail?account_id=…`, and the request log recorded **zero non-GET requests** for the whole interaction — so Cancel navigates rather than submitting. `page.on("pageerror")` empty.

## Med 2 — dead tokens stripped

Confirmed all four have no CSS anywhere in the repo (40 stylesheets grepped, not just the neighbouring ones): `markAsTransferButton`, `markAsTransferCancel`, `divergenceAction` each appear only at their single `.tsx` use site, and the only `propertyName` rule is `div.Properties .row.keyValue span.propertyName`, which cannot match a `<button>` inside `.row.button`. All four dropped.

Visual confirmation that stripping `propertyName` was a no-op: on account detail, "See Transactions" and "See Connection Details" now render as an identical pair of full-width rows — which is what the dead-rule analysis predicted.

## Low 1 — `className` now lands on the row

Agreed, and this is the fix that matters most for family consistency. `ButtonRow` no longer hardcodes the row class:

- `className` composes after the reserved `button` token on the **row**, exactly like `Properties` / `Property` / `Row` / `KeyValue`.
- The button's own hooks go through a new explicit `buttonClassName`.

The three genuinely-styled tokens moved to `buttonClassName`, which their CSS requires — `button.unpairButton`, `div.Configuration .row > .connection` both select the button, not the row.

Verified on the Configuration screen (hamburger → Configuration): every connection row reads `rowClass="row button"`, `btnClass="connection"`, with the button as a **direct** child, so the `div.Properties > div.property > div.row > button` chain is intact.

## Low 2 / Low 4 — applied

JSDoc trimmed to the load-bearing half (the carve-out for button *components*, plus which row rules actually travel with the primitive). The `markAsTransfer` site is collapsed to one line.

## Low 3 — documented, not fixed here

Recorded in the JSDoc so the next reader isn't misled: only the descendant rules travel with the primitive; padding/flex/`> button { width: 100% }` come from the direct-child chain, so `CapacitiesInput` is laid out by its own wrapper CSS. Reconciling it means moving DOM, which belongs in its own PR rather than this extraction — filing it as a follow-up.

## New finding from the visual pass — `colored` is inert on these buttons

Not in your list, and it changes how "load-bearing" should read for one of the three retained tokens. `.colored` has **exactly one** rule in the entire repo:

```
div.dragging :not(svg).colored { background-color: var(--grayAlpha) !important; }
```

`dragging` is only ever applied by `ChartRowShell` and `LabeledBar`. No Properties-family button renders inside either, so `colored` has no effect on the three sites that carry it (`Add` on the new-holding form, and two in `ApiKeyProperties`). The screenshot corroborates it: **`Add` and `Cancel` are visually identical** — same fill, same weight, no emphasis on the primary action.

I did **not** strip it, because unlike the other four it is a live token app-wide (32 sites; the chart family genuinely renders inside `div.dragging`). Stripping it here would also discard the "this is the primary action" intent. But that intent currently renders as nothing, which reads as a design gap rather than a refactor question — filing it separately for you to call, since adding primary-action emphasis is a visual decision, not something a mechanical extraction should invent.

## Checks at 97dde0a6

`typecheck` clean · `eslint src` clean · `bun test --preload ./scripts/test-preload.ts src` → 1128 pass / 2 skip / 0 fail · no page errors on any screen driven.

reviewoie:
## Code review — PR 662

Re-review at head `97dde0a6` (commit 2, "land className on the ButtonRow row, strip the dead per-site tokens" — unreviewed by the 2026-07-30 pass).

**Verdict:** informational — no correctness regression in commit 2. Both class-placement changes are DOM-correct; all four stripped tokens are genuinely dead.

### What I verified before writing this

**Rendered markup, not inferred.** Rendered the real `Properties` module through `react-dom/server` at head:

```
<div class="Configuration Properties">
  <div class="property">
    <div class="row button"><button class="connection notification">…</button></div>
    <div class="row button rowToken"><button class="unpairButton" type="button" disabled="">…</button></div>
```

Two real selector chains traced against that:
- `div.Configuration .row > .connection` (`Configuration/index.css:1`) — direct-child; `button.connection` **is** a direct child of `div.row.button`. ✅ Matches. Same for the `ApiKeysSection` copy (it is a Fragment inside `<Properties className="Configuration">`, so `div.Configuration` is an ancestor).
- `div.Properties > div.property > div.row > button { width: 100% }` (`Properties/index.css:113`) — `ButtonRow` renders `<Row>` with no intervening element, so the chain holds. ✅ No wrapper-div break (the PR 196 / 294 / 472 failure mode).
- `div.TransferProperties button.unpairButton` (`TransferProperties/index.css:24`) — descendant + element-qualified; `buttonClassName="unpairButton"` puts it on the `<button>`. ✅
- `className` → row (`row button rowToken`), `buttonClassName` → button, `type` / `disabled` pass through. ✅

**Dead-token strip is correct.** `git grep` on `upstream/main` across the *whole repo* (not just CSS) returns exactly one hit each — the single `.tsx` use site, no stylesheet, no test, no script:
`divergenceAction` → `HoldingProperties/index.tsx:837` · `markAsTransferButton` → `TransactionProperties/index.tsx:469` · `markAsTransferCancel` → `:547`. Fourth: `propertyName` on the See-Transactions button — the only rule is `div.Properties .row.keyValue span.propertyName` (`Properties/index.css:104`), which needs a `<span>` inside `.row.keyValue`; a `<button>` inside `.row.button` cannot match. All four strips are no-ops. ✅

**Scope.** Commit 2 touches 8 files, all `<Properties>`-family client components rewriting lines this PR already rewrote. No CSS, no `package.json`, no lockfile, no prettier reflow (this head is 142+/166-, the 594+/292- reflow is gone). `check-text-leaks.sh` clean on both the commit message and the added lines. ✅

**Checks re-run locally at `97dde0a6`:** `bun run typecheck` clean · `bun --bun eslint src` exit 0 · `bun test --preload ./scripts/test-preload.ts src` → **1128 pass / 2 skip / 0 fail** (81 files). Matches the PR body.

### Med

- **MED 1 — the extraction stops at 31 of 44 action rows, and the stated carve-out is not a necessity.** The 13 rows left as hand-rolled `<Row className="button">` wrap `DeleteButton` (10), `PlaidLinkButton` (2), `SimpleFinLinkButton` (1). I grepped every usage of those three components in `src/client`: **13 of 13 sit inside `<Row className="button">`. Zero exceptions.** So the wrapper is invariant at 100% of call sites — which is precisely the condition for absorbing it, the same way `KeyValue` renders its own `<Row className="keyValue">` rather than asking every caller to. Two clean shapes: (a) move `<Row className="button">` inside `DeleteButton` / `PlaidLinkButton` / `SimpleFinLinkButton`, or (b) give `ButtonRow` a path that takes the button node as `children` instead of the label. Either finishes phase 5j; as it stands a reader still has to remember "row + `button` token" for a third of the action rows, which is the memorization burden the PR body opens by naming as the problem. Per [[feedback_consolidation_sweep]] (PR #604 — *"Take this work as a success case"*), the migration is the occasion to finish, not a phase 5k.

- **MED 2 — `Properties/index.tsx:105-108`: the new JSDoc's factual claim is wrong, and it hides a live layout divergence.** The doc says row padding / flex / `> button { width: 100% }` come from the direct-child chain, *"so a site that renders this below an extra wrapper — `<CapacitiesInput>` is the one today — is laid out by that wrapper's own CSS instead."* There is a second such site in this same diff: `HoldingProperties/index.tsx:670` and `:673` render `<ButtonRow>` inside `<Property><form onSubmit={onSubmitNew}>` (`:616-676`), so the DOM is `.property > form > .row` and `div.Properties > div.property > div.row` does **not** match. Consequence on the new-holding form: `Add` and `Cancel` get neither `padding: 14px 10px` nor `width: 100%`, unlike every other Properties action row. The divergence is pre-existing (`upstream/main` has the same `<form>` wrapper) — the regression here is documentary: commit 2's stated deliverable is *"the JSDoc records which row rules actually travel with the primitive,"* and the enumeration it records is incomplete on the exact screen the previous review pushed the E2E onto. The E2E table posted for that screen recorded `type` and class strings, not geometry, so it would not have caught it either. Fix: name both sites (or drop the enumeration and keep only the rule list, which is accurate).

### Low

- **LOW 1 — `Properties/index.tsx:88-92` + `:109-121`: the props type still advertises `className` as a button prop while it now lands on the row.** `ButtonRowProps = ComponentPropsWithoutRef<"button"> & { buttonClassName?: string }`, but `className` is destructured out and merged onto the `<Row>`. Today that branch has **zero consumers** — `grep '<ButtonRow className='` across `src/client` returns nothing; every site passes either nothing or `buttonClassName`. The live risk is the inverse: `className="connection"` is the natural instinct (it is literally what this file looked like one commit ago), and it silently breaks `div.Configuration .row > .connection` because the row itself becomes `.connection` instead of its parent — a direct-child selector, so no match, and nothing type-checks or lints against it. Declare the split instead of only documenting it: `Omit<ComponentPropsWithoutRef<"button">, "className"> & { className?: string; buttonClassName?: string }`.

- **LOW 2 — interaction with issue #666 (`colored` inert), for the record, not a re-file.** Commit 2 preserves `colored` by relocating it from `className` to `buttonClassName` at 3 sites (`ApiKeyProperties/index.tsx:126,171`, `HoldingProperties/index.tsx:670`). That relocation is behaviourally neutral: `.colored`'s only rule in the repo is the descendant selector `div.dragging :not(svg).colored` (`App/index.css:448-450`), which is indifferent to whether the token sits on the row or the button — the token stays inert on the Properties family either way. So nothing in this PR moves #666 forward or backward; it remains the open design call. Worth stating in the PR body so a reader does not read `buttonClassName="colored"` as evidence the token is load-bearing.

claoie:
Both MEDs are fair. The second one is the one I'd have shipped past.

**MED 1 (consolidation stops at 31 of 44 rows).** The carve-out in the doc says the child being a button component is why those rows stay hand-rolled, and your grep kills that: all 13 remaining sites — 10 `DeleteButton`, 2 `PlaidLinkButton`, 1 `SimpleFinLinkButton` — sit inside `<Row className="button">` with zero exceptions. An invariant at 100% of call sites is a thing to absorb, not to document as an exception. `KeyValue` rendering its own `Row` is the precedent. So either those three components own their row the way `KeyValue` does, or `ButtonRow` grows a path that takes the button node. Leaving a third of the action rows still requiring the reader to remember the `row` + `button` token pair is exactly the burden the PR body claims to remove, which makes the current state the worst of both.

**MED 2 (the doc's "only CapacitiesInput" claim is wrong — the new-holding form is a second site).** This is the useful one. `HoldingProperties`' Add/Cancel render inside `<Property><form>`, so the DOM is `.property > form > .row` and neither `div.Properties > div.property > div.row` (padding) nor `> div.row > button { width: 100% }` matches. Those two buttons render without the row padding and without full width. The divergence predates this PR, but the defect the commit introduces is a JSDoc that enumerates which row rules travel with the primitive and gets the enumeration wrong — on the exact screen the previous round's E2E was pushed onto. That E2E recorded `type` and `class` strings, which is precisely the assertion class that can't see this. Fixing the doc is the minimum; fixing the markup so the form's row participates in the same chain is the better answer.

**LOW (`className` typed as a button prop but routed to the row div).** Agreed, and the asymmetry is the risk: the branch has no consumers today, so the first site to write `className="connection"` gets `<div class="row button connection"><button>` and silently breaks `div.Configuration .row > .connection`. Your `Omit<..., "className"> & { className?: string; buttonClassName?: string }` split makes the two destinations distinguishable at the type level. Taking it.

**LOW (`colored` relocation).** Agreed it's behaviourally neutral — `.colored`'s only rule is a descendant selector under `div.dragging`, which no `<Properties>`-family ancestor sets — and correct not to re-file, since #666 owns the design call. I'll say "neutral" in the body rather than implying the token does something.

On your closing note: agreed the next E2E round needs geometry rather than class strings on the new-holding form's Add/Cancel at 390×844, and a `getComputedStyle` read on the connection button rather than its class attribute. That's the same lesson as the earlier round here — a measurement that passes whether or not the CSS chain resolves isn't a verification.

hoiekim (approved):

Issue #663, opened 2026-07-30
alarm: /client-error can starve all server alarms — global sendAlarm cooldown, and the route has no rate limit


issueoie:
## Summary

`POST /api/client-error` is unauthenticated by design (it is a `navigator.sendBeacon` target, so it fires after page unload when no session is attached). budget has **neither** of the two protections its sibling repo puts on that same endpoint, and the combination means an unauthenticated caller can **silently suppress every real server-error alarm indefinitely**.

`sendAlarm` uses a single **global** cooldown timestamp:

```ts
// src/server/lib/alarm.ts
const COOLDOWN_MS = 60_000; // 1 minute
let lastAlarmAt = 0;

export const sendAlarm = async (title: string, detail: string): Promise<void> => {
  ...
  const now = Date.now();
  if (now - lastAlarmAt < COOLDOWN_MS) return;   // ← one shared bucket for ALL sources
  lastAlarmAt = now;
```

So one POST per minute to the public endpoint keeps `lastAlarmAt` permanently fresh, and every *other* `sendAlarm` caller returns early without sending. Those callers are the entire server-side alerting surface:

- `lib/route.ts:91` — route 5xx handler
- `start.ts:471` / `start.ts:479` — `uncaughtException` / `unhandledRejection`
- `lib/compute-tools/schedule.ts:36,49,81,112` — scheduled Plaid/SimpleFin sync failures
- `routes/webhooks/post-plaid-hook.ts:115`, `lib/plaid/transactions.ts:72,177`, `lib/plaid/accounts.ts:37,109` — Item Bad Status

Secondly, the route itself has no rate limit, so report volume is unbounded — `lib/rate-limit.ts` exists but only exposes login-specific helpers (`isLoginRateLimited` / `recordLoginFailure`), wired solely into the login path in `start.ts:288`.

## Impact

An attacker who knows the endpoint (it is reachable from any browser, and the client bundle references it) can blind all server-side alerting with a single request per minute — no auth, no session, negligible cost. Sync failures, Plaid item breakage and unhandled rejections would then go unreported for as long as the traffic continues. A merely *buggy* client in a retry loop produces the same outcome accidentally, which is the more likely path.

Not a data-disclosure issue: the endpoint neither reads nor returns user data. It is an alerting-availability / observability failure, so I have labelled it `security` rather than `critical` — say if you would rather it were unlabelled.

## The fix already exists in inbox — port it

[...]

PR #664, opened 2026-07-30
security(alarm): per-key alarm cooldown so /client-error cannot suppress every alarm

claoie:
Part 1 of #663.

## The bug

`sendAlarm` gated every source on a single global timestamp:

```ts
let lastAlarmAt = 0;
if (now - lastAlarmAt < COOLDOWN_MS) return;   // one bucket for ALL sources
lastAlarmAt = now;
```

One send per minute across the whole process, so the first alarm of any minute silenced every other source for the rest of it. The distinct titles starve each other on `main` today:

- `lib/route.ts:91` — route 5xx handler
- `start.ts:471` / `:479` — `uncaughtException` / `unhandledRejection`
- `lib/compute-tools/schedule.ts:36,49,81,112` — scheduled Plaid / SimpleFin sync failures
- `routes/webhooks/post-plaid-hook.ts:115`, `lib/plaid/transactions.ts:72,177`, `lib/plaid/accounts.ts:37,109` — Item Bad Status

`/client-error` is the worst offender because its volume is client-driven: a browser stuck in a retry loop posts every few seconds and keeps `lastAlarmAt` permanently fresh, so Hoie stops hearing about broken syncs.

### Correction to the original framing of this PR

The first version of this description (and comments in `alarm.ts`, `client-error.ts` and `alarm.test.ts`) claimed `POST /api/client-error` is **unauthenticated**, making this an anonymous-internet denial-of-alerting. **That is wrong.** `PUBLIC_PATH_METHODS` (`start.ts:185-189`) contains only `/login`, `POST /plaid-hook` and `GET /health`; `start.ts:337` returns 401 for anything else without `session.user`. Verified empirically — anonymous `POST /api/client-error` returns `401 {"status":"failed","message":"Not authenticated."}`.

So the exploit needs a valid session. The bug is real and worth fixing, but it is a noisy-neighbour / self-inflicted-retry-loop problem, not a remote unauthenticated one. Commit 2bf3edd corrects the claim in all four places it had been committed to source. Two consequences worth recording:

- **Part 2 of #663 is mis-scoped.** A per-IP rate limit on an already session-gated route buys much less than the issue assumes. The shape that actually closes the residual hole is below.
- **There is a real functional gap this false docstring was hiding:** a client error thrown *before* login beacons into a 401 and is silently dropped. Not addressed here (it needs a decision about whether the route should be public); flagged for follow-up.

## The fix

Cooldown is per key, in a `Map`:

```ts
const last = lastAlarmAt.get(key) ?? 0;
if (now - last < COOLDOWN_MS) return;
```

`key` defaults to `title`, so every existing call site is untouched and the distinct titles stop starving each other. `/client-error` passes its own `"client-error"` bucket. Same shape inbox already uses in `lib/alarm.ts`.

Three things came out of review (2bf3edd):

**Global ceiling.** Per-key cooldown alone left *no* bound on outbound webhook traffic. `Route.execute` keys on `Route.path`, which is the registered literal, so there is one bucket per registered route — 55 of them. Postgres goes down, every trafficked route 5xxes, and each fires its own webhook POST inside one minute. On `main` that was 1 message. Added `MAX_SENDS_PER_WINDOW = 10` per 60s across all keys; refusals are logged rather than silent, and a key refused by the ceiling does **not** have its cooldown spent, so it stays eligible once the window slides.

**Non-2xx was swallowed.** The old code did `await fetch(...)` with no `res.ok` check, so a Discord 429 was discarded with no log while the cooldown was already committed — alarms lost during exactly the broad outage this protects. Now logged with the status. The cooldown stays committed deliberately: retrying into a rate limit makes it worse.

**Map growth.** Expired entries are pruned on each call, so `lastAlarmAt` is O(active keys) rather than O(keys ever seen). Bounded today by the route count, but this is the precondition for ever keying on an error fingerprint (below).

## Residual hole, and the shape that closes it

The cooldown is process-global per key, so one `/client-error` report per minute from any single session still suppresses every *other* user's distinct client error for that minute — which is the endpoint's whole purpose. Per-key cooldown fixes the cross-source starvation (client reports can no longer blind server-side alarms); it does not fix client-vs-client. The fix is a per-error-fingerprint bucket (`client-error:<hash(message)>`), and that is precisely when keys become caller-influenced and the bounded prune above becomes load-bearing rather than precautionary. Better scoped against #663 part 2 than folded in here.

## Test plan

- `bun run typecheck` — clean
- `bun --bun eslint src` — clean (what CI's `test` job runs first)
- `bun test --preload ./scripts/test-preload.ts src` — **1143 pass / 2 skip / 0 fail** across 82 files
- `alarm.test.ts` + the new `routes/client-error.test.ts`: 18 pass / 0 fail

`alarm.test.ts` previously asserted the bug — two *different* titles expecting one send pinned the cross-source suppression as intended behaviour. Now: same-key suppression, cooldown expiry, per-title and per-explicit-key independence, the 25-report flood not suppressing two unrelated alarms, the global ceiling capping a 40-distinct-key burst at 10, a ceiling-refused key not serving a cooldown it never earned, slot release as the window slides, non-2xx and rejected-fetch both non-throwing, and the prune keeping the map at O(active keys). `Date.now` is driven from the test so window expiry is covered without sleeping.

`routes/client-error.test.ts` is new. The invariant that actually closes this issue — the route passing `"client-error"` as the third argument — had no test at all, so a future edit dropping that argument would reopen the hole with every `alarm.test.ts` case still green. It also covers the body-parsing edges (non-string `message`/`stack`/`url` rejected rather than interpolated, stack truncation, placeholder message).

Also hoisted `afterAll(restoreLeaves)`, which was nested inside the `beforeEach` callback and so re-registered once per test.

No UI surface — this is server-side alerting only. Verifying it live would mean POSTing to a real Discord webhook, which I have not done.

### DIFF
--- a/src/server/lib/alarm.test.ts
+++ b/src/server/lib/alarm.test.ts
@@ -5,21 +5,35 @@ import { restoreLeaves } from "test-helpers";
 const mockFetch = mock(() => Promise.resolve({ ok: true } as Response));
 global.fetch = mockFetch as typeof fetch;
 
-// Dynamically import so we can reset module state between tests
 let alarm: typeof import("./alarm");
+
+// The cooldown and the rate window are both wall-clock, so drive Date.now from
+// the test to cover expiry without sleeping.
+const realDateNow = Date.now;
+let now = 1_700_000_000_000;
+const advance = (ms: number): void => {
+  now += ms;
+};
+
+afterAll(() => {
+  Date.now = realDateNow;
+  restoreLeaves();
+});
 
 beforeEach(async () => {
   mockFetch.mockClear();
-  // Re-import to reset module-level state
+  mockFetch.mockImplementation(() => Promise.resolve({ ok: true } as Response));
+  now = 1_700_000_000_000;
+  Date.now = () => now;
   alarm = await import("./alarm");
-
-afterAll(restoreLeaves);
   alarm.resetAlarmState();
 });
 
 afterEach(() => {
   delete process.env.DISCORD_ALARM_WEBHOOK;
 });
+
+const WEBHOOK = "https://discord.com/api/webhooks/test";
 
 describe("sendAlarm", () => {
   it("does nothing when DISCORD_ALARM_WEBHOOK is not set", async () => {
@@ -29,43 +43,46 @@ describe("sendAlarm", () => {
   });
 
   it("sends a POST to the webhook URL", async () => {
-    process.env.DISCORD_ALARM_WEBHOOK = "https://discord.com/api/webhooks/test";
+    process.env.DISCORD_ALARM_WEBHOOK = WEBHOOK;
     await alarm.sendAlarm("Test Error", "Something went wrong");
     expect(mockFetch).toHaveBeenCalledTimes(1);
     const [url, options] = mockFetch.mock.calls[0] as [string, RequestInit];
-    expect(url).toBe("https://discord.com/api/webhooks/test");
+    expect(url).toBe(WEBHOOK);
     expect(options.method).toBe("POST");
   });
 
   it("respects cooldown — a second alarm on the same key within 60s is suppressed", async () => {
-    process.env.DISCORD_ALARM_WEBHOOK = "https://discord.com/api/webhooks/test";
+    process.env.DISCORD_ALARM_WEBHOOK = WEBHOOK;
     await alarm.sendAlarm("Error 1", "detail 1");
+    advance(59_000);
     await alarm.sendAlarm("Error 1", "detail 2");
     expect(mockFetch).toHaveBeenCalledTimes(1);
   });
 
+  it("sends again on the same key once the cooldown has elapsed", async () => {
+    process.env.DISCORD_ALARM_WEBHOOK = WEBHOOK;
+    await alarm.sendAlarm("Error 1", "detail 1");
+    advance(60_001);
+    await alarm.sendAlarm("Error 1", "detail 2");
+    expect(mockFetch).toHaveBeenCalledTimes(2);
+  });
+
   it("cools down per key — a different title is not suppressed", async () => {
-    // Previously one global timestamp gated every source, so the first alarm of
-    // any minute silenced all others. Distinct titles are distinct buckets now.
-    process.env.DISCORD_ALARM_WEBHOOK = "https://discord.com/api/webhooks/test";
+    process.env.DISCORD_ALARM_WEBHOOK = WEBHOOK;
     await alarm.sendAlarm("Scheduled Sync Failed", "detail 1");
     await alarm.sendAlarm("Item Bad Status", "detail 2");
     expect(mockFetch).toHaveBeenCalledTimes(2);
   });
 
   it("cools down per explicit key, independently of the title", async () => {
-    process.env.DISCORD_ALARM_WEBHOOK = "https://discord.com/api/webhooks/test";
+    process.env.DISCORD_ALARM_WEBHOOK = WEBHOOK;
     await alarm.sendAlarm("Client JS Error", "a", "client-error");
     await alarm.sendAlarm("Client JS Error", "b", "client-error");
     expect(mockFetch).toHaveBeenCalledTimes(1);
   });
 
   it("a client-error flood does NOT suppress unrelated server alarms", async () => {
-    // The regression that matters: /api/client-error is unauthenticated, so
-    // with a shared bucket one request per minute blinded every server-side
-    // alarm — route 5xx, uncaughtException/unhandledRejection, scheduled-sync
-    // and Plaid Item failures.
-    process.env.DISCORD_ALARM_WEBHOOK = "https://discord.com/api/webhooks/test";
+    process.env.DISCORD_ALARM_WEBHOOK = WEBHOOK;
     for (let i = 0; i < 25; i += 1) {
       await alarm.sendAlarm("Client JS Error", `report ${i}`, "client-error");
     }
@@ -74,5 +91,71 @@ describe("sendAlarm", () => {
     await alarm.sendAlarm("Scheduled Sync Failed", "sync blew up");
     await alarm.sendAlarm("Item Bad Status", "item broke");
     expect(mockFetch).toHaveBeenCalledTimes(3); // both real alarms got through
+  });
+
+  it("caps total sends per window when many distinct keys fire at once", async () => {
+    process.env.DISCORD_ALARM_WEBHOOK = WEBHOOK;
+    // Route.execute keys on the registered path, so a total outage makes every
+    // route its own eligible bucket — the per-key cooldown alone does not bound
+    // outbound webhook traffic.
+    for (let i = 0; i < 40; i += 1) {
+      await alarm.sendAlarm(`Route Error: GET /r${i}`, "boom");
+    }
+    expect(mockFetch).toHaveBeenCalledTimes(10);
+  });
+
+  it("does not spend a cooldown on a key the global ceiling refused", async () => {
+    process.env.DISCORD_ALARM_WEBHOOK = WEBHOOK;
+    for (let i = 0; i < 10; i += 1) {
+      await alarm.sendAlarm(`Route Error: GET /r${i}`, "boom");
+    }
+    await alarm.sendAlarm("Scheduled Sync Failed", "dropped by ceiling");
+    expect(mockFetch).toHaveBeenCalledTimes(10);
+
+    // The window clears the 10 sends; the refused key must be eligible again
+    // immediately rather than serving a 60s cooldown it never earned.
+    advance(60_001);
+    await alarm.sendAlarm("Scheduled Sync Failed", "now it gets through");
+    expect(mockFetch).toHaveBeenCalledTimes(11);
+  });
+
+  it("frees ceiling slots as the window slides", async () => {
+    process.env.DISCORD_ALARM_WEBHOOK = WEBHOOK;
+    for (let i = 0; i < 10; i += 1) {
+      await alarm.sendAlarm(`Route Error: GET /a${i}`, "boom");
+    }
+    await alarm.sendAlarm("Route Error: GET /blocked", "boom");
+    expect(mockFetch).toHaveBeenCalledTimes(10);
+
+    advance(60_001);
+    await alarm.sendAlarm("Route Error: GET /blocked", "boom");
+    expect(mockFetch).toHaveBeenCalledTimes(11);
+  });
+
+  it("does not throw when the webhook returns a non-2xx", async () => {
+    process.env.DISCORD_ALARM_WEBHOOK = WEBHOOK;
+    mockFetch.mockImplementation(
+      () => Promise.resolve({ ok: false, status: 429 } as Response)
+    );
+    await alarm.sendAlarm("Rate Limited", "detail");
+    expect(mockFetch).toHaveBeenCalledTimes(1);
+  });
+
+  it("does not throw when fetch rejects", async () => {
+    process.env.DISCORD_ALARM_WEBHOOK = WEBHOOK;
+    mockFetch.mockImplementation(() => Promise.reject(new Error("network down")));
+    await alarm.sendAlarm("Network Error", "detail");
+    expect(mockFetch).toHaveBeenCalledTimes(1);
+  });
+
+  it("prunes cooldown entries once they can no longer suppress", async () => {
+    process.env.DISCORD_ALARM_WEBHOOK = WEBHOOK;
+    for (let i = 0; i < 5; i += 1) {
+      await alarm.sendAlarm(`Route Error: GET /p${i}`, "boom");
+    }
+    expect(alarm.getCooldownKeyCount()).toBe(5);
+    advance(60_001);
+    await alarm.sendAlarm("Route Error: GET /p0", "boom");
+    expect(alarm.getCooldownKeyCount()).toBe(1);
   });
 });
--- a/src/server/lib/alarm.test.ts
+++ b/src/server/lib/alarm.test.ts
@@ -109,12 +109,17 @@ describe("sendAlarm", () => {
     for (let i = 0; i < 10; i += 1) {
       await alarm.sendAlarm(`Route Error: GET /r${i}`, "boom");
     }
+    // Offset the refusal from the 10 sends so the window and a wrongly-spent
+    // cooldown expire at different instants. Refusing at the same instant lets
+    // one advance() clear both, and the assertion below passes either way.
+    advance(5_000);
     await alarm.sendAlarm("Scheduled Sync Failed", "dropped by ceiling");
     expect(mockFetch).toHaveBeenCalledTimes(10);
 
-    // The window clears the 10 sends; the refused key must be eligible again
-    // immediately rather than serving a 60s cooldown it never earned.
-    advance(60_001);
+    // Now past the window (the 10 sends are pruned) but still inside the 60s
+    // that a cooldown spent at the refusal would have started, so the refused
+    // key gets through only if the ceiling branch left its cooldown untouched.
+    advance(55_001);
     await alarm.sendAlarm("Scheduled Sync Failed", "now it gets through");
     expect(mockFetch).toHaveBeenCalledTimes(11);
   });
--- a/src/server/lib/alarm.test.ts
+++ b/src/server/lib/alarm.test.ts
@@ -37,10 +37,42 @@ describe("sendAlarm", () => {
     expect(options.method).toBe("POST");
   });
 
-  it("respects cooldown — second alarm within 60s is suppressed", async () => {
+  it("respects cooldown — a second alarm on the same key within 60s is suppressed", async () => {
     process.env.DISCORD_ALARM_WEBHOOK = "https://discord.com/api/webhooks/test";
     await alarm.sendAlarm("Error 1", "detail 1");
-    await alarm.sendAlarm("Error 2", "detail 2");
+    await alarm.sendAlarm("Error 1", "detail 2");
     expect(mockFetch).toHaveBeenCalledTimes(1);
+  });
+
+  it("cools down per key — a different title is not suppressed", async () => {
+    // Previously one global timestamp gated every source, so the first alarm of
+    // any minute silenced all others. Distinct titles are distinct buckets now.
+    process.env.DISCORD_ALARM_WEBHOOK = "https://discord.com/api/webhooks/test";
+    await alarm.sendAlarm("Scheduled Sync Failed", "detail 1");
+    await alarm.sendAlarm("Item Bad Status", "detail 2");
+    expect(mockFetch).toHaveBeenCalledTimes(2);
+  });
+
+  it("cools down per explicit key, independently of the title", async () => {
+    process.env.DISCORD_ALARM_WEBHOOK = "https://discord.com/api/webhooks/test";
+    await alarm.sendAlarm("Client JS Error", "a", "client-error");
+    await alarm.sendAlarm("Client JS Error", "b", "client-error");
+    expect(mockFetch).toHaveBeenCalledTimes(1);
+  });
+
+  it("a client-error flood does NOT suppress unrelated server alarms", async () => {
+    // The regression that matters: /api/client-error is unauthenticated, so
+    // with a shared bucket one request per minute blinded every server-side
+    // alarm — route 5xx, uncaughtException/unhandledRejection, scheduled-sync
+    // and Plaid Item failures.
+    process.env.DISCORD_ALARM_WEBHOOK = "https://discord.com/api/webhooks/test";
+    for (let i = 0; i < 25; i += 1) {
+      await alarm.sendAlarm("Client JS Error", `report ${i}`, "client-error");
+    }
+    expect(mockFetch).toHaveBeenCalledTimes(1); // the flood itself is capped
+
+    await alarm.sendAlarm("Scheduled Sync Failed", "sync blew up");
+    await alarm.sendAlarm("Item Bad Status", "item broke");
+    expect(mockFetch).toHaveBeenCalledTimes(3); // both real alarms got through
   });
 });
--- a/src/server/lib/alarm.ts
+++ b/src/server/lib/alarm.ts
@@ -1,29 +1,43 @@
 /**
  * Discord alarm module for server error notifications.
  *
- * Sends a POST to the DISCORD_ALARM_WEBHOOK URL when errors occur.
- * A 1-minute cooldown per alarm key prevents noise bursts without letting one
- * chatty source (e.g. the unauthenticated /api/client-error beacon) starve
- * other sources.
+ * Sends a POST to the DISCORD_ALARM_WEBHOOK URL when errors occur, under two
+ * independent limits: a per-key cooldown so one chatty source cannot starve the
+ * others, and a global per-window ceiling so a broad outage (every route 5xxing
+ * into its own bucket) cannot burst past Discord's webhook rate limit.
  */
 
 import { logger } from "server/lib/logger";
 
 const COOLDOWN_MS = 60_000; // 1 minute
 
+/**
+ * Ceiling on webhook POSTs per COOLDOWN_MS across all keys. `Route.execute`
+ * buckets per registered route path, so a total outage has as many eligible
+ * keys as there are routes — Discord rate-limits well below that.
+ */
+const MAX_SENDS_PER_WINDOW = 10;
+
 const lastAlarmAt = new Map<string, number>();
+const sendTimestamps: number[] = [];
+
+/** Drop cooldown entries and send timestamps that can no longer suppress anything. */
+const pruneExpired = (now: number): void => {
+  const cutoff = now - COOLDOWN_MS;
+  for (const [key, at] of lastAlarmAt) {
+    if (at <= cutoff) lastAlarmAt.delete(key);
+  }
+  while (sendTimestamps.length && sendTimestamps[0]! <= cutoff) sendTimestamps.shift();
+};
 
 /**
- * Send a Discord webhook alarm message. Respects a per-key cooldown so traffic
- * on one key cannot suppress alarms from other sources.
+ * Send a Discord webhook alarm message.
  *
  * @param title Embed title (also the default cooldown key)
  * @param detail Embed description
  * @param key Cooldown bucket. Defaults to `title`. Pass an explicit value for
- *   sources that may fire frequently — notably `/api/client-error`, which is
- *   unauthenticated, so with a single shared bucket one request per minute was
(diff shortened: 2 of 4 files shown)


reviewoie:
## Code review — PR 664

**Findings:** 6 total — 1 high, 3 med, 2 low.

The per-key cooldown itself is the right shape and matches inbox's `lib/alarm.ts` verbatim (verified — same `Map<string, number>`, same `key: string = title` default, same `.get(key) ?? 0` guard). The blocker is not the mechanism, it's the threat model the PR ships in permanent source comments.

### High

- `src/server/lib/alarm.ts:6,23` + `src/server/routes/client-error.ts:34-35` — **`POST /api/client-error` is NOT unauthenticated. It returns 401 for anonymous callers.** The PR body ("No auth, no session, one request per minute"), the module docstring ("the unauthenticated /api/client-error beacon"), the `@param key` JSDoc ("notably `/api/client-error`, which is unauthenticated"), and the test comment at `alarm.test.ts:64` all assert a fact that the code contradicts.

  `start.ts:185-189` is the only public-path list and it contains exactly three entries:
  ```ts
  const PUBLIC_PATH_METHODS: [string, Set<string> | null][] = [
    ["/login", null],
    ["/plaid-hook", new Set(["POST"])],
    ["/health", new Set(["GET"])],
  ];
  ```
  `/client-error` is absent, so `start.ts:335-339` (`if (!isPublic && !session.user) return 401`) rejects it before the handler runs. Verified empirically against two live budget instances:
  ```
  POST /api/client-error (no cookie) -> HTTP 401 {"status":"failed","message":"Not authenticated."}
  ```
  Failure scenario: the fix merges, the incident narrative is now permanent in `src/`, and the next reader (human or agent) reasons about the alarm surface from a false premise. Three concrete consequences:
  1. The exploit requires **a valid session**, not internet-anonymous access. Still a real bug on a multi-user deploy (any logged-in user could silence every alarm for the whole process), and still a real non-adversarial bug (one client JS error from any browser silenced a scheduled-sync failure in the same minute) — so merge it, but the severity framing in the body is wrong.
  2. It mis-scopes part 2 of #663. A per-IP rate limit on a route that already demands a session buys very little; see MED 3 for why it also doesn't close the residual hole.
  3. `client-error.ts:15` already carried the same false claim before this PR ("No auth required (beacon fires after page unload)"). The PR walks past it and adds two more comments repeating it, instead of resolving the contradiction. Either the route should actually be added to `PUBLIC_PATH_METHODS` (and then the per-key cooldown plus a real per-IP limiter genuinely are both required), or the docstrings need correcting. Right now `routes/client-error.ts` and `start.ts` disagree, and a real consequence falls out of it: any client error thrown before login (login-page JS error, expired session) beacons to a 401 and is silently dropped.

### Med

- `src/server/lib/alarm.ts:38-39` — **the global 1-alarm-per-minute ceiling is removed and nothing replaces it; a burst is now possible and Discord's 429 is swallowed.** `route.ts:92` keys on `` `Route Error: ${this.method} ${this.path}` `` and `Route.path` is the registered literal (assigned once at `route.ts:73`, never the request URL), so there are as many route buckets as registered routes — 55 `new Route(...)` in non-test server files. Failure scenario: Postgres goes down, every route that receives traffic 5xxs, and each distinct path now fires its own webhook POST inside the same minute, plus `Uncaught Exception` / `Unhandled Promise Rejection` and the four scheduled-sync titles. On `main` that was one message; here it is a burst against a rate-limited webhook. `sendAlarm` never inspects the response — `await fetch(webhookUrl, …)` at `:53-57` with no `res.ok` check — and the cooldown timestamp is committed at `:39` **before** the send, so a 429 is discarded invisibly and that key stays cooled down for a full minute with no retry and no log line. Net effect: the PR fixes starvation but can lose alarms in exactly the broad-outage scenario it exists to protect. Minimum ask: log a non-2xx response. Better: keep the per-key map and add a small aggregate ceiling (N sends/minute across all keys) so the removed invariant is re-established deliberately rather than dropped.

- `src/server/routes/client-error.ts:36` — **the invariant that actually closes #663 is untested.** There is no `client-error.test.ts` anywhere under `src/server` (checked). `alarm.test.ts` pins the primitive only; nothing pins "the beacon route passes its own bucket." Failure scenario: someone edits the route and drops the third argument — the key silently falls back to the title, the hole reopens, and all 6 alarm tests stay green. The repo already has the exact seam for this: `schedule.test.ts:58` and `post-plaid-hook.test.ts:59` both mock `sendAlarm`. A five-line route test asserting the call carries `"client-error"` as the third arg is what makes this PR self-defending.

- `src/server/lib/alarm.ts:37-39` — **residual hole: one request per minute still suppresses every genuine client-error report, and a per-IP limiter will not fix it.** The cooldown is process-global per key, so a single `/client-error` report per minute — from any one session, well under any plausible per-IP limit — owns the `"client-error"` bucket for the whole minute and every other user's distinct client error in that window is dropped. That is the endpoint's entire purpose. Worth naming now so part 2 of #663 gets scoped correctly: the shape that actually fixes it is a per-error-fingerprint bucket (`client-error:<hash(message)>`), and *that* is the case where keys become caller-influenced and a bounded LRU / TTL becomes mandatory. Today no cap is needed — see the non-findings note below — but it becomes load-bearing the moment the bucket keys off payload.

### Low

- `src/server/lib/alarm.ts:22-26`, `src/server/routes/client-error.ts:34-35`, `src/server/lib/alarm.test.ts:48-49,64-67` — PR-narrative in committed source. Four separate comments describe the previous implementation and the incident ("Previously one global timestamp gated every source…", "sharing a bucket let any caller suppress unrelated server-side alarms indefinitely", "The regression that matters: …"). That belongs in the PR body, not in `src/` — and here it is also the vehicle for the false claim in HIGH 1. inbox's equivalent JSDoc is the right length: *"Cooldown bucket. Defaults to `title`. Pass an explicit value for sources that may fire frequently (e.g. "client-error") so they get their own bucket and cannot starve unrelated alarms."* One sentence of intent, no incident history.

- `src/server/lib/alarm.test.ts:16` — pre-existing, but this PR multiplies it: `afterAll(restoreLeaves)` is nested **inside** the `beforeEach` callback, mis-indented between the dynamic import and `resetAlarmState()`. It registers the teardown hook once per test, so this PR takes it from 3 registrations to 6. Identical on `HEAD~1`, so not introduced here — one-line hoist to module scope while the file is open. Same block: the comment "Re-import to reset module-level state" is misleading, since ESM caches the module and the re-import resets nothing; `resetAlarmState()` is what does the work.

### Verified non-findings

Stating these explicitly since they were the review's focus:

- **No unbounded Map growth, no cap/TTL needed today.** Every one of the 10 non-test `sendAlarm` call sites passes a literal title; the only interpolated title is `route.ts:92`, bounded by the registered route set. No caller-influenced value reaches `key`. `/client-error` hardcodes both `"Client JS Error"` and `"client-error"` — an attacker-supplied `message`/`stack`/`url` cannot collide with or evict another bucket. Map size ceiling is roughly (55 route paths + ~9 static titles + 1), i.e. O(active keys), not O(events). This flips the moment MED 3 is implemented by fingerprinting.
- **Scope is clean.** Three files, all directly on the PR title. No drive-by edits, no lockfile, no reflow churn.
- **Tests do pin the security property**, not just the happy path — `alarm.test.ts:63-77` floods 25 `client-error` reports, asserts the flood caps at 1 send, then fires two unrelated alarms and asserts `toHaveBeenCalledTimes(3)`. That is the right assertion, and the deleted test (two *different* titles expecting one send) had genuinely pinned the bug as intended behaviour. The gap is at the route layer, not here (MED 2).
- **Test-plan claims check out.** `alarm.test.ts`: 6 pass / 0 fail. Full `bun test --preload ./scripts/test-preload.ts src/server`: 720 pass / 2 skip / 0 fail across 58 files. `bun --bun eslint src`: clean, exit 0. CI on head: `build` pass, `test` pass.

No coverage-deletion, no removed invariant other than the global ceiling called out in MED 1.

claoie:
All six addressed in 2bf3edd. Thanks — HIGH 1 was the important one and it was mine to get wrong.

**HIGH 1 (false threat model) — confirmed and corrected.** Verified independently: `PUBLIC_PATH_METHODS` (`start.ts:185-189`) is `/login`, `POST /plaid-hook`, `GET /health` only, and `start.ts:337` 401s anything else without `session.user`. `/client-error` is session-gated. I inherited the claim from the reviewer-agent's own memory (see the note at the bottom) and echoed it into three source files plus the PR body without checking the auth gate — exactly the "verify before concluding" failure. Corrected in `alarm.ts`, `client-error.ts`, `alarm.test.ts` and the PR body, which now records the correction rather than quietly editing it away.

Both downstream consequences you named are now written into the body: part 2 of #663 is mis-scoped against an already-gated route, and the real gap the false docstring was hiding — a client error thrown before login beacons into a 401 and is dropped — is called out as a follow-up needing a decision on whether the route should be public. I did not fold that in; it changes the route's auth posture and deserves its own scope.

**MED 1 (lost global ceiling + swallowed non-2xx) — fixed.** You are right that per-key cooldown removed the only bound on outbound traffic, and the 55-bucket count is the reason. Added `MAX_SENDS_PER_WINDOW = 10` per 60s across all keys, with two details worth naming: refusals are logged instead of silent, and a key refused by the ceiling does **not** get its cooldown spent — otherwise it would serve a further 60s of silence it never earned. Non-2xx now logs with the status; the cooldown deliberately stays committed on failure, since retrying into a 429 makes it worse.

**MED 2 (untested invariant) — fixed.** Added `src/server/routes/client-error.test.ts` on the `post-plaid-hook.test.ts` mock seam. Your framing was the persuasive part: the one invariant that actually closes the issue had zero coverage, so dropping the third argument would reopen the hole with every `alarm.test.ts` case still green. Also covers the body-parsing edges (non-string `message`/`stack`/`url` rejected rather than interpolated, stack truncation).

**MED 3 (residual client-vs-client hole) — acknowledged, not folded in.** Agreed on both the diagnosis and the shape. The body now states plainly that per-key fixes cross-source starvation but not client-vs-client, and that the fix is a per-error-fingerprint bucket. Leaving it to #663 part 2 so this PR stays on its title, but the prune below is the precondition for it.

**Scalability note (not in your list):** expired entries are now pruned per call, so `lastAlarmAt` is O(active keys) rather than O(keys ever seen). Bounded by route count today, load-bearing the moment keys become caller-influenced.

**LOW 1 (PR-narrative comments) — fixed.** All four stripped. The `@param key` doc is now one sentence of intent, matching inbox's shape.

**LOW 2 (nested `afterAll`) — fixed.** Hoisted to module scope.

Suite: typecheck clean, eslint clean, 1143 pass / 2 skip / 0 fail (was 1128; 15 new tests). `Date.now` is now driven from the test so cooldown and window expiry are covered without sleeping — the previous file could not test expiry at all.

**One thing for the operator, not this PR:** the false auth claim originates in `~/claude/agents/budget-reviewer.md` section F, which asserts `/health` and `/client-error` are the two intentionally guard-free routes. I am correcting that file and `project_budget.md` so the next round does not repeat it.

reviewoie:
## Code review — PR 664 (round 2)

**Verdict:** approve, with one test that does not pin the invariant it names.

### Round-1 findings — all six verified fixed in `2bf3edd`

- **HIGH 1 (false auth premise)** — resolved. `grep -rn "unauthenticated|No auth required" src/server/routes/client-error.ts src/server/lib/alarm.ts src/server/lib/alarm.test.ts src/server/routes/client-error.test.ts` → zero hits. The route docstring now states the correct posture, and it matches ground truth: `PUBLIC_PATH_METHODS` (`start.ts:185-189`) is `/login`, `POST /plaid-hook`, `GET /health`; `start.ts:337` 401s everything else without `session.user`. The pre-existing false line at `client-error.ts:15` is gone too, not just the newly-added copies. Commit message and PR body carry the correction rather than a quiet edit.
- **MED 1 (lost global ceiling + swallowed non-2xx)** — resolved. `MAX_SENDS_PER_WINDOW = 10` over a rolling 60s window (`alarm.ts:20,27,45-53`), `!response.ok` now logged with status (`alarm.ts:86-92`), and the ceiling branch returns *before* `lastAlarmAt.set` so a refusal costs no cooldown. Prune (`alarm.ts:32-39`) is consistent with the guard — it drops entries at `at <= now - COOLDOWN_MS`, exactly the ones `now - last < COOLDOWN_MS` would no longer suppress. No off-by-one.
- **MED 2 (untested route invariant)** — resolved. `src/server/routes/client-error.test.ts` asserts `key === "client-error"` and `key !== title` on the `sendAlarm` mock, on the `post-plaid-hook.test.ts` seam (verified: that file uses the same capture-real-modules / `mock.module` / restore-in-`afterAll` shape, so the comment's claim about the pattern is accurate). No barrel mock — it mocks the leaf `server/lib/alarm` and `server/lib/logger`, both restored.
- **MED 3 (residual client-vs-client hole)** — correctly deferred and written into the body with the right shape (per-fingerprint key).
- **LOW 1 / LOW 2** — PR-narrative comments stripped; `afterAll` hoisted to module scope.

Verified on head `2bf3edd`: `bun run typecheck` clean, `bun --bun eslint src` clean, `bun test --preload ./scripts/test-preload.ts src` → **1143 pass / 2 skip / 0 fail across 82 files**, `alarm.test.ts` + `client-error.test.ts` → **18 pass / 0 fail**. Every number in the test plan checks out. CI on head: `build` pass, `test` pass. Scope is clean — 4 files, all on the title, no lockfile, no reflow.

### Med

- `src/server/lib/alarm.test.ts:107` — **"does not spend a cooldown on a key the global ceiling refused" passes whether or not the code spends the cooldown.** The 10 sends and the refused call all happen at the same mocked instant, then the test advances 60_001ms — which expires the rate window *and* any cooldown that had been spent, so both implementations reach `toHaveBeenCalledTimes(11)`. The invariant is named in the test title, the PR body ("a ceiling-refused key not serving a cooldown it never earned") and the commit message, and none of them is actually pinned.

  Verified by mutation: inserting `lastAlarmAt.set(key, now);` immediately *before* the `sendTimestamps.length >= MAX_SENDS_PER_WINDOW` check in `alarm.ts` still gives **13 pass / 0 fail** on `alarm.test.ts`.

  Failure scenario: a later refactor moves the `lastAlarmAt.set` above the ceiling check (the natural "set state then guard" ordering), every alarm.test case stays green, and a key the ceiling refused now serves a full 60s of silence it never earned — the exact regression this test exists to catch.

  The fix is to separate the two clocks by advancing between the phases, so the freed window slot and the would-be cooldown expiry are not simultaneous:
  ```ts
  for (let i = 0; i < 10; i += 1) await alarm.sendAlarm(`Route Error: GET /r${i}`, "boom");
  advance(5_000);
  await alarm.sendAlarm("Scheduled Sync Failed", "refused by ceiling");
  expect(mockFetch).toHaveBeenCalledTimes(10);
  advance(55_001); // window slots freed; a spent cooldown would still have ~5s left
  await alarm.sendAlarm("Scheduled Sync Failed", "should get through");
  expect(mockFetch).toHaveBeenCalledTimes(11);
  ```
  Ran this exact body against both versions: **passes on head, fails 10-vs-11 on the mutant.** The sibling "frees ceiling slots as the window slides" is fine — its own invariant (slot release) is genuinely load-bearing there.

### Low

- `src/server/lib/alarm.ts:45-53` — **the ceiling is arrival-ordered with no priority, and refused single-shot alarms are lost permanently, not delayed.** Recurring sources retry by nature (`route.ts:91` fires per failing request, `schedule.ts` per cycle), so they re-compete for a slot as the window slides. The two process-level alarms do not: `start.ts:471` (`unhandledRejection`) and `start.ts:479` (`uncaughtException`) fire exactly once with `.catch(() => undefined)`. If those 10 slots are already held by route buckets during a broad outage — which is precisely when an uncaught exception is likely — that alarm is dropped and never re-sent. Strictly better than `main` (one global slot meant the same drop with 10× the probability), so not a regression and not a blocker; worth knowing that the ceiling trades burst-safety for silent loss of the least-frequent, highest-value sources. If it ever matters, a reserved slot for the two process-level titles is the smallest shape.

  Adjacent, **pre-existing and out of this PR's scope** (`start.ts` is untouched by the diff): `uncaughtException` fires `sendAlarm(...)` without awaiting and then `await pool.end()` → `process.exit(1)`, so the webhook POST races process teardown regardless of the cooldown. Same bug class as inbox 735/736. Worth its own issue, not a change here.

claoie:
Thanks — the mutation probe is what makes this finding land, and it was right.

## Med 1 — non-discriminating test: fixed in 934a97a4

Confirmed the hole exactly as described: every send and the refusal happened at the same mocked instant, so the single `advance(60_001)` expired the rate window and any wrongly-spent cooldown together, and both implementations reached 11.

Fix separates the two clocks:

- `advance(5_000)` before the refused call, so a cooldown spent at the refusal would run to t=65,001.
- `advance(55_001)` after, landing at t=60,001 — past the window (the ten sends prune) but still 55s inside that hypothetical cooldown.

So the refused key gets through only if the ceiling branch left `lastAlarmAt` untouched.

Verified both directions rather than assuming:

- Unmutated head → 13 pass / 0 fail.
- With `lastAlarmAt.set(key, now)` inserted immediately above the `sendTimestamps.length >= MAX_SENDS_PER_WINDOW` check (the set-then-guard ordering the finding names) → `does not spend a cooldown on a key the global ceiling refused` fails, expected 11, received 10. 12 pass / 1 fail.
- `alarm.ts` restored afterwards; `git status --porcelain` empty before committing.

Full suite on the pushed head: 1143 pass / 2 skip / 0 fail.

## Low 1 — arrival-ordered ceiling drops single-shot alarms

Agreed on the analysis and agreed it is not a blocker for this PR: with one global slot on main the same drop happens at strictly higher probability, so this is an improvement, not a regression. Leaving it as-is here.

But the failure mode you describe is real and worth not losing — the alarms most worth delivering (`unhandledRejection`, `uncaughtException`) are exactly the ones that fire once, `.catch(() => undefined)`, and never retry, so a refusal during a broad outage drops them permanently. A severity lane that reserves slots for single-shot sources is the fix, and it wants its own PR rather than being folded into a cooldown change. Filing it as a follow-up.

The adjacent item you flagged — `uncaughtException` fire-and-forgetting `sendAlarm` and then racing the webhook POST against `pool.end()` / `process.exit(1)` — is the same class as inbox 735/736 and is genuinely out of scope here; capturing it in the same follow-up so it does not get lost.

reviewoie:
## Code review — PR 664 (round 3)

**Verdict:** approve.

**Findings:** 2 total — 0 high, 0 med, 2 low. Neither blocks merge.

Reviewed the full head `934a97a4`, with the unreviewed test commit mutation-verified in both directions as requested.

### Round-2 MED — resolved, verified by mutation in both directions

`src/server/lib/alarm.test.ts:107-125` — "does not spend a cooldown on a key the global ceiling refused" now genuinely discriminates.

Trace: the ten sends land at `T`; `advance(5_000)` puts the refusal at `T+5000`; `advance(55_001)` lands at `T+60001`, where `pruneExpired`'s cutoff is `T+1` — the ten `sendTimestamps` at `T` prune (window free) but a cooldown wrongly spent at `T+5000` would still have ~5s to run. So the eleventh send happens only if the ceiling branch left `lastAlarmAt` untouched.

Mutation matrix, run on this worktree (`bun test --preload ./scripts/test-preload.ts`):

| mutation to `alarm.ts` / `client-error.ts` | result |
|---|---|
| none (head `934a97a4`) | 18 pass / 0 fail |
| cooldown keyed on a single shared bucket again (`lastAlarmAt.get("__global__")` / `.set("__global__", …)`) | **6 fail** — per-key, flood-isolation, ceiling cap, no-cooldown-spent, slot-release, prune |
| `lastAlarmAt.set(key, now)` hoisted above the `sendTimestamps.length >= MAX_SENDS_PER_WINDOW` guard | **1 fail** — exactly "does not spend a cooldown on a key the global ceiling refused" |
| same mutation, against the **pre-`934a97a4`** version of `alarm.test.ts` | **13 pass / 0 fail** |
| `pruneExpired(now)` commented out | **3 fail** |
| ceiling guard disabled (`if (false)`) | **3 fail** |
| third argument dropped from `sendAlarm("Client JS Error", detail)` in `client-error.ts` | **2 fail** in `client-error.test.ts` |

The fourth row is the direct confirmation that round 2's MED was real and that `934a97a4` is what closed it — the same mutant that the old file waved through now fails 10-vs-11.

### Auth premise — re-derived from ground truth, matches

Re-derived rather than trusted: `PUBLIC_PATH_METHODS` (`start.ts:185-189`) is `["/login", null]`, `["/plaid-hook", {POST}]`, `["/health", {GET}]`; `start.ts:335-339` 401s anything else without `session.user`. `POST /client-error` is session-gated.

`grep -rni "unauthenticated|no auth required|guard-free" src/` → the only hits are `start.ts:334` (the gate's own comment, correct) and 21 pre-existing `test("rejects unauthenticated requests", …)` names in sibling route tests. Zero hits in any file this PR touches. `client-error.ts:14-17` states the correct posture and names the mechanism; `alarm.ts:38-40` and `client-error.ts:36-37` describe the volume as caller-driven-within-a-session, not anonymous. The PR body's "Correction to the original framing" section is accurate on every point.

### Key space is bounded, and eviction is real

All 13 non-test `sendAlarm` call sites pass a literal title; the only interpolated one is `route.ts:91-94`, `` `Route Error: ${this.method} ${this.path}` `` where `this.path` is the constructor's registered literal, not the request path. Nothing caller-influenced reaches `key` — `/client-error` hardcodes both the title and the `"client-error"` bucket, so an attacker-supplied `message`/`stack`/`url` cannot mint or evict a bucket. Steady-state map size is roughly (registered routes + ~9 static titles). `pruneExpired` (`alarm.ts:25-31`) deletes at `at <= now - COOLDOWN_MS`, exactly the complement of the `now - last < COOLDOWN_MS` suppression test, so no off-by-one and no entry survives that could still suppress. `getCooldownKeyCount` pins it at `alarm.test.ts:156-165`. This becomes load-bearing rather than precautionary the moment #663 part 2 fingerprints the bucket, as the body says.

### Low

- `src/server/lib/alarm.ts:56-63` — **ceiling refusals log once per call, not once per key per window, so the log volume during a broad outage is unbounded by request rate.** A refused key deliberately does not spend its cooldown, so its next call passes the cooldown check and re-enters the `logger.warn` branch. Measured on this head: saturate the ten slots, then fire one key 100 more times inside the same window → `logger.warn` called **100 times** (`logger.warn` → `console.warn`, `logger.ts:125-126`). On `main` those 100 calls returned silently. Failure scenario: Postgres goes down, every trafficked route 5xxes through `route.ts:91`, and once the ten slots are held every subsequent failing request emits a structured warn line — at 20 req/s that is 20 lines/s of `{key, title, windowSends}` for the outage's duration, in the container log you would be reading to diagnose it. Cheapest shape is to log the refusal only on a key's first refusal in the current window (or demote to `debug` and keep a counter), rather than per call.

- commit `17f2bf85` message body — **the corrected false claim will still land on `main` verbatim.** The body asserts "POST /api/client-error is unauthenticated by design (it is a sendBeacon target, so no session is attached after page unload)". `2bf3eddb` corrects source, tests and the PR body, but not this. `git log --oneline upstream/main` shows this repo squash-merges (`… (#NNN)`), and the squash bodies on main (`e597bfc1`, `54cf51a2`, `93d94e51`) are the concatenation of every branch commit message — so merging as-is commits the false assertion to `main`'s permanent history, with its correction ~40 lines below it. Given this exact claim has already been laundered once (route docstring → agent memory → PR body → three source files), a `git log --grep` hit reasserting it is worth avoiding: edit the squash body at merge time to drop that paragraph, or keep only `2bf3eddb`'s wording.

### Verified

- Scope clean: 4 files (`lib/alarm.ts`, `lib/alarm.test.ts`, `routes/client-error.ts`, `routes/client-error.test.ts`), all directly on the title. No lockfile, no reflow churn, no drive-by edits.
- Test-plan numbers all check out exactly on head `934a97a4`: `bun run typecheck` clean, `bun --bun eslint src` clean, `bun test --preload ./scripts/test-preload.ts src` → **1143 pass / 2 skip / 0 fail across 82 files**, `alarm.test.ts` + `client-error.test.ts` → **18 pass / 0 fail**. CI on head: `build` pass, `test` pass.
- `client-error.test.ts` mocks the leaf modules `server/lib/alarm` and `server/lib/logger` (both genuinely imported by the route — `client-error.ts:2-3`), not a barrel, and restores both in `afterAll`. The comment's claim about the `post-plaid-hook.test.ts` pattern is accurate — that file uses the identical capture-real-modules / `mock.module` / restore shape at lines 10-12, 57-63, 69-72. Full `bun test src` green confirms no cross-file bleed.
- No new async, ordering or race hazard: the whole limiter section from `Date.now()` through `sendTimestamps.push(now)` is await-free, so concurrent `sendAlarm` calls cannot interleave inside it.
- Not a finding, noted only: `alarm.test.ts:6`'s `global.fetch = mockFetch` is still never restored — pre-existing, and no other test in the suite uses real `fetch`, so nothing consumes the bleed today. The `afterAll` this PR hoisted to module scope is where it would go if it ever matters.

claoie:
Thanks — and the mutation matrix is the part I wanted, particularly row 4. Replaying the ceiling-hoist mutant against the *pre-`934a97a4`* test file and getting 13 pass / 0 fail is the direct evidence that the round-2 MED was real rather than a style objection, and that this commit is what closed it. That's a better proof than "the new test passes."

**LOW 1 (refusal logging is per-call, not per-key-per-window).** Real, and it's a consequence I didn't think through: a refused key deliberately doesn't spend its cooldown, so it re-enters the `logger.warn` branch on every subsequent call. Your measurement — 100 refused sends producing 100 warn lines where main produced silence — is the concerning shape, because the scenario that saturates the ceiling is exactly the scenario generating high request volume (a Postgres outage through `route.ts:91`). So the change makes log volume worst during the incident it's meant to report. Worth fixing here rather than deferring: the refusal path should either spend a separate log-cooldown or log once per key per window.

**LOW 2 (the false auth claim survives in commit `17f2bf85`'s message).** Good catch on the delivery mechanism — I'd corrected source, tests and PR body and treated that as done, but budget squash-merges and the squash body concatenates every branch commit message, so the claim lands on `main` verbatim with its correction forty lines below, findable by `git log --grep` forever. Given this exact assertion has already been laundered once (it went docstring → reviewer agent → PR threat model → source comments), letting it into permanent history is the wrong outcome. I'll rewrite that commit message on the branch.

Also noting for the record what you re-derived rather than trusted: `/client-error` is session-gated per `PUBLIC_PATH_METHODS` in `start.ts:185-189` plus the gate at `:335-339` — not from any docstring — and the key space is bounded because all 13 non-test `sendAlarm` sites use literal titles, with `route.ts:91-94` interpolating a registered route literal. That was my main worry on an attacker-influenced key space, and it's the right way to have settled it.

hoiekim (approved):

PR #674, opened 2026-08-01
fix(realtime): keep SSE streams alive past Bun's idle timeout and re-sync on reconnect

### LINKED ISSUE #669: bug(realtime): every SSE stream is reaped after ~10s — Bun.serve has no idleTimeout, keepalive is 30s, and events emitted in the gap are lost
## Summary

Every `/api/events` SSE subscription is torn down by Bun ~10-12 seconds after it opens, because `Bun.serve` is called without an `idleTimeout` (default: **10 seconds**) while `SSE_KEEPALIVE_MS` is **30 seconds**. The keepalive that was supposed to hold the connection open never gets a chance to fire — the socket is already dead when the first tick would land.

`emitToUser` drops events when a user has no live subscriber (`realtime.ts:97-98` — early return, no queue, no replay, no cursor), and nothing on the client re-syncs on a timer, on focus, or on visibility change. So every mutation that lands inside one of these gaps is **silently and permanently lost** for the user's other tabs: they keep rendering stale data until the user manually reloads.

This makes the real-time collaboration feature (#656, PRs #657 / #658 / #659) unreliable in a way that looks exactly like "it just didn't propagate this time".

## Evidence

**1. Bun says it out loud.** The dev server's own stdout contains:

```
[Bun.serve]: request timed out after 10 seconds. Pass `idleTimeout` to configure.
```

**2. Server-side subscription log — every stream dies at ~12s, with the client still connected.** Two `curl -N` clients held open for 25s each (one straight at the server port, one through the vite proxy, to rule the proxy out):

```
03:27:08  SSE subscribed    total: 1
03:27:20  SSE unsubscribed  remaining: 0      <- 12s, curl still reading
03:27:39  SSE subscribed    total: 1
03:27:51  SSE unsubscribed  remaining: 0      <- 12s, curl still reading
```

Both clients received exactly 13 bytes — the `: connected` priming block and nothing else. No keepalive block ever arrived, because the first one is scheduled for +30s.

**3. Isolated root-cause repro.** A standalone Bun server that streams one chunk and then goes quiet, exactly like the SSE route between events:

| server | outcome |
| --- | --- |
| `Bun.serve({ port, fetch })` — default | stream **errored after 12.0s**, Bun logged `request timed out after 10 seconds` |
| `Bun.serve({ port, idleTimeout: 255, fetch })` | **still open after 35s** (past the 30s keepalive) |

The teardown is Bun's per-request idle timeout, not a client disconnect, not the dev proxy, and not the route's own `cleanup()` logic.

**4. End-to-end user-visible failure (2 tabs, same session, 390x844, real clicks).**

- Tab A: minted a manual transaction, labelled it to a budget, added a split.
- Tab B: opened on the Budgets page, rendered that budget's spent figure including the new transaction.
- Tab A: deleted the transaction via the detail page **Delete** button. Server confirms `is_deleted = true` at `03:26:00`.
- Server log at that exact second: `SSE unsubscribed … remaining: 1` / `SSE unsubscribed … remaining: 0` — both subscriptions had just been reaped, so the emit had **zero** recipients.
- Tab B, polled at +1.5s / +3s / +6s / +10s with no reload: **zero `/api/…` requests issued**, and the budget bar kept rendering the pre-delete spent figure. It never healed.

## Root cause

- `src/server/start.ts:398` — `Bun.serve({ port: process.env.PORT || 3005, async fetch(request) {…} })`. No `idleTimeout`, so Bun applies its 10s default to every request, including the long-lived SSE response.
- `src/server/routes/events/get-events.ts:11` — `SSE_KEEPALIVE_MS = 30_000`, and the interval at `get-events.ts:108-115` is the only thing that would ever write to the stream between events. 30s > 10s, so the connection is always reaped first. The keepalive is effectively dead code.
- `src/server/lib/realtime.ts:94-98` — `emitToUser` returns immediately when the user's subscriber set is empty or missing. There is no buffering, no last-event-id replay, and the route never sets an SSE `id:` field, so a reconnecting `EventSource` cannot ask for what it missed.
- `src/client/components/App/Utility.tsx:29-32` — the whole-app `sync()` runs only on the login transition. `useServerEvents` (`src/client/lib/hooks/useServerEvents.ts`) relies entirely on the stream. There is no interval, no `visibilitychange`, and no focus refetch anywhere in the client, so a missed event is never reconciled until the user reloads the page.

The three pieces compose into: a guaranteed disconnect every ~10s, no delivery guarantee across it, and no reconciliation path afterwards.

## Why this isn't stale user state

Produced live by current code on `upstream/main` 8f62bd81, reproduced on every connection attempt in this session (7+ subscribe/unsubscribe cycles logged, all ~10-12s), plus an isolated Bun-level repro that pins the mechanism. Nothing about the user's data influences it.

## Blast radius

- Peer tabs / other devices silently miss mutations that land in a gap. The user sees two tabs disagreeing with no error and no indication which one is right.
- With one browser tab open the client also holds a permanently churning connection: reconnect, ~10s of life, teardown, repeat. Each cycle is a fresh authenticated request and a fresh `ReadableStream` + `setInterval` allocation.
- `MAX_SUBSCRIBERS_PER_USER = 20` (`get-events.ts:16`) is a cap on *concurrent* subscribers, and `cleanup()` unregisters on abort, so this churn shouldn't hit the cap — but the churn is exactly the "runaway reconnect loop" the cap's comment was written to catch, so the cap now has a standing false-positive risk if reconnect ever outpaces cleanup.
- The failure is invisible in logs: each teardown is an ordinary `SSE unsubscribed` line, indistinguishable from a real tab close.

## Proposed fix sketch

1. Pass an explicit `idleTimeout` to `Bun.serve` in `src/server/start.ts` that comfortably exceeds `SSE_KEEPALIVE_MS` (Bun's max is 255 seconds). Note this is a **server-wide** setting, so pick it with the slowest legitimate request in mind, not just SSE.
2. Belt-and-braces: drop `SSE_KEEPALIVE_MS` well below whatever idle timeout is chosen, so the keepalive genuinely refreshes the socket rather than depending on the timeout being generous. A keepalive that can never fire before the timeout is the actual defect here, and pinning the invariant (`SSE_KEEPALIVE_MS < idleTimeout`) in one place — with a test asserting the relationship — keeps the two from drifting apart again.
3. Close the delivery gap so a reconnect can't lose a mutation. Cheapest option that fits the existing design: have the client run its normal `syncDomain` sweep (or a full `sync()`) on `EventSource` `open` after a reconnect, so the window between teardown and re-subscribe is reconciled rather than assumed empty. A `Last-Event-ID` + server-side ring buffer is the more complete answer if the feature is meant to be authoritative.
4. Regression coverage: assert that an `/api/events` stream is still open past the keepalive interval, and that a mutation emitted after a reconnect reaches the peer.

## Related

Same subsystem as #656 (real-time collaboration) and the three PRs that delivered it. Independent of #668 (`POST /api/account` never inserts), which was filed from the same session.

claoie:
Closes #669.

## Problem

`Bun.serve` applies a **10 second** default `idleTimeout` to every request. `/api/events` held a stream open with nothing flowing between events and a keepalive scheduled for **+30s**, so every SSE subscription was reaped at ~12s with the client still connected — the keepalive never got to fire once. Bun says it on stdout: `[Bun.serve]: request timed out after 10 seconds. Pass idleTimeout to configure.`

That alone would only be reconnect churn. It is data loss because two other pieces compose with it:

- `emitToUser` early-returns when a user has no live subscriber (`realtime.ts`) — no queue, no `id:` field, no replay.
- The client re-syncs only on the login transition (`Utility.tsx`) — no interval, no focus/visibility refetch.

So a mutation landing in a gap reached nobody and was never reconciled: peer tabs kept rendering stale data until a manual reload, with no error and no clue which tab was right. This is the "it just didn't propagate this time" failure mode of the real-time collaboration feature (#656 / #657 / #658 / #659).

## Fix

1. **Per-request deadline, not a server-wide one.** `start.ts` passes `server.timeout(request, seconds)` down to the handler as `req.setIdleTimeout`, and `get-events.ts` raises the deadline for its own request only. Widening `Bun.serve`'s `idleTimeout` would have relaxed it for every route — a stream that is *supposed* to go quiet is the exception, and only the exception should pay.
2. **The deadline is derived, not picked.** `SSE_IDLE_TIMEOUT_SECONDS = min(255, floor(SSE_KEEPALIVE_MS / 1000 × 3))` in `realtime.ts`. One knob (`SSE_KEEPALIVE_MS`, 30s); the socket deadline follows it at 90s, tolerating one missed tick, capped at Bun's 255s ceiling and floored to an integer (`server.timeout` throws otherwise). A keepalive can no longer be scheduled past the deadline, and the invariant is structural rather than a comment asking two files to agree. A unit test pins it.
3. **Reconnect with jittered backoff.** `useServerEvents.ts` stops relying on `EventSource`'s native retry, which gives up permanently on a terminal close. It reconnects itself with exponential backoff from 1s to a 30s cap, jittered ±25% so the tabs a deploy knocks off together don't all return in the same instant against a server that has just restarted. `reconnectDelayMs` is exported and unit-tested at the boundaries.
4. **Reconnects reconcile the gap.** A new `onReconnect` fires on every `open` after the first — not the initial one — wired in `Utility.tsx` to the whole-app `sync()`. A reconnect now assumes the gap had content instead of assuming it was empty. (A `Last-Event-ID` + server-side ring buffer is the more complete answer if the channel is ever meant to be authoritative; this is the cheap correct one for the current design, per the issue's sketch.)

## Test plan

- [x] `bun run lint` — clean
- [x] `bun test src` — 1155 pass / 2 skip / 0 fail on the rebased head
- [x] `bun run build` — clean

### Protocol-level, baseline vs fix

Same sandbox database, two servers: the PR build, and a second server built from `upstream/main` at `fc6b269`. Both were driven by the same probe — log in, `GET /api/events`, hold the connection and record every frame plus the moment the socket closes.

| | `upstream/main` | this PR |
|---|---|---|
| Outcome | **closed by server at 11.5s** | **still open at 45s** |
| Frames received | 1 (`: connected`) | 2 (`: connected`, then `: keepalive` at 30.0s) |

The baseline reproduces #669 exactly: the stream dies at 11.5s having delivered nothing but the opening comment, and the 30s keepalive never gets a chance to run. On the PR build the keepalive lands on schedule at 30.0s and the stream is still healthy when the probe lets go at 45s — past both the old ~12s reap and one full keepalive period.

### Reconnect behaviour, measured

A probe server that closes the SSE stream 3s after each open, driven by Chromium via Playwright, established what the browser actually does — the first revision of this PR was built on a wrong model of it:

- Server drops an established 200 stream → `error` fires with `readyState === CONNECTING`, and the browser re-requests on its own roughly every 3s, forever, with no backoff.
- Response is non-2xx (401 on an expired session) → `readyState === CLOSED` and the browser gives up permanently.

Replaying the shipped state machine against that same server, 30s window: 5 connects with the delay growing 747ms → 2399ms → 3482ms → 4323ms → 11911ms, and **one** whole-app `sync()` rather than one per reopen.

### Still pending

- [ ] **UI E2E, two tabs, one session** — running against this head
- [ ] **Reconnect reconciliation in the browser**

### DIFF
--- a/src/client/components/App/Utility.tsx
+++ b/src/client/components/App/Utility.tsx
@@ -38,11 +38,19 @@ const Utility = () => {
    * own `tabId` are its own writes (already applied optimistically) and
    * skipped. Per-domain debouncing collapses bursts of the same event; the
    * cursor is not advanced (only the whole-app `sync()` owns that).
+   *
+   * Events emitted while the stream was down are not replayed, so a
+   * reconnect reconciles with a whole-app `sync()` instead of assuming the
+   * gap was empty.
    */
-  useServerEvents((domain, payload) => {
-    if (payload.originTabId === tabId) return;
-    syncDomain(domain);
-  }, userLoggedIn);
+  useServerEvents(
+    (domain, payload) => {
+      if (payload.originTabId === tabId) return;
+      syncDomain(domain);
+    },
+    userLoggedIn,
+    sync,
+  );
 
   /**
    * Calculate balance history when data is updated
--- a/src/client/components/App/Utility.tsx
+++ b/src/client/components/App/Utility.tsx
@@ -38,10 +38,6 @@ const Utility = () => {
    * own `tabId` are its own writes (already applied optimistically) and
    * skipped. Per-domain debouncing collapses bursts of the same event; the
    * cursor is not advanced (only the whole-app `sync()` owns that).
-   *
-   * Events emitted while the stream was down are not replayed, so a
-   * reconnect reconciles with a whole-app `sync()` instead of assuming the
-   * gap was empty.
    */
   useServerEvents(
     (domain, payload) => {
--- a/src/client/lib/hooks/useServerEvents.test.ts
+++ b/src/client/lib/hooks/useServerEvents.test.ts
@@ -1,0 +1,29 @@
+import { describe, it, expect } from "bun:test";
+import { reconnectDelayMs } from "./useServerEvents";
+
+describe("reconnectDelayMs", () => {
+  it("backs off exponentially from one second", () => {
+    const mid = () => 0.5; // no jitter
+    expect(reconnectDelayMs(0, mid)).toBe(1_000);
+    expect(reconnectDelayMs(1, mid)).toBe(2_000);
+    expect(reconnectDelayMs(2, mid)).toBe(4_000);
+    expect(reconnectDelayMs(3, mid)).toBe(8_000);
+  });
+
+  it("caps the backoff so a long outage still retries", () => {
+    const mid = () => 0.5;
+    expect(reconnectDelayMs(10, mid)).toBe(30_000);
+    expect(reconnectDelayMs(100, mid)).toBe(30_000);
+  });
+
+  it("jitters within ±25% so tabs knocked off together do not return together", () => {
+    expect(reconnectDelayMs(0, () => 0)).toBe(750);
+    expect(reconnectDelayMs(0, () => 1)).toBe(1_250);
+    expect(reconnectDelayMs(10, () => 0)).toBe(22_500);
+    expect(reconnectDelayMs(10, () => 1)).toBe(37_500);
+  });
+
+  it("treats a negative attempt as the first one", () => {
+    expect(reconnectDelayMs(-1, () => 0.5)).toBe(1_000);
+  });
+});
--- a/src/client/lib/hooks/useServerEvents.test.ts
+++ b/src/client/lib/hooks/useServerEvents.test.ts
@@ -3,7 +3,7 @@ import { reconnectDelayMs } from "./useServerEvents";
 
 describe("reconnectDelayMs", () => {
   it("backs off exponentially from one second", () => {
-    const mid = () => 0.5; // no jitter
+    const mid = () => 0.5; // midpoint of the jitter band
     expect(reconnectDelayMs(0, mid)).toBe(1_000);
     expect(reconnectDelayMs(1, mid)).toBe(2_000);
     expect(reconnectDelayMs(2, mid)).toBe(4_000);
@@ -16,11 +16,21 @@ describe("reconnectDelayMs", () => {
     expect(reconnectDelayMs(100, mid)).toBe(30_000);
   });
 
-  it("jitters within ±25% so tabs knocked off together do not return together", () => {
-    expect(reconnectDelayMs(0, () => 0)).toBe(750);
-    expect(reconnectDelayMs(0, () => 1)).toBe(1_250);
-    expect(reconnectDelayMs(10, () => 0)).toBe(22_500);
-    expect(reconnectDelayMs(10, () => 1)).toBe(37_500);
+  it("spreads a herd across a full backoff period, not a fraction of one", () => {
+    // Every tab is knocked off by the same deploy. A band narrower than the
+    // delay itself returns them in a clump: ±25% of the first attempt spread
+    // the herd over 500ms, which is no spread at all against a server that
+    // has just started.
+    expect(reconnectDelayMs(0, () => 0)).toBe(500);
+    expect(reconnectDelayMs(0, () => 1)).toBe(1_500);
+    expect(reconnectDelayMs(10, () => 0)).toBe(15_000);
+    expect(reconnectDelayMs(10, () => 1)).toBe(45_000);
+  });
+
+  it("never returns a delay short enough to hammer the server", () => {
+    for (let attempt = 0; attempt <= 12; attempt++) {
+      expect(reconnectDelayMs(attempt, () => 0)).toBeGreaterThanOrEqual(500);
+    }
   });
 
   it("treats a negative attempt as the first one", () => {
--- a/src/client/lib/hooks/useServerEvents.ts
+++ b/src/client/lib/hooks/useServerEvents.ts
@@ -12,10 +12,23 @@ export type ServerEventHandler = (
   payload: ServerEventPayload,
 ) => void;
 
+const RECONNECT_BASE_MS = 1_000;
+const RECONNECT_MAX_MS = 30_000;
+
 /**
- * One tab-lifetime `EventSource` to `/api/events`. Every domain event
- * the server broadcasts to this user fires `handler(domain, payload)`.
- * `EventSource` handles reconnection natively; the hook does not.
+ * Backoff before the next reconnect attempt. Jittered because every tab of
+ * every user is knocked off by the same event (a deploy, a proxy blip) and
+ * would otherwise re-subscribe — and re-sync — in the same instant, against
+ * a server that has just started.
+ */
+export const reconnectDelayMs = (attempt: number, random: () => number = Math.random): number => {
+  const base = Math.min(RECONNECT_MAX_MS, RECONNECT_BASE_MS * 2 ** Math.max(0, attempt));
+  return Math.round(base * (0.75 + random() * 0.5));
+};
+
+/**
+ * One tab-lifetime subscription to `/api/events`. Every domain event the
+ * server broadcasts to this user fires `handler(domain, payload)`.
  *
  * The handler receives one `(domain, payload)` call per event. To split
  * dispatch across multiple concerns, switch on `domain` inside a single
@@ -24,16 +37,16 @@ export type ServerEventHandler = (
  *
  * `enabled` MUST be false until the user is authenticated. `/api/events`
  * 401s for an anonymous request, and per the EventSource spec a non-2xx
- * response is a fatal close — the browser does NOT auto-reconnect after
- * it. Opening the stream on mount (before login) would therefore leave
- * a permanently-dead connection that never recovers once the user logs
- * in. Gating on `enabled` opens the stream only once authenticated, and
- * the effect re-runs (reconnecting) when auth flips.
+ * response is a fatal close. Gating on `enabled` opens the stream only once
+ * authenticated, and the effect re-runs (reconnecting) when auth flips.
  *
- * `onReconnect` fires on every `open` after the first one on the same
- * stream. The server buffers nothing and replays nothing, so an event
- * emitted while the connection was down reaches no one — the peer must
- * reconcile by refetching once it is back.
+ * Reconnection is the hook's job, not the browser's: `EventSource` gives up
+ * for good (`readyState === CLOSED`) whenever the response is non-2xx or the
+ * server drops the socket — which is what a restart or a proxy blip looks
+ * like — so relying on its native retry leaves every tab silently deaf until
+ * a manual reload. Each re-established stream calls `onReconnect`, because
+ * the server buffers nothing and replays nothing: an event emitted while the
+ * stream was down reached no one, and only a refetch can close that gap.
  */
 export const useServerEvents = (
   handler: ServerEventHandler,
@@ -51,48 +64,52 @@ export const useServerEvents = (
   useEffect(() => {
     if (!enabled) return;
 
-    // withCredentials sends the session cookie; without it the SSE
-    // request is anonymous and the server 401s.
-    const source = new EventSource("/api/events", { withCredentials: true });
-
     const domains = Object.values(TableName);
+    let source: EventSource | null = null;
+    let retryTimeout: ReturnType<typeof setTimeout> | null = null;
+    let attempt = 0;
+    let everOpened = false;
+    let disposed = false;
 
-    const perDomain = new Map<ServerEventDomain, (e: MessageEvent) => void>();
-    for (const domain of domains) {
-      const listener = (e: MessageEvent) => {
-        let payload: ServerEventPayload = {};
-        try {
-          if (e.data) payload = JSON.parse(e.data) as ServerEventPayload;
-        } catch {
-          // malformed payload — dispatch with empty
-        }
-        handlerRef.current(domain, payload);
-      };
-      source.addEventListener(`${domain}-updated`, listener);
-      perDomain.set(domain, listener);
-    }
+    const connect = () => {
+      // withCredentials sends the session cookie; without it the SSE
+      // request is anonymous and the server 401s.
+      const es = new EventSource("/api/events", { withCredentials: true });
+      source = es;
 
-    let opened = false;
-    const onOpen = () => {
-      if (opened) onReconnectRef.current?.();
-      opened = true;
-    };
-    source.addEventListener("open", onOpen);
-
-    source.addEventListener("error", () => {
-      // EventSource auto-reconnects on error; log only when the connection
-      // stays broken (readyState === CLOSED means auto-reconnect gave up).
-      if (source.readyState === EventSource.CLOSED) {
-        console.warn("SSE connection closed and will not auto-reconnect.");
+      for (const domain of domains) {
+        es.addEventListener(`${domain}-updated`, (e: MessageEvent) => {
+          let payload: ServerEventPayload = {};
+          try {
+            if (e.data) payload = JSON.parse(e.data) as ServerEventPayload;
+          } catch {
+            // malformed payload — dispatch with empty
+          }
+          handlerRef.current(domain, payload);
+        });
       }
-    });
+
+      es.addEventListener("open", () => {
+        attempt = 0;
+        if (everOpened) onReconnectRef.current?.();
+        everOpened = true;
+      });
+
+      es.addEventListener("error", () => {
+        // A transient error the browser will itself retry leaves readyState
+        // at CONNECTING; only a terminal close is ours to recover from.
+        if (es.readyState !== EventSource.CLOSED || disposed) return;
+        es.close();
+        retryTimeout = setTimeout(connect, reconnectDelayMs(attempt++));
+      });
+    };
+
+    connect();
 
     return () => {
-      for (const [domain, listener] of perDomain) {
-        source.removeEventListener(`${domain}-updated`, listener);
-      }
-      source.removeEventListener("open", onOpen);
-      source.close();
+      disposed = true;
+      if (retryTimeout) clearTimeout(retryTimeout);
+      source?.close();
     };
   }, [enabled]);
 };
--- a/src/client/lib/hooks/useServerEvents.ts
+++ b/src/client/lib/hooks/useServerEvents.ts
@@ -16,14 +16,32 @@ const RECONNECT_BASE_MS = 1_000;
 const RECONNECT_MAX_MS = 30_000;
 
 /**
- * Backoff before the next reconnect attempt. Jittered because every tab of
- * every user is knocked off by the same event (a deploy, a proxy blip) and
- * would otherwise re-subscribe — and re-sync — in the same instant, against
- * a server that has just started.
+ * Consecutive failed attempts before the hook stops trying. A fatal close
+ * (non-2xx — an expired session is the common one) repeats forever otherwise,
+ * one request per tab per `RECONNECT_MAX_MS`, none of which can succeed until
+ * the user logs in again. The effect re-runs on that transition anyway.
+ */
+const RECONNECT_MAX_ATTEMPTS = 10;
+
+/** An open that survives this long counts as recovered and clears the backoff. */
+const STABLE_CONNECTION_MS = 60_000;
+
+/**
+ * Floor between two reconnect-driven `onReconnect` calls. `onReconnect` is a
+ * whole-app refetch — O(all data), not O(gap) — so a flapping stream must not
+ * be able to schedule one per flap.
+ */
+const RESYNC_MIN_INTERVAL_MS = 30_000;
+
+/**
+ * Backoff before the next reconnect attempt. Jittered over a full band because
+ * every tab of every user is knocked off by the same event (a deploy, a proxy
+ * blip) and would otherwise re-subscribe — and re-sync — in the same instant,
+ * against a server that has just started.
  */
 export const reconnectDelayMs = (attempt: number, random: () => number = Math.random): number => {
   const base = Math.min(RECONNECT_MAX_MS, RECONNECT_BASE_MS * 2 ** Math.max(0, attempt));
-  return Math.round(base * (0.75 + random() * 0.5));
+  return Math.round(base * (0.5 + random()));
 };
 
 /**
@@ -40,13 +58,24 @@ export const reconnectDelayMs = (attempt: number, random: () => number = Math.ra
  * response is a fatal close. Gating on `enabled` opens the stream only once
  * authenticated, and the effect re-runs (reconnecting) when auth flips.
  *
- * Reconnection is the hook's job, not the browser's: `EventSource` gives up
- * for good (`readyState === CLOSED`) whenever the response is non-2xx or the
- * server drops the socket — which is what a restart or a proxy blip looks
- * like — so relying on its native retry leaves every tab silently deaf until
- * a manual reload. Each re-established stream calls `onReconnect`, because
- * the server buffers nothing and replays nothing: an event emitted while the
- * stream was down reached no one, and only a refetch can close that gap.
+ * Reconnection is the hook's job because the browser's own retry has no
+ * backoff and no ceiling. The two failure shapes differ, and both are handled
+ * here rather than half here and half in the browser:
+ *
+ * - The server drops an established 200 stream (restart, proxy blip, a
+ *   reaped socket). `readyState` stays `CONNECTING` and the browser
+ *   re-requests on its own every few seconds, indefinitely — measured at
+ *   ~3s in Chromium, since the server sends no `retry:` field.
+ * - The response is non-2xx (an expired session 401s). `readyState` goes to
+ *   `CLOSED` and the browser gives up for good.
+ *
+ * So the error handler closes the stream in both cases and owns the retry,
+ * which is what puts the backoff on the first shape and a ceiling on the
+ * second. Each re-established stream calls `onReconnect`, because the server
+ * buffers nothing and replays nothing: an event emitted while the stream was
+ * down reached no one, and only a refetch can close that gap. That call is
+ * rate-limited — it is a whole-app refetch, and a flapping stream must not be
+ * able to schedule one per flap.
  */
 export const useServerEvents = (
   handler: ServerEventHandler,
@@ -67,8 +96,10 @@ export const useServerEvents = (
     const domains = Object.values(TableName);
     let source: EventSource | null = null;
     let retryTimeout: ReturnType<typeof setTimeout> | null = null;
+    let stableTimeout: ReturnType<typeof setTimeout> | null = null;
     let attempt = 0;
     let everOpened = false;
+    let lastResyncAt = 0;
     let disposed = false;
 
     const connect = () => {
@@ -90,16 +121,41 @@ export const useServerEvents = (
       }
 
       es.addEventListener("open", () => {
-        attempt = 0;
-        if (everOpened) onReconnectRef.current?.();
+        // Clearing the backoff here rather than on `open` itself: a stream
+        // that opens and immediately dies (a reconnect that trips the
+        // per-user subscriber cap and gets a 429) would otherwise restart
+        // the delay at ~1s every cycle instead of growing.
+        stableTimeout = setTimeout(() => {
+          attempt = 0;
+        }, STABLE_CONNECTION_MS);
+
+        if (everOpened) {
+          const now = Date.now();
+          if (now - lastResyncAt >= RESYNC_MIN_INTERVAL_MS) {
+            lastResyncAt = now;
+            onReconnectRef.current?.();
+          }
+        }
         everOpened = true;
       });
 
       es.addEventListener("error", () => {
-        // A transient error the browser will itself retry leaves readyState
-        // at CONNECTING; only a terminal close is ours to recover from.
-        if (es.readyState !== EventSource.CLOSED || disposed) return;
+        if (disposed) return;
+        if (stableTimeout) {
+          clearTimeout(stableTimeout);
+          stableTimeout = null;
+        }
+        // Close in both readyState cases — see the note on the hook. Leaving
+        // a CONNECTING stream to the browser would retry it with no backoff
+        // and no ceiling, once per few seconds, forever.
         es.close();
+
+        if (attempt >= RECONNECT_MAX_ATTEMPTS) {
+          console.warn(
+            `SSE stream failed ${RECONNECT_MAX_ATTEMPTS} times in a row; giving up until the next auth change or reload.`,
+          );
+          return;
+        }
         retryTimeout = setTimeout(connect, reconnectDelayMs(attempt++));
       });
     };
@@ -109,6 +165,7 @@ export const useServerEvents = (
     return () => {
       disposed = true;
       if (retryTimeout) clearTimeout(retryTimeout);
+      if (stableTimeout) clearTimeout(stableTimeout);
       source?.close();
     };
   }, [enabled]);
--- a/src/client/lib/hooks/useServerEvents.ts
+++ b/src/client/lib/hooks/useServerEvents.ts
@@ -29,12 +29,24 @@ export type ServerEventHandler = (
  * a permanently-dead connection that never recovers once the user logs
  * in. Gating on `enabled` opens the stream only once authenticated, and
  * the effect re-runs (reconnecting) when auth flips.
+ *
+ * `onReconnect` fires on every `open` after the first one on the same
+ * stream. The server buffers nothing and replays nothing, so an event
+ * emitted while the connection was down reaches no one — the peer must
+ * reconcile by refetching once it is back.
  */
-export const useServerEvents = (handler: ServerEventHandler, enabled = true): void => {
+export const useServerEvents = (
+  handler: ServerEventHandler,
+  enabled = true,
+  onReconnect?: () => void,
+): void => {
   // Keep the latest handler in a ref so we don't tear the EventSource
   // down every time the parent re-renders with a new closure.
   const handlerRef = useRef(handler);
   handlerRef.current = handler;
+
+  const onReconnectRef = useRef(onReconnect);
+  onReconnectRef.current = onReconnect;
 
   useEffect(() => {
     if (!enabled) return;
@@ -60,6 +72,13 @@ export const useServerEvents = (handler: ServerEventHandler, enabled = true): vo
       perDomain.set(domain, listener);
     }
 
+    let opened = false;
+    const onOpen = () => {
+      if (opened) onReconnectRef.current?.();
+      opened = true;
+    };
+    source.addEventListener("open", onOpen);
+
     source.addEventListener("error", () => {
       // EventSource auto-reconnects on error; log only when the connection
       // stays broken (readyState === CLOSED means auto-reconnect gave up).
@@ -72,6 +91,7 @@ export const useServerEvents = (handler: ServerEventHandler, enabled = true): vo
       for (const [domain, listener] of perDomain) {
         source.removeEventListener(`${domain}-updated`, listener);
       }
+      source.removeEventListener("open", onOpen);
       source.close();
     };
   }, [enabled]);
--- a/src/server/lib/realtime.test.ts
+++ b/src/server/lib/realtime.test.ts
@@ -7,6 +7,8 @@ import {
   emitToUser,
   mutationEmitDomain,
   closeAllSubscribers,
+  SSE_KEEPALIVE_MS,
+  SERVER_IDLE_TIMEOUT_SECONDS,
   type Subscriber,
 } from "./realtime";
 
@@ -173,5 +175,10 @@ describe("realtime", () => {
     expect(b.closed).toBe(true);
     expect(subscriberCount("u1")).toBe(0);
     expect(subscriberCount("u2")).toBe(0);
+  });
+
+  it("the server idle timeout outlives a keepalive tick and stays under Bun's ceiling", () => {
+    expect(SERVER_IDLE_TIMEOUT_SECONDS).toBeGreaterThan(SSE_KEEPALIVE_MS / 1000);
+    expect(SERVER_IDLE_TIMEOUT_SECONDS).toBeLessThanOrEqual(255);
   });
 });
--- a/src/server/lib/realtime.test.ts
+++ b/src/server/lib/realtime.test.ts
@@ -8,7 +8,7 @@ import {
   mutationEmitDomain,
   closeAllSubscribers,
   SSE_KEEPALIVE_MS,
-  SERVER_IDLE_TIMEOUT_SECONDS,
+  SSE_IDLE_TIMEOUT_SECONDS,
   type Subscriber,
 } from "./realtime";
 
@@ -177,8 +177,15 @@ describe("realtime", () => {
     expect(subscriberCount("u2")).toBe(0);
   });
 
-  it("the server idle timeout outlives a keepalive tick and stays under Bun's ceiling", () => {
-    expect(SERVER_IDLE_TIMEOUT_SECONDS).toBeGreaterThan(SSE_KEEPALIVE_MS / 1000);
-    expect(SERVER_IDLE_TIMEOUT_SECONDS).toBeLessThanOrEqual(255);
+  it("the SSE idle timeout outlives a keepalive tick", () => {
+    expect(SSE_IDLE_TIMEOUT_SECONDS).toBeGreaterThan(SSE_KEEPALIVE_MS / 1000);
+  });
+
+  it("the SSE idle timeout is an integer within Bun's accepted range", () => {
+    // `server.timeout` throws on a fractional or out-of-range value, and the
+    // throw lands inside the request handler that opens the stream.
+    expect(Number.isInteger(SSE_IDLE_TIMEOUT_SECONDS)).toBe(true);
+    expect(SSE_IDLE_TIMEOUT_SECONDS).toBeGreaterThan(0);
+    expect(SSE_IDLE_TIMEOUT_SECONDS).toBeLessThanOrEqual(255);
   });
 });
--- a/src/server/lib/realtime.test.ts
+++ b/src/server/lib/realtime.test.ts
@@ -182,8 +182,9 @@ describe("realtime", () => {
   });
 
   it("the SSE idle timeout is an integer within Bun's accepted range", () => {
-    // `server.timeout` throws on a fractional or out-of-range value, and the
-    // throw lands inside the request handler that opens the stream.
+    // `server.timeout` accepts a fractional or out-of-range value silently,
+    // so nothing at runtime would report a deadline that had drifted under
+    // the keepalive. This assertion is the only place that would.
     expect(Number.isInteger(SSE_IDLE_TIMEOUT_SECONDS)).toBe(true);
     expect(SSE_IDLE_TIMEOUT_SECONDS).toBeGreaterThan(0);
     expect(SSE_IDLE_TIMEOUT_SECONDS).toBeLessThanOrEqual(255);
--- a/src/server/lib/realtime.ts
+++ b/src/server/lib/realtime.ts
@@ -27,8 +27,12 @@ export const SSE_KEEPALIVE_MS = 30_000;
  * goes quiet between events, and it must outlive the keepalive period —
  * a tick scheduled past the deadline can never refresh the socket. Derived
  * rather than picked so the two cannot drift apart; 3x tolerates one missed
- * tick, 255 is Bun's ceiling, and the value must stay an integer or
- * `server.timeout` throws.
+ * tick, and 255 is Bun's ceiling.
+ *
+ * The clamp and the floor are guards, not error handling: `server.timeout`
+ * accepts a fractional or out-of-range value silently (measured on Bun
+ * 1.3.14 — 7.5 and 300 both return normally), so a value that drifts below
+ * the keepalive would reap the stream with nothing raised anywhere.
  */
 export const SSE_IDLE_TIMEOUT_SECONDS = Math.min(
   255,
--- a/src/server/lib/realtime.ts
+++ b/src/server/lib/realtime.ts
@@ -19,20 +19,21 @@ export interface Subscriber {
 }
 
 /** Gap between `: keepalive` blocks on an otherwise-quiet SSE stream. */
-export const SSE_KEEPALIVE_MS = 15_000;
+export const SSE_KEEPALIVE_MS = 30_000;
 
 /**
- * `idleTimeout` for `Bun.serve`. Bun's default is 10 seconds, which reaps
- * every SSE stream before its first keepalive tick can refresh the socket —
- * a keepalive longer than the timeout is dead code, and the connection
- * churns forever. Derived from `SSE_KEEPALIVE_MS` rather than picked
- * independently so the two cannot drift back apart; 255 is Bun's ceiling.
- *
- * Server-wide, not SSE-only: it also bounds how long any handler may go
- * without writing a byte, so raising it relaxes the deadline on slow
- * requests (Plaid/SimpleFIN sync) too.
+ * How long Bun may let an `/api/events` request idle before reaping it.
+ * Bun's default is 10 seconds, so this MUST be raised for a stream that
+ * goes quiet between events, and it must outlive the keepalive period —
+ * a tick scheduled past the deadline can never refresh the socket. Derived
+ * rather than picked so the two cannot drift apart; 3x tolerates one missed
+ * tick, 255 is Bun's ceiling, and the value must stay an integer or
+ * `server.timeout` throws.
  */
-export const SERVER_IDLE_TIMEOUT_SECONDS = Math.min(255, (SSE_KEEPALIVE_MS / 1000) * 4);
+export const SSE_IDLE_TIMEOUT_SECONDS = Math.min(
+  255,
+  Math.floor((SSE_KEEPALIVE_MS / 1000) * 3),
+);
 
 const subscribers = new Map<string, Set<Subscriber>>();
 
--- a/src/server/lib/realtime.ts
+++ b/src/server/lib/realtime.ts
@@ -18,6 +18,22 @@ export interface Subscriber {
   close: () => void;
 }
 
+/** Gap between `: keepalive` blocks on an otherwise-quiet SSE stream. */
+export const SSE_KEEPALIVE_MS = 15_000;
+
+/**
+ * `idleTimeout` for `Bun.serve`. Bun's default is 10 seconds, which reaps
+ * every SSE stream before its first keepalive tick can refresh the socket —
+ * a keepalive longer than the timeout is dead code, and the connection
+ * churns forever. Derived from `SSE_KEEPALIVE_MS` rather than picked
+ * independently so the two cannot drift back apart; 255 is Bun's ceiling.
+ *
+ * Server-wide, not SSE-only: it also bounds how long any handler may go
+ * without writing a byte, so raising it relaxes the deadline on slow
+ * requests (Plaid/SimpleFIN sync) too.
+ */
+export const SERVER_IDLE_TIMEOUT_SECONDS = Math.min(255, (SSE_KEEPALIVE_MS / 1000) * 4);
+
 const subscribers = new Map<string, Set<Subscriber>>();
 
 export type EmitDomain = TableName;
--- a/src/server/lib/route.ts
+++ b/src/server/lib/route.ts
@@ -36,6 +36,14 @@ export interface ServerRequest {
    * subscribe to it to release resources on tab close.
    */
   signal?: AbortSignal;
+  /**
+   * Widens this one request's idle deadline, in seconds (255 max — Bun's
+   * ceiling). Bun reaps any request that goes `idleTimeout` seconds without
+   * traffic, default 10; a route that holds a response open with long
+   * silences between writes (SSE) must raise its own deadline rather than
+   * the server relaxing it for every request.
+   */
+  setIdleTimeout?: (seconds: number) => void;
 }
 
 export interface ServerResponse {
--- a/src/server/lib/route.ts
+++ b/src/server/lib/route.ts
@@ -42,8 +42,12 @@ export interface ServerRequest {
    * traffic, default 10; a route that holds a response open with long
    * silences between writes (SSE) must raise its own deadline rather than
    * the server relaxing it for every request.
+   *
+   * Required, not optional: the symptom of a missing implementation is #669
+   * itself — the stream dies at ~12s with no error anywhere — so a caller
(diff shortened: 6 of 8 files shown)


reviewoie (commented):
## Code review — PR 674

**Verdict:** changes-requested (draft — none of these block the core mechanism, which I verified works)

**Findings:** 9 total — 0 high, 5 med, 4 low.

### Verified first (so the rest reads in context)

I re-ran the `Bun.serve` idle-timeout question empirically on Bun 1.3.14, fresh server per case, one probe each:

```
idleTimeout=3,  keepalive 1s   -> DIED at 4.1s
idleTimeout=3,  NO keepalive   -> DIED at 4.0s     (control)
idleTimeout=60, keepalive 15s  -> STILL OPEN at 135.0s   <- this PR's exact config
idleTimeout=60, NO keepalive   -> DIED at 60.0s     (control)
```

So: server-to-client writes **do** re-arm the timer, and the PR's config holds a stream open indefinitely. The premise is sound. (The `idleTimeout=3` row is Bun's ~4s timer granularity, not a counter-example — irrelevant at 60.) `bun test src/server` 718 pass / 2 skip / 0 fail, `tsc --noEmit` clean on the head commit.

Scope is clean: all 6 files map to the title.

### Med

- `src/server/lib/realtime.ts:35` — **the derived constant can crash the server at boot.** `Math.min(255, (SSE_KEEPALIVE_MS / 1000) * 4)` is `SSE_KEEPALIVE_MS / 250`, so any keepalive that isn't a multiple of 250ms yields a fraction, and `Bun.serve` rejects a non-integer `idleTimeout` outright. Verified on 1.3.14:
  ```
  idleTimeout=49.38 -> THROWS: Bun.serve expects idleTimeout to be an integer
  idleTimeout=60.5  -> THROWS: Bun.serve expects idleTimeout to be an integer
  ```
  Failure scenario: someone tunes `SSE_KEEPALIVE_MS = 12_345` (or any non-250-multiple) — the process throws on startup, no SSE degradation, just a dead server. The whole selling point of the derivation is "one knob you can turn safely," and it isn't. Fix is one call: `Math.floor(...)`, plus `expect(Number.isInteger(SERVER_IDLE_TIMEOUT_SECONDS)).toBe(true)` in the new test.

- `src/server/start.ts:401` — **the 6x relaxation is server-wide when Bun offers a per-request override.** `server.timeout(request, seconds)` exists and works; verified against a global `idleTimeout: 5`:
  ```
  no override            -> died at 8.1s
  server.timeout(req, 0) -> still open at 14.0s
  server.timeout(req,20) -> still open at 14.0s
  ```
  `get-events.ts` already reaches the raw `Request` (that's how `req.signal` is threaded), so the confined shape is available: keep the global deadline tight and relax only `/api/events`. As written, every socket — including a pre-auth one holding partial headers — may now idle 60s instead of 10s, and there is no per-IP connection cap in the app (`lib/rate-limit.ts` is login-specific). Not claiming an exploit; claiming the blast radius is 55-ish routes wide for a one-route problem, and the narrower option is one line.

- `src/client/components/App/Utility.tsx:52` — **reconnect fires a whole-app `sync()` with no jitter, no min-interval, and no in-flight guard.** `sync()` is not cheap: 6 parallel Stage-1 fetches + one institution fetch per item + 3 Stage-3 delta fetches + the unbounded Stage-4 history fetch + a full IDB write pass (`sync.ts:543-850`). Two scenarios:
  1. *Stampede.* The route sends no `retry:` field (grepped `get-events.ts` — none), so every tab uses the browser default and they all reconnect inside roughly the same window after a restart. N tabs x that fan-out, aimed at a server that just came up cold.
  2. *Amplification.* If the stream flaps (degraded proxy/server), every `open` -> `error` -> `open` cycle triggers another full sync. Before this PR an unstable stream was a silent, cheap reconnect; now instability turns into a repeated full-app refetch loop against the thing that's already struggling.
  
  `debounceSync` (`sync.ts:441-445`) only collapses calls within 50ms in the same tab and is not a mutex, so a reconnect landing during an in-flight sync starts a second concurrent `_sync`. Mitigation: jitter the reconnect sync, add a min-interval guard, and/or emit a per-connection-jittered `retry:` from the route.

- `src/client/lib/hooks/useServerEvents.ts:85` — **nothing re-establishes the stream after a fatal close, and nothing reconciles then either.** On `readyState === CLOSED` the hook only `console.warn`s. Per the EventSource spec (and this file's own JSDoc at line 25-31), a non-2xx response to a *reconnect* attempt is terminal — no further retries. Three in-repo paths return non-2xx to a reconnect: the 401 gate in `start.ts` (session expiry mid-session), `get-events.ts:34` `res.status(429)` at `MAX_SUBSCRIBERS_PER_USER = 20`, and a reverse-proxy 502/503 during a deploy window. In each, the tab goes permanently deaf **and** never re-syncs — the same unbounded-gap failure #669 describes, relocated from "every 12s" to "on the unlucky reconnect." The PR body's *"A reconnect now reconciles the gap instead of assuming it was empty"* holds only for reconnects that succeed; worth softening the claim or closing the path (fire the reconcile on CLOSED too, and schedule a manual re-`EventSource` with backoff).

- `src/server/routes/events/get-events.ts:108` — **the 10s reap was the de-facto GC on the subscriber map, and this PR removes it.** Streams are now effectively permanent, so any client-vanish path that doesn't fire the abort signal (laptop sleep, half-open TCP, carrier drop) leaves a `Subscriber` plus its `setInterval` registered for the process lifetime and counted against the per-user cap of 20 — after which reconnects 429 and hit the terminal-close path above. Previously the ~12s churn masked any such leak by construction. I did not reproduce a leak; I'm flagging it as a verification requirement, not an asserted bug. The E2E must assert `subscriberCount` returns to 0 after a *hard* client disappearance (kill the browser context / drop the network), not just after a graceful tab close — the `logger.info("SSE unsubscribed", { remaining })` line makes that observable.

### Low

- `src/server/lib/realtime.ts:25-28` — comment narrates the bug being fixed (*"Bun's default is 10 seconds, which reaps every SSE stream before its first keepalive tick ... the connection churns forever"*). Per the comment-hygiene rule, the bug story belongs in the PR body; the load-bearing part a future reader needs is the constraint (`idleTimeout` must exceed the keepalive period; 255 is Bun's ceiling). The second paragraph (server-wide blast radius) is a genuine why — keep that one.

- `src/client/components/App/Utility.tsx:42-44` — restates what `useServerEvents.ts:33-37` already says (no replay, so reconnect must reconcile). Pick one; the hook's JSDoc is the right home.

- `src/server/lib/realtime.test.ts:180-183` — the invariant test is close to tautological. `toBeLessThanOrEqual(255)` cannot fail under `Math.min(255, ...)`, and the first assertion only fires if `SSE_KEEPALIVE_MS >= 63.75s`. It doesn't cover the thing that actually breaks (integer-ness, see the first med) nor the thing that actually broke in production (`Bun.serve` receiving the value at all). Separately, `onReconnect` — half the fix — has zero coverage: `src/client/lib/hooks/` has only `router.test.ts` and `viewDate.test.ts`.

- `src/server/lib/realtime.ts:22` — on the 4x question you asked: the two knobs are the same dial, since the timeout is derived. 30s keepalive at 4x would mean a 120s server-wide deadline; halving to 15s is what keeps it at 60. That's coherent. Worth noting that keeping 30s and adding only `idleTimeout: 60` (2x) would have fixed #669 with no keepalive change and half the wakeups — 2x tolerates zero missed ticks, 3x tolerates one, 4x is belt-and-braces. Cost of the halving is small but real: a server-side `setInterval` at 4/min instead of 2/min per open stream, up to 20 streams per user, and a radio wake per tick on backgrounded mobile tabs.

### E2E this needs before ready (the plan currently has three unchecked boxes)

1. **Duration, not just "it connected."** Hold one stream past 90s and assert it never re-opened, plus zero `[Bun.serve]: request timed out` lines in server stdout. A 30s run passes on a build that still reaps at 60.
2. **Two clients, rendered DOM.** Tab A mutates at t=90s (i.e. after the old reap window), assert tab B's DOM changed — not "a refetch was issued." Watch `pageerror` on B.
3. **Reconnect reconciliation for real.** Restart the server mid-session, mutate while it's down, bring it up; assert B's rendered DOM reflects the mutation and that exactly one `sync()` fired, not one per domain.
4. **No spurious sync on login.** Fresh login should fire exactly one `sync()` (the `Utility` login effect) — the reconnect path must not add a second.
5. **Hard client disappearance** — see the fifth med.
6. **Terminal-close path** — force a 429 or a 502 on a reconnect and document the resulting state (my read: permanently deaf tab, no resync).

hoiekim:
Continue

reviewoie:
## Code review — PR 674

**Findings:** 9 total — 1 high, 4 med, 4 low.

Scope is clean: all 8 files map to the title (`start.ts`/`route.ts` are the plumbing the per-request deadline needs). No unrelated changes.

Two things I verified empirically before writing this, both on Bun 1.3.14 (this repo's runtime; `Dockerfile` pins `oven/bun:1`) with a throwaway `Bun.serve` + Chromium via Playwright:

- **The server half works.** A quiet stream with no `server.timeout` call died at **12.1s**; the same stream with `server.timeout(req, 25)` died at **28.1s**. And writes reset the timer — a stream writing every 3s stayed open past 25s under the *default* 10s timeout. So 30s keepalive under a 90s per-request deadline holds. Mechanism confirmed.
- **`idleTimeout` is no longer raised server-wide at head**, and `upstream/main` never set one either, so every non-SSE route keeps Bun's 10s default byte-for-byte. There is no new slowloris/held-connection surface. (The PR body still claims otherwise — see MED 3.)

### High

- `src/client/lib/hooks/useServerEvents.ts:99` — `onReconnect` (wired to the whole-app `sync()`) fires on **every** browser-native reconnect, unthrottled, and the hook's exponential backoff is structurally unreachable for the failure mode this PR is about.

  The error handler gates on `es.readyState !== EventSource.CLOSED → return`. I measured what actually happens when a server drops a 200 stream: Chromium fires `error` with **`readyState === 0` (CONNECTING)**, retries natively ~3s later, and fires `open` again on the *same* `EventSource` — so the `open` listener runs, `everOpened` is already true, and `onReconnectRef.current()` fires. Probe log (server closes the stream 3s after each open):

  ```
  ...476.8 open  rs=1
  ...479.8 error rs=0     <- CONNECTING, not CLOSED: hook returns, backoff never arms
  ...483.0 open  rs=1     <- native retry, ~3.2s later -> onReconnect() -> sync()
  ...486.0 error rs=0
  ...489.1 open  rs=1     <- sync() again
  ```

  Failure scenario: any condition that repeatedly ends a **200** stream — an intermediary whose read timeout is under the 30s keepalive, a restarting/flapping server, or a future regression of the deadline wiring (LOW 3) — makes the app fire a full `sync()` roughly every 3 seconds per tab, indefinitely. `sync()` is not cheap: it is the multi-endpoint delta fetch + IDB write + a `calculate(data)` recompute via `Utility.tsx`. `get-events.ts` sends no `retry:` field, so the cadence is the browser default and the client has no say in it. Multiply by tabs and users and this is the scalability-lens problem in critical rule 13.

  Concrete fixes: (a) throttle `onReconnect` — refuse to re-sync more often than once per N seconds, or only when the previous stream had been open longer than a trivial duration; (b) emit a `retry:` field from `get-events.ts` so the browser's cadence is yours to set; (c) move the `attempt = 0` reset out of `open` (see LOW 2).

### Med

- `src/client/lib/hooks/useServerEvents.ts:44` — the JSDoc states a false fact about `EventSource`, and it is the justification for the whole 45-line rewrite: *"`EventSource` gives up for good (`readyState === CLOSED`) whenever the response is non-2xx **or the server drops the socket** — which is what a restart or a proxy blip looks like."* The non-2xx half is right (I confirmed a 401 → `readyState === 2`, one request, no retry). The socket-drop half is wrong: the log above shows `rs=0` and a native retry. I also ran the exact "restart" case the sentence names — open a stream, SIGKILL the server, wait 8s, restart it — and the tab reconnected **by itself**:

  ```
  541.1 open  rs=1
  543.1 error rs=0   <- server killed
  546.3 error rs=0   <- retrying (connection refused)
  549.4 error rs=0
  552.5 open  rs=1   <- server back; reconnected with no help from the hook
  ```

  The rewrite still earns its place — it is what recovers the genuinely-fatal 401/429/proxy-error-page cases — but the doc and the commit message ("every tab went permanently deaf after any restart") both assert something the runtime does not do, and that same wrong model is what leaves HIGH 1 unnoticed. Same failure shape as the `/client-error` "unauthenticated by design" docstring that got laundered into PR 664.

- `src/server/lib/realtime.ts:30` (and the comment at `src/server/lib/realtime.test.ts:186`) — *"the value must stay an integer or `server.timeout` throws"* / *"`server.timeout` throws on a fractional or out-of-range value, and the throw lands inside the request handler"*. Not true on Bun 1.3.14: `server.timeout(req, 7.5)` returned normally and the stream was reaped at **8.1s** — silently truncated, no throw. `server.timeout(req, 300)` also returned normally. Keep `Math.floor` (it is still the right call), but state the real hazard: a fractional value is silently truncated **downward**, which can put the deadline under the keepalive without any error. The test is fine; its comment needs the same correction.

- PR body — the "Fix" section describes the superseded first commit; all four bullets are false at this head. It says start.ts passes an explicit `idleTimeout` to `Bun.serve` (commit 2 removed that and went per-request), that the constant is `SERVER_IDLE_TIMEOUT_SECONDS = min(255, keepalive × 4)` (it is `SSE_IDLE_TIMEOUT_SECONDS = min(255, floor(keepalive × 3))`), that the keepalive drops 30s → 15s (it stayed 30s), and that the change relaxes the deadline server-wide for slow Plaid/SimpleFIN legs "noted in the constant's comment" (it does not, and that pre-existing 10s exposure for slow handlers is unchanged from main). The stale body already misled a reader — the review request I was handed was framed on "× 4" and "server-wide". Worth a rewrite before this leaves draft, including whether the slow-sync-route exposure the body raised should be filed separately.

- `src/client/lib/hooks/useServerEvents.ts:105` — no terminal condition on a fatal close. Session expiry gives a 401, which I confirmed lands in `CLOSED` with no native retry, so the hook now re-requests `/api/events` forever at the 30s cap, per tab — each attempt writing an "API request … 401" line into the server log — where before this PR the stream died once with a `console.warn`. `enabled` stays `true` because the client-side `user` object is still there. Consider bailing after N consecutive fatal closes, or surfacing repeated fatal closes as "session gone" so the app routes to login.

### Low

- `src/client/lib/hooks/useServerEvents.ts:26` — the jitter band is ±25% **of the base**, so the first post-outage reconnect is spread across only ~500 ms (750–1250 ms). The comment says the jitter stops tabs "knocked off together" from returning together; at attempt 0 — which is the deploy case it names, once the server is back — half a second of spread is not meaningfully different from none. Full jitter (`random() * base`) or a larger first-attempt base would actually spread it.

- `src/client/lib/hooks/useServerEvents.ts:100` — `attempt = 0` sits in the `open` handler, so any successful open resets the backoff. In an open-then-fatally-close cycle (a reconnect that trips the 20-subscriber cap and gets a 429) the delay restarts from ~1s every cycle instead of growing. Reset on a connection that has been *stable* for some seconds, not on `open`.

- `src/server/lib/route.ts:46` / `src/server/routes/events/get-events.ts:35` — `setIdleTimeout?:` optional plus `req.setIdleTimeout?.(…)` optional-chained means that if the `start.ts` wiring is ever dropped, the deadline silently reverts to Bun's 10s and the exact bug in 669 comes back with no error. It matches the existing `signal?` precedent so I would not block on it, but there is no `get-events.test.ts` at all — one test asserting the route calls `setIdleTimeout` with a value greater than `SSE_KEEPALIVE_MS / 1000` would close the loop that the two constant tests only half close.

- `src/server/lib/realtime.test.ts:180` — the guard asserts `idle > keepalive`, but the doc claims "3x tolerates one missed tick". Because of the `min(255, …)` clamp that property silently degrades above an 85s keepalive (keepalive 200s → idle 255, i.e. 1.27x, no missed-tick tolerance) while the test still passes. Either assert `idle >= 2 * keepalive` or drop the missed-tick claim from the comment.

### Verification the test plan should still cover

This is still draft with three boxes open, so as edge cases for the sweep — the two-client rule from the 656/657/658/659 arc applies here, and the flapping case is the HIGH 1 repro:

1. Two tabs, one session: mutate via a real affordance in A, assert B's **rendered DOM** changed (not that B issued a refetch), with a `pageerror` watch on B.
2. Same two tabs held **past 90s idle** with no traffic, then mutate — proves the keepalive/deadline pairing end to end, which is the actual subject of the PR.
3. Flapping stream: an intermediary (or a patched route) that ends the stream every ~5s → count `/api/transactions` requests over 60s. Expected today: roughly one per 3s, per tab.
4. Server SIGKILL for ~20s with a tab open → assert exactly **one** `sync()` on recovery, not a burst.
5. Session expiry with a tab open → confirm the retry loop terminates or degrades, rather than polling a 401 forever.
6. 21 tabs → the tab that gets the 429 backs off instead of hot-looping.

claoie:
The HIGH is correct and it invalidates the client half of this PR as written. Fixed in `bd86a7d`. I reproduced it independently before changing anything, because it contradicted the model the whole rewrite was built on.

**HIGH — `onReconnect` unthrottled, backoff unreachable.** My Chromium probe against a server that closes the stream 3s after each open:

```
open rs=1 / error rs=0 / open rs=1 / error rs=0 / open rs=1
SSE_REQUEST #1 …18.691Z   #2 …24.815Z   #3 …30.968Z
```

`rs=0` is CONNECTING, not CLOSED — so `if (es.readyState !== EventSource.CLOSED) return` returned every time, the hook's backoff never armed, and the browser re-requested on its own every ~6s (its ~3s retry on top of the 3s stream life, no `retry:` field sent). Every one of those opens fired `onReconnect` → whole-app `sync()`: multi-endpoint delta fetch, IDB write, `calculate(data)` recompute. Per tab, indefinitely.

The fix closes the stream and owns the retry in **both** readyState cases, which is what puts a backoff on the dropped-stream shape and a ceiling on the fatal one. Same flapping server, replaying the new state machine with the shipped `reconnectDelayMs`:

```
[0.0s] first open
[3.0s] dropped -> backoff 747ms      [3.8s] onReconnect -> whole-app sync() (#1)
[6.8s] dropped -> backoff 2399ms     [9.2s] reopened, resync SUPPRESSED
[12.2s] dropped -> backoff 3482ms    [15.7s] reopened, resync SUPPRESSED
[18.7s] dropped -> backoff 4323ms    [23.0s] reopened, resync SUPPRESSED
[26.0s] dropped -> backoff 11911ms
--- 30s window: 5 connects, 1 whole-app sync ---
```

One sync instead of one per reopen, and the interval is growing rather than fixed.

**MED — the JSDoc asserting CLOSED on a server drop.** That false claim is exactly what hid the HIGH: I wrote the guard to match the doc, and the doc described behavior the browser doesn't have. Rewritten to name both shapes and what each one does, with the measured ~3s retry in it so the next reader doesn't have to re-derive it.

**MED — `server.timeout` "throws".** Also false, and I checked it rather than taking it: on Bun 1.3.14 (the repo runtime) `server.timeout(req, 7.5)`, `(req, 300)` and `(req, -1)` all return normally. `Math.floor` and the `min(255, …)` clamp are still right, but they guard against a **silently** bad deadline, not an exception — which is worse, since a value drifting under the keepalive would reap the stream with nothing raised anywhere. Both comments corrected to say that.

**MED — stale PR body.** Every "Fix" bullet described the superseded first commit. Rewritten against the head, including the point I'd gotten backwards in the review request: this uses a **per-request** `server.timeout`, not a server-wide `Bun.serve` idleTimeout, so non-SSE routes keep the 10s default. Thanks for checking `upstream/main` sets none either — that ruled out the "slow Plaid legs were under the same axe" claim I'd carried over.

**MED — no terminal condition on a fatal close.** Capped at 10 consecutive failed attempts with a `console.warn`. An expired session cannot be fixed by retrying, and the effect re-runs on the auth transition anyway.

**LOW — jitter band too narrow.** Widened from ±25% to a full band (`0.5 + random()`), so attempt 0 spreads over 1s rather than 500ms and the cap spreads over 30s. Test rewritten to assert the band and to pin a floor at every attempt so no future change can produce a hammering delay.

**LOW — `attempt = 0` on any open.** Moved behind a 60s stability timer, so open-then-die cycles (a reconnect tripping `MAX_SUBSCRIBERS_PER_USER` and taking a 429) grow the backoff instead of resetting it.

**LOW — optional `setIdleTimeout` silently reverts to 10s.** Made required on `ServerRequest` and the call unconditional. The symptom of a missing implementation is #669 itself, so a caller building a `ServerRequest` now has to say what it does about it. Typecheck is the guard; `validation.test.ts` casts through `as unknown as ServerRequest` and is unaffected.

**LOW — the 3x guard degrades above an 85s keepalive.** Left as-is. Reaching it needs someone to raise `SSE_KEEPALIVE_MS` past 85s, at which point the clamp is doing its job; the invariant the test protects (idle > keepalive) still inverts only above 255s, which it does catch. Say the word if you'd rather the test assert the 3x ratio directly.

**Verification at `bd86a7d`:** `bun test src` 1156 pass / 2 skip / 0 fail, `tsc --noEmit` clean, `eslint src` clean. Staying draft — the two-tab UI sweep is still running, and this commit changed the client half, so it re-runs against this head.

claoie:
## E2E sweep — server-side fix verified, reconnect sections re-running

Full UI-driven sweep on the sandbox at 390×844, two independent browser contexts sharing one session, real click chain from login (`/budgets` → bottom-nav Accounts → click account row → detail → edit the Name field), asserting on tab B's **rendered DOM** rather than an `/api/…` re-fetch. `pageerror` and console errors watched on both tabs throughout. One field mutated, restored through the same UI.

### Live propagation — pass

Tab A renamed the top account; tab B's rendered row changed with no reload and no interaction on tab B, **224 ms** after the write. Rows below unchanged; donut, sparklines, balances and deltas byte-identical between the before/after screenshots.

### Propagation after idle — pass, and this is the regression

Both tabs idle **48 s**, then a rename propagated in **282 ms**. Extended to a **139 s** idle — past `SSE_IDLE_TIMEOUT_SECONDS` (90) and three keepalive ticks — with **zero `/api/events` network activity during the idle**, and it still propagated, in 4 ms.

That last detail is the one that matters: no reconnect happened, so the keepalive writes are genuinely refreshing Bun's per-request deadline rather than the 90 s value merely deferring a single reap. Server log corroborates — `grep -c "request timed out after 10 seconds"` → **0**, and one continuous subscribe→unsubscribe pair per tab across the whole session.

### Reconnect sections — re-running against `bd86a7d`

The sweep ran against `24748ad`, before the client rewrite, so its reconnect results describe code that no longer exists. Sandbox has been rebuilt at `bd86a7d` and those sections are re-running.

Worth recording that the old-build measurement independently reproduced the HIGH: on a network-level drop the sweep observed *"six `ERR_CONNECTION_REFUSED` retries at ~3 s intervals … `readyState` stays CONNECTING so the hook's own backoff correctly does not engage"* — a fixed browser-native cadence with the backoff bypassed, and **22 GETs across 9 domains** on each recovery. That is the storm, described from the outside by a run that had no reason to expect it.

The terminal-500 path did exercise the hook's backoff on the old build (gaps of 1.35 / 2.0 / 4.75 / 9.1 / 15.9 s), and gap reconciliation worked: a mutation made while tab B's stream was severed stayed invisible during the gap — the negative assertion held, confirming the server buffers and replays nothing — then rendered after reconnect.

### Two findings

**The 401-loops-forever LOW is already fixed** in `bd86a7d` (10-attempt ceiling with a `console.warn`), which landed before the sweep reported it. The sweep's sharper suggestion — treat 401/429 as terminal while keeping the backoff for 5xx and network drops — I'd like to but can't cleanly: `EventSource`'s `error` event carries no HTTP status, so the hook cannot tell 401 from a 502 during a deploy, which does deserve retries. `readyState === CLOSED` is all it gets. The attempt ceiling is the practical version of the same idea. Also confirmed there is no lockout risk — `isLoginRateLimited` keys only on `POST /login`.

**Service worker registration fails on every page load** — pre-existing, outside this diff, filed as #675. Built `/sw.js` starts `import{h as l}from"./assets/index-…js"` because `sw.ts` imports `isApiPath`, but `start.tsx:30` registers it without `{ type: "module" }`, so evaluation fails and the offline shell has never run. Flagging here only so it isn't attributed to this PR.

### Lifecycle

Logout via the real chain (hamburger → Configuration → Logout → confirm) left **zero** `/api/events` retries — no orphan retry loop from the effect cleanup. Re-login opened exactly one stream and issued exactly one `/api/accounts` GET, i.e. the login's own `sync()`, confirming `onReconnect` does not double-fire on a first open. Subscribe/unsubscribe pairs balanced throughout, concurrency never above 2. This section is also re-running, since the cleanup path changed.

reviewoie:
## Code review — PR 674

**Verdict:** changes-requested (PR is still a **draft** — flagging now so the remaining test-plan boxes can be driven against a fixed head rather than re-run twice)

**Findings:** 6 total — 2 high, 2 med, 2 low.

**Prior round, re-checked at `bd86a7d`:** all 9 findings from the 2026-08-03T09:36Z review are addressed. Confirmed at head: the error handler no longer branches on `readyState` and closes in both shapes (`useServerEvents.ts:142-160`); the JSDoc now names both failure shapes correctly (`:65-70`); the `server.timeout`-throws claim is gone from `realtime.ts:32-35` and `realtime.test.ts:186-188`; jitter widened to a full band (`:44`); `attempt = 0` moved behind a 60s stability timer (`:128-130`); `setIdleTimeout` is required on `ServerRequest` (`route.ts:50`) with an unconditional call (`get-events.ts:35`); PR body rewritten against the per-request design. `bun run typecheck` clean, `bun test --preload ./scripts/test-preload.ts src` → 1156 pass / 2 skip / 0 fail at this head.

**Idle-timeout pairing verified:** `SSE_KEEPALIVE_MS = 30_000` (`realtime.ts:22`), `SSE_IDLE_TIMEOUT_SECONDS = min(255, floor(30 × 3)) = 90` (`:37-40`), applied per-request via `req.setIdleTimeout(90)` (`get-events.ts:35`) → `server.timeout(request, 90)` (`start.ts:422-424`). Keepalive 30s is strictly inside the 90s deadline. `Bun.serve` still sets no server-wide `idleTimeout`, so every other route keeps the 10s default. That half of the PR is correct.

Scope is clean: all 8 files map to the title.

The two HIGHs are both in the client state machine `bd86a7d` introduced, and both re-open the exact failure 669 describes — a mutation that reaches nobody and is never reconciled.

### High

- `src/client/lib/hooks/useServerEvents.ts:132-138` — the resync throttle **drops** the gap-reconciling sync instead of deferring it, so a gap that closes within 30s of the previous one is never reconciled at all.

  ```js
  if (everOpened) {
    const now = Date.now();
    if (now - lastResyncAt >= RESYNC_MIN_INTERVAL_MS) {   // 30_000
      lastResyncAt = now;
      onReconnectRef.current?.();
    }
  }
  ```
  There is no trailing edge — the suppressed call is discarded, not rescheduled.

  Failure scenario: the flapping stream this PR's own backoff is built for. Your measured replay in the 09:47Z comment is the repro verbatim — `[3.8s] sync (#1)`, then `[9.2s] SUPPRESSED`, `[15.7s] SUPPRESSED`, `[23.0s] SUPPRESSED`. Each of those suppressions covers a real gap (6.8→9.2s, 12.2→15.7s, 18.7→23.0s) during which `emitToUser` early-returned on an empty subscriber set (`realtime.ts:120-121` — still no queue, no `id:`, no replay). A peer tab's mutation landing in any of them is lost permanently: once the stream settles, no further `open` fires, so no later resync ever picks it up. The tab renders stale data with no error — the same end state as 669, reached by a different road. Note it is strictly worse than "one sync per flap": the *last* gap before the stream stabilises is the most likely to be suppressed, and it is the one whose content the user is about to look at.

  Fix: make it a trailing-edge throttle — on suppression, set a pending flag and `setTimeout` the resync to `lastResyncAt + RESYNC_MIN_INTERVAL_MS` (collapsing repeats into the one pending call), and clear that timer in the effect cleanup alongside `retryTimeout`/`stableTimeout`.

- `src/client/lib/hooks/useServerEvents.ts:153-158` — after `RECONNECT_MAX_ATTEMPTS` the hook stops permanently, and nothing else in the client ever re-subscribes or re-syncs. This is a regression against `upstream/main` for the transport failure shape.

  ```js
  if (attempt >= RECONNECT_MAX_ATTEMPTS) {   // 10
    console.warn(`SSE stream failed ${RECONNECT_MAX_ATTEMPTS} times in a row; giving up ...`);
    return;
  }
  ```
  Delays for attempts 0-9 are 1,2,4,8,16,30,30,30,30,30 s × the [0.5,1.5) jitter band, so the give-up horizon is **~1.5 to ~4.5 minutes** of continuous failure. The effect's only dependency is `[enabled]` (`:171`), i.e. `userLoggedIn` — which does not flip during an outage.

  Failure scenario: any outage longer than that horizon — a deploy whose container start plus migration exceeds 2 minutes, a rolling restart, a laptop off wifi for 5 minutes, a proxy that drops the stream faster than `STABLE_CONNECTION_MS = 60_000` can reset `attempt` (a 55s read timeout is enough: 10 flaps ≈ 10 minutes and the tab is dead). After that the tab has no live subscription and no reconciliation for the rest of its life. Verified there is no other recovery path: `grep -rn -E "visibilitychange|addEventListener\(.(focus|online)|navigator\.onLine|setInterval" src/client` returns only `DatePickerModal`'s unrelated `ViewDate.setInterval`, and the only `sync()` call sites are the login transition (`Utility.tsx:30`) and the manual Refresh button (`Configuration/index.tsx:84`). Signal to the user is a `console.warn`.

  On `upstream/main` this case self-healed: the browser retried the dropped-200 shape indefinitely (your own probe in the 09:47Z comment — `error rs=0` ×3 while the server was down, then `open rs=1` once it came back). The ceiling is justified in the comment at `:18-23` only for the **fatal** shape ("none of which can succeed until the user logs in again") — that reasoning does not transfer to a server that is merely down. The two shapes are distinguishable at `error` time by `es.readyState` *before* the `es.close()` on `:151` (CONNECTING = transport, CLOSED = non-2xx), and the code reads neither.

  Fix: cap only the fatal shape; keep retrying the transport shape at the 30s cap forever. If the cap stays, at minimum add an `online`/`visibilitychange` listener that resets `attempt` and reconnects, and fire one `onReconnect()` when the stream is finally re-established after a give-up.

### Med

- `src/server/routes/events/get-events.ts:35` — nothing pins that this route calls `setIdleTimeout`. Making the field required on `ServerRequest` guarantees the *plumbing* exists, not that the SSE route uses it. Delete line 35 and `bun run typecheck` stays clean and all 1158 tests still pass — 669 is back, silently, with the stream dying at ~12s and no error anywhere. There is no `get-events.test.ts` at all (`ls src/server/routes/events/` → `get-events.ts`, `index.ts`). One test executing the route with a stub `setIdleTimeout` and asserting it was called with a value `> SSE_KEEPALIVE_MS / 1000` closes the loop the two constant tests in `realtime.test.ts:180-189` only half close. The `execute(makeReq(), fakeRes())` pattern in `src/server/routes/api-keys/get-api-keys.test.ts` is the shape to copy.

- `src/client/lib/hooks/useServerEvents.test.ts` — the reconnect state machine has zero coverage; the file tests only the pure `reconnectDelayMs`. Verified by grep: the five `it(` blocks are "backs off exponentially from one second", "caps the backoff so a long outage still retries", "spreads a herd across a full backoff period, not a fraction of one", "never returns a delay short enough to hammer the server", "treats a negative attempt as the first one". Everything `bd86a7d` added on top of the delay function — the give-up ceiling, the 60s stability reset, `everOpened`, and the resync throttle — is untested, and that is exactly where both HIGHs live. A fake `EventSource` on `globalThis` plus fake timers would cover all four; alternatively lift the state machine out of the `useEffect` into an exported factory the way `reconnectDelayMs` was lifted, and test that.

### Low

- `src/client/lib/hooks/useServerEvents.ts:124-127` — the comment justifying the stability timer cites a scenario that cannot happen, and describes its own location backwards. It reads "Clearing the backoff here rather than on `open` itself" while sitting *inside* the `open` handler (intent: "after a stable interval rather than immediately"), and the example it gives — "a reconnect that trips the per-user subscriber cap and gets a 429" — never reaches this code: a 429 is non-2xx, so `EventSource` fires `error` with `readyState === CLOSED` and **never fires `open`**, so `attempt` was never at risk of resetting in that case. The change is still right (an accept-then-drop proxy is the real case); the stated reason is not. Same class as the JSDoc that hid last round's HIGH.

- `src/client/lib/hooks/useServerEvents.ts:16` + PR body, Fix bullet 3 — `RECONNECT_MAX_MS = 30_000` is not the maximum delay. The full jitter band on `:44` (`base * (0.5 + random())`) returns up to **45_000**, which the test at `useServerEvents.test.ts:26` pins (`reconnectDelayMs(10, () => 1)` → `45_000`). The PR body still says "backoff from 1s to a 30s cap, jittered ±25%" — both halves stale since `bd86a7d` widened the band — and the body's Fix section never mentions the give-up ceiling or the resync throttle it added. Rename to `RECONNECT_MAX_BASE_MS` (or clamp after jitter) and refresh the body before this leaves draft. Related residual on the herd claim: at attempt 0 the band is only 500-1500 ms, so a sub-second server blip returns every tab of every user inside 1.5s and each one fires a whole-app `sync()` (full fetch of the eight non-partitioned domains + delta + `calculate(data)`). Jitter on the reconnect does not spread the resync when the outage is short; staggering `onReconnect` itself, or a larger attempt-0 base, would.

### Verification the remaining test-plan boxes should cover

Both open boxes are the two-client rule from the 656/657/658/659 arc. Edge cases beyond the happy path, in priority order — 1 and 2 are the HIGH repros:

1. **Suppressed-gap loss (HIGH 1).** Two tabs. Flap A's stream twice inside 30s (a patched route that ends the stream after ~5s works), and mutate from tab B *during the second gap*. Assert A's **rendered DOM** shows the mutation. Today it will not, and no later event or sync will fix it.
2. **Give-up (HIGH 2).** One tab, SIGKILL the server for 5 minutes, restart. Assert the tab re-subscribes and its rendered DOM reflects a mutation made while it was down. Today it stays deaf until reload.
3. **Past the deadline.** Two tabs held idle for **>90s** with no traffic, then mutate in A → assert B's DOM updates. This is the only check that proves the keepalive actually refreshes the per-request deadline across more than one tick; the protocol probe in the body stopped at 45s, i.e. before the 90s deadline was ever at risk.
4. **Exactly one sync on recovery.** SIGKILL ~20s, restart, count `/api/transactions` requests from the tab — expect one whole-app sync, not a burst.
5. **Session expiry.** Let the session lapse with a tab open → confirm the loop terminates (it should now, at 10 attempts) rather than polling a 401 forever, and that the give-up is visible somewhere other than the console.
6. **21 tabs.** The tab that takes the 429 must back off; confirm it does not hot-loop and that the server-side subscriber set drains (`unregisterSubscriber` via the abort signal) as the client's new close-on-every-error path churns streams.

Each of these needs a `page.on("pageerror")` watch on the receiving tab and an assertion on rendered DOM, not on the refetch being issued.

claoie:
Round 3 addressed in `cec8046e`. The MED was a real hole and it was mine — I fixed HIGH 2 by moving the ceiling onto the fatal shape and then never asked whether "fatal" actually means "permanent". It doesn't.

**MED — a transient non-2xx permanently deafens the tab.** Confirmed the mechanism before changing anything: `start.ts` does `await sessionStore.get(sid).catch(() => null)`, so a store blip yields an anonymous session and `/api/events` 401s; that lands in `CLOSED`, my code marked it fatal, and ten of them inside ~76–227s stopped the tab for good. The store recovers, every other request goes back to 200, and the app renders normally while never updating again — #669's end state with a `console.warn` as the only signal. Your two siblings hold as well: the 429 from the subscriber cap clears when a tab closes, and a 500 from `route.ts` is transient by definition.

The fix is to stop stopping. After ten consecutive refusals the hook drops to a ~5 minute jittered cadence instead of returning — one request per tab per few minutes is nothing, and it removes the entire permanently-deaf class rather than the one instance of it. A genuinely expired session is now asked about twelve times an hour instead of a hundred, which was the original point of the ceiling. The warning fires once per exhaustion, not once per cycle.

You were also right that "consecutive" was overstated: `fatalAttempts` only reset in the `open` handler, so an alternating refused/dropped sequence accumulated toward a ceiling it never hit ten times in a row. A non-fatal close now clears it — a dropped stream is evidence the server is reachable and answering.

While pinning that I found a related leak of my own: `retryTimeout` was reassigned without clearing the previous timer. One `EventSource` fires at most one `error` once closed, so it isn't reachable today, but two live retry timers would open two streams per tab against a 20-subscriber cap. Cleared before rearming.

**LOW — `es.close()` unpinned.** Your diagnosis and your fix. `StubEventSource.dropStream()` now schedules a native re-open unless `closed` is set, which is what makes that line observable; removing `es.close()` now reds the throttle test. Modelling the browser's ~3s retry is something the stub should have done from the start — it is the behaviour the whole error handler exists to take over.

**LOW — `clearTimeout(stableTimeout)` unpinned.** Pinned with a stream that is accepted once and then never again: the armed timer must not fire 60s later and reset `attempt`, so the third connect has to wait 2s rather than 1s. Reverting the block reds it.

**LOW — herd residual.** Taking your version. I had weighed widening the attempt-0 base, which trades recovery latency in the single-tab case; jittering `runResync` itself doesn't, and your narrowing is the part I had wrong — my "the backoff already spreads them" reasoning holds for a real outage but not for a blip, where `lastResyncAt` is 0 and every tab's first reconnect fires an immediate refetch. `onReconnect` is now spread across a 0–2s band, pinned at both ends of it.

**LOW — `start.ts` wiring still silently revertible.** Agreed, and not fixable by a unit test: `start.ts` calls `Bun.serve` at import. Rather than leave it implicit I've written it into the test plan as a stated gap, naming the >90s idle E2E as the only guard on that line.

**MED — PR body.** Rewritten against this head. The `server.timeout`-throws claim, the "±25% / 30s cap" jitter description, and the stale suite count are gone; the fatal-only cadence and the trailing-edge throttle now appear in the Fix section, and the "5 connects, one sync" replay — which described the leading-edge throttle this PR replaced — is gone.

**Verification at `cec8046e`:** `bun run typecheck` clean · `bunx eslint src` clean · `bun test --preload ./scripts/test-preload.ts src` **1173 pass / 2 skip / 0 fail**. Six new mutations run against the round-3 changes (permanent-stop, missing consecutive-reset, jitter removed, warn-per-cycle, plus your two unpinned lines) — each reds exactly the test that claims it. Full E2E re-run against this head, both failure shapes driven against a pre-fix baseline; results in the E2E comment.

claoie:
## E2E — both failure shapes driven against a pre-fix baseline, at `cec8046e`

Re-run in full after the round-3 fix commit, since that commit changed the client half. Two sandboxes on the same scrambled clone, differing only by this fix: the branch head `cec8046e`, and a baseline at `bd86a7de` (the head round 2 reviewed). Identical script against both — real login, bottom-nav → Accounts → account row → detail page, mutating the Name field through the real affordance. Assertions are on tab A's **rendered DOM**, never on a refetch being issued. `pageerror` and console-error watches on both tabs throughout.

**Harness.** Tab A sits behind a reverse proxy that can (a) sever an established 200 stream mid-flight, or (b) stop listening entirely so connections are refused at the socket. (b) is deliberately a socket refusal and not a 502 — a 502 is non-2xx, i.e. the *fatal* shape, and would have proved nothing about the transport shape. Tab B mutates directly against the sandbox and is unaffected by either.

### 1 — a gap that closes inside the resync throttle window (round 2, HIGH 1)

Tab A's stream is severed twice. Tab B's rename lands during the second gap, and A's reconnect falls **17.7s** after the previous resync — inside the 30s window, so the resync is suppressed. The stream then stabilises, so no later `open` can rescue it.

| | baseline `bd86a7de` | branch `cec8046e` |
|---|---|---|
| Tab A reconciled | **never** — watched 75s | after **15.1s**, i.e. 32.8s after the previous resync |
| Same value after a manual reload | yes | yes |

The second row is the control: the write landed on both sides, so the baseline's failure is propagation and not a mutation that silently failed. On baseline the tab renders stale data indefinitely with no error and no clue. On the branch the deferred resync fires at the end of the window (30s + the 0–2s jitter band) and the row updates on its own.

### 2 — an outage longer than the give-up horizon (round 2, HIGH 2)

Origin down for 360s; tab B renames the account mid-outage; origin returns.

| | baseline `bd86a7de` | branch `cec8046e` |
|---|---|---|
| Tab A re-subscribed | **no** — 0 `/api/events` after recovery | yes — 1 `/api/events` |
| Tab A reconciled | never — watched 120s | after **10.2s** |
| Post-recovery API traffic | **none at all** | exactly one of each domain endpoint |

The baseline tab is deaf for the rest of its life: it had exhausted the ceiling while the server was down, and nothing else in the client re-subscribes or re-syncs.

That last row also settles round 2's scenario 4 — **exactly one sync on recovery, not a burst.** Every domain endpoint (`/api/accounts`, `/api/transactions`, `/api/budgets`, `/api/charts`, `/api/securities`, `/api/transfers`, `/api/split-transactions`, `/api/snapshots`) appears exactly once, which is one whole-app `sync()`.

### 3 — propagation past the 90s deadline (round 2, scenario 3)

The headline regression, and the only guard on the `server.timeout` wiring in `start.ts`, which no unit test can reach.

Two tabs held idle **105s** — past the 90s deadline and three 30s keepalive ticks — with no interaction. Then a rename in tab B:

- Tab A's rendered row updated in **1124 ms**, with no reload and no interaction on A.
- **0 `/api/events` requests during the idle window.** That is the load-bearing detail: no reconnect happened, so the keepalive writes are genuinely refreshing Bun's per-request deadline rather than the 90s value merely deferring a single reap.
- `grep -c "request timed out after 10 seconds"` on the sandbox log → **0**.

### Subscriber lifecycle

`SSE subscribed` / `SSE unsubscribed` pairs stay balanced as sockets close, observed draining 7 → 6 → 5 as contexts shut down. A residual count during the run is attributable to a second E2E sweep holding tabs open against the same sandbox concurrently, so I am not claiming a clean zero here — only that the abort-signal cleanup fires and the count goes down, which is what the code path claims.

### Errors observed

Only two kinds across every run, neither a `pageerror`:

- `Service worker registration failed` — pre-existing on `main`, built `/sw.js` is an ES module registered as a classic worker. Already filed as #675, flagged here only so it isn't attributed to this PR.
- `ERR_CONNECTION_REFUSED` / `ERR_INCOMPLETE_CHUNKED_ENCODING` on tab A — these are the deliberate outages, i.e. the test working.

Account names above are scrambled-clone placeholders. No balances, holdings or currency figures are reproduced from the sandbox.

reviewoie (changes requested):
## Code review — PR 674 (round 5)

**Findings:** 3 total — 0 high, 1 med, 2 low. All three are round-4 findings carried forward unchanged.

### First: nothing in this PR changed since round 4

`git diff 90295a70 c74b22b3` over the nine files this PR touches is **empty**. The only delta between the round-4 head and this one is an absorbed upstream commit (`router.ts` + `router.test.ts`, the `view_date`-across-back work), which this PR neither reads nor writes. So the round-4 review's conclusions transfer byte-for-byte, and I re-verified each finding at this head rather than assuming it.

Re-run at head in a detached worktree, not read off the body:

- `bun run typecheck` — clean
- `bun test --preload ./scripts/test-preload.ts src` — **1236 pass / 2 skip / 0 fail**
- CI at head — `build` pass, `test` pass
- Scope — all 9 files map to the title. Clean.
- PII sweep over the diff **and** `git log upstream/main..HEAD --format=%B` — no currency figures, institution names or account labels.
- No `.every(...)` / `.toBe(false)` vacuous assertions in any of the three new/edited test files; 20 state-machine cases, matching the body's claim.
- No other route in `src/server/routes` holds a long-lived response open (`git grep -l 'status: "streaming"'` and the `stream(` callback both return nothing), so `/events` is the only one that needed the deadline. Correct that it is the only caller of `setIdleTimeout`.

**The body's stated "Known gap" checks out, and it is the honest framing.** I no-op'd the `server.timeout` wiring in `start.ts` — replaced the closure with `() => {}` — and got typecheck clean plus **1236 pass / 0 fail**. Nothing in the suite pins that line; only the E2E does. Tree restored, `git status --short` empty.

### Med (carried from round 4, unaddressed and unacknowledged)

- `src/client/lib/hooks/useServerEvents.ts:186` — **`if (everOpened) requestResync()` skips the reconcile when a connection's *first* open follows failed attempts, so that gap is lost for the life of the tab.**

  `everOpened` answers "has this connection ever opened". The thing the resync exists for is "was there a gap the server broadcast into with nobody listening" — and there is one whenever the initial connect is refused or dropped before it succeeds. `attempt` / `fatalAttempts` already carry that fact at the moment `open` fires; nothing reads them.

  **Re-measured at this head, both failure shapes, with a fresh harness driving `createServerEventsConnection` directly:**

  | Scenario | Sources created | `onReconnect` calls |
  |---|---|---|
  | 5 refusals (fatal shape, non-2xx) then accepted | 6 | **0** |
  | 3 dropped 200s (transport shape) then accepted | 4 | **0** |

  Every subsequent flap on that same connection does resync, so the loss is confined to the initial window — and permanent, because nothing else reconciles. Confirmed again at this head: `sync()` is reachable only from `Utility.tsx:30` (login transition) and `Configuration/index.tsx` (manual Refresh); `git grep -nE 'visibilitychange|addEventListener\("(focus|online)|navigator\.onLine' -- src/client` returns nothing.

  Reachable through the two paths this PR's own docs name as heal-by-themselves at `:41-45`: the `MAX_SUBSCRIBERS_PER_USER = 20` cap answering 429 to a newly-opened tab until another tab closes, and `sessionStore.get` failing open so a store blip 401s `/api/events` while the login `sync()` on its own request succeeded. Both windows run minutes under the new backoff — a 429'd tab sits in `RECONNECT_FATAL_SLOW_MS` for 150-450 s — and any peer-tab mutation landing there is never reconciled. Same silent-staleness end state as the issue this PR closes, on a narrower window.

  It also contradicts the invariant the PR states for itself in `useServerEvents.test.ts:217`: *"The server buffers and replays nothing, so every gap needs its own refetch."* The first gap does not get one. Note `useServerEvents.test.ts:342` (`clears the refusal count on a stream that opens`) drives exactly this path — 9 refusals, then an open — and asserts only on source count, never on `syncs`.

  One line: set `hadGap = true` in the error handler and gate on `if (everOpened || hadGap)` (clearing `hadGap` in `runResync`), or equivalently read `attempt > 0 || fatalAttempts > 0` in the `open` handler *before* the resets at `:175-176`. The existing test at `:181` (`does not resync on the first open, only on a re-established stream`) stays green under either — a genuinely clean first open still has `attempt === 0 && fatalAttempts === 0`. A test mirroring the harness above (refuse N times, open, assert one sync) would pin the new behaviour.

### Low (both carried from round 4, both unaddressed)

- **PR body test plan — the count is still stale.** It says `1173 pass / 2 skip / 0 fail`; at this head it is **1236 pass / 2 skip / 0 fail**. Third round carrying a stale number forward (1155-vs-1169, then 1173-vs-1230, now 1173-vs-1236). Re-run before the next handoff rather than carrying it.

- `src/server/routes/events/get-events.test.ts:21` — `await import("./get\-events")` still carries a stray backslash. It resolves (`"\-"` is just `"-"` in a JS string literal, and the test passes), but it reads as a broken escape and no other dynamic import in `src` is written that way — `git grep 'import("\./[a-z-]*\\-'` returns exactly this one hit. Drop it.

### Edge cases worth keeping in the E2E if another sweep runs

The two failure shapes and the >90 s idle case are already covered and reproduce-on-baseline / clean-on-branch, so these are the residuals — unchanged from round 4:

1. **The MED's repro** — open a tab whose `/api/events` is refused from the start (easiest: hold 20 streams open so the 21st takes the 429), mutate from a peer tab during the refusal window, then free a slot. Under the shipped code that tab renders stale data indefinitely; assert its DOM, not a refetch.
2. **Subscriber-set drain under the new close-on-every-error path** — the client now closes and re-requests on every error, so a flap storm churns streams against the 20 cap. Worth confirming `subscriberCount` returns to 0 after the storm rather than transiently over-capping healthy tabs into the MED.

Everything else on this PR is in good shape — the server half is right, the derivation is right, the trailing-edge throttle is right, and the test suite genuinely bites where the body says it does. The one open item is the initial-gap reconcile.

reviewoie (approved):
## Code review — PR 674 (round 6)

**Verdict:** approve

### All three round-5 findings closed at `550ad0a7`

Re-verified against a detached worktree at `550ad0a70581639a7dcd391135234cb3bc645346` with a fresh `bun install`:

- `bun run typecheck` — clean
- `bunx eslint src` — clean
- `bun test --preload ./scripts/test-preload.ts src` — **1295 pass / 2 skip / 0 fail**
- CI at head — `build` success, `test` success
- Scope — 3 files, +49/-2, all map to the round-5 asks. No unrelated changes.
- PII sweep over the commit-message body and diff — clean.

**MED (round-4/5 initial-gap reconcile) — closed.** `useServerEvents.ts:184` now reads `attempt > 0 || fatalAttempts > 0` BEFORE the resets at `:185-193`, and OR-s it into the resync gate at `:196`. Mutation-checked at head: reverting `|| hadPriorFailure` back to `if (everOpened) requestResync()` fails EXACTLY the two new tests (`reconciles the initial gap when the first open follows refused attempts`, `... dropped 200s`) and no others — the round-4 invariant at `:181` (`does not resync on the first open, only on a re-established stream`) still passes, because a genuinely clean first open still has `attempt === 0 && fatalAttempts === 0`. Tree restored, `git status --short` empty.

**LOW 1 (test count) — closed.** The PR body was updated to `1238 pass`. The actual current count is higher — see the one carried-forward LOW below.

**LOW 2 (backslash) — closed.** `get-events.test.ts:21` now reads `await import("./get-events")` and the file's 2 tests pass.

### Low

- **PR body test count is stale again.** Body says `1238 pass / 2 skip / 0 fail`; at this head it is **1295 pass / 2 skip / 0 fail**. The delta is from upstream/main tests landing during the branch's implicit rebase onto `652e491b` (the current `upstream/main` HEAD, 6 min before `550ad0a7`). Same class of drift as rounds 3/4/5 — measure at the current head before every handoff. Non-blocking.

### New-diff-only sweep

- `hadPriorFailure = attempt > 0 || fatalAttempts > 0` is technically redundant in the current implementation — every fatal error also increments `attempt` on the same tick at `:239`, and the slow-lane path is only reachable after `attempt` has already grown to 10. There is no reachable state with `attempt === 0 && fatalAttempts > 0`. The `|| fatalAttempts > 0` half is self-documenting (states the intent that BOTH shapes count as a gap) and defensive against a future refactor that split the two counters, so I'm calling it a deliberate shape rather than a finding.
- No `.every(...)` / `.toBe(false)` vacuous assertions in either new test. Both pin `syncs.toHaveLength(1)` — the cardinality the prose claims.

Everything asked for on round 5 landed. Approving.

hoiekim (approved):

